PR #42, opened 2015-10-24
Add paket.lock and CommonAssembly as ignored, fixed wrong using statement

mdissel:
Add paket.lock and CommonAssembly as ignored, fixed wrong using
statement


jeremydmiller:
@mdissel Got it in, just had to do some manual merging first. This makes you only the second non-bot contributor. Thanks!

mdissel:
You're welcome!

Issue #59, opened 2015-11-09
Load should return objects from unit of work


tim-cools:
Scenario:
- session.Store(student.id = 1)
- session.Load<Student>(1) returns null
(SaveChanges not called yet)

I expect that the second call returns the object stored by the first call from a cache. What do you think?

jeremydmiller:
@tim-cools That probably needs to be part of a larger Identity Map feature. I've gone back and forth on whether or not we should do that underneath DocumentSession.

I can be convinced either way. I can think of a scenario or two within the system at work we're building Marten for that could use it.

tim-cools:
I agree that it normally doesn't happen. But I'm creating an implementation of http://www.asp.net/identity backed-up by PG and I face this situation when I plug in the library for storage.

I also think you need an Identity map to detect dirty objects when doing a SaveChanges to the store if you want to support automatically updates like ravendb and the orm's does.

jeremydmiller:
Done, but there's a separate issue for Select()

PR #63, opened 2015-11-09
Load objects from identity map

tim-cools:
I tried to plug the library into a MVC Identity implementation I'm working on that will us PG. For this I needed to support the loading of not yet stored items as described here: https://github.com/JasperFx/Marten/issues/59

The changes in the pull-request are the changes that I did locally to make my implementation work and the tests turn green. This is not completed and I don't expect you to merge this direct but I created this pull-request to make it easier to discuss whether this is something that you can consider implementing in the library or not.

It currently supports:
- When a document is add to a session by the `Store` method then it can be retrieved from the session by calling the Load method. Even when the SaveChanges is not yet called.
- When a document is added to a session but there is already a document with the same Id and type then an exception is thrown.

Things that need to be done:
- Support for all other DocumentSession methods (Load, Query, ...) should be added
- The Id is fetched from the entity by using a dynamic object. This can probably be improved by using Rosyln generated code.
- This code will also fail when no Id is assigned to an entity yet

Let me know what you think?


tim-cools:
I updated the pull request so that:

- Updates are now tracked by comparing the original json and the new json (copy + tests from https://github.com/jenspettersson/Marten/commit/6876b96558e9d222d303318a61a8ddd3f178cef9) 
- Support for DocumentSession Store, Load, Load by ids and Query(sql) methods
- Code is generated for the retrieval of a document id. The dynamic method is not used anymore

Todo:

- Tracking is not yet supported for the Linq query. Is this something we want? 
- Maybe we should also implement an option so that the tracking can be disabled/enabled? 
- The original and updated documents are compared by json. This can be improved for performance. I thing we could generate code to extract the state from an object. This in a form of and array of property name and value:
"Id":"12",
"Name:":"Tim"
"Address.Street":
Or it could be enough to use the JilSerializer. Probably we need some benchmarks if we want to make this performant...

PR #67, opened 2015-11-11
Use Roslyn packages from stable NuGet feed instead of nightly MyGet feed

danbarua:
As per discussion in Gitter.


jeremydmiller:
Thanks @danbarua

PR #87, opened 2015-11-20
Support Last() and LastOrDefault() queries (closes #41)

### LINKED ISSUE #41: Support Last() and LastOrDefault() in Linq support
Laters...

dpen2000:
Basically ensure first tests pass when switched to last but am setting flag property and checking for this to ensure it really does return last one

Hope this is okay.


jeremydmiller:
@dpen2000 Cool and thanks. I'll have this pulled in sometime on Monday. I'm going to open a separate issue to optimize both the First() and Last() by using LIMIT 1. W/ the Last(), maybe we could try to reverse the ordering, then do LIMIT 1. Or maybe there's a postgres trick I don't know yet.

Issue #89, opened 2015-11-23
Optimize the First() & Last() mechanics by limiting the results


jeremydmiller:
Instead of doing a complete fetch and then grabbing the first or last in that array, let's try to make the DB query only return the correct value

ekulakov:
@jeremydmiller If someone need `Last()` then he should use `OrderByDescending()` + `First()` instead of `Last()`.

From PostgreSQL docs: When using LIMIT, it is important to use an ORDER BY clause that constrains the result rows into a unique order. Otherwise you will get an unpredictable subset of the query's rows. You may be asking for the tenth through twentieth rows, but tenth through twentieth in what ordering? The ordering is unknown, unless you specified ORDER BY. [Source](http://www.postgresql.org/docs/8.1/static/queries-limit.html)

BTW EF is not supported `Last()` too.

jeremydmiller:
@ekulakov That's a pretty easy argument to accept. First() only, and Last() will throw an exception telling you to order by descending, then first.

jeremydmiller:
This should be marked as "closed" by commit https://github.com/JasperFx/marten/commit/ba0d06f4d5f1cbeed294ef59d8e1e6929969b1b6

PR #90, opened 2015-11-23
Optimize first

dpen2000:
@jeremydmiller, A first stab at implementing #89 

I've made the first tests slightly harder to pass by checking the returned result has flag set but this may not be such a good move as we actually guarantee postgresql will return the results in insertion order (http://www.postgresql.org/docs/9.1/static/queries-order.html says results will come back in "unspecified order" if no sort is set). What do you think?

Last is a little trickier to push down. Either as you suggested in #87, we reverse the order but we can't do this if there's no user provided order. Or we can do the wrap the original query and skip the count of this subquery minus one but that's quite a big change to the query construction.

### DIFF
--- a/src/Marten.Testing/Linq/invoking_queryable_through_first_Tests.cs
+++ b/src/Marten.Testing/Linq/invoking_queryable_through_first_Tests.cs
@@ -44,28 +44,28 @@ namespace Marten.Testing.Linq
                 .ShouldBeNull();
         }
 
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Sweet, I'll get this in tomorrow.

dpen2000:
Cool. I might have a try implementing Last optimized soon via the second method I outlined.

PR #93, opened 2015-11-25
Session request counter

### LINKED ISSUE #85: Track number of requests per session
In RavenDB they have `MaxNumberOfRequestsPerSession` on the DocumentSession that's defaulted to 30. If more than 30 requests is made on one session, an exception is thrown.

This is a really good way of preventing easy mistakes like registering something as singleton which also has a dependency to the session. That session would live forever and that's probably not a good idea.

The only negative thing about this is that it might work great locally for a developer with low usage, but then throw in production when the usage goes up. But in my opinion, it's worth it.

One solution could be to not throw exception and instead log warnings, but that might be missed and/or easily ignored.

jenspettersson:
Simple request counter with the possibility to invoke a predefined action if configured threshold is exceeded. This is to help find sessions that might be used incorrectly (i.e. live way too long, do way too much).

**Note!** By default no threshold is set! If you want this feature you have to explicitly define a threshold and an action to invoke if that threshold is exceeded. We don't want any exceptions to occur in production just because the load was higher than in dev/test. If no threshold is set, you can still see the number of requests when debugging. Haven't made the number of requests publically available as I didn't find a fitting place where I could expose it (if it's even needed).

I extended the `StoreOptions` with `WithRequestThreshold(RequestCounterThreshold threshold)` and stored the threshold in the Advanced property on the DocumentStore. Not sure this is the best place to have it.

Usage:

```
var documentStore = DocumentStore.For(opt =>
{
    opt.Connection(ConnectionSource.ConnectionString);
    opt.WithRequestThreshold(new RequestCounterThreshold(5, () =>
    {
        throw new Exception("Number of requests exceeded configured threshold of 5!");
    }));
});
```

Closes #85


jeremydmiller:
@jenspettersson Hey, what if we don't even apply the decorator if the request threshold isn't applied to a document store? Maybe cut a new method in DocumentStore for `getCommandRunnerForSession()` or something that builds up the runner for you? So you'd return the wrapped version for request thresholds, but just use the normal CommandFactory otherwise?

jenspettersson:
@jeremydmiller Yeah, sounds good. Agree that it's probably unnecessary to apply it always. Will try to look into it tomorrow.

jenspettersson:
@jeremydmiller Changed so the `RequestCounter` only wraps the `ICommandRunner` if it's defined. So if you don't define it, it will always be the default `ICommandRunner` that's used.

I made a rather naive `GetCommandRunnerForSession()` method that just checks if a threshold is set. Keeping it simple, a bit hesitant to introduce possible complexity in a, still, rather foreign code base ;) Feel free to change/trash it any way you'd like!

jenspettersson:
Now when Marten isn't required to use a IoC container I kind of have to leave a footprint in the DocumentStore. If using a container you could just decorate the ICommandRunner and the DocumentStore wouldn't know a thing, which would be nice. But I really like the non-IoC-requirement you put in recently...

src/Marten/DocumentStore.cs
```diff
@@ -55,7 +55,7 @@ public DocumentStore(StoreOptions options)
             _serializer = options.Serializer();
 
             var cleaner = new DocumentCleaner(_runner, Schema);
-            Advanced = new AdvancedOptions(cleaner);
+            Advanced = new AdvancedOptions(cleaner, options.RequstThreshold());
```
phillip-haydon: `RequstThreshold` ? Spelling mistake?

src/Marten/StoreOptions.cs
```diff
@@ -35,10 +34,20 @@ public void Serializer<T>() where T : ISerializer, new()
...
+        {
+            _threshold = threshold;
+        }
+        
         public readonly MartenRegistry Schema = new MartenRegistry();
 
         public bool AutoCreateSchemaObjects = false;
+        private RequestCounterThreshold _threshold = RequestCounterThreshold.Empty;
 
+        internal RequestCounterThreshold RequstThreshold()
```
phillip-haydon: Ah here it is. Yup spelling mistake :)

PR #103, opened 2016-01-04
Add async methods

ekulakov:
Added async versions: Any, Count/LongCount, First/FirstOrDefault, Single/SingleOrDefault, Last/LastOrDefault.
Code of extensions was taken from EF7 [EntityFrameworkQueryableExtensions](https://github.com/aspnet/EntityFramework/blob/0b96c132f41442a10064156798d18d2526c3c0a7/src/EntityFramework.Core/EntityFrameworkQueryableExtensions.cs).
Is it a problem?


jeremydmiller:
@ekulakov I think that's perfectly fine w/ their license now. I'll get this in shortly.

PR #111, opened 2016-01-29
Reduced unnecessary `await` statements and added missing `ConfigureAwait(false)`

danielmarbach:
* Reduces the amount of async statemachines created and therefore reduces the allocations and GC pressure
* Adds necessary `ConfigureAwait(false)


jeremydmiller:
@danielmarbach Got it, and thank you. Guess you can see that we opted for the "Something"Async() methods instead of a separate interface ala Raven.

PR #123, opened 2016-02-01
Syntax spike for Compiled Queries

jeremydmiller:
Just demonstrating one possibility for the CompiledQuery syntax. This is somewhat inspired by [EF's support](https://msdn.microsoft.com/library/bb896297(v=vs.100).aspx).

Another alternative that would be a touch trickier behind the scenes would be:

``` csharp

IDocumentQuery<IList<User>, string, string> query = CompiledQuery
    .Against<User, string, string>()
    .For((users, first, last) => users.Where(x => x.FirstName == first && x.LastName == last));

// Against an IQuerySession, you'd get:
var users = query.Execute(session, "Lebron", "James");

```

I'm assuming that there'd be some overloads for selecting first/single/count and whatever else makes sense.

What do you think?

### DIFF
--- a/src/Marten.Testing/CompileQueryDemonstrator.cs
+++ b/src/Marten.Testing/CompileQueryDemonstrator.cs
@@ -1,0 +1,35 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Baseline;
+using Marten.Linq;
+using Marten.Testing.Documents;
+using StructureMap;
+using Xunit;
+
+namespace Marten.Testing
+{
+    public class CompileQueryDemonstrator
+    {
+        [Fact]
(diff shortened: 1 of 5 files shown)


jeffdoolittle:
There's part of me that likes the static class syntax and part of me that doesn't. It limits discoverability when you use the static class. But it highlights the "disconnected" nature of the query - that it's not tied to a session.

CoreyKaylor:
From gitter:

I like that you favored the interface over the func return type.
Wonder if there’s a deterministic way to key the expression at runtime and provide a memoized query plan / translation.

jeremydmiller:
@CoreyKaylor put a different alternative on the board I like a lot better than this one. I'm closing this PR.

Issue #125, opened 2016-02-02
improved naming conventions for nested classes


jeffdoolittle:
Ability to override table names or specify conventions for naming. For a scenario like the code below, there are name collision issues.

```
public class Foo
{
    public class Aggregate {}
    public class State {}
}

public class Bar
{
    public class Aggregate {}
    public class State {}
}
```

jeremydmiller:
@jeffdoolittle What's your preference here? I do vote for an attribute for the table name (might already exist, I've slept since then.) For nested types, what if the example above were to be `mt_doc_foo_aggregate` and `mt_doc_bar_aggregate`?

jeffdoolittle:
I think that's the sensible default.

jeffdoolittle:
@jeremydmiller I can work on this if you want to assign me the issue

jeremydmiller:
@jeffdoolittle I'm taking you up on that;)

jeremydmiller:
You might need to assign yourself, or I can add you to the GH team

PR #126, opened 2016-02-02
closes JasperFX/marten#125 nested document naming

jeffdoolittle:
@jeremydmiller ready for your review

### DIFF
--- a/src/Marten/Schema/DocumentMapping.cs
+++ b/src/Marten/Schema/DocumentMapping.cs
@@ -165,12 +165,24 @@ namespace Marten.Schema
 
         public static string TableNameFor(Type documentType)
         {
-            return "mt_doc_" + documentType.Name.ToLower();
+            return "mt_doc_" + DocumentPersistentName(documentType);
         }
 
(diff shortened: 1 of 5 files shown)


src/Marten/Schema/DocumentMapping.cs
```diff
@@ -165,12 +165,24 @@ public PropertySearching PropertySearching
...
         {
-            return "mt_upsert_" + documentType.Name.ToLower();
+            return "mt_upsert_" + DocumentPersistentName(documentType);
+        }
+
+        private static string DocumentPersistentName(Type documentType)
+        {
+            var parts = new List<string>();
+            parts.Add(documentType.Name.ToLower());
+            if (documentType.IsNested)
```
jeffdoolittle: not sure if we'd want to go recursively to another level of nesting or now. for now I figured one level deep was sufficient.

jeremydmiller: I think that's fine for now. YAGNI and all that;)

PR #131, opened 2016-02-04
Fix for #130

### LINKED ISSUE #130: Enable case insensitive custom sql queries
When you try to query a table with custom sql, let's say 
`Session.Query<Account>("SELECT * FROM mt_doc_account") `
it fails on the **SELECT** command if it's uppercase.

ekulakov:
Closes #130


jeremydmiller:
Got it, and thanks!

PR #147, opened 2016-02-10
added support for modulo where queries

nieve:
Not 100% sure about my implementation- should I have gone for a ModuloWhereFragment rather than re-appropriating the standard WhereFragment?
Would rather do it right- don't hesitate to say this sucks :)

Also, I've got `autocrlf = false` on my git config. Is that going to unleash endline hell on you guys? If so, let me know, I'll fix it.

### DIFF
--- a/src/Marten.Testing/Linq/query_with_modulo_Tests.cs
+++ b/src/Marten.Testing/Linq/query_with_modulo_Tests.cs
@@ -1,0 +1,28 @@
+using System.Linq;
+using Marten.Services;
+using Marten.Testing.Fixtures;
+using Xunit;
+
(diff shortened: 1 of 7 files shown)


jeremydmiller:
Hey @nieve, I think this is gonna take a little rework (plus line ending problems you already caught). Can I catch up with you early next week on this?

nieve:
Cool, just let me know when. In the mean time I'll fix the line ending issue and take another stub at it, as I can see there are conflicts...

nieve:
btw- if it's easier for you, feel free to drop a few lines here whenever you've got the time, and I'll take it from there. No rush.

PR #149, opened 2016-02-11
added support for modulo where queries

nieve:
I am *a lot* happier with this implementation :-P
btw- I think MartenExpressionParser starts to look like a class in need for refactoring maybe..?


jeremydmiller:
@nieve,

Cool, I'll pull this in when Teamcity is done with it.

"I think MartenExpressionParser starts to look like a class in need for refactoring maybe..?" - Ya think? ;-)

Cleaning up the Linq parsing code is on my TODO list.

Issue #148, opened 2016-02-11
General purpose logging mechanism for Marten


jeremydmiller:
I'm proposing a new interface that you can plug into Marten to receive logged events for:

* Opening or disposing a new session
* Each NpgsqlCommand executed by Marten, and hopefully in a way so that you can easily correlate
  requests to a session
* Command failures
* Command performance?

phillip-haydon:
I'm not sure if its possible, but would be nice to extract `EXPLAIN` from queries too. Which probably falls under Command performance.

jeremydmiller:
Before we put any effort into this, there is a facility in Npgsql to log commands. Might be worth either using that or even chipping into that project. It's just brute force logging with no context, but still.

jeremydmiller:
Meh, I've tried a couple different concepts for this so far and there's too many issues with each. Npgsql's facility is limited, but at least it would catch everything that's done to the open connection. Right now, I'm leaning toward talking to that team and maybe trying to do some pull requests to them so that we could swap out their logger per command/connection so we could do session-based, contextual logging.

PR #151, opened 2016-02-12
Spike of what IMartenLogger might look like

jeremydmiller:
This is in regards to https://github.com/JasperFx/marten/issues/148.

I'm proposing these two new, pluggable interfaces as a mechanism to basically listen for any kind of information at runtime. This would allow the request governor approach or performance metrics or just exceptions from the sql being executed.

*My* big wish from the diagnostics is being able to associate DB activity with semantic elements of the application like "what HTTP URL was this" or "what message was the service bus handling?"

### DIFF
--- a/src/Marten/IMartenLogger.cs
+++ b/src/Marten/IMartenLogger.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using Npgsql;
+
+namespace Marten
+{
+    // Use this to inject default session loggers into DocumentSessions when opened
+    public interface IMartenLogger
+    {
+        // Somehow, I'd like to be able to associate the description
+        // with what the system is doing. In our web based systems, I'd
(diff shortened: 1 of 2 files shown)


jenspettersson:
Might as well comment here as well as in the gitter chat.

I like it... The naming: is it a SessionLogger or a CommandLogger? Isn't the command just a part of the Session (that might hold more than the command) and the interface just wraps the command atm?

Or is it "a Session against the db" and that is just the command?

jeremydmiller:
@jenspettersson I'm not locked in on the name, but I wanted it clear that that logger was supposed to live for the lifecycle of the IDocumentSession.

jenspettersson:
Ok, I really don't have any issues with the naming, was just a bit curious. It's a good thing it's clear that it's the lifecycle of the IDocumentSession.

Issue #175, opened 2016-02-25
`DateTime`s incorrectly mapped to `date`


trullock:
`Marten.Util.TypeMappings` is treating `DateTime`s as `date`s,

see line ~21 and and ~62

Suggest these are changed to `timestamp`s

trullock:
@jeremydmiller said:
```
private static readonly Dictionary<Type, string> PgTypes = new Dictionary<Type, string>
        {
            {typeof (int), "integer"},
            {typeof (long), "bigint"},
            {typeof (Guid), "uuid"},
            {typeof (string), "varchar"},
            {typeof (Boolean), "boolean"},
            {typeof (double), "double precision"},
            {typeof (decimal), "decimal"},
            {typeof (DateTime), "date"}, // We need to change this to timestamp, right?
            {typeof(float), "decimal" },
            {typeof (DateTimeOffset), "timestamp with time zone"}
        };
```

trullock:
> {typeof (DateTime), "date"}, // We need to change this to timestamp, right?

I think so, and what about the other line: 

```DateTime => NpgSqlType.Date?```

jeremydmiller:
@trullock Probably the same. Npgsql itself has this exact functionality, but it's buried behind private/internal scoping. Might be worthwhile either getting them to take a PR for that or just cheating and grabbing things through reflection (shudder).

jeremydmiller:
This was addressed in https://github.com/JasperFx/marten/pull/177

PR #177, opened 2016-02-29
Change DateTime mapping to timestamp without time zone

mholland:
Fixes issue #175 

Changed `DateTime` mapping to `timestamp without time zone` and removed manual override from the `Npgsql.TypeHandlerRegistry` which returns `Timestamp`


jeremydmiller:
Got it, I think I'll have this one in this morning. Thanks!

Issue #150, opened 2016-02-12
Formal concept of foreign key relationships between documents


jeremydmiller:
Partially to make the Include() feature more efficient (GH-146), but also to take advantage of relational integrity checks in the database, let's support foreign key relationships between documents. Implementation wise, this would probably mean a special kind of "Searchable"/duplicated field that also has a foreign key constraint generated by the DDL. Hopefully we'd be able to take advantage of this foreign key in making the joins when doing Include.

jeremydmiller:
To add to this, I think we could either allow users to just refer to the key of another document:

```
public Guid DocumentId {get; set;}
```

and configure that via something like:

```
For<SomeDocument>().ForeignKey<TOtherDoc>(x => x.Other.Id, CanBeNull = true/false);
```

That would just end up creating a searchable field with a foreign key relationship to the other document type. Easy enough.

We could also do this:

```
public OtherDocType Other {get;set;}
```

and assuming that `OtherDocType` is a document type with an Id in the Marten store, we could also do the foreign key + searchable field for `Other.Id`.

That begs a couple new questions though, 

Do we want to make loading and attaching the OtherDocType reference automatic that obviates the need for doing Include() on that relationship? Or do we make that configurable? Or do we say that the load with an Include() has to be explicit?

jeffdoolittle:
Perhaps these could be configurable options on the DocumentStore when it is spun up, and the overridable on a per document basis? Although I lean toward the sensible default being that the relationships are not included by default, but others may feel differently.

jeremydmiller:
Absolutely, yes to having it be done per document/relationship.

PR #188, opened 2016-03-08
Concept of foreign key relationships between documents

ekulakov:
I tried to implement #150 
There are several problems:
- unable to save Issue without User in db (store for foreign key type is not populated)
- order of entities while persisting


jeremydmiller:
@ekulakov I need to look closer at this, but for the ordering of inserts/updates, we had some dependency tree ordering functionality in FubuCore that I was thinking we could port to Baseline that Marten uses. If we had that in place, we'd force the UnitOfWork class into applying all of its work in the dependency order.

jeremydmiller:
This was pulled in w/ a rebase today.

PR #194, opened 2016-03-09
Failing test for appendingevents to a newstream

tim-cools:
Failing test when appending events to a new stream:

It throws:

Marten.Testing.Events.end_to_end_event_capture_and_fetching_the_stream_Tests.capture_events_to_a_non_existing_stream_and_fetch_the_events_back [FAIL]
Npgsql.NpgsqlException : 23502: null value in column "type" violates not-null constraint

Where "type" is stream_type in this case.

This is because, unlike in StartStream, the AggregateType is not set:

StartStream: 
var stream = new EventStream(id, events)
            {
                AggregateType = typeof(T)
            };

AppendEvents:
var stream = new EventStream(id, events);

I like to support this scenario because I can write infrastructure logic that is unaware of the fact whether a stream (aggregate) already exists. 

But this can only be solved it we add the Aggregate type as Generic parameter. Is this a valid solution for you?

I can also implement the void Append<T>(Guid stream, T @event) where T : IEvent method. Maybe it should be renamed to AppendEvent ?

### DIFF
--- a/src/Marten.Testing/Events/end_to_end_event_capture_and_fetching_the_stream_Tests.cs
+++ b/src/Marten.Testing/Events/end_to_end_event_capture_and_fetching_the_stream_Tests.cs
@@ -10,17 +10,7 @@ namespace Marten.Testing.Events
         [Fact]
         public void capture_events_to_a_new_stream_and_fetch_the_events_back()
         {
-            var store = DocumentStore.For(_ =>
-            {
-                _.AutoCreateSchemaObjects = AutoCreate.All;
-
-                _.Connection(ConnectionSource.ConnectionString);
-
-                _.Events.AddEventType(typeof (MembersJoined));
-                _.Events.AddEventType(typeof (MembersDeparted));
-            });
-
-            store.Advanced.Clean.DeleteAllEventData();
+            var store = InitStore();
 
             using (var session = store.OpenSession())
             {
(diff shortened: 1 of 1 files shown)


jeremydmiller:
I've been going back and forth on whether or not the stream type/Aggregate type has to be captured. I think there's some value in having the concept for being able to do snapshots and per stream projections, but I can be talked into making it be non required. I vote for making the aggregate type optional, and letting stream_type be a nullable field.

tim-cools:
Ok, I agree to make it optional. Streams are not always linked to an aggregate type anyway.

What I do in EventStore for aggregate streams is using aggregate type as prefix in the id by convention. So it would be 'product/123456-789456'. But the Id is a string there.
And in EventStore you also have projected streams that contain events linked from multiple aggregates. In this case you don't have an aggregate type but does the stream id represent the projection partition. Eg stream id: 'category-someCategory'. The type of stream might be used for this. But it feels a bit hackish to use type for this. Do you plan to support such projection streams to? Or do you plan to project always to document(s)? It is fancy but I don't really need projections to streams to be honest.

I don't mind updating this. I can make the field optional in the sql script and support both a generic an non-generic version of the AppendEvent method. We can even support a version wher you can provided the stream type as string later.
Do you want to support some kind of db upgrade? I don't really need it yet

jeremydmiller:
This is in, and I made the one line change to fix the test;)

tim-cools:
Thx :)

PR #198, opened 2016-03-10
Elementary support for projections: fetch only the json fields used i…

jokokko:
…n a given projection.

Expose IDocumentStore from IQuerySession - this should allow for more flexible extension methods


CoreyKaylor:
From gitter...

I’m thinking the projection support should behave the same as the LINQ Select(x => new ProjectionTarget { ThirdValue = x.ThirdValue, Value = x.Value });

Probably should also allow for non-matching property names in the select statement mapping if possible.

PR #200, opened 2016-03-10
Add document loaded/addedForStorage interception methods

tim-cools:
I added two methods to the IDocumentSessionListener to intercept when documents are loaded/stored.

void DocumentLoaded(object id, object document);
void DocumentAddedForStorage(object id, object document);

This interception is currently implemented in the default and DirtyChecking IdentityMap.

I was tempted to split the IDocumentSessionListener into a IDocumentSessionListener and IDocumentListener. The latter would contain the new methods. But I decided to go for the simplest implementation for now. Let me know what your thoughts are about this?

If you agree I can add this also to the documentation if you want.

### DIFF
--- a/src/Marten/Services/IdentityMap.cs
+++ b/src/Marten/Services/IdentityMap.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Baseline;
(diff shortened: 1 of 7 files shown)


ekulakov:
@tim-cools  Maybe I'm wrong but I think you could write a decorator and reuse it for all IdentityMap's

```
    public class IdentityMapListenersDecorator : IIdentityMap
    {
        private readonly IIdentityMap _identityMap;
        private readonly IEnumerable<IListener> _listeners;

        public IdentityMapListenersDecorator(IIdentityMap identityMap, IEnumerable<IListener> listeners)
        {
            _identityMap = identityMap;
            _listeners = listeners;
        }

        public T Get<T>(object id, Func<FetchResult<T>> result) where T : class
        {
            var obj = _identityMap.Get(id, result);
            _listeners.Each(x => x.DocumentLoaded(id, obj));
            return obj;
        }

        ...
    }
```

tim-cools:
@ekulakov thanks for the feedback!This is however not possible because the listeners are only called when a document is loaded the first time and not when the document is returned from the cache. 

https://github.com/tim-cools/Marten/commit/306bb1767ee1396d7b5ff3df64b956eebc0c40cc#diff-99cac1569732f4ed1d27aceb9e8f52ebR64

A decorator is unaware of the internals of the decorated class. We might consider pulling out a generic base class for both IdentityMaps however. I will look into that later...

src/Marten/Services/IdentityMap.cs
```diff
@@ -9,18 +11,25 @@ namespace Marten.Services
...
+        private readonly IEnumerable<IDocumentSessionListener> _listeners;
 
         private readonly Cache<Type, ConcurrentDictionary<object, object>> _objects
             = new Cache<Type, ConcurrentDictionary<object, object>>(_ => new ConcurrentDictionary<object, object>());
 
-        public IdentityMap(ISerializer serializer)
+        public IdentityMap(ISerializer serializer, IEnumerable<IDocumentSessionListener> listeners)
         {
             _serializer = serializer;
+            _listeners = listeners?.Any() == true ? listeners : null;
```
jeremydmiller: @tim-cools I'd rather we used a "nullo" empty list here instead of having to remember to always use the "?" operator. Nit pick-y, I know.

jeremydmiller:
I've got a minor comment about using a nullo for the list of listeners, but other than that I think this looks good. The build passed, and I'm pulling this in now.

src/Marten/Services/IdentityMap.cs
tim-cools: no problem, will update the request. it was a silly attempt of premature optimization...

PR #202, opened 2016-03-11
added sum queries

### LINKED ISSUE #157: Support aggregate functions possible?
From a little bit of googling, it appears postgresql will let you do aggregate functions from jsonb documents, i was hoping marten could do simple queries like the following test:

```
using (var session = store.OpenSession())
{
	session.Store(new Order() { Id = "order/1", Total = 10 });
	session.Store(new Order() { Id = "order/2", Total = 12 });
	session.SaveChanges();
}

using (var session = store.OpenSession())
{
	var result = session.Query<Order>().Sum(a => a.Total);
	result.ShouldBe(22);
}
```

However currently this returns the following error:
Marten.Schema.InvalidDocumentExceptionCould not determine an 'id/Id' field or property for requested document type System.Double

nieve:
Fixes #157 
I refactored some code in MartenQueryExecutor- feel free to let me know if this could be done better :-)

### DIFF
--- a/src/Marten/Linq/MartenQueryExecutor.cs
+++ b/src/Marten/Linq/MartenQueryExecutor.cs
@@ -8,7 +8,6 @@ using Marten.Schema;
 using Marten.Services;
 using Npgsql;
 using Remotion.Linq;
-using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.ResultOperators;
 using Remotion.Linq.Parsing.Structure;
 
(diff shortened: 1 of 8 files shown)


src/Marten/Linq/MartenQueryExecutor.cs
```diff
@@ -49,35 +49,16 @@ private IResolver<T> resolver<T>()
...
-                documentQuery.ConfigureForCount(countCommand);
-
-                return _runner.Execute(countCommand, c =>
-                {
-                    var returnValue = c.ExecuteScalar();
-                    return Convert.ToInt32(returnValue).As<T>();
-                });
-            }
-
-            throw new NotSupportedException();
```
nieve: Should I move this up into a field by replacing the whole generic results with an object?

PR #207, opened 2016-03-14
Added case insensitive string support

JeffN825:


jeremydmiller:
@JeffN825 Code looks fine, but it needs tests.

jnevins-gcm:
I must admit, I'm super confused by the layout of your tests. I can't seem to figure out what the naming convention is, or what tests belong where. I was at first thinking to use LinqTests (which is where I put the code), but I don't see that used anywhere. I added a test to the storyteller xml...is that what you're referring to?

jeremydmiller:
Aw, I'm sorry, I missed the extra Storyteller assertion. That's good enough for me.

We're organizing the tests more by use case & functionality than by classes. Most of the tests are integration tests so far.

jeremydmiller:
It's in after a quick rebase. Thanks!

PR #208, opened 2016-03-14
Added no-op id generation support for non public id members.

JeffN825:


jeremydmiller:
@JeffN825 Cool, hadn't really thought of this case. Can you add some tests thought? Build a DocumentMapping w/ that type and see what Id generation strategy is assigned maybe?

jnevins-gcm:
I must admit, I'm super confused by the layout of your tests. I can't seem to figure out what the naming convention is, or what tests belong where. I guess I'd add this to Testing.Generation? And the model class to the Examples namespace?

jeremydmiller:
I'd put this in the `Marten.Testing.Schema` namespace, and just put the sample doc class in the same file as your test file. Just something descriptive like `uses_no_id_generation_for_non_public_id_Tests` for the name.

JeffN825:
Added a test.

PR #209, opened 2016-03-14
2nd Spike for compiled query api

CoreyKaylor:
This is here for feedback on a new way to do compiled queries. I think this approach has a few benefits.

1. Doesn't hugely deviate from existing API in a way that would make it unappealing to use.
1. Encourages query re-use by type instead of instance of a compiled func.
1. Low-bar on complexity of use.
1. Minimizes generic parameter explosion for overloads.
1. Should be easier to play nice with async, batch, etc.


CoreyKaylor:
More current implementation
https://github.com/JasperFx/marten/pull/263

Issue #179, opened 2016-03-03
IdentityMap may generate the same HashCode for different ids


jenspettersson:
In the IdentityMap (https://github.com/JasperFx/marten/blob/master/src/Marten/Services/IdentityMap.cs#L69) the generated HashCode is not always unique.

Really hard to provide a failing unit test as the generated hashcode differs from my lab project to a test in Marten.Testing. From MSDN:

> However, there is not a unique hash code value for each unique string value. Different strings can return the same hash code.

and

> ...they should never be used as key fields in a collection...

It's now used as a key in the dictionary in the IdentityMap and may or may not throw exceptions when trying to store two completely different objects in the same session just because their ids generated the same hashcode.

jeremydmiller:
Meh, okay.

PR #210, opened 2016-03-14
Extract base identity map

tim-cools:
DirtyTrackingIdentityMap and IdentityMap now share most of their logic in a base class.

This probably fixed a bug with duplicated id's in DirtyTrackingIdentityMap by using object instead of int (id.GetHashcode())
See: https://github.com/JasperFx/marten/issues/179

+ Plus use a default list of listeners instead of a null value (see previous PR)


jeremydmiller:
@tim-cools Hey Tim, can you make sure that's rebased on master? We did switch the identity map to using `object` because of some problems that @jenspettersson ran into.

tim-cools:
@jeremydmiller true, but that was only fixed in IdentityMap but not in DirtyTrackingIdentityMap (https://github.com/JasperFx/marten/pull/181)

jeremydmiller:
Ugh, okay. Thanks for doing this, I'll get it in my tomorrow morning.

jenspettersson:
Damn, my bad. I only ran into #181 in the IdentityMap, never tried the DirtyTrackingIdentityMap and missed it completely. Sry guys

jenspettersson:
@jeremydmiller @tim-cools want me to fix the `DirtyTrackingIdentityMap` in master or shall we just wait until this gets pulled in?

tim-cools:
No worries, I only ran into this because I noticed that they both used a different key type. That is the nice thing about looking at the code with many different people. 
And it should be fixed with this pull-request, so I would not bother fixing it on master. They now share a base class with the main logic so we only have to maintain it in one place.

jeremydmiller:
Got it!

Issue #205, opened 2016-03-13
Add support for max queries


nieve:
This fails:
```
[Fact]
public void get_max()
{
    theSession.Store(new Target { Color = Colors.Blue, Number = 1 });
    theSession.Store(new Target { Color = Colors.Red, Number = 2 });
    theSession.Store(new Target { Color = Colors.Green, Number = 3 });
    theSession.Store(new Target { Color = Colors.Blue, Number = 42 });

    theSession.SaveChanges();
    var maxNumber = theSession.Query<Target>().Max(t => t.Number);
    maxNumber.ShouldBe(42);
}
```

Looking into it :)

nieve:
I'll be also happy to pick up Min and Average afterwards :)

ekulakov:
You can copy IQueryable extensions from [EF Extensions](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs)

jeremydmiller:
Addressed with https://github.com/JasperFx/marten/pull/212

PR #211, opened 2016-03-14
added support for min, max & average

nieve:
#205


jeremydmiller:
We used GH-212 instead.

PR #213, opened 2016-03-14
Temp workaround for .net core code generation

tim-cools:
Do you mind pulling this too? It is the temp workaround for .net core code generation. I have no dnx tests in this solution but it works in a dnx app I'm working on.


jeremydmiller:
@tim-cools This breaks a unit test for hierarchical storage. Can you check it out before I pull this in? http://build.fubu-project.org/viewLog.html?buildId=47149&tab=buildResultsDiv&buildTypeId=Marten_Master

Thanks,

Jeremy

tim-cools:
@jeremydmiller updated

PR #214, opened 2016-03-14
Async usage

danielmarbach:
* A few `ConfigureAwait(false)` missing
* A few micro-optimizations to return instead of `await`
 * I used two times `AsEnumerable` to be able to return instead of `await`. `AsEnumerable` directly returns the source enumerable. This change is questionable since it could lead to more allocations. But I figured it can be neglected because we are returning an enumerable anyway on the public method and I think `await` is actually more expensive. BUT, I didn`t measure. So I could be wrong.


jeremydmiller:
It's in, thanks Daniel!

PR #215, opened 2016-03-15
Support for IEnumerable.Contains

robinvanderknaap:
First attempt: I've created another method call parser, called IsInGenericEnumerable. Basically the same as IsOneOf, but now you can use IEnumerable.Contains.


jeremydmiller:
It's in, thanks!

PR #216, opened 2016-03-15
ResolveAsync all the way

danielmarbach:
Introduces `ResolveAsync`

Caveat: I'm not a DbReader expert. So I could be horribly wrong!

Based on my quick research the indexers of the DbReader and SqlDbReader do internal IO-bound calls if the value is not already fetched. This PR introduces a fully async path to resolve the json documents with `GetFieldValueAsync`.

### DIFF
--- a/src/Marten/Events/EventMapping.cs
+++ b/src/Marten/Events/EventMapping.cs
@@ -159,6 +159,14 @@ namespace Marten.Events
             return map.Get<T>(id, json);
         }
 
+        public async Task<T> ResolveAsync(DbDataReader reader, IIdentityMap map, CancellationToken token)
+        {
+            var id = await reader.GetFieldValueAsync<Guid>(0, token).ConfigureAwait(false);
+            var json = await reader.GetFieldValueAsync<string>(1, token).ConfigureAwait(false);
+
+            return map.Get<T>(id, json);
+        }
+
         public T Build(DbDataReader reader, ISerializer serializer)
         {
             return serializer.FromJson<T>(reader.GetString(0));
--- a/src/Marten/Linq/DeserializeSelector.cs
+++ b/src/Marten/Linq/DeserializeSelector.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
 using Marten.Schema;
 using Marten.Services;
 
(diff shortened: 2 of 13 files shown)


src/Marten/Events/EventMapping.cs
```diff
@@ -159,6 +159,14 @@ public T Resolve(DbDataReader reader, IIdentityMap map)
             return map.Get<T>(id, json);
         }
 
+        public async Task<T> ResolveAsync(DbDataReader reader, IIdentityMap map, CancellationToken token)
+        {
+            var id = await reader.GetFieldValueAsync<Guid>(0, token).ConfigureAwait(false);
+            var json = await reader.GetFieldValueAsync<string>(1, token).ConfigureAwait(false);
```
danielmarbach: @jeremydmiller Can you tell me why event mapping has the fields reserved? So in the other classes the id is on index 1 and the json on index 0. See for example https://github.com/JasperFx/marten/pull/216/files#diff-bcb01cdb23b98acc4aefa1d606ef994bR76

jeremydmiller:
@danielmarbach Having the indexes reversed isn't purposeful. I'm not thrilled with having to have an alternative async mechanism for reading a single row of a data reader. Is this really necessary? It's going to cause a **lot** of extra code down the line for `ISelector` and `IReceiver` implementations.

danielmarbach:
@jeremydmiller the indexer approach does internal IO calls when necessary, hooks up timeouts etc. so you are essentially providing an asynchronous API which then calls into potentially blocking IO operations. That seems quite surprising for a user of your API since the expectation would be that the async operations are of non-blocking nature. On the plus side the async APIs enable cancellation token support.

jeremydmiller:
That royally sucks. I'd like to push this one off a little bit (just until next week) if you don't mind. I'm going to try to move some stuff that's in the generated code today into some generic code before we do this. There's a lot more code that touches data readers we'll have to also touch. With even more on the way today.

src/Marten/Events/EventMapping.cs
jeremydmiller: It's not on purpose, just some sloppy inconsistency;-)

Event storage is its own thing that I wrote very early, and I just barely finished pulling it closer to how the documents get persisted.

jeremydmiller:
@danielmarbach You're gonna have to get it in the DocumentStorageBuilder too where it generates code. I'm afraid you're going to have to run the whole test suite.

jeremydmiller:
@danielmarbach Shoot, and we'll have to have an async path for `ISelector<T>` as well.

danielmarbach:
Ok I'll give it a go and touch all points and get the test running

danielmarbach:
@jeremydmiller how about this? Sorry I'm still learning about your architecture.

jeremydmiller:
@danielmarbach I'm *finally* going to pick this one up today, but I'll probably re-apply your changes here locally. As part of the work for #280, I'm going to introduce new patterns to read DbDataReader's asynchronously field by field and correct that here.

danielmarbach:
Ping me when I can help

jeremydmiller:
All of these changes are in master, thanks @danielmarbach!

Issue #35, opened 2015-10-21
Optimistic locking with version tracking on documents


jeremydmiller:

jeremydmiller:
Either by an auto-incrementing version number in the db or by some kind of etag

mdissel:
Using Etag is complexer, but when implemented as in RavenDb (global incrementing Etags) could help in other area's of your application like (http) caching, syncing, etc.

kiliman:
You can also use PostreSQL's txid support. txid_current() [1] returns a 64-bit integer that increments with each transaction (so guaranteed not to roll over).

Note: I think you'll want to bring in this value into the identity map, so it can be exposed to the client. Also, it will make a quick optimization. If a user tries to store a document without a matching txid/etag, then you can throw an exception right there without even hitting the database.

Also, it would help if you can query against this value. I use Raven's Etags for offline sync support. I can then query for all documents where Etag > last etag synced.

[1] http://www.postgresql.org/docs/9.1/static/functions-info.html#FUNCTIONS-TXID-SNAPSHOT

jeremydmiller:
My thought right now is to start this by using the sequential guid's that @tim-cools built into Marten. We already have the version column, just nothing that does the concurrency check. That should satisfy when @kiliman is talking about for finding newer things.

For the concurrency exception, I want to batch up all possible errors into one exception, so no throwing exceptions in the upsert functions themselves.

We'd need at minimum a new method like this:

```
IDocumentSession.Store<T>(T doc, Guid etag) 
```

Otherwise, my thought is that if the concurrency checking is turned on for a document, that you track the current version whenever you load that document and keep it on the IdentityMap somehow (and yes, we'll make this work w/ NulloIdentityMap too). When you issue the upsert, it would be smart enough to know to use the existing version + a new one to make the updates.

jvdvleuten:
For the Event store, it would look something like this?

`void Append(Guid stream, int exptectedCurrentVersion, params object[] events);`

Right now, I am using `StartStream` to both start *and* append an aggregate with events... Maybe cheating somewhat there ;)

jeremydmiller:
"Right now, I am using StartStream to both start and append an aggregate with events..." -- we've talked a little bit about eliminating the distinction. It's a minor performance optimization for new streams because you'd know where the version was and know that aggregates built in line can be fetched by just new'ing up one.

For the event store, we could do that. Might not be that bad to pull off.

PR #220, opened 2016-03-15
Conditional store spike.

jokokko:


CoreyKaylor:
Thinking this could help with https://github.com/JasperFx/marten/issues/35

Issue #219, opened 2016-03-15
IQueryable extensions to retrieve the raw JSON 


jeremydmiller:
In many cases in web apps, you wouldn't want to bother deserializing the Json just to immediately serialize it back out to Json in the HTTP response. What if you could just skip that and get the raw Json? Even if there's a Select() in there.

- [ ] FirstJson() : string
- [ ] FirstOrDefaultJson() : string
- [ ] SingleJson() : string
- [ ] SingleOrDefaultJson() : string
- [ ] ToListJson() : string // Think this would wrap it in "[ ]"

- [ ] WriteJsonResults(Stream) // maybe a more efficient way to do this

and async of all of the above.

phillip-haydon:
I love the idea of `ToListJson()` but my request is that at bare minimum if the result is empty, that the list returns `[]` to simulate an empty array, because `.ToList()` always creates atleast an empty list. It shouldn't be empty or null.

jeremydmiller:
@phillip-haydon I definitely agree.

nieve:
I wouldn't mind giving this a go- unless you've already got this.

jeremydmiller:
Go @nieve! My next Marten trick is the Includes() feature.

nieve:
So, following from Gitter- should I focus on FirstOrDefault and ToList and then we'll see about the Single/OrDefault/First?

jeremydmiller:
I think that's fair. It should be rinse and repeat after those first tow. But a lot of rinse & repeat:(

nieve:
I've got these left:
SingleJson()
SingleJsonAsync()
SingleOrDefaultJson()
SingleOrDefaultJsonAsync()
WriteJsonResults(Stream)
WriteJsonResultsAsync(Stream)

jeremydmiller:
Gonna call this one closed via https://github.com/JasperFx/marten/pull/229

PR #229, opened 2016-03-18
added support for SingleJson & SingleOrDefaultJson (a/sync)

nieve:
continues #219


ekulakov:
@nieve better to avoid `async void` methods. [Details](http://haacked.com/archive/2014/11/11/async-void-methods/)

nieve:
Well spotted @ekulakov ! Thanks a lot, will fix it :)

Issue #152, opened 2016-02-12
Add EXPLAIN preview to the diagnostics


jeremydmiller:
See http://www.postgresql.org/docs/9.2/static/sql-explain.html. I don't think it's going to be too crazy to fetch the EXPLAIN plan behind a Linq query.

Might make an extension method off of IQueryable instead. This might get a little awkward when we introduce Include()

nieve:
I can give that a go, if no one is looking into it (at least for non include queries). Should we add the optional parameters as well?

jeremydmiller:
I don't think the code would be much different for the Includes. Just make sure we centralize the code for creating a DocumentQuery inside of the MartenQueryable, and you'd get the NpgsqlCommand object.

I'm drawing a blank, what do you mean about optional parameters?

PR #234, opened 2016-03-23
Added support for Explain queries

nieve:
targeting #152 
I just noticed the issue talks about adding the explain to the **diagnostics** - I've got a feeling I missed the point.
This executes explain (and returns a QueryPlan object with all the data) on non-aggregates, non-scalar queries- do we need explain on count/single/first/sum/min/max etc'..?

Let me know if it's not exactly what you had in mind.


nieve:
Yeah, this cannot be what was meant by the issue 152. Am looking into it.

PR #235, opened 2016-03-24
Added support for Explain on MartenQueryable 

nieve:
Ok, so I do believe this is what you had in mind for making 

> an extension method off of IQueryable

Not sure how explain would fit the Diagnostics, would discuss on gitter.

### DIFF
--- a/src/Marten.Testing/Linq/explain_query.cs
+++ b/src/Marten.Testing/Linq/explain_query.cs
@@ -1,0 +1,62 @@
+﻿using System;
+using System.Linq;
+using Marten.Linq;
+using Marten.Services;
+
+namespace Marten.Testing.Linq
+{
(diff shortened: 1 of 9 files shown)


jeremydmiller:
@nieve Shoot, would you mind taking a look at this and maybe doing a rebase? Some other changes came in that are impacting this.

I personally like having the extension method to show the explain plan. I'm not entirely sure how you'd do it for aggregate functions w/o having ExplainPlanForSum(), etc. overloads.

I think the extension method will be easier to use than the `Diagnostics` property off of DocumentStore.

nieve:
yeah, no worries, I'll look into rebasing this work.
As for aggregates, I'll play around with it, see what I can come up with. If I come up with anything worthwhile, I'll open a new issue.

PR #236, opened 2016-03-24
Make database schema configurable

tim-cools:
I implemented the support of database schema's. 

It is now possible to configure the default schema used by the document store:

`options.DatabaseSchemaName = "other";`

It is also possible to override the database schema per document:

`options.MappingFor(typeof(User)).DatabaseSchemaName = "yet_another";`
           
And it is possible to put the eventstore db objects in a different schema:

`options.Events.DatabaseSchemaName = "event_store";`

It also ensures that a database schema is created if it doesn't exist yet. Any feedback welcome!


jeremydmiller:
@tim-cools Wow, that looks like a lot of work. I might be a little slow getting this in, gimme until Monday night at least.

tim-cools:
@jeremydmiller sure no problem. I know it is pita to review :-) 
I hyink I forgot the configuration from MartenRepository. I will look into this later...

jeremydmiller:
@tim-cools Hey, so there's no surprises later, I'm reversing some of the small stylistic things here. I'm putting `DocumentMapping.For<T>()` back. I'm not wild about having the static factory classes. Mostly me being Linus here though;). 

I'm working through this now. Might be later today before I can push though.

jeremydmiller:
Pulled in w/ a rebase first.

tim-cools:
@jeremydmiller no problem, you are the owner :-)

I removed the DocumentMapping.For<T> because it is only used in Test code and I don't think consumers of the Marten API needs to call it directly. I would even argue that they should not be able to call it at all....

I noticed you are quit flexible about styling etc which I understand and many people probably appreciate. But it might be useful to document the preferred style briefly in a GitHub wiki page, even if it  only used as a guideline. I'm very pragmatic about the used style, but I like it when the same style is used in the same project/code-base. Just a suggestion :)

Issue #238, opened 2016-03-25
The containment operators in Linq query generation aren't using the table alias


jeremydmiller:
See this conversation from Gitter:

```
Looks like includes are failing with a .Where() clause, it references data rather than d.data by the looks of it.
data is ambigious
select d.data, d.id,
account_id.data,
account_id.id
from mt_doc_audio as d
INNER JOIN mt_doc_account as account_id ON d.data ->> 'AccountId' = account_id.id
where (data @> '{\"Tags\":[\"Tech House Competition 2015\"]}')
and (CAST(d.data ->> 'Status' as integer) = 1)
and (CAST(d.data ->> 'Visible' as boolean) = True)
order by CAST(d.data ->> 'DateAdded' as timestamp with time zone)
desc LIMIT 24;
```

Easy fix, but add tests for it too.

nieve:
Am on it.

jeremydmiller:
@nieve fixed this w/ https://github.com/JasperFx/marten/pull/239.

PR #239, opened 2016-03-25
fixed include with a where containment (normal or any) clause

nieve:
fixes issue #238


jeremydmiller:
I pulled it in w/ a rebase just now.

PR #242, opened 2016-03-28
added failing test for OrderBy when using .Include

phillip-haydon:
When using `.OrderBy(` and `.OrderByDescending(`, both will fail with 

> NpgsqlException: 42702: ORDER BY "id" is ambiguous


jeremydmiller:
Got it in and fixed w/ a rebase

PR #259, opened 2016-04-03
Custom id generation and comb guid

tim-cools:
Added support for CombGuid. From first tests with bulk insert has CombGuid the same perf as Int32 id's now. The difference this is not spectacular and only a couple % faster as the default Guid. 
I also implemented a IdentityKeyGenerator based on the current HiLo.
I will do more perf tests and post a report of the read/write perf differences when using the different ID types later this week.


jeremydmiller:
In with a rebase this morning.

PR #262, opened 2016-04-04
Projection Spike

jeremydmiller:
This is here strictly to gather feedback and provide a discussion forum

### DIFF
--- a/src/Marten/Marten.csproj
+++ b/src/Marten/Marten.csproj
@@ -88,6 +88,7 @@
     <Compile Include="Linq\TargetObject.cs" />
     <Compile Include="Linq\WhereClauseVisitor.cs" />
     <Compile Include="Linq\WholeDocumentSelector.cs" />
+    <Compile Include="Projections\ProjectionClasses.cs" />
     <Compile Include="QuerySession.cs" />
     <Compile Include="Schema\ColumnValue.cs" />
     <Compile Include="Schema\CombGuidIdGeneration.cs" />
--- a/src/Marten/Projections/ProjectionClasses.cs
+++ b/src/Marten/Projections/ProjectionClasses.cs
@@ -1,0 +1,171 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Baseline;
+using Marten.Events;
+
+namespace Marten.Projections
+{
+
+    // May be just a marker
+    public interface ITransform
+    {
+        Type FromType { get; }
+        Type ToType { get; }
+    }
+
+    // For mapping a single event to a different representation
+    public interface ITransform<TFrom, TTo> : ITransform
+    {
+        TTo Transform(TFrom from);
+    }
+
+    public interface IUpdater<TAggregate>
+    {
+        Type EventType { get; }
+    }
+
+    // TODO -- support partial updates later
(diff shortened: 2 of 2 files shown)


jeremydmiller:
Hey folks, I’m working a bit on the projections support today to try to get the event store half of things moving. For the first wave of things, I’m thinking that this is the functionality:
* Start with .Net projections first, then go onto the equivalent functionality w/ PLV8 in the database
* Support projections calculated either “Inline” during event capture, “Async” by some kind of async daemon that you would need to leave running in your application, or “Live” that’s can be calculated on the fly at any time from the latest data.
* Leave partial document updates for a later phase
* Support a single event to a single document projection (per Event)
* Support an aggregation over all the events in a Stream. This would later get reused as the mechanism to create aggregate snapshots

For aggregates, we can do things in a couple different ways. We can:
1. Allow you to define and plug in a bunch of Func<TAggregate, TEvent, TAggregate> functions that would transform an aggregate model
1. Allow you to define a concrete type with Apply(Event1), Apply(Event2) methods that mutate itself based on event types. My thought here is to generate the Func’s from the preceeding item behind the scenes for this approach. I was iffy on this the first time I saw this pattern, but it's nice to have everything segregated.
1. We’ll have a high level aggregator interface you could use to do something completely different.
1. Dunno how this will go over, but we could also have an approach to say `IEvent<TAggregate>`, where that interface has a `void Apply(TAggregate)` function on it so that the event class itself would "know" how to transform the aggregate.

jeffdoolittle:
I'm not thrilled with (4) - events should be pocos with no behavior.

jeremydmiller:
@jeffdoolittle I did that on a previous system. Makes the codebase very easy to extend because you don't need any kind of duplicated hierarchy of "one class for this, one other matching class for that."

jeffdoolittle:
of course I'm open to learning new things. :) but when I think of an event, I think of something you might be publishing to another service where having behavior on the event class would be odd/weird.

CoreyKaylor:
Yeah, not fond of 4 either. Might accuse you of staring at NServiceBus code for too long. :smile:

tim-cools:
- I prefer a dedicated class per Projection which and the output a separated POCO with several IHandleEvent<X>. There is no need to encapsulate the data here so I prefer to have the responsibilities separated. 

- But I also used the self mutating projection on a project which also worked out fine. Maybe we can support both. In the end this is the same as the first option but only with `return this;` as output.

- Is some situation it might be useful to have a projection that subscribes to all the events.

- And I also find it useful to be able to restart a projection from scratch a. This will probably be difficult with the sycnprojections .

- For event sourced aggregates I currently use something with AggregateEvent objects. I like that the publish and subscribe of an event are combined in a single object. The wiring is a bit more complicated but there is no need for a special base class of to inject a service to publish the events. (Gist: https://gist.github.com/tim-cools/694ba34dbb93b7a0cdf710886aa29275)

- 4 might work if you have one projection per event, but I guess it becomes messy when you have multiple projections per event. I prefer to group the handler per projection type instead of per event type.

jeremydmiller:
To make the point clear that I didn't before, none of those options are mutually exclusive and everything would boil down to `IAggregator.Apply(TAggregate, TEvent)` behind the scenes.

I get the theoretical objections to "OMG, putting code into an Event class", but the reality is that that creates a very easily extensible mechanism for modeling state changes with fewer moving parts.

jeffdoolittle:
:scream: methods on events! ;)

jeremydmiller:
Closing this as the work is already under way and usable.

PR #266, opened 2016-04-06
Subclass inheritance

### LINKED ISSUE #233: Inheritance Problems
I'm in learning stage so I'd setup a test case where:
* IResource - Interface for a resource
 * Id : Guid
* Resource - Base type implementing the IResource interface
* ZoneResource - Base type for zone resources
  * Zone : int
  * Kiln
* Kiln - Derives from Resource
* CFan - Derives from ZoneResource

I've started out doing a interface for the ZoneResource since both ZoneResource and Resource are IResource (but that did not work at all).

Now with "flatternd" inheritance / implementation chain I explicitly setup subclassing (tried to just add IResource when all resources where implementing that but it just created a table for each resource):

      For<Resource>()
        .AddSubclass<Ce>()
        .AddSubclass<DigitalInputPort>()
        .AddSubclass<Kiln>()
        .AddSubclass<Sawmill>();

      For<ZoneResource>().AddSubclass<CFan>();

I add a set of resources and do a initial query - session.Query<CFan>().Where(x => x.Name = "CFAN1").ToList(); and it works perfectly.

In another class I do:

      var kiln = Kilns().First();
      var cf = _session.Query<CFan>().Where(c => c.Kiln == kiln.Id);
      if (_config.KilnZone != -1)
        return cf.Where(c => c.Zone == _config.KilnZone).ToList();

      return cf.ToList();

I see in the log that it creates a new table 'mt_doc_cfan' and queries against that instead of table 'mt_doc_zoneresource' as earlier.

What I'm I doing wrong or is this a bug?

Is the intention that it shall be possible to do a For<IResource> and then it shall use all subtypes even if the inheritance chain is 3 or more on the same table?

Cheers,
 Mario

nieve:
fixes #233 
I didn't remove the AddSubclass methods as this PR started getting too big. Can do it in another PR.
The last commit added the ability to add the whole hierarchy + alias per type in one method call. It felt though a bit clunky, so if this needs fixing, just let me know.
Otherwise if it's all good, I can add docs on this.


nieve:
Just removed the clunky code I referred to yesterday with a nice refactoring.
I think it's good to go. Will add another PR to remove the non/generic AddSubclass methods if that's still on the table.

jeremydmiller:
@nieve,

Hey, what do you think about just making the `IntegratedFixture` class force the Thread.Culture to "en_US" across the board? So you don't really have to worry about that in your tests? And yes, I know that makes me a provincial, "ugly American" ;)

And now I'm going to have the Smurfs theme song stuck in my head all day. I'm reviewing this while I let the biggest, craziest build script you've ever seen set up what I'm supposed to be doing right now...

src/Marten.Testing/Linq/query_with_inheritance.cs
```diff
@@ -0,0 +1,133 @@
...
+    public class BrainySmurf : PapaSmurf{}
+
+    public class query_with_inheritance_and_aliases : DocumentSessionFixture<NulloIdentityMap>
+    {
+        public query_with_inheritance_and_aliases()
+        {
+            StoreOptions(_ =>
+            {
+                _.Schema.For<ISmurf>()
+                    .AddSubclassHierarchy(
```
jeremydmiller: I like this syntax. Someday we might wanna add some way, or just examples, to do this with type scanning or some other sort of dynamic Type discovery.

nieve:
Awesome- that makes a lot more sense. Will sort it out later on today.
Also have a think about the other AddSubclass methods- they enforce a different behaviour since they don't support what this work fixes! Either we rename them to reflect that or they go?

src/Marten.Testing/Linq/query_with_inheritance.cs
nieve: Only thing is it'll have to be done upwards the hierarchy tree I suppose. It'll be dead easy.

nieve: What I said was nonsense.
You mean giving an example like AddSubclassHierarchy(assembly.GetTypes().Where(typeIsSubclass)) ? Would be better doing this in the docs, where we could add this should be avoided in big projects.

PR #269, opened 2016-04-09
added ability to dynamically add subclass hierarchy & docs

nieve:
I started this work just doing the doc side of things, then wanted to show an example of how this could be done by using dynamic type discovery as you suggested [here](https://github.com/JasperFx/marten/pull/266#discussion_r58878614). 
This turned out a bit messy and I decided to add this to the framework while adding intelliSense comments that advise users to beware of using this feature willy-nilly. The sample in the docs also gives out this warning.
If you're not happy with this feature, I can easily remove it; the docs should be in there though I believe.


jeremydmiller:
Got it w/ a rebase.

Issue #270, opened 2016-04-10
Remove IEvent and IAggregate interface to prevent tight coupling


jvdvleuten:
Remove IEvent and IAggregate interface to prevent tight coupling. Just like we don't have an IDocument interface for the DocumentStorage itself, we should not have an IEvent for Events.

Probably these things have to be done:

- Remove the IEvent and IAggregate interfaces
- Streams should be identified by their Id's (and could be of any class type?)
- Encapsulate the Event data in an new internal Event object, which holds and generates the Id of the Event.
- Store the body of the Event just like we do it now, with the correct event_type
- The encapsulation of the Event object could in theory also hold other information (e.g. headers) which can be persisted as other columns in the row. (CoreyKaylor)

jeremydmiller:
@jvdvleuten This commit: https://github.com/JasperFx/marten/commit/4a740704da39f866ab2a1d1c7d9d8c37b7e3d778, should move the Id member validation out of your way so that `EventMapping` can cleanly inherit from `DocumentMapping`. There's probably some cleanup later on that hierarchy after the next big release.

jvdvleuten:
I removed IAggregate in this commit: https://github.com/jvdvleuten/marten/commit/cdd8b12b15760a96dfdc266fcc94ba291ad8e611

Going onward with IEvent

jvdvleuten:
Started to remove IEvent in this commit: https://github.com/jvdvleuten/marten/commit/2420f7e60373e5910736cee38687f5caed857f9c

Saving an Event works, now moving on to change `fetchStream(reader)` to correctly reconstruct the Event object with correct Body and Body Type

jvdvleuten:
From Jeremy:
> Okay, so my thoughts after talking over this w/ @CoreyKaylor this morning:
> Let’s scrap the idea of making event types be registered upfront
> Let’s say that event querying against the raw data only happens through session.Events.Query<TEvent>(), which enables us to sidestep the “is it an event or not” problem and tip off the user that “hey, this is a different operation than normal document queries that is potentially much more expensive”
> Let’s keep the EventStream fetching exactly the same for now, but come back with a follow up story to return an EventStream object w/ a collection of Event objects that wrap the raw event data.
> Also follow up w/ a later story on session.Events.Load(Guid) : Event maybe
[...]

PR #287, opened 2016-04-13
Remove IEvent and IAggregate interface to prevent tight coupling

jvdvleuten:
Issue #270


jeremydmiller:
It's in w/ a rebase and some little fixes on top of it.

Issue #291, opened 2016-04-14
Fetch stream should return Event wrapper objects


jeremydmiller:
So you get the version, the id, and potentially header information

jeremydmiller:
Closed by https://github.com/JasperFx/marten/pull/304

PR #304, opened 2016-04-16
Fetch stream should return Event wrapper objects

jvdvleuten:
#291 

Renamed Body to Data to better reflect mt_events column name.
Also fixed storyteller tests.


jeremydmiller:
Got it w/ a rebase.

Issue #302, opened 2016-04-16
Be able to fetch raw JSON for Select() transformations


jeremydmiller:
Follow on to #301

PR #323, opened 2016-04-23
Added support for Asjson / ToJsonArray compiled queries

### LINKED ISSUE #312: Support for Json() in compiled query
This fails:
``` csharp
    public class FindJsonUserByUsername : ICompiledQuery<User, string>
    {
        public string Username { get; set; }
        public Expression<Func<IQueryable<User>, string>> QueryIs()
        {
            return query =>
                    query.Where(x => Username == x.UserName)
                        .Json().Single();

        }
    }

        [Fact]
        public void json()
        {
            var user = theSession.Query(new FindJsonUserByUsername() { Username = "jdm" });
            user.ShouldNotBeNull();
            user.ShouldBe("jdm");
        }
```

nieve:
fixes #312 
Pleas code review thoroughly; it works, but if feels quite convoluted.

### DIFF
--- a/src/Marten/Linq/QueryHandlers/JsonQueryHandler.cs
+++ b/src/Marten/Linq/QueryHandlers/JsonQueryHandler.cs
@@ -1,0 +1,63 @@
+using System;
+using System.Data.Common;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Baseline;
+using Marten.Schema;
+using Marten.Services;
+using Marten.Util;
+using Npgsql;
+using Remotion.Linq;
+
+namespace Marten.Linq.QueryHandlers
+{
+    public class JsonQueryHandler : IQueryHandler<string>
+    {
+        private readonly IDocumentSchema _schema;
+        private readonly QueryModel _query;
+        private readonly ISelector<string> _selector;
+        private readonly IDocumentMapping _mapping;
+
+        public JsonQueryHandler(IDocumentSchema schema, QueryModel query)
+        {
+            _mapping = schema.MappingFor(query);
+            _schema = schema;
+            _query = query;
+
+            var selector = _schema.BuildSelector<string>(_mapping, _query);
+
+            _selector = selector;
+        }
+
+
+        public Type SourceType => _query.SourceType();
+        public void ConfigureCommand(NpgsqlCommand command)
+        {
+            var sql = _selector.ToSelectClause(_mapping);
+            var @where = _schema.BuildWhereFragment(_mapping, _query);
+
+
+
+            sql = sql.AppendWhere(@where, command);
+
+            var orderBy = _query.ToOrderClause(_mapping);
+            if (orderBy.IsNotEmpty()) sql += orderBy;
+
+            sql = _query.AppendLimit(sql);
+            sql = _query.AppendOffset(sql);
+
+            command.AppendQuery(sql);
+        }
+
+        public string Handle(DbDataReader reader, IIdentityMap map)
+        {
+            return $"[{_selector.Read(reader, map).ToArray().Join(",")}]";
+        }
+
+        public async Task<string> HandleAsync(DbDataReader reader, IIdentityMap map, CancellationToken token)
+        {
+            return await _selector.ReadAsync(reader, map, token).ContinueWith(t=> $"[{t.Result.Join(",")}]", token).ConfigureAwait(false);
+        }
+    }
+}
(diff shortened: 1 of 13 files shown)


src/Marten/Linq/QueryHandlers/QueryHandlerFactory.cs
```diff
@@ -157,7 +162,7 @@ private IQueryHandler<T> tryFindSingleQuery<T>(QueryModel model, IIncludeJoin[]
 
             var setters = findSetters(queryType, expression);
 
-            var model = MartenQueryParser.Flyweight.GetParsedQuery(invocation);
+            var model = MartenQueryParser.ForCachedQuery.GetParsedQuery(invocation);
```
jeremydmiller: That static public field is accessed in other places than just for the cached queries, and isn't really specific to cached queries anyway. Could you rename it back to what it was, or something that makes more sense to you if you want? It is just a GoF "flyweight", but feel free to call it "Instance" or something if you prefer

src/Marten/Linq/QueryHandlers/JsonQueryHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+            _query = query;
+
+            var selector = _schema.BuildSelector<string>(_mapping, _query);
+
+            _selector = selector;
+        }
+
+
+        public Type SourceType => _query.SourceType();
+        public void ConfigureCommand(NpgsqlCommand command)
```
jeremydmiller: @nieve You *might* wanna think about having this class use an inner ListQueryHandler<string> that configures the command and fetches the raw data.

jeremydmiller:
@nieve I made some minor comments on the PR code, but I think this looks fine to me. I didn't know about the customizations to Relinq. I think this is going to make #302 easier.

src/Marten/Linq/QueryHandlers/QueryHandlerFactory.cs
nieve: I had to add another instance- for a compiled query. Flyweight is still there: https://github.com/JasperFx/marten/pull/323/files#diff-11bbc3e3b8d500d2e3747762350d973dL11

src/Marten/Linq/QueryHandlers/JsonQueryHandler.cs
nieve: I tried to use ListQueryHandler but couldn't since I was returning a `string` rather than a `List<string>` which made me have to do that.

nieve:
Have a look at the Flyweight bit- this is the worst bit of my code. It seemed to me I had to do this bit :worried: 

And yes, it will make #302 easier.

src/Marten/Linq/QueryHandlers/QueryHandlerFactory.cs
nieve: renamed.

src/Marten/Linq/QueryHandlers/JsonQueryHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        public string Handle(DbDataReader reader, IIdentityMap map)
+        {
+            return $"[{_selector.Read(reader, map).ToArray().Join(",")}]";
+        }
+
+        public async Task<string> HandleAsync(DbDataReader reader, IIdentityMap map, CancellationToken token)
+        {
+            return await _selector.ReadAsync(reader, map, token).ContinueWith(t=> $"[{t.Result.Join(",")}]", token).ConfigureAwait(false);
```
ekulakov: IMO better not to mix async\await and Task. `ContinueWith` is redundant
```
var results = await _selector.ReadAsync(reader, map, token).ConfigureAwait(false);
return $"[{results .Join(",")}]";
```

src/Marten/Linq/QueryHandlers/JsonQueryHandler.cs
jeremydmiller: Yeah, but you could use an inner ListQueryHandler to build the SQL and even fetch the raw data w/ a JsonSelector, then just delegate to that in your Handle() methods. Honestly though, it's not worth the effort to change IMO. If we see any more duplication of the build SQL though, we'll come up with some more formal sql builder

jeremydmiller:
@nieve I say to just deal w/ what @ekulakov is saying about the async/Task, then let's get this puppy in. Any chance you'd be up for updating the docs w/ a few more samples to show this off? It's a killer feature IMO.

nieve:
Yeah, I'll fix that and will then send another PR for the docs.

nieve:
Done :smile:

PR #340, opened 2016-04-29
fix 339 async cache issue

### LINKED ISSUE #339: Serialization fails on async cache hits for a DirtyTrackedSession identity map
The following test will fail with an exception

```chsarp
        [Fact]
        public async Task pending_with_dirty_checks_async()
        {
            var user1 = new User();

            using (var session1 = theStore.LightweightSession())
            {
                session1.Store(user1);
                session1.SaveChanges();
            }

            using (var session2 = theStore.DirtyTrackedSession())
            {
                var user12 = await session2.LoadAsync<User>(user1.Id);
                var breakThings = await  session2.LoadAsync<User>(user1.Id);
            }
        }
```

The exception is as follows:

Test 'Marten.Testing.UnitOfWork_PendingChanges_Functionality_Tests.pending_with_dirty_checks_async' failed: System.AggregateException : One or more errors occurred.
---- System.InvalidCastException : Unable to cast object of type 'Marten.Services.TrackedEntity' to type 'Marten.Testing.Documents.User'.

This appears to occur when there is a cache hit on the GetAsync method of IdentityMap.cs. The following code fails at ``return dict[id].As<T>();``

```csharp
        public async Task<T> GetAsync<T>(object id, Func<CancellationToken, Task<FetchResult<T>>> result, CancellationToken token = default(CancellationToken)) where T : class
        {
            var dict = Cache[typeof(T)];

            if (dict.ContainsKey(id))
            {
                return dict[id].As<T>();
            }

            var fetchResult = await result(token).ConfigureAwait(false);
            if (fetchResult == null) return null;

            var document = fetchResult.Document;

            dict[id] = ToCache(id, typeof(T), document, fetchResult.Json);

            _listeners.Each(listener => listener.DocumentLoaded(id, document));

            return document;
        }
```

This was found using basically the current code of Marten - I synced about two hours ago. I can also replicate this using the latest NuGet package with a Raven application that I'm converting over as a POC. The following code from that application will also fail with the same exception:

```csharp
        public async Task<string> TestMarten()
        {
            using (var session = MvcApplication.MartenStore.DirtyTrackedSession())
            {
                var userInfo = await session.LoadAsync<UserInfo>("76bb4089-24aa-e311-8f56-d4ae529037b1/zzmarten2");

                var userInfo2 = await session.LoadAsync<UserInfo>("76bb4089-24aa-e311-8f56-d4ae529037b1/zzmarten2");

                return "pass";

            }

            return "test";
        }
```

I'm trying to fix this for you and make a pull request. I haven't quite deconstructed everything that is going on in my head.

Wanted to document the issue for the sake of completeness.

tim-cools:
fixes #339

### DIFF
--- a/src/Marten.Testing/Bugs/Bug_339_async_cache_problem.cs
+++ b/src/Marten.Testing/Bugs/Bug_339_async_cache_problem.cs
@@ -1,0 +1,28 @@
+﻿using System.Threading.Tasks;
+using Marten.Services;
+using Marten.Testing.Documents;
+using Xunit;
+
+namespace Marten.Testing.Bugs
+{
+    public class Bug_339_async_cache_problem : DocumentSessionFixture<NulloIdentityMap>
+    {
+        [Fact]
+        public async Task pending_with_dirty_checks_async()
+        {
+            var user1 = new User();
+
+            using (var session1 = theStore.LightweightSession())
+            {
+                session1.Store(user1);
+                session1.SaveChanges();
+            }
+
+            using (var session2 = theStore.DirtyTrackedSession())
+            {
+                var user12 = await session2.LoadAsync<User>(user1.Id);
+                var breakThings = await session2.LoadAsync<User>(user1.Id);
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@tim-cools I think a file got missed by .gitignore maybe. "Bugs\Bug_339_async_cache_problem.cs" isn't in the commit.

tim-cools:
@jeremydmiller done. got to love those PR builds...

jeremydmiller:
I'm pulling it in now.

PR #342, opened 2016-04-29
Create schema on initialization

tim-cools:
I used the new AllSchemaNames method to create the database schema when the store is initialized. I think it is cleaner this way instead of checking this for each document. I also updated the generation of the ddl files(s).

I was temped to extract the whole export in a DatabaseObjectExporter class, but I don't want to pull off bigger refactorings without feedback first...


jeremydmiller:
@tim-cools I actually do like the idea of pulling exporting out of DocumentSchema, or at least by hanging it off the schema class maybe. DocumentSchema is too big as it is, even after I pulled some stuff out into a new DbObjects class.

PR #345, opened 2016-05-03
Removed unnecessary awaits and added `ConfigureAwait(false)`

danielmarbach:
@jeremydmiller here we go ;)


jeremydmiller:
The point here is to eliminate unnecessary allocations, right? I'm pulling this one in now.

danielmarbach:
Yes. It avoids the statemachine generation and therefore allocations, context capturing...

PR #346, opened 2016-05-04
Asynchronous path for document cleaning

danielmarbach:
For symmetry reasons introduced the async methods on the document cleaner API as well.

### DIFF
--- a/src/Marten.Testing/Schema/DocumentCleanerTests.cs
+++ b/src/Marten.Testing/Schema/DocumentCleanerTests.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using Baseline;
 using Marten.Events;
 using Marten.Schema;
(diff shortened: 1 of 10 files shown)


danielmarbach:
@jeremydmiller This is done from my point of view

jeremydmiller:
@danielmarbach I should have shut this one down earlier. I think this is just unnecessary complexity and clutter in the codebase. I'd rather we didn't take this in. Document cleaner is strictly for testing.

danielmarbach:
Well then the API belongs into a different package. But even then, for example in our acceptance testing solutions we are purely async. The complexity argument: Well that is driven by the design decision to support sync and async overloads. So you either say: production lib is sync and async and then all APIs follow that guideline, which means the cleaner has to follow that rule as well OR you move it into separate package and come up with the rule where you say: testing code will not be async but sync only because of reasons.

CoreyKaylor:
Now that the API churn has settled down a bit, rebase off the latest master if you're still interested in getting this in. Sorry if this causes some inconvenience.

Issue #309, opened 2016-04-18
Add Include() support to compiled queries


jeremydmiller:
Just what it says, make sure you can include related documents as part of a compiled query. My initial thought is to make an additional interface that you would use to express includes like:

```
public interface ICompiledQueryWithIncludes<TDoc, TResult> : ICompiledQuery<TDoc, TResult>
{
    // this will be a cut down version of IMartenQueryable just to get Include() methods
    void Include(IMartenQueryable<TDoc> queryable);
}
```

I think that we would calculate the includes upfront to build out the SQL query, then on each execution, we would execute the Include again, build out the `IIncludeJoin` objects that point to the current object and use that to actually execute the query handler.

nieve:
Going to give this a go, unless someone else is on  it.

jeremydmiller:
It'd be very helpful if somebody besides me were familiar with the Include() internals. So, go @nieve !

jeremydmiller:
Fixed by https://github.com/JasperFx/marten/pull/349

PR #349, opened 2016-05-08
Added support for compiled query with include/s

nieve:
currently support one or multiple includes. For #309


jeremydmiller:
@nieve I think you need to be careful, because you're quickly becoming the goto Linq guru for Marten;) I'm pulling this in locally now.

PR #366, opened 2016-05-16
Bug 365 compiled query with constant fails

tim-cools:
+ fix 2 two inconsistent filenames
+ remove unused SetParameter(object query, NpgsqlParameter parameter)

ps: still 2 regression tests failing: Lifecycle: Regression, 21 successful, 2 failed
ps2: the faster tests are great!


jeremydmiller:
Got it, thank you Tim.

Issue #272, opened 2016-04-11
Build a Sample App in the codebase


jeremydmiller:
Do it by stealing a sample app from EF somewhere and converting that? Or RavenDb. 

I've never been good at this kind of thing, and I'd actually prefer this be done by folks that aren't necessarily deep into the Marten internals so we can use this exercise to flush out usability problems or things that are confusing.

jeremydmiller:
The current thinking is to use https://github.com/jchannon/DinnerParty as the basis. Probably port it to ASP.net MVC (boo!) at some point for more reach into the community.

CodingGorilla:
I'm working on this issue, while trying to keep up with the on-going development; @jeremydmiller is your vision that this sample project be included in the main solution (Marten.sln), and builds along with everything else?

jeremydmiller:
@CodingGorilla At least at first, I think we should keep it in the same repo for exactly the reason you said. I've been beaten up for over a decade for not having enough samples in my OSS projects and it'd be nice to not have that issue so much with Marten;) Thanks for doing this.

jeremydmiller:
Closed by https://github.com/JasperFx/marten/pull/371. Thanks @CodingGorilla !

PR #371, opened 2016-05-18
Brings in a port the DinnerParty project as a sample application for marten

CodingGorilla:
Addresses #272, would still like to port from NancyFX to ASP.NET MVC and eventually maybe even ASP.NET MVC Core.


jeremydmiller:
Awesome! I'll get this in one way or another today.

PR #373, opened 2016-05-20
Support of private ID property setters

tim-cools:
JSON.NET support serialization of properties with private setters. So I believe we should do so tou for ID's 'to support correct encapsulation.

I added support for auto assignment of properties ID with a private ID.

I know someone introduced the NoOpIdGeneration for Id with private Id's, but I can't find who and when. The NoOpIdGeneration is now only used in case of a read-only Id properties, I hope this will do for the dev who added this NoOpIdGeneration?


jeremydmiller:
Got it w/ a rebase. Will be in 0.9.4 very soon.

PR #374, opened 2016-05-20
support serialization of immutable events

tim-cools:
I like that my events are immutable like this:

```
+    public class ImmutableEvent
+    {
+        public Guid Id { get; private set; }
+        public string Name { get; private set; }
+
+        public ImmutableEvent(Guid id, string name)
+        {
+            Id = id;
+            Name = name;
+        }
+    }
```
This was not supported due to a generic constraint.  So I removed the constraint and added a test.


jvdvleuten:
I had the same problem yesterday +1

jeremydmiller:
Got it w/ a rebase, thanks @tim-cools!

PR #408, opened 2016-06-03
Support for configuration on .Index(...)

phillip-haydon:
- [x] Can mark index as Unique
- [x] Can mark index as Concurrent
- [x] Support for Where Clauses
- [x] Support for naming the index
- [x] Support for casing of the value

This is my first cut at supporting some arguments on the `.Index` 

Notes:

- Cannot add unit test for Concurrent index as concurrent only exists for the creation of the index. It's not concurrent after the creation. 
- Named index's will be pre-fixed with `mt_` to ensure that we can pull them from the database as 'marten related indexes' and not adhoc indexes put into the system outside of marten.
- Where clauses take a string, they work but its a bit fiddly. 

I also made `Duplicate` arguments optional.

### DIFF
--- a/src/Marten.Testing/Acceptance/computed_indexes.cs
+++ b/src/Marten.Testing/Acceptance/computed_indexes.cs
@@ -47,7 +47,48 @@ namespace Marten.Testing.Acceptance
         }
 
         [Fact]
-        public void create_unique()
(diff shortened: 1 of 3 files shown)


phillip-haydon:
@jeremydmiller when you get a chance can you please review this.

Issue #438, opened 2016-06-21
HiLoSequence Increment is not used


mdissel:
What is the purpose of the Increment option in https://github.com/JasperFx/marten/blob/15056c05f53a3e917ef4060403ebe1b528f11e71/src/Marten/Schema/Identity/Sequences/HiLoSequence.cs#L26
At the moment it's not used.

jeremydmiller:
@mdissel I'd be lying if I told you I remember. It was theoretically for folks who wanted the id version to go up by more than 1. I'm inclined to just delete it.

mdissel:
Agreed, just delete the increment property

jeremydmiller:
I take pull requests.

jeremydmiller:
Closed by #440

PR #440, opened 2016-06-22
Remove unused identity property in HiLoSequence

mdissel:
Remove unused Identity property in HiLowSequence, see #438 
and removed some unused variables (thanks to coderush)


jeremydmiller:
Pulled in via a rebase.

PR #453, opened 2016-07-15
Add remove from collection patch

jimgolfgti:
Equality on complex child collections compares all fields, could possibly inspect for identity field as an optimisation.


jeremydmiller:
@jimgolfgti You might just be the new gold standard for PR's. Thanks for doing this.

PR #454, opened 2016-07-15
FetchAll added to EventStore

khalidabuhakmeh:
This should allow users to retrieve **all** events regardless
of event type. They also have the option of using `before` and
`after` datetimes which allow for event windows. Finally, they
can still use versions to retrieve versioned events.


jeremydmiller:
@khalidabuhakmeh Can you add some tests for this first before we pull it in?

khalidabuhakmeh:
Sure, can you point me to a comparable test?

PR #461, opened 2016-07-19
Update EventsQueryHandler exception parameter naming

billboga:
Previous behavior used `before` parameter name with throwing an exception for the `after` parameter.
Also, remove trailing whitespace from related test.


jeremydmiller:
Cool, got it.

Issue #459, opened 2016-07-18
Improve query generation when querying against Dictionary


baronfel:
When a Linq query is comparing against `Dictionary<K,V>.Contains(KeyValuePair<K,V>)` or `dictionary[key] == value`, we should be able to optimize the comparison by generating the dictionary for the comparison and then converting that the json.

Example:

`Session.Query<Doc>().Where(d => d.Guys.Contains(new KeyValuePair<string,string>("foo", "bar")))`  
should result in a comparison json of `{ "Guys" : { "foo" : "bar" } }`. 
Right now we give `{"Guys": [{"Key":"foo", "Value":"bar"}]}` instead

Per Jeremy's pointer this can probably be done using http://jasperfx.github.io/marten/documentation/documents/advanced/customizing_linq/

baronfel:
So I'm working on this and I think there's a small problem in the WhereFragment code with regards to the operator I want to use. Specifically, I want to use the `?` operator to check if a top-level text element exists in a jsonb object, but the escaping in the WhereFragment is obliterating my first `?` symbol.  

Specifically I want to write the sql query `{fieldLocator} ? ?`, where fieldLocator is the found field and the second `?` is my key parsed from the expression.  

I think we should change the token used to signal a variable in the query DSL, especially given that more of the json operators contain special symbols like `|`, `?`, `@` and `&`.

CoreyKaylor:
Going to close this now. Feel free to open a new issue if you think the other issues you ran into that were "hairy" need to be discussed further.

PR #463, opened 2016-07-19
Address Dictionary query generation

baronfel:
Addresses #459 

There are some dramatic changes in this.
Applying fixes for the `IDictionary<,>.ContainsKey('tkey)` and `ICollection<KVP<,>>.Contains(KVP<,>)` cases was easy enough because those fit into the existing method-call based plugin architecture. There was a wrinkle in that the operator that I want to use in postgres is `?`, which is already used by the `WhereFragment` expansion, so I also took the opportunity to make a version of WhereFragment that has a configurable separator.

The drastic thing was adding in a pluggable `IBinaryExpressionParser` interface.  This was required so that I could easily get at cases where we're indexing into a dictionary on the right or left sides to pull out the keys and values to construct the comparison dictionary.  As you'd expect a ton of things broke right off the bat, so I spent some time tweaking the `Matches` condition on the `SimplePropertyBinaryExpressionParser`. Please  please please please please please sanity check me on that.  The tests pass on my local box now but there's no way they're exhaustive in this circumstance.

- [x] Should we add a publicly-facing extension point for `IBinaryExpressionParser` instances like there are for `IMethodCallParser` instances?
- [x] I'd also like to add some more tests actually storing and querying for data based on these dictionaries before I take the [WIP] off of this.
- [x] Also, thoughts about generic dictionary access?  I think it's safe to say that the most extreme we could go is `IDictionary<string, TValue>`, because json maps must have string keys, so is that worth going ahead and doing here?
- [x] Also, I want to buff out the binary parsing some more to accept more than dict["key"] == "value".

### DIFF
--- a/src/Marten.Testing/Linq/query_for_json_format.cs
+++ b/src/Marten.Testing/Linq/query_for_json_format.cs
@@ -36,7 +36,7 @@ namespace Marten.Testing.Linq
 }}, 
 ""UserName"": ""{UserName}"", 
 ""Birthdate"": ""{Birthdate.ToString("s")}""
-}}".Replace("\r\n", "");
+}}".Replace("\r\n", "").Replace("\n", "");
         }
     }
 
--- a/src/Marten.Testing/Linq/dictionary_is_translated.cs
+++ b/src/Marten.Testing/Linq/dictionary_is_translated.cs
@@ -1,4 +1,5 @@
 ﻿using Marten.Services;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Xunit;
@@ -13,8 +14,8 @@ namespace Marten.Testing.Linq
             var query = theSession.Query<Target>().Where(t => t.StringDict.ContainsKey("foo"));
             var command = query.ToCommand(Marten.Linq.FetchType.FetchMany);
             var dictParam = command.Parameters[0];
-            dictParam.DbType.ShouldBeTheSameAs(System.Data.DbType.Object);
-            dictParam.Value.ShouldBeTheSameAs(new Dictionary<string, string> { { "foo", null } });
+            (dictParam.DbType == System.Data.DbType.String).ShouldBeTrue();
+            (dictParam.Value.ToString() == "foo").ShouldBeTrue();
         }
 
         [Fact]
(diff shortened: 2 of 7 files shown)


jeremydmiller:
Oh man, how upset w/ me would you be if I asked you to solve the "?" problem a different way? Leave "?" substitution the way that it was, and maybe use custom IWhereFragment's for the places where you need that operator? The question mark thing is a common usage and I'd prefer to keep it.

And would you also mind rebasing this on master?

baronfel:
I'm actually in the middle of a rebase (kinda scary looking, with csproj changes :)

I could also do a custom WhereFragment, easy enough to revert that commit. No hurt feelings at all, that's why the [WIP]!

jeremydmiller:
@baronfel Actually, @CoreyKaylor is in the process of removing all the csproj files and putting us squarely on project.json. You might have a better time waiting for that.

*A* way we could maybe do this is have the substitution wildcard not be hardcoded inside of WhereFragment. Keep "?" as the default, but let you override that on an individual WhereFragment?

baronfel:
Sounds like I'll take a breather then and wait for the csproj churn to finish :D
I can do a separate PR for the configurable WhereFragment as a smaller chunk while I'm waiting for all that to settle.

CoreyKaylor:
Already in master.

baronfel:
Thanks, rebasing and making the WhereFragment configurable.

src/Marten.Testing/Linq/query_for_json_format.cs
```diff
@@ -36,7 +36,7 @@ public string ToJson()
 }}, 
 ""UserName"": ""{UserName}"", 
 ""Birthdate"": ""{Birthdate.ToString("s")}""
-}}".Replace("\r\n", "");
+}}".Replace("\r\n", "").Replace("\n", "");
```
baronfel: needed this to make the tests green up on my machine

CoreyKaylor:
Looks like quite a few failing tests on the build server for this branch.

http://build.fubu-project.org/viewLog.html?buildId=47864&tab=buildResultsDiv&buildTypeId=Marten_Master&guest=1

baronfel:
Oh that's expected. I've been having trouble running all the tests successfully on my machine (my poor docker Postgres instance seems to be getting swamped), but I'm still in the process of cleaning things up on my end.

CoreyKaylor:
Just making sure we weren't the ones dropping the ball on getting this taken in. :smile:

baronfel:
I'm going to reel this back a bit and skip doing the `dict[key] == value` pattern right now, because that's a TON of extra work around making sure that we flatten/resolve the left and/or right expressions correctly to try and pull the value out of them to construct the json dictionary correctly.  So just looking at the Contains(KVP) and ContainsKey(K) methods right now.

baronfel:
Ok, I removed the [WIP] on this PR.  We now handle `ICollection<KeyValuePair<string,T>>.Contains(KeyValuePair<string,T>>` and `IDictionary<string,T>.ContainsKey(string)` cases generically, and optionally have the ability to set the NpgsqlDbType on a sql parameter, which was necessary when using the `@>` operator to check for the presence of a key-value pair in a jsonb object.   Separate PRs can address the dictionary index question, because that junk is HAIRY!

CoreyKaylor:
Thanks

PR #481, opened 2016-08-12
Daemon: Fix support for other schema

tim-cools:
I started to play around with the async daemon and noticed that events on another schema were not yet supported properly.


jeremydmiller:
Good catch.

PR #483, opened 2016-08-12
Fix event stream fetchall with before|after

jimgolfgti:
Tests can_fetch_all_events_after_now and can_fetch_all_events_before_now were repeatedly failing running in BST timezone.

The Column `mt_events.timestamp` is defined as `timestamp without time zone` and is always populated with insertion time taken from the UTC time zone. Also EventsQueryHandler enforces that the DateTime given for either `before` or `after` parameters must have `DateTimeKind` of UTC.

However, the parameters for the query were prepared as `NpgsqlDbType.TimestampTZ` which causes Postgres to adjust the given value appropriate to the server's time zone.

### DIFF
--- a/src/Marten.Testing/Acceptance/patching_api.cs
+++ b/src/Marten.Testing/Acceptance/patching_api.cs
@@ -1,8 +1,8 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Cool, thanks! I'll pull it in shortly.

PR #488, opened 2016-08-17
Use PG object comments to record metadata (IDocumentStore FQN)

jokokko:
Record Marten related metadata in PG object comments - first for tables created (so I can e.g. 
```
SELECT relname, description from pg_description
join pg_class on pg_description.objoid = pg_class.oid
where description like 'origin:Marten.IDocumentStore, Marten, Version=1.0.0.51046%'
```

### DIFF
--- a/src/Marten.Testing/Schema/add_origin_Tests.cs
+++ b/src/Marten.Testing/Schema/add_origin_Tests.cs
@@ -1,0 +1,39 @@
+﻿using Marten.Testing.Documents;
+using Marten.Util;
+using Xunit;
+
+namespace Marten.Testing.Schema
+{
+    public class add_origin_Tests
+    {     
+        [Fact]
+        public void origin_is_added_to_tables()
+        {
+            var user1 = new User { FirstName = "Jeremy" };
+            var user2 = new User { FirstName = "Max" };
+            var user3 = new User { FirstName = "Declan" };
+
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@jokokko I'll get this one tomorrow morning

jeremydmiller:
@jokokko Except it does need some tests first. I like the idea though.

bradyclifford:
I have a similar PR that injects the JSON schema into the PH meta data comments.

jokokko:
@jeremydmiller Oops, had the tests in stash and forgot to apply them to my commit.

@bradyclifford That's clever. My idea of using PG comments was to easily query the db for Marten related objects and see from which version they originate.

bradyclifford:
Ya I will probably instead just have the JSON Schema in normal script comments. My database team is requiring that they review the JSON schema that will be persisted in the data column before they will approve the DDL script.

jeremydmiller:
@jokokko @bradyclifford I'm going to hold off on this one just until Brady's changes are ready just in case there's any conflicts

bradyclifford:
@jeremydmiller decided that it probably is better that we don't include the Json schema in the table comments.  Instead, we can inject it in after the DDL script is created via our own project CLI.  Unless, you feel it would still be beneficial to have for the entire Marten community.

Issue #23, opened 2015-10-21
Command line management tool?


jeremydmiller:
I think we need some kind of facility to generate the database schema objects for document types. Think about this one more. Is it a console tool on top of the library support? Migrations?

Possible features would be:

1. Dump the full DDL for the entire schema
1. Generate patches for any places where the schema is different than the Marten configuration
1. Bulk data loading?
1. Wipe out Marten schema objects?
1. Load the Javascript projections?

jeremydmiller:
Here's my proposal. For right now, I'm proposing that we have a separate **library** to hold the command line support. In usage, you'd build your own console app so you could build out the `StoreOptions` any which way you want, then delegate to the Marten CLI support. That quickly sidesteps the issues of needing to use separate AppDomain's and "finding" the right StoreOptions if we distributed our own console app.

For usage, I've got:

```
yourtool ddl [file] --> writes all the DDL for the StoreOptions to the designated file
yourtool patch [file] --> writes the patch (if any) and drop files to the designated location
yourtool apply --log [file] --> applies all outstanding changes at one time
yourtool assert --log [file] --> asserts that the db schema matches the Marten configuration
yourtool sqitch [folder] --> See https://github.com/JasperFx/marten/issues/493



```

jeremydmiller:
This is done, just needs to be doc'd. I'm closing it for now.

PR #489, opened 2016-08-17
Toy cli - maybe a skeleton for #23

jokokko:
Created a little skeleton for CLI tool...


jeremydmiller:
@jokokko Before you go down that any farther, peek at this one: https://github.com/JasperFx/oakton. It's the CLI parsing tooling we've used for years in the fubu projects in a new, smaller library.

jeremydmiller:
@jokokko Actually, if you don't mind, I'd like to tackle this one myself tomorrow. One of our teams actually needs it pretty soon. The only one that I think will be time consuming is the Sqitch integration, which is of course the one they need.

jokokko:
@jeremydmiller sure thing.

PR #497, opened 2016-08-23
Allow switching the default IAggregator to be used when no explicit r…

jokokko:
…egistration exists.

Adds one signature to the public API, namely.
EventGraph.UseAggregatorLookup(IAggregatorLookup aggregatorLookup)


jeremydmiller:
Got it.

PR #508, opened 2016-09-01
Fix The UTC time represented when the offset is applied must be betwe…

tim-cools:
…en year 0 and 10,000 in tests

Got the error: The UTC time represented when the offset is applied must be between year 0 and 10,000.
while running theses tests. The fix below...


jeremydmiller:
Bueno, thank you Tim.

Issue #512, opened 2016-09-06
Allow for limiting the number of results returned when querying an event stream


jeffdoolittle:
When synchronizing append-only event streams between different systems, it's helpful to have control over the number of events returned when querying. 

Add a `limit` parameter to the `IEventStore.FetchStream` and `IEventStore.FetchStreamAsync` methods.

jeffdoolittle:
I think we're good to just query for `IEvent`s with `Skip/Take`

Issue #469, opened 2016-08-03
Linq support for IEvent


Tazer:
Possibility to query after events with linq.

For example something like this:

`session.Query<IEvent>().Where(x => x.Stream == ? && x.Version > #)`

So we are not only locked down to the IEventStore methods.

Tazer:
Awesome :)

PR #513, opened 2016-09-06
closes GH-512 option to limit event stream results

jeffdoolittle:
@jeremydmiller our team needs this to aid in controlling event stream querying and synchronizing event streams across systems.

### DIFF
--- a/src/Marten/Events/EventQueryHandler.cs
+++ b/src/Marten/Events/EventQueryHandler.cs
@@ -10,9 +10,10 @@ namespace Marten.Events
         private readonly EventSelector _selector;
         private readonly Guid _streamId;
         private readonly DateTime? _timestamp;
+        private readonly int? _limit;
         private readonly int _version;
 
-        public EventQueryHandler(EventSelector selector, Guid streamId, int version = 0, DateTime? timestamp = null)
+        public EventQueryHandler(EventSelector selector, Guid streamId, int version = 0, DateTime? timestamp = null, int? limit = null)
             : base(selector)
         {
             if (timestamp != null && timestamp.Value.Kind != DateTimeKind.Utc)
@@ -24,6 +25,7 @@ namespace Marten.Events
             _streamId = streamId;
             _version = version;
             _timestamp = timestamp;
+            _limit = limit;
         }
 
         public override Type SourceType => typeof(IEvent);
@@ -48,6 +50,12 @@ namespace Marten.Events
             }
 
             sql += " order by version";
+
+            if (_limit.HasValue)
+            {
+                var limitParam = command.AddParameter(_limit.Value);
+                sql += " limit :" + limitParam.ParameterName;
+            }
 
             command.AppendQuery(sql);
         }
--- a/src/Marten/Events/EventStore.cs
+++ b/src/Marten/Events/EventStore.cs
@@ -77,16 +77,16 @@ namespace Marten.Events
             return StartStream<TAggregate>(Guid.NewGuid(), events);
         }
 
-        public IList<IEvent> FetchStream(Guid streamId, int version = 0, DateTime? timestamp = null)
+        public IList<IEvent> FetchStream(Guid streamId, int version = 0, DateTime? timestamp = null, int? limit = null)
         {
-            var handler = new EventQueryHandler(_selector, streamId, version, timestamp);
+            var handler = new EventQueryHandler(_selector, streamId, version, timestamp, limit);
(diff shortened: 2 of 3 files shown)


jeffdoolittle:
Tried to add storyteller tests to verify functionality but wasn't able to get it working. I keep getting exceptions like this:

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> Npgsql.NpgsqlException: 42883: function mt_append_event(stream => uuid, stream_type => character varying, event_id => uuid, event_type => character varying, body => jsonb) does not exist
   at Npgsql.NpgsqlConnector.DoReadSingleMessage(DataRowLoadingMode dataRowLoadingMode, Boolean returnNullForAsyncMessage, Boolean isPrependedMessage)
   at Npgsql.NpgsqlConnector.ReadSingleMessage(DataRowLoadingMode dataRowLoadingMode, Boolean returnNullForAsyncMessage)
   at Npgsql.NpgsqlCommand.Execute(CommandBehavior behavior)
   at Npgsql.NpgsqlCommand.ExecuteNonQueryInternal()
   at Marten.Services.UpdateBatch.<>c.<Execute>b__7_0(NpgsqlCommand c) in C:\projects\Marten\src\Marten\Services\UpdateBatch.cs:line 46
   at Marten.Services.ManagedConnection.Execute[T](NpgsqlCommand cmd, Func`2 func) in C:\projects\Marten\src\Marten\Services\ManagedConnection.cs:line 110
   at Marten.Services.UpdateBatch.Execute() in C:\projects\Marten\src\Marten\Services\UpdateBatch.cs:line 46
   at Marten.Services.UnitOfWork.ApplyChanges(UpdateBatch batch) in C:\projects\Marten\src\Marten\Services\UnitOfWork.cs:line 120
   at Marten.DocumentSession.SaveChanges() in C:\projects\Marten\src\Marten\DocumentSession.cs:line 128
   at Marten.Testing.Fixtures.EventStore.EventStoreFixture.ForNewQuestStream(String name, DateTime date) in C:\projects\Marten\src\Marten.Testing\Fixtures\EventStore\EventStoreFixture.cs:line 44
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at StoryTeller.Grammars.Reflection.MethodInvocation.<Invoke>d__19.MoveNext() in C:\BuildAgent\work\cafc4cf00d277d9c\src\StoryTeller\Grammars\Reflection\MethodInvocation.cs:line 81
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at StoryTeller.Grammars.LineStep.execute(ISpecContext context) in C:\BuildAgent\work\cafc4cf00d277d9c\src\StoryTeller\Grammars\LinePlan.cs:line 18
   at StoryTeller.Grammars.LineStepBase.Execute(SpecContext context) in C:\BuildAgent\work\cafc4cf00d277d9c\src\StoryTeller\Grammars\LineStepBase.cs:line 40
```

jeremydmiller:
@jeffdoolittle Before we take this in (and I won't without tests), can you take a look at this functionality: https://github.com/JasperFx/marten/issues/469? It would allow you to do Skip() and/or Take().

jeffdoolittle:
Totally agree regarding tests which is why I was trying to get ST working. If you can point me in the right direction to get it running, that would be great.

Regarding the request at hand, we were trying to avoid querying the raw stream based on warnings in the docs: 

http://jasperfx.github.io/marten/documentation/events/streams/#sec3

We got a bit timid reading the docs at first since it says flat out that direct querying is not a good idea. But as I'm looking at the code for #469 it looks like it's possible to still filter by streamid and get an index hit, so this could work for us. 

Thoughts?

jeremydmiller:
It shouldn't be that bad if you're querying against some combination of stream id and sequence. Both of those fields would be indexed. You *can* add additional indexes. Not sure how much difference that would make to write performance.

The EventQueryHandler does go at the raw data too, so it's 6 of one, half a dozen of another here.

PR #527, opened 2016-09-21
Adds an example of using ViewProjection<T>

daveaglick:
As promised, here's an example of `ViewProjection<T>` for aggregating across streams. cc @tim-cools

### DIFF
--- a/src/Marten.Testing/Events/Projections/WeaponNamesByQuest.cs
+++ b/src/Marten.Testing/Events/Projections/WeaponNamesByQuest.cs
@@ -1,0 +1,108 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Marten.Events.Projections;
+
+namespace Marten.Testing.Events.Projections
+{
+    // SAMPLE: WeaponNamesByQuest
+// Aggregates
+
+public class Quest
+{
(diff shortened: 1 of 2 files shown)


src/Marten.Testing/Events/Projections/WeaponNamesByQuest.cs
```diff
@@ -0,0 +1,108 @@
...
+
+public class WeaponNamesByQuestProjection : ViewProjection<WeaponNamesByQuest>
+{
+    public WeaponNamesByQuestProjection()
+    {
+        ProjectEvent<WeaponCreated>(
+            (session, evt) => session.Load<Person>(evt.HolderId).QuestId,
+            (view, evt) => view.WeaponNamesLookup[evt.Id] = evt.Name);
+        ProjectEvent<WeaponNameChanged>(
+            (session, evt) => session.Load<Person>(session.Load<Weapon>(evt.Id).HolderId).QuestId,
```
jeremydmiller: You can use an Include() there to make it more efficient. Uglier code sample though. I think it might be worth going back at that ViewProjection to find ways of making it more efficient on the data fetching so that it could batch things up as needed.

jeremydmiller:
Cool, I'll get this up on the site soon. Thanks!

PR #542, opened 2016-10-04
Ensure existing functions are not regenerated

jimgolfgti:


jeremydmiller:
@jimgolfgti Thanks for doing this. I'll roll this one in after CI gets a chance to run the build for this.

PR #546, opened 2016-10-05
Support for Apply(Event<T>) aggregation

jmezach:
As briefly discussed in the Gitter room I took a stab at adding support for Apply(EventOfT) style aggregation. This would allow the aggregate to use raw event data (such as timestamp).

Would love some feedback on this.

### DIFF
--- a/src/Marten/Events/Projections/Aggregator.cs
+++ b/src/Marten/Events/Projections/Aggregator.cs
@@ -22,7 +22,7 @@ namespace Marten.Events.Projections
                 {
                     object step = null;
                     var eventType = method.GetParameters().Single<ParameterInfo>().ParameterType;
-                    if (eventType.GetTypeInfo().IsGenericType && eventType.GetGenericTypeDefinition() == typeof(Event<>))
+                    if (eventType.Closes(typeof(Event<>)))
                     {
                         eventType = eventType.GetGenericArguments().Single();
                         step = typeof(EventAggregationStep<,>)
--- a/src/Marten/Events/Projections/Aggregator.cs
+++ b/src/Marten/Events/Projections/Aggregator.cs
@@ -20,9 +20,19 @@ namespace Marten.Events.Projections
                 .Where(x => x.Name == ApplyMethod && x.GetParameters().Length == 1)
                 .Each(method =>
                 {
+                    object step = null;
                     var eventType = method.GetParameters().Single<ParameterInfo>().ParameterType;
-                    var step = typeof (AggregationStep<,>)
-                        .CloseAndBuildAs<object>(method, typeof (T), eventType);
+                    if (eventType.GetTypeInfo().IsGenericType && eventType.GetGenericTypeDefinition() == typeof(Event<>))
+                    {
+                        eventType = eventType.GetGenericArguments().Single();
+                        step = typeof(EventAggregationStep<,>)
+                            .CloseAndBuildAs<object>(method, typeof(T), eventType);
+                    }
+                    else
+                    {
+                        step = typeof(AggregationStep<,>)
+                            .CloseAndBuildAs<object>(method, typeof(T), eventType);
+                    }
 
                     _aggregations.Add(eventType, step);
                 });
--- a/src/Marten.Testing/Events/Projections/AggregationStepTests.cs
+++ b/src/Marten.Testing/Events/Projections/AggregationStepTests.cs
@@ -33,21 +33,19 @@ namespace Marten.Testing.Events.Projections
         [Fact]
         public void can_build_aggregation_step_for_an_event_apply_method()
         {
-            Expression<Action<QuestParty, MembersDeparted>> apply = (p, j) => p.Apply(new Event<MembersDeparted>(j));
+            Expression<Action<QuestPartyWithEvents, MembersJoined>> apply = (p, j) => p.Apply(new Event<MembersJoined>(j));
 
 
-            var departed = ReflectionHelper.GetMethod<QuestParty>(x => x.Apply(new Event<MembersDeparted>(new MembersDeparted())));
+            var joined = ReflectionHelper.GetMethod<QuestPartyWithEvents>(x => x.Apply(new Event<MembersJoined>(new MembersJoined())));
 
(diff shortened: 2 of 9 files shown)


src/Marten/Events/Projections/Aggregator.cs
```diff
@@ -20,9 +20,19 @@ public Aggregator()
                 .Where(x => x.Name == ApplyMethod && x.GetParameters().Length == 1)
                 .Each(method =>
                 {
+                    object step = null;
                     var eventType = method.GetParameters().Single<ParameterInfo>().ParameterType;
-                    var step = typeof (AggregationStep<,>)
-                        .CloseAndBuildAs<object>(method, typeof (T), eventType);
+                    if (eventType.GetTypeInfo().IsGenericType && eventType.GetGenericTypeDefinition() == typeof(Event<>))
```
jeremydmiller: If you want to shortcut that code, there's an extension method in Baseline called `Closes(type)` that would do the same logic.

jeremydmiller:
Actually, I think it looks good as is in the real code, but in the tests, could you make a separate `QuestParty2` aggregate (or some better name) that does `Apply(Event<T>)` and leave the original tests that use the old `Apply(event body type)` signatures? So that we've got coverage for both? 

I'm coincidentally getting 1.1 ready this morning and I need to push it out when I can. If you agree, I could take in the PR and do the little changes to the test coverage.

jeremydmiller:
@jmezach Oh, and thank you for doing this one;)

src/Marten/Events/Projections/Aggregator.cs
jmezach: You mean if (eventType.Closes(typeof(Event<>))?

jmezach:
@jeremydmiller No problem. Loved to do it. Quite like Marten so far, but discovered a scenario where this could be really useful so I thought I'd might as well add it myself since its open source ;).

I did break the IAggregation<T, TEvent> interface since that now requires an Apply(T, Event<TEvent>) method instead of Apply(T, TEvent), so technically that's a breaking change ;).

jeremydmiller:
Gotcha. This one's going to have to wait then. Let's find a way to pull this off additively so we don't have to break the IAggregation interface.

jmezach:
Agreed, that would be better solution. And I think I already have a solution that doesn't require breaking the interface. Need to run some tests though ;).

jeremydmiller:
Cool, I'm gonna see if I can put off the 1.1 thing for a couple hours. I do need to push out the connection/transaction hygiene improvements to my internal team today though.

jmezach:
After looking at this some more I think it might be a bit tricky to pull off without breaking things. My first attempt was to use the **IAggregation&lt;T, TEvent>** interface as is, but use *Event&lt;TEvent>* as the type for *TEvent*. Unfortunately there is also the **IAggregator&lt;T>** interface which has an *AggregatorFor&lt;TEvent>()* method that returns an **IAggregation&lt;T, TEvent>**. That would then also have to change to return an **IAggregation&lt;T, TEvent>** which is a breaking change as well.

Perhaps it would be possible to make **Event&lt;TEvent>** directly convertible to *TEvent*, but I'm not quite sure if thats enough to make this non-breaking.

Any thoughts on this?

jeremydmiller:
Jump in a time machine, go back 6 months, and do it the way you had it at first?

My thought is that you keep two different collections of `IAggregation<T, TEvent>` and `IAggregationWithEventData<T, TEvent>` inside of aggregator. Then just try to hide that detail from the user and allow them to interchange Apply(T) and Apply(Event<T>)?

jeremydmiller:
I am going to hold off on this until 1.2, but that's going to be right around the corner anyway

jmezach:
That sounds reasonable.

Problem I still have though is that the aggregator needs to have access to the Event&lt;T> for both cases, which it currently doesn't have. Event&lt;T> calls AggregatorFor&lt;T> which returns an object that only has an Apply(T) method, but not an Apply(Event) method.

jmezach:
Did some more experimenting on this, but one way or the other we need access to the actual event data (ie. **Event&lt;TEvent>**). So what I did right now is add another method to the **IAggregator&lt;T>** interface called *AggregatorForEvent&lt;TEvent>()* which returns an IAggregation&lt;T, Event&lt;TEvent>> and I marked the existing *AggregatorFor&lt;TEvent>()* method as obsolete.

Of course this is still breaking an interface, since implementors of **IAggregator&lt;T>** will now have to implement another method. I've already made an **EventAggregationTranslator** class that translates from **IAggregation&lt;T, Event&lt;TEvent>>** to **IAggregation&lt;T, TEvent>** which could help with implementing the new method.

Would that be more acceptable than breaking the IAggregation interface?

jeremydmiller:
@jmezach Do you mind if I take this the last little bit today or tomorrow? I've got a backlog of things to go out in a 1.2 release, so now's a good time to bet it bundled in.

jmezach:
Sure, no problem. I do still have some uncommitted changes which were moving in the direction I mentioned in my last update. I believe that works for Marten itself, but I still had to fix some compile errors in the tests. I can probably fix those quickly by using the EventAggregationTranslator class, but I wasn't sure if you want to go there?

jeremydmiller:
@jmezach I finished it off in https://github.com/JasperFx/marten/commit/c337bd476bddcba33397b126db0c62024b55aec7. See what I did in the `Event<T>` class. I'm not terribly proud here, but it should serve until we do some kind of 2.0 and adjust the public API.

jmezach:
Well, I guess that works yes ;). Nice to see that this got in and we can always fix the API issues later. API design is hard ;). Looking forward to 1.2.

jeremydmiller:
In retrospect, I wish I'd cut the event store to a separate addon package so it could be rev'd independently. Didn't do that at the time because I didn't want the extra complexity of dealing with the plugin system.

jmezach:
I guess that would have been neat from a usage perspective since you could just use Marten as a document database and not have to bother with the event store stuff. Then again, I quite like how both the event store side of things and the document database side of things combine together. I'm sure that also brings some complexity if it was separated out.

Issue #548, opened 2016-10-06
LINQ boolean And error


lngr:
The following query returns documents:
 
`           var query = theSession.Query<Target>().Where(item => item.Flag && false);
`

Please see the attached pull request with a testcase that fails on my machine (tested on a Mac with dotnet 1.0.0-preview2-003131 after we encountered the problem in our actual Windows app).

jeremydmiller:
@lngr What is it that you're trying to do to query there? What's the point of the `&& false`?

lngr:
@jeremydmiller This was actually the simplest test case that I came up with after some testing why our queries returned wrong results.

Our real queries look somewhat like that:

`.Where(x => x.IsActive && x.Candidates.Intersect(all).Any())`

where `Candidates` is a `List<Guid>` and `all` is a `List<Guid>` of precomputed  candidates.  (Essentially, we want to fetch all documents where either of the candidates we are interested in is in the document's `Candidates` collection.)

Now it happened that this particular Intersection-Query was completely ignored, i.e., we received results for all documents having `x.IsActive` true, not only those with a non-empty intersection.

After some manual exploring and testing, I figured that even a `.Where(x => x.IsActive && false` yields a list of all documents having a true `x.IsActive`, and therefore I assumed the underlying error is within the combination of a boolean field (`x.IsActive`) with a boolean result.  In fact, if you replace the `x.IsActive` with another expression that yields true (e.g. `x.Field == "value" && false`), the query works as expected.

jeremydmiller:
Marten doesn't support `Intersect()` in the Linq provider. Pretty well every possible way you can use Linq has to be explicitly coded in Marten with some way to translate the Linq to postgres flavored SQL, and that one hasn't come up before. The combination of Intersect(), then Any() would be rough. Might be easier after I push through the `SelectMany()` support.

Honestly, because the Linq work is just so damn complicated for this kind of thing, I might be pushing to say folks should use https://github.com/JasperFx/marten/issues/514 or calculated fields for anything like this. What would you think about that? The JS function can be used from a postgres calculated index, so all the perf hit is on the writes.

lngr:
Oh, ok, it did not throw an exception, so I assumed it was supported (`Where(x => all.Intersects(x.Candidates).Any())` did throw).

[...]

PR #549, opened 2016-10-06
Testcase for issue 548

lngr:
Testcase for the issue described in
https://github.com/JasperFx/marten/issues/548


jeremydmiller:
@lngr Got it, and thank you for writing up the test case. Makes fixing problems go a lot faster on my end. It'll be at least tomorrow though.

PR #556, opened 2016-10-14
Add basic intersection support via the ?| jsonb operator

lngr:
This PR adds basic intersection support to Arrays/Lists that serialize to a "array of strings" representation in JSON (this is a limitation/requirement of the `?|` jsonb-operator). As of now only `string` and `Guid` lists/arrays are supported, other types are not tested and yield a `NotSupportedException`.

This PR should provide functionality for a quite common use case for a document store: Suppose you have a document with a collection of "tags":
`public class Article {
    public string[] Tags { get; set }
}`

and you want to query all documents that have one or more "tags" in common with your "interest": `var interests = new string[] { "finance", "astrology" };`

You can now do the following:
`theSession.Query<Article>()
                .Where(x => x.Tags.Any(s => interests.Contains(s)))
`

This is essentially an intersection query with a funny syntax but matches the way the `?|` json operator works.

Note that due to the tight coupling of the query to the `?|`, (as of now) the query in fact must look like specified above; in particular, the "query parameter" `interests` must be a `string[]` and the `.Any`-predicate must be a single `Contains` condition.

### DIFF
--- a/src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
+++ b/src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
@@ -10,7 +10,6 @@ using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Clauses.ResultOperators;
 using System.Reflection;
-using System.Collections;
 
 namespace Marten.Linq
 {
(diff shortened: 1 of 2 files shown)


src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
```diff
@@ -112,5 +130,78 @@ private static void gatherSearch(BinaryExpression x, Dictionary<string, object>
...
+
+            return $"data->'{pathTo(member)}' ?| :{fromParam.ParameterName}";
+        }
+
+        private void throwNotSupportedContains()
+        {
+            throw new NotSupportedException($"The Contains() operator on subqueries within Collection.Any() searches only supports constant array/lists of {string.Join(" or ", supportedTypes.Select(t => t.Name))} expressions");
+        }
+
+        private string pathTo(MemberExpression memberExpression)
```
jeremydmiller: I think you probably want to use `FindMembers` here to do this

src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
```diff
@@ -112,5 +130,78 @@ private static void gatherSearch(BinaryExpression x, Dictionary<string, object>
...
+            {
+                var propInfo = memberExpression.Expression.GetType().GetProperty("Member");
+                var propValue = propInfo.GetValue(memberExpression.Expression, null) as PropertyInfo;
+                path.Add(propValue.Name);
+                memberExpression = memberExpression.Expression as MemberExpression;
+            }
+            return path.AsEnumerable().Reverse().Join("'->'");
+        }
+
+        private bool isListOrArrayOf(Type value, Type valid)
```
jeremydmiller: There might be an `IsEnumerable()` type extension somewhere that effectively does the same thing.

jeremydmiller:
@lngr Nice!. I think you probably got to see this, but that `CollectionAnyContainmentWhereFragment` class is about the ugliest code in all of Marten. I'm going to let CI catch up to build this PR, then I'll pull it in.

The comments aren't terribly important, and I might try to take care of them later when I have to go back into this kind of code for the `SelectMany()` support.

src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
lngr: Ah, great, thanks, PR updated. What is your stance on git history here, do you want me to merge both commits to one?

src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
lngr: I took a quick search and could not find anything suitable. If anyone knows a better way to check that, I'll be glad to change it.

src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
jeremydmiller: I'm pretty easy going on that score, as is is just fine with me. I generally don't worry about squashing commits in most cases.

PR #557, opened 2016-10-15
Add Delete and Duplicate to Patch API

### LINKED ISSUE #434: Patching API should allow adding and removing properties
Common schema migration operations done in bulk would be nice to handle in the patching API:

- Add new property with a common specified value (ie session.Patch<Foo>.Add("newprop", 4242))
- Delete an existing property (session.Patch<Foo>.Delete("oldprop"))

jimgolfgti:
See docs for functionality, I've updated to state available from v1.2.

This should close #434, @mattzink can you want to take a look and confirm if this covers what you were requesting.


jeremydmiller:
Dude, you're awesome!

Issue #562, opened 2016-10-17
Add LINQ support for Any() with no predicate


ericgreenmix:
Ex.
```
session.Query<ClassWithChildCollections>().Where(x => x.Companies.Any());
```

jeremydmiller:
Closed by https://github.com/JasperFx/marten/pull/564

PR #564, opened 2016-10-18
Possible implementation for #562, albeit not overly elegant. Speciali…

jokokko:
…zed case for .Any() with no predicate -> uses JSONB_ARRAY_LENGTH.

Does not generalize (e.g. for nested arrays) as executed only on paths where expression.QueryModel.BodyClauses is empty.


jeremydmiller:
@jokokko Nice, I'll take it. I was going to try this week to clean up and generalize the child collection Linq support after all the bugs came in last week, but your new where fragment will carry right over into that. Thank you.

PR #568, opened 2016-10-20
Unify aggregator lookup in ProjectionCollection to use the same aggre…

jokokko:
…gator lookup as EventStore.AggregateStream (instead of hardcoded Aggregator).

Allows for immutable projections (as demonstrated in unit test).


jeremydmiller:
I see what you're saying now and this looks fine. Would you be up for doing some kind of improvement on registering the lookup factory or adding your example to the event aggregator docs maybe?

jokokko:
Sure. I'll PR something tomorrow / over weekend.

Issue #577, opened 2016-10-22
AsyncDaemon: Rebuild after projections not working


tim-cools:
In my application I try to add Rebuild projections feature in the admin page. When I try to rebuild with the daemon that I also use to project my projections, the projections are not rebuild. The data is removed from the database, but the projections are not rebuild. I believe it is because as _lastEncountered in the Fetcher is not reset but stays at the event number of the previous projections.

The same behavior is happening when you rebuild twice from the same daemon. I added this in a test that fails.

jeremydmiller:
@tim-cools The easy workaround is to just build a new daemon object each time you do that. I'm trying to make a fix for this now, but I'm running into trouble w/ postgres itself.

PR #578, opened 2016-10-22
Add failing test for 577 AsyncDaemon: Rebuild after projections not w…

tim-cools:
…orking

Failing test for #577

### DIFF
--- a/src/Marten.Testing/AsyncDaemon/async_daemon_end_to_end.cs
+++ b/src/Marten.Testing/AsyncDaemon/async_daemon_end_to_end.cs
@@ -199,6 +199,31 @@ namespace Marten.Testing.AsyncDaemon
         }
 
         [Fact]
+        public async Task do_a_complete_rebuild_of_the_active_projects_from_scratch_twice_on_other_schema()
+        {
+            _fixture.LoadAllProjects();
+
+            StoreOptions(_ =>
+            {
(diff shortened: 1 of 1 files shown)


jeremydmiller:
@tim-cools I'll try to look at this tomorrow. Today flew by too fast for me.

jeremydmiller:
@tim-cools Thought I had this fixed, then I think I've hit a different issue with it.

My first question is "why would you do that?"

tim-cools:
@jeremydmiller good question. Currently I host the Daemon my web app, not sure if that is a good idea. And Daemon sounded like it is something expensive to create, so I keep only one instance. So I thought to reuse the single instance to both run the projections and to rebuild the projections. So I stop the Daemon before the rebuild and start it again afterwards. I suspected it might be a problem that two running Daemons start to interact with each-other.

I also had a case where the projections didn't start again after rebuilding. But I couldn't create a tests to reproduce it at first attempt. If it occurs again, I will try to reproduce it in a test again.

jeremydmiller:
@tim-cools Actually, that makes perfect sense to me. I think what we need though is a totally different method for semantics of "restarting" a projection:

1. Delete all existing projected docs
1. Rollback the event encountered counter to 0
1. Process all existing events and continue to run the projection after it catches up

See https://github.com/JasperFx/marten/issues/596 for the new issue.

Issue #579, opened 2016-10-23
CombGuid implementation not correct for Postgres


barryhagan:
Currently the CombGuid generator does not produce uuids that will be sequential or ordered correctly in Postgres.

Postgres stores all uuids as big endian, in fact, it simply converts the string representation into a byte array without considering whether the system is big/little endian.

Given that the string representation of a Guid is always assumed to be big endian, Marten should be setting the timestamp portion of the comb in a way that will sort correctly in string representation rather than the byte order.  This will allow Postgres to sort correctly, and in .Net, the default Equality comparison of Guids can be used rather than byte order, which is more natural when doing things like .OrderBy() anyway.

jeremydmiller:
This was addressed by #580

PR #580, opened 2016-10-23
#579 Change CombGuid to be based on unix time, and sort properly in Postgres

barryhagan:


jeremydmiller:
Yikes, thank you for doing that.

PR #581, opened 2016-10-23
#579 - Fix for big endian.

barryhagan:
Big endian should be a straight copy of the 48 bits of the timestamp we care about.


barryhagan:
apologies, forgot to handle the big endian conversion.

PR #607, opened 2016-11-02
Bug in subinterface query

lngr:
If an `IPapaSmurf` is queried, `SubClassDocumentStorage` would try to deserialize to objects of type `IPapaSmurf` instead of the actual types.

This bug was not triggered by the ST test case in the documentation because `.Count()` does not actually instantiate the classes.

The attached commits borrows the type resolving from `HierarchicalResolver` by delegating to `SubClassMapping`.  (I've seen the TODOs in `SubClassDocumentStorage.cs` but that's a larger architectural change that's currently beyond my abilities in the Marten codebase :) )

Note: A plethora of tests fails on my local machine mainly due to (appearently) broken PLV8 extension on my dev machine. I could therefore not reliably confirm that the change breaks no other tests.

### DIFF
--- a/src/Marten.Testing/Linq/query_with_inheritance.cs
+++ b/src/Marten.Testing/Linq/query_with_inheritance.cs
@@ -142,5 +142,22 @@ namespace Marten.Testing.Linq
             theSession.Query<IPapaSmurf>().Count().ShouldBe(3);
         }
         // ENDSAMPLE
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@lngr Got it, and everything passes just fine. Code looks good and thank you for doing this.

I did add another UT that just exercised the async path from your test. This will be in 1.2.

lngr:
@jeremydmiller  lightning fast, as usual :)

Issue #624, opened 2016-11-20
Async Daemon Projection Rebuild breaks when seq_id of event is over 101 greater than last event seq_id


ericgreenmix:
So this is a strange issue. I have been able to consistently reproduce it. If I have multiple events of the same type in the `mt_events` table, and the `seq_id` of one of those events is 101 greater than the `seq_id` of a previous event of that same type. The streams of that type, starting with the one that was 101 greater than the previous, will left out of the call to `ApplyAsync` (from the `IProjection` interface)

Ex. that fails
```
seq_id | type
8      | user_created
109    | user_created     -> this one will not be passed to `ApplyAsync`
```

Ex. that works
```
seq_id | type
8      | user_created
108    | user_created   
```

ericgreenmix:
@jeremydmiller did you mean to close #632 instead of #624 with 88cd64b ?

jeremydmiller:
@ericgreenmix No, I was pushing that branch up to deal w/ #624. I had been trying to work through a couple different daemon issues at one time. I'm going to let you close #632 if you're okay with the explanation. Or add the "docs" tag and make it a "go make the documentation better" issue

jeremydmiller:
This will be in 1.2.4

PR #625, opened 2016-11-22
Passing and failing tests for #624

ericgreenmix:


jeremydmiller:
@ericgreenmix I'm looking at this right now.

jeremydmiller:
Ah, okay, I get it. I think the way I'd want to approach this is to make the daemon on a rebuild to first look for the latest sequence id, then keep the fetching going until then.  It's getting defeated right now because the gap in sequence is bigger than the page, and the daemon is stopping when it hits an empty set.

ericgreenmix:
Ah that makes sense. Sounds like a good solution to me

jeremydmiller:
@ericgreenmix Okay, the next question is "how did you get the event store into this state in the first place?"

ericgreenmix:
Eh, what exactly do you mean by that? We have tons of events going into our event store. When an event is stored and then over 100 different event types get stored, then that same event type gets stored again, that is when we start seeing the issue.

jarroda:
@jeremydmiller looking at the `Fetcher` code, it looks like the issue might be the `fetchNextPage`.  We don't actually have gaps in our event store (there are events for every sequence number), but the query 
```
select seq_id from {_selector.Events.DatabaseSchemaName}.mt_events where seq_id > :last and seq_id <= :limit and age(transaction_timestamp(), {_selector.Events.DatabaseSchemaName}.mt_events.timestamp) >= :buffer order by seq_id;
{_selector.ToSelectClause(null)} where seq_id > :last and seq_id <= :limit and type = ANY(:types) and age(transaction_timestamp(), {_selector.Events.DatabaseSchemaName}.mt_events.timestamp) >= :buffer order by seq_id;
```
seems to be filtering on type and grabbing the seq_id window at the same time.  So if there aren't any events of the particular type in that window, you'll get 0 results.  This I think is what is causing the issue to manifest itself in our system.

jeremydmiller:
@jarroda Yeah, I bumped into that w/ Eric's test. I'll add another test that runs continuously with big gaps somehow. This is turning into a much bigger issue than I was anticipating.

Issue #39, opened 2015-10-21
Full text searching support on the documents?


jeremydmiller:
More research on this one.

jeffdoolittle:
Part 2 of of the @robconery series on postgres as document store may be relevant to this issue:

http://rob.conery.io/category/postgres-document-api/

danbarua:
You could do this the same way you've done searchable fields, just persist it as a ```tsvector```.

jeremydmiller:
@danbarua Yeah, but I haven't thought about how you might use or expose that from the client. You always have the ability to go straight at SQL I guess. Any suggestions on that angle?

danbarua:
I'll have a play with it and see if I can come up with anything.

danbarua:
Just leaving this here as a reminder:

```csharp
class Foo
{
  [FullText]
  public string Name {get; set;}
}

db.Query<Foo>.Where(x => x.Name.MatchesTextQuery("your_full_text_query"))
```

robconery:
You have to persist the values at save - unfortunately triggers won't work on documents (which is a bummer). What we did for Moebius (Elixir db tool) was to define a function that would index a set of fields;

```ex
db(:users)
  |> searchable([:name, :email])
  |> save(...)
```

I would imagine here you could do the same with an attribute on the document itself:

```cs
[SearchableDocument("name","email")]
class Foo {
  //...
}
```

The when save is called check the attributes and update them in a separate call after save:

```sql
-- spin up this SQL string based on the attributes
update foos set(search) = to_tsvector(concat(name, ' ', email, ' '));
```

Anyway - that's what we do :).

jeremydmiller:
@robconery Thanks for the input Rob. 

In Marten's internals, we've got a class called `DocumentMapping` that models everything about how to persist a document type (what's the id, which fields are duplicated for searching, etc), and from that model Marten codegen's the db table, upsert sproc, and even the C# glue code that goes from a document object to calling the right sproc.

In this case I think we'd add the full text searchable fields to the DocumentMapping model and enhance the codegen stuff to pass that along to upsert sprocs that also set up the full text mapping.

robconery:
Sounds like it fits right in - nice job :).

danbarua:
That's pretty much the approach I'm trying :+1:

superlogical:
Here's another approach that uses Trigram's and the pg_trgm extension http://www.postgresql.org/docs/current/static/pgtrgm.html

[...]

PR #630, opened 2016-11-28
Elementary full text search implementation

jokokko:
Could this serve as a basis for #39?

### DIFF
--- a/src/Marten.Testing/FullTextSearch/search_by_phrase_Tests.cs
+++ b/src/Marten.Testing/FullTextSearch/search_by_phrase_Tests.cs
@@ -1,0 +1,83 @@
+﻿using System;
+using System.Text;
+using System.Linq;
+using System.Threading.Tasks;
+using Marten.Services;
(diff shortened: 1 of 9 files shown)


jeremydmiller:
@jokokko,

Sorry I've been so slow getting around to this one. I'd be happy to take this one in today. I do think we need to add some additional tests for the database object lifecycle (clean out the DB, see it get reestablished, different schema auto-creation rights).

What else do you think needs to happen to get this production ready?

And docs, but that's no big deal.

jokokko:
@jeremydmiller,

Yeah, the things you mention need to be there... I don't know if the UsingVector is the most elegant way to go, but on the other hand it gives explicit control over naming (maybe omitting that could default to reflecting over the member name and using as part of the name).

I'll hope to find the time to refine this PR soon.

jeremydmiller:
I'm closing because there's no activity here. I think the consensus seems to be to just wait for Postgresql 10.

JonCanning:
Now? I could really use this!

Issue #480, opened 2016-08-11
Pagination Problem


ngonzalezromero:
Hi.
I have 11 registers in database

With this code

```
   public async Task<PagingResult<Message>> PagingAsync(Expression<Func<Message, bool>> condition, string columnOrder, bool orderDescending, int pageIndex, int pageSize)
    {
        var session = _documentStore.OpenSession();
        PagingResult<Message> data = new PagingResult<Message>();
        IList<Message> list;

        try
        {
            QueryStatistics stats = null;
            var order = _apiUtil.GetLambda<Message>(columnOrder);

            var query = session.Query<Message>().Stats(out stats).Where(condition);

            if (orderDescending)
            {
                list = await query.OrderByDescending(order).Skip(pageIndex).Take(pageSize).ToListAsync();
            }
            else
            {
                list = await query.OrderBy(order).Skip(pageIndex).Take(pageSize).ToListAsync();
            }

            if (list != null && list.Any())
            {
                data.List = list;
                data.Count = stats.TotalResults;
            }
        }
        catch (Exception ex)
        {
            _logger.LogCritical(_apiUtil.GetExceptionDetails(ex));
        }
        finally
        {
            session.Dispose();
        }
        return data;
    }
```

if i send  pageIndex =1 and  pageSize =10, Marten return 10 registers.
But when pageIndex =2 and  pageSize =10, Marten return 9 register, but this is no correct, because should be return only 1( the page 2 of 11 element only contains 1 element)

cc: @jeremydmiller

jeremydmiller:
I think you need something like:

```
Skip((pageIndex - 1) * pageSize);
```

The Skip() uses the total number of items to skip. It doesn't have any logic around paging per se.

ngonzalezromero:
Thank you @jeremydmiller 
Nice project !!!
Regards

jeremydmiller:
@ngonzalezromero If you're bored and need a little OSS side thing, it wouldn't hurt to have a better paging model in Marten where you could do something more like:

```
session.Query<Document>().Page(page #, page size);
```

but just have it delegate to Skip() & Take() internally.

ngonzalezromero:
Uf.. It would be nice !!

if you want.. I could do it ..
Just please give me some instruction where I must add the code and then I will send the PR..

I would love to contribute to this project

mdissel:
And optionally return a PagedResult with recordcount, numberofpages, etc.

[...]

PR #639, opened 2016-12-12
Implemented IQueryable extension methods PagedResult and PagedResultAsync for paginated queries

mysticmind:
Implemented IQueryable extension methods PagedResult and PagedResultAsync for paginated queries. Refer #480.


src/Marten.Testing/Pagination/invoking_pagedquery_Tests.cs
```diff
@@ -0,0 +1,101 @@
...
+            Func<IQueryable<Target>, IQueryable<Target>> order = null;
+            order = q => q.OrderBy(m => m.Date);
+
+            var result = theSession.Query<Target>().PagedQuery(pageIndex, pageSize, order);
+
+            result[0].Date.ShouldBeLessThan(result[result.Count - 1].Date);
+        }
+
+        [Fact]
+        public void can_return_paged_result_async()
```
jeremydmiller: Can you use async Task tests instead of relying on GetAwaiter().GetResult() here?

jeremydmiller:
I'm good with this other that the tests. I'd prefer that you didn't bother with all the `this.[prop name]`, but I don't really care that much. I take the attitude of "try not to discourage folks from contributing" in these cases.

I might open a separate issue with this afterward. What you're doing won't work OOTB for compiled queries -- and that's way, way too much work to do over vacation, so January;)

mysticmind:
Thanks for reviewing the PR. I have removed usage of this.[prop name] in code and removed usage of GetAwaiter().GetResult() in unit tests and used async Task tests as per your review comments.

jeremydmiller:
@mysticmind I *think* that I'd vote to merge the Metadata class and logic into the PagedList so that everything is easily available and discoverable to users.

That being just a preference thing, the only real problem is that I think it needs a bit more testing. I think you need more tests on fetching a page in the middle, the "has next page" / "has previous page", and the page counts.

mysticmind:
@jeremydmiller I have taken your suggestions, re-looked at the implementation and refactored the whole feature with better abstractions and unit tests covering all functionality. I am quite happy on how the implementation has evolved to the current form. Please check and let me know your thoughts.

Issue #640, opened 2016-12-15
Buffer pooling for marten


Scooletz:
Below you can find a description of providing buffer pooling for Marten.

---

Currently, `EventStreamAppender` uses `JsonBodies` func to append bodies as an array of strings ([here](https://github.com/JasperFx/marten/blob/2f18d09fa2034cbc647f48a74bbf3bbb8ea51116/src/Marten/Events/EventStreamAppender.cs#L37)).
According to `Npgsql` driver, it's possible to pass a `char[]` instead of string and use the parameter size to limit the length of the appended data ([here](https://github.com/npgsql/npgsql/blob/4ef74fa78cffbb4b1fdac00601d0ee5bff5e242b/src/Npgsql/TypeHandlers/TextHandler.cs#L269-L276). 

We can't use it as we pass the array parameter, but we could 
- enable passing `ArraySegment<char>` as `NpgsqlDbType.Jsonb` in Npgsql driver first
- enable serializers to serialize to a predefined `char[]` (+ possibly reallocation if needed)
- pass result `ArraySegment<char>` below.

Having this, we could leverage that for other operations, like storing documents (`DocJsonBodyArgument` uses serializer as well). Actually, **we could start with docs first** as they issue a single `Jsonb` parameter (hence, we can use `Size`) and later, decide do we want it for events.

I'm aware that these allocations are short living. On the other hand we can't tell is it worth it before trying it out.

What do you think?

jeremydmiller:
@Scooletz It's a little bit related to https://github.com/JasperFx/marten/issues/610. I know that we're doing more allocations than is strictly necessary internally in Marten. I chickened out of trying to do this for 1.2, but I'd like to see this happen. 

Hey, in a busy system, this might be well worth the effort. I got stopped last time on the identity map abstractions, but that's a problem that we can beat.

Scooletz:
A little bit. My proposal would be to cache `TextWriter` or `char[]` to allow pass it dipper via params.

Don't 🐔 , I'll sketch something by the end of weekend ;-)

Scooletz:
The pooling has been added.

PR #642, opened 2016-12-17
[WIP] Buffer pooling for JSON serialization

Scooletz:
**Work in progress**

Addresses #640 

This PR Introduces `CharArrayTextWriter` that by provides access to its `char[]`. Additionally changes the way that driver parameters are created, enables reusing memory and reducing allocations for serializing.

Plan
- [x] change `DocJsonBodyArgument` to use pooled writer
- [x] ~change `EventStreamAppender` to use pooled writers~
- [x] test buffering
- [x] decide what's the best buffering strategy: 
  - one pool per app (current approach) 
  - one per `DocumentStore`
  - additional pool for session to keep leased writers for reuse during the same session and return at the end of the session
  - another one?
- [x] decide should this be `opt-in` or `opt-out` behavior

**Notes:**
- The `EventStreamAppender` can't be changed now as Npgsql does not support writing `ArraySegment<char>` as `jsonb`, `json` nor `string`. I took a look at the implementation of the driver at it would require a lot of work as the [WriteBuffer](https://github.com/npgsql/npgsql/blob/dev/src/Npgsql/WriteBuffer.cs#L331-L33378) does not have overloads for this. If we had it, then appending events would be easy.

### DIFF
--- a/src/Marten.Testing/performance_tuning.cs
+++ b/src/Marten.Testing/performance_tuning.cs
@@ -22,6 +22,11 @@ namespace Marten.Testing
         private readonly Options _options 
             = new Options(dateFormat: DateTimeFormat.ISO8601, includeInherited:true);
 
+        public void ToJson(object document, TextWriter writer)
+        {
+            JSON.Serialize(document, writer, _options);
+        }
+
         public string ToJson(object document)
         {
             return JSON.Serialize(document, _options);
--- a/src/Marten/ISerializer.cs
+++ b/src/Marten/ISerializer.cs
@@ -7,6 +7,13 @@ namespace Marten
     public interface ISerializer
     {
         /// <summary>
+        /// Serialize the document object into <paramref name="writer"/>.
+        /// </summary>
+        /// <param name="document"></param>
+        /// <param name="writer"></param>
+        void ToJson(object document, TextWriter writer);
+
+        /// <summary>
         /// Serialize the document object into a JSON string
         /// </summary>
         /// <param name="document"></param>
--- a/src/Marten/StoreOptions.cs
+++ b/src/Marten/StoreOptions.cs
@@ -95,6 +95,11 @@ namespace Marten
         public int UpdateBatchSize { get; set; } = 500;
 
         /// <summary>
+        /// Configures the store to use char buffer pooling, greatly reducing allocations for serializing documents and events.
+        /// </summary>
+        public bool UseCharBufferPooling { get; set; } = false;
+
+        /// <summary>
         ///     Set the default Id strategy for the document mapping.
         /// </summary>
         public Func<IDocumentMapping, StoreOptions, IIdGeneration> DefaultIdStrategy { get; set; }
--- a/src/Marten.Testing/Schema/deriving_delete_sql_for_document_type.cs
+++ b/src/Marten.Testing/Schema/deriving_delete_sql_for_document_type.cs
@@ -13,7 +13,7 @@ namespace Marten.Testing.Schema
             var mapping = DocumentMapping.For<User>();
             mapping.DeleteStyle = DeleteStyle.Remove;
 
-            var resolver = new Resolver<User>(new TestsSerializer(), mapping);
+            var resolver = new Resolver<User>(new TestsSerializer(), mapping, false);
 
             resolver.DeleteByIdSql.ShouldBe("delete from public.mt_doc_user where id = ?");
         }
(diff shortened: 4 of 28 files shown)


src/Marten.Testing/performance_tuning.cs
```diff
@@ -22,6 +22,11 @@ public class JilSerializer : ISerializer
         private readonly Options _options 
             = new Options(dateFormat: DateTimeFormat.ISO8601, includeInherited:true);
 
+        public void ToJson(object document, TextWriter writer)
```
Scooletz: @jeremydmiller Can we change the seam? From semver perspective this is going to be a breaking change (major). We could provide a new interface that would be implemented by serializers. For now I assume it's ok to change this.

src/Marten/ISerializer.cs
```diff
@@ -7,6 +7,13 @@ namespace Marten
     public interface ISerializer
     {
         /// <summary>
+        /// Serialize the document object into <paramref name="writer"/>.
+        /// </summary>
+        /// <param name="document"></param>
+        /// <param name="writer"></param>
+        void ToJson(object document, TextWriter writer);
```
Scooletz: @jeremydmiller Can we change the seam? From semver perspective this is going to be a breaking change (major). We could provide a new interface that would be implemented by serializers. For now I assume it's ok to change this.

jeremydmiller: It's not a major change if it's additive. I say yes.

Scooletz: It broke `DocJsonBodyArgument` as it was selecting the method by name ;-)

src/Marten/StoreOptions.cs
```diff
@@ -95,6 +95,11 @@ public string DatabaseSchemaName
         public int UpdateBatchSize { get; set; } = 500;
 
         /// <summary>
+        /// Configures the store to use char buffer pooling, greatly reducing allocations for serializing documents and events.
+        /// </summary>
+        public bool UseCharBufferPooling { get; set; } = false;
```
jeffdoolittle: Should the default be to use char buffer pooling? I'm not a fan of having a `bool` default to true so perhaps consider flipping the name and logic around to something like `SuppressCharBufferPooling`.

Scooletz: In majority of cases I prefer to use `opt-in`,  then `opt-out` in a next version of a product. I'm not changing the behavior now as it's easier to test with no caching and focus on a few tests with caching. I'll add this as the point to this PR though.

Scooletz:
@jeremydmiller The initial phase of implementation is done in let's say 60%. I use only one test for testing the augmented `UpdateBatch` behavior. Could you give me some feedback on the approach I took? What's the convention for tests as well? I used only one `Fact` with a few assertions, which might be not good enough.

Also, I've just found `Resolver<T>.RegisterUpdate` method accepting json as a parameter. It looks that it isn't used at all. Additionally, the overload without JSON serializers the entity. This means that it's not needed. Am I missing something or is it useless?

src/Marten/StoreOptions.cs
jeffdoolittle: That sounds like a good plan

Scooletz:
I followed the way Marten handles dirty checking. This lead me to fetching documents which again uses strings. If we wanted to make no allocations that would require to change the fetch mechanism as well and use the char buffers in there.

☝️ This is a some kind of note to myself. I'll try to play a little bit with it and post my findings in here.

Scooletz:
Ok, I went through the API again. Basically, `IResolver<T>` and other parts (especially dirty checking) would require to replace `string json` with the new char writer. We could even read from the reader using `GetChars` instead of the current `GetString`.

I'm looking forward to hearing from you as it requires a lot of effort.

jeremydmiller:
@Scooletz I'm pulling this into master as much as anything to make rebase's not be a problem. The one and only change I made was to turn on character pooling by default. That exposed a single test failure, but that one was strictly a testing problem.

@Scooletz 

I say it looks good as is. The next thing I'd love to do is to add some profiling w/ Benchmark.Net to check the allocations with and without the pooling turned on. After that, I'd like to retrofit some tests that hit this w/ multiple threads pretty hard.

Correct me if I'm wrong, but this is only applied to writing documents in the `DocumentSession.SaveChanges() / UpdateBatch` mechanics, right? Doesn't look like it applies to bulk inserts yet, and we still need to do something completely different for reads?

Scooletz:
👍 for `Benchmarks.Net`. This could be a separate PR on it's own.

Yes, this is only applied to writing documents in the `DocumentSession.SaveChanges()` / `UpdateBatch` mechanics. Yes and yes :) For reads, this will involve replacing `string json` with something else. For events, I'd need to send a PR to `npgsql` to allow `ArraySegment<char>` in the following lines: https://github.com/npgsql/npgsql/blob/4ef74fa78cffbb4b1fdac00601d0ee5bff5e242b/src/Npgsql/TypeHandlers/TextHandler.cs#L269-L276

PR #657, opened 2017-01-18
Last modified extensions

jimgolfgti:
Enable the use of the built in `mt_last_modified` column without dropping to raw queries.

Also adds mapping/registry/attribute option to configure index on the storage column.


jeremydmiller:
Cool! I'm pulling it in now. It'll be in a 1.3 release soon-ish. I'll start a conversation in the Gitter room about what that's going to end up being.

PR #658, opened 2017-01-21
Deleted At Extensions

jimgolfgti:
DateTimeOffset.DateTime yields a DateTime with Unspecified kind, I think this could be the issue but I've failed to repro, tried changing the timezone of my postgres instance forward and backward and changing timezone of OS forward and backward.

This test was never supposed to have a DateTime passed, I think I only did it to verify the implicit cast would work and left it in by mistake.

See if TeamCity can passes this build.


jimgolfgti:
@jeremydmiller This build passed with just the first commit present.

I've now added querying on documents using soft deletes and configuration of a partial index on the mt_deleted_at column.

jeremydmiller:
Got it. This PR passed on TeamCity, so I'm pulling it in now.

PR #660, opened 2017-01-24
couple of fixes pertaining to run and publish docs

mysticmind:
1.Fixed to copy embed.js from ST tools to documentation/content which shows a 404 not found when accessing documentation
2.Changed paket restore to use paket.exe rather than paket,cmd which was not running the next steps within run-docs.cmd


jeremydmiller:
@mysticmind Oh man, thanks for doing this, but I was in the midst of completely converting the docs to use the next gen `dotnet stdocs` and I just deleted those two cmd files.

It's `rake docs` to run the documentation website, or `rake publish` to publish the docs to http://jasperfx.github.io/marten (if you have git permissions to do that).

PR #661, opened 2017-01-24
refactored AggregationLookupStrategy to clean up the code

migajek:
Although it worked perfectly, the existing solution was cumbersome as it required the usage of extension method.  The entire code was completely redundant in my opinion. 

To be honest I've spent around 15 minutes digging through the codebase trying to figure out why the documentation code snippet doesn't work until I found out I should use extension method instead of the instance method.


jeremydmiller:
Okay. I'll pull this one in when CI finishes up. Thank you.

migajek:
Great, thank you :)

Issue #516, opened 2016-09-07
Add projection rebuilding and async daemon running to the CLI support


jeremydmiller:
cc/ @jeffdoolittle @ericgreenmix 

Adding this for tracking purposes. I'm going to leave it up to Jeff and Eric to define what this looks like based on what they need.

jeremydmiller:
This can be done at any time since it's in a separate nuget. I'm pulling off the milestone.

jeremydmiller:
Some of this is done, but I'm not sure it's doc'd and released. Make it so in the 1.5 timeframe

jeremydmiller:
Nobody's asked for it by now, and it's easy to roll your own. I'm going to close this as "not gonna do it"

PR #667, opened 2017-02-01
GH-516 initial projection rebuild and async daemon running in Marten.CommandLine

migajek:
I've re-organized files layout to make it more readable, hope it's ok

this change introduces two new commands, 

* projection inline|async - which rebuilds either inline or async projections
* rundaemon which runs the async projection daemon

to be honest, I'm not fully convinced to using Daemon for rebuilding inline projections. It's very tempting because all the paging etc is already there, and therefore I've used it, at least for now.
On the other hand, I'm afraid of possible side effects, the tracking information being stored for inline projections - to name one. 
Maybe it'd be better to use Daemon (ProjectionTracks, in fact) in "no-tracking" mode, or even better - extract the paged fetching/applying out of the ProjectionTrack ?

### DIFF
--- a/src/Marten.CommandLine/Commands/ApplyCommand.cs
+++ b/src/Marten.CommandLine/Commands/ApplyCommand.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using Oakton;
 
-namespace Marten.CommandLine
+namespace Marten.CommandLine.Commands
 {
     [Description("Applies all outstanding changes to the database based on the current configuration")]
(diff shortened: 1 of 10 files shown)


jeremydmiller:
@migajek Sorry I've been slow on this one. I'm pulling this one in now. I do think we'd come back with a `projection --view [view type name]` so you can cherry pick one view type at a time. Or it could be `--views [name1], [name2]` I suppose.

PR #672, opened 2017-02-08
Fix a LANGUAGE sample formatting

miroslavpopovic:
Just a fix for a small issue with `LANGUAGE` sample not being correctly formatted in markdown.


jeremydmiller:
Thank you.

PR #675, opened 2017-02-10
Allow local documentsessionlisteners for a single session.

wastaz:
As you said in Gitter, it probably wouldn't be too hard to add `DocumentSessionListeners` to the `SessionOptions`, and it wasnt ;)

I've made a quite simple implementation, so if you have both global and local `DocumentSessionListeners` then the global listeners always run before the local ones. I think this should be good enough for most usages and if something more advanced is needed then that can be done later if someone actually has a use case for it :)


jeremydmiller:
Got it w/ a rebase. Thanks @wastaz!

Issue #687, opened 2017-02-21
FetchStreamState does not ensure basic event stream storage existance


vinicola:
Calling **session.Events.FetchStreamState(...)** on an empty database (before doing any other calls, such as Append and/or StartStream) will make marten/npgsql throw an exception as it tries to execute a query against the non existing table **mt_streams**.

I would expect the call to return null if the stream does not exist.

vinicola:
Failing test

```
public class fetching_stream_state_on_an_empty_database : DocumentSessionFixture<NulloIdentityMap>
    {
        [Fact]
        public void should_return_null_state()
        {
            theSession.DocumentStore.Advanced.Clean.CompletelyRemoveAll();

            var state = theSession.Events.FetchStreamState(Guid.NewGuid());

            state.ShouldBe(null);
        }
    }
```

wastaz:
Not only `FetchStreamState`, I have the same problem with `AggregateStreamAsync` in #681 (although for some reason Jeremy couldnt reproduce it, I still cant for my life figure out why). From looking at the code in `EventStore` it seems only the `StartStream` and `Append` methods causes the tables to be created. I'd expect that calling any method at all in `EventStore` would create these tables.

wastaz:
I made another PR that adds that fix for all public methods in `EventStore` which should fix the issue for all methods and not just `FetchStreamState`.

jeremydmiller:
@vinicola I'm going to take in @wastaz 's PR that covers this one and several others. Look for it in 1.4.1 in a bit.

vinicola:
That's even better, thanks!

PR #688, opened 2017-02-21
Fix for issue #687

vinicola:
Quick fix for issue #687, please review and let me know if it looks ok. If not, that's fine. Let me know if something is missing or if anything else is out of order.

Thanks!


jeremydmiller:
I took in the other patch. Thank you for doing this though.

PR #701, opened 2017-03-14
Add ability to invoke transform across all docs

jrios:
## Proposed Changes

Add the ability to take a TransformFunction and generate a SQL file that uses PLV8 to execute the transform across all documents within a table.

## Why

The ability to run transforms using the currently built in functionality is great for development time. Generating a SQL file to handle the transform across a table of documents allows people to run the SQL script outside of any code changes. This can help in changing data types in a mapped .NET type. It also affords more capability of automating migrations using something like [Sqitch](http://sqitch.org/) or [Roundhouse](https://github.com/chucknorris/roundhouse).

## How

The TransformFunction class already has the ability to take a JavaScript file and generate a PLV8 sql script. A new method was added to take an individual transform and generate a new transform execution script that does the following:

1. create the transform definition using the ``TransformFunction.GenerateFunction`` method
2. select all the documents in the provided table
3. create an execute transform function
    - loop over all the documents found
        - execute the transform on the document
        - update the document in the table
4. Optionally perform the execute transform function

Here's what the generated sql looks like:

```sql

CREATE OR REPLACE FUNCTION test.mt_transform_get_fullname(doc JSONB) RETURNS JSONB AS $$

  var module = {export: {}};

  // SAMPLE: get_fullname.js
module.exports = function (doc) {
    return {fullname: doc.FirstName + ' ' + doc.LastName};
}
// ENDSAMPLE

  var func = module.exports;

  return func(doc);

$$ LANGUAGE plv8 IMMUTABLE STRICT;

CREATE OR REPLACE FUNCTION test.execute_transform_mt_transform_get_fullname()

RETURNS VOID as $$

  var docs = plv8.execute('select id, data from test.test_table');
  docs.forEach(function(doc) {
    var transformedDoc = plv8.execute('SELECT test.mt_transform_get_fullname($1)', doc.data);
    plv8.execute('UPDATE test.test_table SET data = $1 WHERE id = $2', [transformedDoc[0]["mt_transform_get_fullname"], doc.id]);
  });

$$ LANGUAGE PLV8 IMMUTABLE STRICT;
PERFORM test.execute_transform_mt_transform_get_fullname();
```

There is also a new method that will generate that sql and pump it into a sql file using the existing SchemaPatch functionality.

### DIFF
--- a/src/Marten.Testing/Transforms/TransformFunctionTests.cs
+++ b/src/Marten.Testing/Transforms/TransformFunctionTests.cs
@@ -179,6 +179,69 @@ namespace Marten.Testing.Transforms
             patch.RollbackDDL.ShouldNotContain(func.Function.QualifiedName);
         }
 
+
+
+        [Fact]
+        public void generates_execute_transform_function_file()
+        {
+            var func = TransformFunction.ForFile(new StoreOptions(), _getFullnameJs);
+            var tableName = "mt_test_table";
+            var patch = new SchemaPatch(new DdlRules());
+
+            var directory = AppContext.BaseDirectory.AppendPath("bin", "transforms");
+
+            var fileSystem = new FileSystem();
+            fileSystem.DeleteDirectory(directory);
+            fileSystem.CreateDirectory(directory);
+
+            var fileName = Path.GetFileNameWithoutExtension(_getFullnameJs);
+
+            var filePath = directory.AppendPath($"execute_transform_{fileName}.sql");
+
+            func.WritePatchForAllDocuments(patch, tableName, filePath);
+
+            fileSystem.FileExists(directory.AppendPath($"execute_transform_{fileName}.sql"));
+        }
+
+        [Fact]
+        public void generated_script_includes_function_body()
+        {
+            var func = getTransformationFunction();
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@jrios I like the general idea, but can we pull this off w/ two functions, or a function and a straight up sql command:

1. A PLV8 transformation function
1. A 2nd command that does a `set data = function(data)` update? 

I'm a tad concerned about the performance by doing the looping within plv8

jrios:
yeah, the looping isn't great. There's probably a better way to approach the update part of this because it will chug on large data sets.

jrios:
What do you think about approaching the updates with something like this? [See number 7](http://engineering.tilt.com/7-postgresql-data-migration-hacks/ )

jeremydmiller:
I can buy that. Do you want something that generates that from a call like:

```
var script = store.Advanced.WriteTransformScriptFor<TDocument>(string transformName);
```

We could have that script contain both the plv8 function to do the transformation & a separate func or block that did the batch migration recipe from that blog post.

jrios:
I like putting it into the ``AdvancedOptions`` class. I'm going to see what I can do to get the batch updating working.

jeremydmiller:
Guessing this one never gets done.

PR #703, opened 2017-03-15
Updated documentation regarding Marten assigning an identity value only in case it has no value

mysticmind:


mysticmind:
I am closing this PR since the commits show several changes (looks polluted) even though I had committed just one change pertaining to doco, will try and send another PR.

Issue #629, opened 2016-11-25
Patch Append - possible to treat list as a set?


cocowalla:
The [docs](http://jasperfx.github.io/marten/documentation/documents/advanced/patch_api/) explain how you can patch a document to append an element to a child array.

Is there some way to treat the child array as a set, such that the element will only be added if it's not already there? The aim is to prevent duplicates being added to the array.

jeremydmiller:
@cocowalla Not today, but that wouldn't be a big task to add that. I'd make that an all new patching operation to make the behavior more apparent, and to not break existing code.

The patching is actually done w/ Javascript running against the JSON data, so it's pretty easy to modify.

jamesfarrer:
@jeremydmiller  this also looks a simple issue. what are you preferred names (I was thinking it should cover both append and insert?)

jeremydmiller:
@jamesfarrer I honestly don't know. @cocowalla Do you have any suggestions for what would make sense for that?

cocowalla:
How about `SetAdd`?

jamesfarrer:
Better to keep Append & Insert in the command imo. Two options :-

`AppendAsSet`, `InsertAsSet` or `AppendIfNotExists` `InsertIfNotExists`

jeremydmiller:
I like `AppendIfNotExists` & `InsertIfNotExists`

jeremydmiller:
@jamesfarrer If I end up pushing the 1.5 nuget before this gets in, I can just turn around and do this one in 1.6

jeremydmiller:
@jamesfarrer Reminded me that this one's done;)

PR #714, opened 2017-03-27
Add AppendIfNotExists and InsertIfNotExists to patching API

jamesfarrer:
Resolves issue #629 

Add two new methods and associated JS functions
AppendIfNotExists
InsertIfNotExists

The premise behind both methods is to only apply the patch if the
element within the collection does not already exist.

In order to check for complex types, JSON.stringify is used, which does
mean the order of elements is important. Assume this is OK for now.

### DIFF
--- a/documentation/documentation/documents/advanced/patch_api.md
+++ b/documentation/documentation/documents/advanced/patch_api.md
@@ -60,7 +60,7 @@ To increment a persisted value in the persisted document, use this operation:
 
 <[sample:increment_for_int]>
 
-By default, the `Increment()` operation will add 1 to the existing value. You can optionally override the increment:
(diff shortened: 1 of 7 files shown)


jeremydmiller:
@jamesfarrer Hey, the end to end coverage looks good, but would you actually wanna stick some JS tests in for the transformation?

jamesfarrer:
@jeremydmiller ah, I missed the separate JS tests. Sure I will add some JS tests

jamesfarrer:
@jeremydmiller all done hopefully

jeremydmiller:
@jamesfarrer Bueno, I'm pulling this in now. It'll be in 1.5 some time this week

Issue #715, opened 2017-03-27
AsyncProjectionDaemon might "miss" the first event


wastaz:
Ive managed to make `AsyncProjectionDaemon` "skip" the first event in my eventstore. This seems to be related to the fact that I did a `DELETE FROM mt_events WHERE seq_id < 4000` to remove a lot of old crap from my dev db that was causing problems for me. 

At this point the `AsyncProjectionDaemon` started skipping the first event in my eventstore, which was highly confusing. I managed to track down the reason why in the source code and there's an incoming PR with a one-line fix.

jeremydmiller:
Fixed by #716

PR #716, opened 2017-03-27
Fix problem where sometimes the asyncprojectiondaemon might "miss" an event

wastaz:
Fix for GH-715.


jeremydmiller:
Pfft, good catch. I'm pulling this in now and it'll be in Marten 1.5 hopefully this week

PR #728, opened 2017-04-10
Added string unique ID Generator (WIP)

mdissel:
Adding a unique string ID generator, copied from https://firebase.googleblog.com/2015/02/the-2120-ways-to-ensure-unique_68.html and https://gist.github.com/kiliman/ca1d9f4135078a6b24c5005113bbeea4
Thanks to @kiliman

### DIFF
--- a/src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
+++ b/src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
@@ -45,14 +45,16 @@ public class FirebasePushIDGenerator
     // Modeled after base64 web-safe chars, but ordered by ASCII.
     const string PUSH_CHARS = "-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz";
 
-    // Timestamp of last push, used to prevent local collisions if you push twice in one ms.
-    private static long lastPushTime = 0;
+		// Timestamp of last push, used to prevent local collisions if you push twice in one ms.
+		[ThreadStatic]
+		private static long lastPushTime = 0;
 
-    // We generate 72-bits of randomness which get turned into 12 characters and appended to the
-    // timestamp to prevent collisions with other clients.  We store the last characters we
-    // generated because in the event of a collision, we'll use those same characters except
-    // "incremented" by one.
-    private static char[] lastRandChars = new char[12];
+		// We generate 72-bits of randomness which get turned into 12 characters and appended to the
+		// timestamp to prevent collisions with other clients.  We store the last characters we
+		// generated because in the event of a collision, we'll use those same characters except
+		// "incremented" by one.
+		[ThreadStatic]
+		private static char[] lastRandChars = new char[12];
 
     // Random number generator
     private static Random rng = new Random();
@@ -84,12 +86,10 @@ public class FirebasePushIDGenerator
         else
         {
 					// If the timestamp hasn't changed since last push, use the same random number, except incremented by 1.
-					lock (lastRandChars) {
-						for (var i = 11; i >= 0 && lastRandChars[i] == 63; i--) {
-							lastRandChars[i] = (char)0;
-						}
-						lastRandChars[11]++;
+					for (var i = 11; i >= 0 && lastRandChars[i] == 63; i--) {
+						lastRandChars[i] = (char)0;
 					}
+					lastRandChars[11]++;
         }
 			var sb = new System.Text.StringBuilder();
         for (var i = 0; i < 12; i++)
--- a/src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
+++ b/src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
@@ -1,0 +1,114 @@
+﻿using System;
+using System.Collections.Generic;
+using Baseline;
+
+namespace Marten.Schema.Identity
+{
+    /// <summary>
+    /// StringUniqueIdGeneration Generation. More info https://firebase.googleblog.com/2015/02/the-2120-ways-to-ensure-unique_68.html
+		/// Copied from Gist https://gist.github.com/kiliman/ca1d9f4135078a6b24c5005113bbeea4
+    /// </summary>
+    public class StringUniqueIdGeneration : IIdGeneration, IIdGenerator<string>
+    {
+        public IEnumerable<Type> KeyTypes { get; } = new[] {typeof(string)};
+
+
+        public IIdGenerator<T> Build<T>(IDocumentSchema schema)
+        {
+            return this.As<IIdGenerator<T>>();
+        }
+
+        public string Assign(string existing, out bool assigned)
+        {
+            if (existing.IsEmpty())
+            {
+                assigned = true;
+								return FirebasePushIDGenerator.GeneratePushID();
+            }
+            assigned = false;
+            return existing;
+        }
+    }
+
+public class FirebasePushIDGenerator
+{
+    /**
+     * Fancy ID generator that creates 20-character string identifiers with the following properties:
+     *
+     * 1. They're based on timestamp so that they sort *after* any existing ids.
+     * 2. They contain 72-bits of random data after the timestamp so that IDs won't collide with other clients' IDs.
+     * 3. They sort *lexicographically* (so the timestamp is converted to characters that will sort properly).
+     * 4. They're monotonically increasing.  Even if you generate more than one in the same timestamp, the
+     *    latter ones will sort after the former ones.  We do this by using the previous random bits
+     *    but "incrementing" them by 1 (only in the case of a timestamp collision).
+     */
+    // Modeled after base64 web-safe chars, but ordered by ASCII.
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Cool, but could you add a test that uses this with an actual document and add some docs for the usage too?

src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
```diff
@@ -0,0 +1,114 @@
...
+     * 3. They sort *lexicographically* (so the timestamp is converted to characters that will sort properly).
+     * 4. They're monotonically increasing.  Even if you generate more than one in the same timestamp, the
+     *    latter ones will sort after the former ones.  We do this by using the previous random bits
+     *    but "incrementing" them by 1 (only in the case of a timestamp collision).
+     */
+    // Modeled after base64 web-safe chars, but ordered by ASCII.
+    const string PUSH_CHARS = "-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz";
+
+    // Timestamp of last push, used to prevent local collisions if you push twice in one ms.
+    private static long lastPushTime = 0;
```
mcarter101: This is static, so may need to add concurrency checks for server-side use.

src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
```diff
@@ -0,0 +1,114 @@
...
+    // "incremented" by one.
+    private static char[] lastRandChars = new char[12];
+
+    // Random number generator
+    private static Random rng = new Random();
+    
+    public static string GeneratePushID()
+    {
+        var now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+        Console.WriteLine("now=" + now);
```
mcarter101: Probably want to remove this.

src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
```diff
@@ -0,0 +1,114 @@
...
+            // If the timestamp hasn't changed since last push, use the same random number, except incremented by 1.
+            for (var i = 11; i >= 0 && lastRandChars[i] == 63; i--)
+            {
+                lastRandChars[i] = (char)0;
+            }
+            lastRandChars[11]++;
+        }
+        for (var i = 0; i < 12; i++)
+        {
+            id += PUSH_CHARS[lastRandChars[i]];
```
mcarter101: Should you convert this to a StringBuilder?

mcarter101:
Thanks for looking at this code. It's pretty much a straight port of the original JavaScript implementation, so there might be some things to make it more idiomatic C#. I've commented on a couple of issues. I'm especially interested in the static lastPushTime. For the most part, the push ID would be generated client-side where thread concurrency isn't much of an issue. However, I could see potential problems on the server if 2 threads (requests) tried to generate an ID in the same millisecond, then you could potentially have duplicate IDs. Perhaps storing the static data in TLS would suffice.

kiliman:
Ugh... just noticed that I was logged in with my work account.

src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
mdissel: of course..

src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
mdissel: or introduce a lock around the if statement if there's a conflict . What do you think?
```
// If the timestamp hasn't changed since last push, use the same random number, except incremented by 1.
lock (lastRandChars) {
	for (var i = 11; i >= 0 && lastRandChars[i] == 63; i--) {
		lastRandChars[i] = (char)0;
	}
	lastRandChars[11]++;
}
```

mdissel:
Added a test to store documents using the generator.

mdissel:
Should we change the default string generator into this one? if there's no value assign a unique string value else use the existing value..

src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
mdissel: Hmm, this will not catch concurrent requests...

kiliman:
Just to note in case it wasn't obvious, but these IDs are case-sensitive, as it encodes the data using both upper and lower-case letters. That's fine if you were to use it for JSON property names (which is primary use-case for Firebase). I know that PostgreSQL is case-sensitive by default, but could be an issue if somebody changed the collation, especially for the ID column.

src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
kiliman: Instead of locking, wouldn't it be better to simply add `[ThreadStatic]` attribute to the 2 static variables `lastPushTime` and `lastRandChars`?

The code generates random chars on each request, so if 2 separate threads request an ID at the same instant, the IDs will still be unique because each thread will have different random chars. It's only if the *same* thread asks for IDs in same millisecond that the random chars are reused and incremented by 1, which this code handles.

mdissel: I updated the code according to your suggestion.. Any idea how to unit test this behaviour?

mdissel:
@kiliman Just hitted the upper limit when adding 500 objects using StoreObjects resulting in "same Id already added to the session."

src/Marten/Schema/Identity/StringUniqueIdGeneration.cs
kiliman: The part that makes this difficult to unit test, especially across threads is the call to `DateTimeOffset.UtcNow`, since it is no longer deterministic.

You could add the following static property to the class that enables you to swap out the time function:

	// Func to replace current time getter useful for unit testing
	private static Func<long> _getUnixTimeMillisecondsFunc = new Func<long>(() => DateTimeOffset.UtcNow.ToUnixTimeMilliseconds());
	public static Func<long> GetUnixTimeMilliseconds
	{
		get { return _getUnixTimeMillisecondsFunc; }
		set { _getUnixTimeMillisecondsFunc = value; }
	}

Then update the call to use it:

	var now = GetUnixTimeMilliseconds();

In your unit test, you can then hard code a specific time that you can then use to determine the proper IDs are generated.

	FirebasePushIDGenerator.GetUnixTimeMilliseconds = new Func<long>(() => 1491859478805);

Once set, you should be able to generate multiple IDs and it should differ only by the last bit.

For testing multi-threading, you can set the specific time, then create 2 threads and generate the ID. It should have the same timestamp portion (first 8 chars), and the last 12 should be random per thread.

Hope this helps.

kiliman:
There's an issue with how it increments the last bit, especially if the previous digit is _ (largest value before rolls over to next digit). I'm looking at the code now.

kiliman:
Change the code to this and it will increment properly.

	// If the timestamp hasn't changed since last push, use the same random number, except incremented by 1.
	int i;
	for (i = 11; i >= 0 && lastRandChars[i] == 63; i--)
	{
	    lastRandChars[i] = (char)0;
	}
	lastRandChars[i]++;

mdissel:
Thanks @kiliman ! Code is updated

kiliman:
Here are some changes that I think would make this more testable.

First, instead of the Func property I suggested in an earlier comment, simply add a default parameter to pass in the time to use.

    public static string GeneratePushID(long unixTimeMilliseconds = 0L)
    {
        var now = unixTimeMilliseconds == 0L 
            ? DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() 
            : unixTimeMilliseconds;

This way, your unit tests can easily test for duplicate time handling, etc.

The other change would be to add a `Reset` method to control the static state as well as optional random number seed. Again this makes it simpler to unit test as you can then generate known values.

    public static void Reset(int seed = 0)
    {
        if (seed != 0)
        {
            rng = new Random(seed);
        }
        lastPushTime = 0L;
    }

So now you can write a test like this:

    FirebasePushIDGenerator.Reset(12345678);
    var id1 = FirebasePushIDGenerator.GeneratePushID(1491859478805L);
    var id2 = FirebasePushIDGenerator.GeneratePushID(1491859478805L);
    var id3 = FirebasePushIDGenerator.GeneratePushID(1491859478805L);

    id1.ShouldBe("-KhOkGZKQ_PEkTbWW_9a");
    id2.ShouldBe("-KhOkGZKQ_PEkTbWW_9b");
    id3.ShouldBe("-KhOkGZKQ_PEkTbWW_9c");

And for multi-thread tests, you can use the same time value and different seeds for each thread (call `Reset(seed)` in each thread) to verify that you'll get a unique ID even if multiple threads create an ID in same millisecond.

Hope this helps!

jeremydmiller:
Hey guys, is there any update on this one? It's going to need some serious rebasing if you want this in for 2.0.

mdissel:
I've switched back to incrementing ID's, so for me the need is not there anymore.
I've got no time to make the adjustments in the tests.. I'm fine with closing the pull request

jeremydmiller:
@mdissel kk

Issue #460, opened 2016-07-19
Add support for including child collection objects


kamranayub:
For post-1.0, add support for allowing `Include` on child collection objects:

```
var items = new List<Item>();
session.Query<User>().Include(u => u.Collection.Select(c => c.ItemId), items);
```

jokokko:
Maybe something like
```sql
SELECT * FROM public.mt_doc_issuelist as il
INNER JOIN public.mt_doc_issue i on (il.Data->'Issues') @> to_jsonb(t.Id)
-- or from Marten object serialization (in this case yielding equal execution plans)
SELECT * FROM public.mt_doc_issuelist as il
INNER JOIN public.mt_doc_issue i on (il.Data->'Issues') @> (t.Data->'Id')
```
```C#
public class IssueList
{
	public List<Guid> Issues { get; set; }
	...
}
query.Query<IssueList>().Include<Issue>(x => x.Issues, i => issuesFromDb.Add(i))
```
By quick glance, the current implementation of IncludeJoin does not seem to favour this, though.

jeremydmiller:
Maybe this would help do it in one call: https://tech.npras.in/postgres-recursive-query/

jokokko:
Well, I hope to take a look at this in the near future. But just by gut feeling & past memory, I can see #834 relating. So addressing how LINQ expressions are evaluated, exposed and translated in the pipeline might provide better & more generalizable benefits overall, instead of attacking the problems case-by-case with a very spefic constructs built around the existing pipeline.

jeremydmiller:
@jokokko Are you up for helping out on this if we really do take it on? Gonna put it off until I clear up a bit and get over the flu first though;-)

jokokko:
@jeremydmiller Yeah I hope to help. Overall, I have to go through code to identify scenarios with dance around queries that the current LINQ provider cannot translate.

jeremydmiller:
Believe it or not we're finally working on this again. This time I want to consider using Common Table Expressions in Postgresql to pull off the fetching

jeremydmiller:
Closed by https://github.com/JasperFx/marten/commit/c0a86fdb6fb61251aea4754afda89345b90b5cbd

PR #740, opened 2017-05-02
A way to handle #460 using the @> operator. Does not use the SqlLocat…

jokokko:
…or from IField as customizing that in JsonLocatorField would break SelectMany implementations :|

### DIFF
--- a/src/Marten.Testing/Services/Includes/end_to_end_query_with_include_collections_Tests.cs
+++ b/src/Marten.Testing/Services/Includes/end_to_end_query_with_include_collections_Tests.cs
@@ -1,0 +1,71 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Marten.Services;
+using Marten.Testing.Documents;
+using Shouldly;
+using Xunit;
+
+namespace Marten.Testing.Services.Includes
+{
+    public class end_to_end_query_with_include_collections_Tests : DocumentSessionFixture<IdentityMap>
+    {
+        public class IssueList
+        {
+            public IssueList()
+            {
+                Id = Guid.NewGuid();
+            }
+
+            public Guid Id { get; set; }
+            public List<Guid> Issues { get; set; }
+        }        
+
+        [Fact]
+        public void simple_include_for_a_collection()
+        {
+            var issues = Enumerable.Range(0, 3).Select(x => new Issue {Title = $"Issue {x}"}).ToList();
+            var issuelisting = new IssueList { Issues = new List<Guid>(issues.Select(x => x.Id)) };
+         
+            theSession.StoreObjects(issues);
+            theSession.Store<object>(issuelisting);
+            theSession.SaveChanges();
+
+            using (var query = theStore.QuerySession())
+            {
+                var issuesFromDb = new List<Issue>();
+
+                var listingFromDb = query.Query<IssueList>()
+                    .Include<Issue>(x => x.Issues, i => issuesFromDb.Add(i))
+                    .ToList();
+
+                issuesFromDb.Count.ShouldBe(3);
+                listingFromDb.Count.ShouldBe(3);
+            }
+        }
+
+        // TODO
+        //[Fact]
+        //public void multiple_include_for_a_collection()
+        //{
+        //    var issues = Enumerable.Range(0, 3).Select(x => new Issue { Title = $"Issue {x}" });
+        //    var issues2 = Enumerable.Range(3, 3).Select(x => new Issue { Title = $"Issue {x}" });
+
+        //    var issuelisting = new IssueList { Issues = new List<Guid>(issues.Select(x => x.Id)) };
+        //    var issuelisting2 = new IssueList { Issues = new List<Guid>(issues2.Select(x => x.Id)) };
+
+        //    theSession.Store<object>(issues, issues2, issuelisting, issuelisting2);
+        //    theSession.SaveChanges();            
+
+        //    using (var query = theStore.QuerySession())
+        //    {                
+        //        var issuesFromDb = new Dictionary<Guid, List<Issue>>();
+        //        var listingFromDb = query.Query<IssueList>()
+        //            .Include<List<Issue>, Guid>(x => x.Issues, issuesFromDb)
+        //            .ToList();
+
+        //        issuesFromDb.Count.ShouldBe(3);
+        //    }
+        //}
+    }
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@jokokko I'd be thrilled if that's all the changes required, but at a minimum it needs quite a bit more testing. We need to check that it isn't spitting out multiple references to the same document, either the primary document or the included document. We also need to test it in combination w/ the QueryStatistics and probably within a compiled query too.

jokokko:
@jeremydmiller Indeed, cannot but agree with every statement. But I wanted to present the idea of going with the @> jsonb operator. E.g. currently the JsonLocatorField makes assumptions that make this impractical, or at the very least require touching a much larger (public) surface area than just my few lines on IncludeJoin. Having these concerns addressed before 2.0 is pushed out could be beneficial.

jeremydmiller:
@jokokko "the JsonLocatorField makes assumptions that make this impractical" -- you'll have to refresh my memory, but what do you mean? JsonLocatorField mostly just remembers the path to an element within the JSON document. 

Totally onboard w/ this being in 2.0.

jokokko:
@jeremydmiller As I understand it, it introduces an unnecessary cast. Say I create a GIN index on the IssueList.Issues. With the CAST from JsonLocatorField I get a Join Filter on IssueList.Issues and sequential scan on IssueList, whereby without the cast I get a bitmap heap scan on IssueList yielding massively lower cost. Still very much anecdotal, I'm by no means a PG expert. But if this is the case, then making the required modifications to JsonLocatorField would require touching existing public contracts, not just internal implementation of IncludeJoin (which I fear is the case and thus suitable target to attack in 2.0).

jokokko:
@jeremydmiller To add to the previous, some logs from my dev box, running PG 9.6 with the defaults, as modest as they are. 2.5k IssueList items, each linking (in the domain model sense of things, not DB foreign keys or somesuch) to 3 Issue Items. GIN index on `doc_issuelist.Data->'Issues'`

```sql
explain analyze verbose
select d.data, d.id, d.mt_version, issues.data, issues.id, issues.mt_version from
  public.mt_doc_issuelist as d INNER JOIN
  public.mt_doc_issue as issues ON d.data -> 'Issues' @> to_jsonb(issues.id)
```
```
Nested Loop  (cost=0.10..64170.25 rows=75000 width=403) (actual time=0.020..68.740 rows=15000 loops=1)
  Output: d.data, d.id, d.mt_version, issues.data, issues.id, issues.mt_version
  ->  Seq Scan on public.mt_doc_issue issues  (cost=0.00..605.00 rows=15000 width=179) (actual time=0.004..1.093 rows=15000 loops=1)
        Output: issues.id, issues.data, issues.mt_last_modified, issues.mt_version, issues.mt_dotnet_type
  ->  Bitmap Heap Scan on public.mt_doc_issuelist d  (cost=0.10..4.19 rows=5 width=224) (actual time=0.003..0.003 rows=1 loops=15000)
        Output: d.id, d.data, d.mt_last_modified, d.mt_version, d.mt_dotnet_type
        Recheck Cond: ((d.data -> 'Issues'::text) @> to_jsonb(issues.id))
        Heap Blocks: exact=15000
        ->  Bitmap Index Scan on idxinissues  (cost=0.00..0.10 rows=5 width=0) (actual time=0.002..0.002 rows=1 loops=15000)
              Index Cond: ((d.data -> 'Issues'::text) @> to_jsonb(issues.id))
Planning time: 0.066 ms
Execution time: 69.004 ms
```

Then with the cast produced from `JsonLocatorField`
```sql
explain analyze
select d.data, d.id, d.mt_version, issues.data, issues.id, issues.mt_version from
  public.mt_doc_issuelist as d INNER JOIN
  public.mt_doc_issue as issues ON CAST(d.data ->> 'Issues' as jsonb) @> to_jsonb(issues.id)
```

```
Nested Loop  (cost=0.00..469177.25 rows=18750 width=403) (actual time=0.020..40645.094 rows=7500 loops=1)
  Join Filter: (((d.data ->> 'Issues'::text))::jsonb @> to_jsonb(issues.id))
  Rows Removed by Join Filter: 18742500
  ->  Seq Scan on mt_doc_issue issues  (cost=0.00..303.00 rows=7500 width=179) (actual time=0.005..4.275 rows=7500 loops=1)
  ->  Materialize  (cost=0.00..130.50 rows=2500 width=224) (actual time=0.000..0.057 rows=2500 loops=7500)
        ->  Seq Scan on mt_doc_issuelist d  (cost=0.00..118.00 rows=2500 width=224) (actual time=0.004..0.326 rows=2500 loops=1)
Planning time: 0.384 ms
Execution time: 40646.347 ms
```

It's entirely possible that I'm missing something obvious, but it would seem that the (redundant) cast originating from `JsonLocatorField` is preventing the query planner from using the previously defined index and we go from index scans to sequential scans, in this scenario producing execution times of ~69 milliseconds and ~40 seconds respectively.

jeremydmiller:
@jokokko I say let's add a completely new, semantically different property to JsonLocatorField for something like `RawJsonLocator` or if it's that specific, `JsonCollectionJoinLocator`. Definitely make it be a completely new concept rather than try to change any of the existing logic.

jokokko:
Some convo about this in Gitter. I think eliminating some of the casting that now propagates to many, many areas from JsonLocatorField could produce great perf gains overall, especially when indices are used. I also gave an example of a case where I immediately see a change in behaviour: `CAST(d.data ->> 'Published' AS BOOLEAN) = TRUE)` vs. `d.data -> 'Published' = to_jsonb(TRUE))` wherewby the first case would produce a match if `Published` was serialized to `"true"`, whereas the latter would not. And to that I'd say that if somebody wanted to serialize their booleans as strings, it is up to them to then also do the matching as strings.

jeremydmiller:
@jokokko Do you know why we closed this one, or did I just make a mistake in Github?

jokokko:
@jeremydmiller Seems like a mistake.

jeremydmiller:
@jokokko I'm sorry, but I pulled this one in today and it's just too problematic. Right now it duplicates documents in the main list from the inner join duplication, and I think it's too far a bridge to brute force change all the selectors handle that, and I'm not enthusiastic about forcing it to always do a Distinct().

I want to go back to my original concept for this feature and just issue multiple commands in a batch to avoid that problem.

As for all the cast optimizations on, can you just raise them on #744.

PR #758, opened 2017-05-10
Linq query with escape character

KopIlya:
Added failing tests


jeremydmiller:
@KopIlya I pulled this in directly to avoid merge problems.

Issue #765, opened 2017-05-21
default index on DocumentTypeColumn


mdissel:
There's no index created on the field DocumentTypeColumn when hiearchies/subclasses are stored in the same table.

jeremydmiller:
Would you mind taking that one? Having it add an index automatically if it's hierarchical?

mdissel:
see #766

mdissel:
The other system fields like Deleted also don't have an index. Should I add an index to those fields?

jeremydmiller:
Deleted, yes, because that's queried so often. I don't think I'd add anything to anything beyond that.

mdissel:
done!

jeremydmiller:
It's in.

PR #766, opened 2017-05-22
Add index on column doc_type, see #765

mdissel:


jeremydmiller:
Hey, could you throw a unit test on `DocumentTable` for "not hierarchical, no index" and "hierarchical, so index"? Otherwise that looks good to me

mdissel:
But if there's no hiearchy the column docType is not created..

jeremydmiller:
@mdissel I just meant the negative case is all. I'm good with this and pulling it now.

Thanks for taking this on.

jeremydmiller:
@mdissel Couple issues with this one that I'm taking care of:

* There was already a separate facility to index the Deleted column, so I just removed the code in DeletedColumn
* The unit test wasn't written correctly and failed. Easy fix though.

mdissel:
Thanks.. i will run all the tests next time.. sorry!

Issue #772, opened 2017-05-24
Indexes that wrap columns via `Expression` never match existing DDL


jmbledsoe:
I have created an index on a duplicated column that indexes the lowercase value of the column, as follows:
```
For<Company>()
    .Duplicate(c => c.Name, configure: id =>
    {
        id.Expression = "lower(?)";
        id.IndexName = id.IndexName + "_lower";
    });
```

(I also added a set of `IMethodCallParser` implementations to use this index, but that's a different issue.)

My production application runs with `AutoCreateSchemaObjects` set to `AutoCreate.None`, since I have multiple servers all connecting to the same database and I don't want them competing to execute schema migrations. In our environment, we use a small console application to execute all Marten data migrations during each deployment. The application applies configured changes to the database and then verifies that they were applied successfully.

```
documentStore.Schema.ApplyAllConfiguredChangesToDatabase();
documentStore.Schema.AssertDatabaseMatchesConfiguration();
```

Immediately after applying configured changes, the call to `AssertDatabaseMatchesConfiguration` throws an exception indicating schema differences.

Working on a PR now, will update the issue once it's ready.

jeremydmiller:
Going at the canonicalize Sql function? That's been problematic from day one. I'll be looking for the PR

jmbledsoe:
Hadn't seen that yet. Was just going to consider `IndexDefinition.Expression` in the `Match` function but we'll see where it leads.

jeremydmiller:
Postgresql frequently reformats or changes the raw DDL we use to generate objects, so when we do the actual DDL vs. expected DDL, you've gotta coerce both sides before doing the comparison. That's where I'd expect the problem to be.

jeremydmiller:
Fixed by #774. Thanks!

PR #773, opened 2017-05-24
#772 Index with Expression Should Match DDL

jmbledsoe:
See #772 

I wasn't able to universally match DDL from the database with DDL generated by code, but I was able to change the `Match` method on `IndexDefinition` to return a positive match if the `Expression` property is added in such a way as to minimize/eliminate reformatting by Postgres. This PR handles my use case, and I believe others, as long as `Expression` values are set appropriately.

I also corrected the error message in the testing `ConnectionSource`, since the environment variable in the exception message and the actual environment variable were different.


jmbledsoe:
Resubmitted as #774 as a patch to 1.5

PR #776, opened 2017-05-26
Implementation of camelCase support for documents. (WIP)

jarroda:
I was selfishly hoping we could pull this off in the 1.5 branch without causing trouble in the 2.0 branch... But if you're against the idea of doing it here, I can update this PR to be against 2.0.


jarroda:
Moving this to 2.0 (master) branch.

Issue #710, opened 2017-03-22
Create database if not exist


KopIlya:
Hello. In EntityFramework we has parameter like databaseInitializer - CreateDatabaseIfNotExists. Can you add parameter like this in StoreOptions? I think it will be convenient for many people. Thanks.

jeremydmiller:
Closed by https://github.com/JasperFx/marten/pull/788

PR #788, opened 2017-06-06
Possible implementation for #710. Note, unit testing this requires ma…

jokokko:
…rten_testing_database cstring user to have privileges to create & drop db (CREATEDB) :|

If this looks okayish, then I can ofc create docs for it.


jeremydmiller:
I think it looks usable as is, I'm pulling this in. I'll probably end up coming in to disable the tests in CI, but let's just see what happens in CI first

jeremydmiller:
It's in with some rebasing. Thanks!

bradyclifford:
@jeremydmiller what new version of Marten will this find its way into?

PR #810, opened 2017-06-28
Not a pretty or complete commit (due to the .NET SDK issues discussed…

jokokko:
… in Gitter).

Bugs: "clean" serializer in the default serializer has to respect casing (otherwise e.g. contains queries [ContainmentWhereFragment] will result in unexpected matches, which is dangerous). This needs to be fixed prior to CamelCasing being pushed to production.

API Changes: lift StorageFeatures to AdvancedOptions.StorageFeatures. Rationale: Mappings, amonst other things, can be accessed this way.

The rest: Start fixing test cases not to assume casing. Notes:
*a subset of tests goes outside of the shared setup -> test session specific customizations need to be applied in these cases too
*a subset of tests assumes string formatting. Some of these are contained within docs -> resort to extensions that adjust expectations
*harness: changed harness to check env variables for customizations -> Rake build script currently runs harness twice: default & CamelCase. (not reflected in build.sh atm)

While I'm not happy with this commit (due to reasons above), at least the bug with containtment needs to be addressed before 2.0 is released.

### DIFF
--- a/rakefile.rb
+++ b/rakefile.rb
@@ -12,7 +12,7 @@ BUILD_NUMBER = build_number
 
 task :ci => [:connection, :version, :default, :storyteller, 'pack']
 
-task :default => [:mocha, :test, :storyteller]
+task :default => [:mocha, :tests, :storyteller]
 
 desc "Prepares the working directory for a new build"
 task :clean do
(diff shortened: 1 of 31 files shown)


jokokko:
To add (based on some previous discussion): it's after all a damn good idea to run the whole harness against store customizations that propagate to query model execution, e.g. serialization, even if it results in reduntant test runs. Rationale through counterargument: saved 100 LOC and 100 ms on harness, lost 1 read from DB.

PR #822, opened 2017-07-24
fixed WhereClauseVisitor to handle constant booleans. Fixes #821

### LINKED ISSUE #821: Where doesn't work with PredicateBuilder
The common usage scenario for PredicateBuilder (http://www.albahari.com/nutshell/predicatebuilder.aspx) is to build the conditions as following:

```
      var predicate = PredicateBuilder.True<AssetProjection>();

            if (!request.NumberStartsWith.IsNullOrEmpty())
                predicate = predicate.And(x => x.Number.StartsWith(request.NumberStartsWith));
```

applying the predicate to the query `DocumentSession.Query<AssetProjection>().Where(predicate).ToList()` doesn't work, i.e. it behaves like there is no filtering at all.

I believe this is due to the way PredicateBuilder works, i.e. the built predicate is 
`True AND <x.Number.StartsWith>`

according to https://github.com/JasperFx/marten/issues/548 constant booleans are not supported.

first of all it'd be great to support that as this is quite common.

it's a shame Marten doesn't throw exception, I strongly believe it should yield loudly when the query is not supported, instead of behaving like everything was ok :)

migajek:
###  Please be careful 

I don't really know what I'm doing :)

I did a quick analysis of the behavior and fixed what seemed to be the problem in my case. I'm not however sure I completely understand the original idea, so I can't tell if the previous implementation was a mistake or had some purpose to behave like it did.

I ran all the tests, some of them failed but they did failed on the latest origin sources as well, so I suppose these are not because of my change. More likely the lack of v8 language and some system encoding problems, I can't investigate it right now

### DIFF
--- a/src/Marten.Testing/Linq/using_const_boolean_in_where_clause_Tests.cs
+++ b/src/Marten.Testing/Linq/using_const_boolean_in_where_clause_Tests.cs
@@ -1,0 +1,49 @@
+﻿using System.Linq;
+using Marten.Services;
+using Shouldly;
+using Xunit;
+
+namespace Marten.Testing.Linq
+{
+    [ControlledQueryStoryteller]
+    public class using_const_boolean_in_where_clause_Tests : DocumentSessionFixture<NulloIdentityMap>
+    {
+        [Fact]
+        public void where_const_false()
+        {
+            var target1 = new Target { Number = 1, String = "Foo" };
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@migajek Got it, and thanks! Honestly, the Linq support for now falls into the category of "if it passes the tests, it's good enough"

PR #828, opened 2017-07-27
Add support for snake case naming convention

ejsmith:


jeremydmiller:
@ejsmith You've got some failing tests in this against the `JsonLocatorField`. Did something get missed?

ejsmith:
Looks like it. I will get it updated.

ejsmith:
Looks like there may be some flakey tests? These tests pass locally for me.

Issue #817, opened 2017-07-18
Passthrough SQL where clause


ejsmith:
How do you feel about a LINQ extension that lets you pass a raw where expression through to the SQL command? In my use case I have some other logic that is generating a where expression and I would like to combine that with other LINQ operators like Skip and Take. So I was thinking that I would add a IMethodCallParser to let me do that. Thoughts? Also, is that something you would be interested in a pull request for?

jeremydmiller:
That *shouldn't* be that crazy to do, and yep, I'd take a PR for that

ejsmith:
Think it would just be an overload of Where that takes a SQL string and parameters?

ejsmith:
It was super easy to get this working:

```csharp
var user = session.Query<User>()
  .FirstOrDefault(x => x.MatchesWhereFragment("first_name = ? and last_name = ?", "Han", "Solo"));
```

What do you think?

jeremydmiller:
Might even go for just `x.MatchesSql()` or something shorter, but yep, go for it

jeremydmiller:
This is done.

PR #830, opened 2017-07-27
Implement #817 to add support for raw SQL where clauses in LINQ

ejsmith:


src/Marten/Linq/MatchesSql/MatchesSqlExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+namespace Marten.Linq.MatchesSql
```
mdissel: Why in a different namespace?

ejsmith: It's following the pattern from [LastModifiedExtensions.cs](https://github.com/JasperFx/marten/blob/master/src/Marten/Linq/LastModified/LastModifiedExtensions.cs). Also, since it has to be an extension method on object, then I would want it to be opt in.

PR #835, opened 2017-07-31
POC for incrementing document versions

ejsmith:
This is just a proof of concept to see what all changes are required to implement a separate versioning strategy that uses incrementing integer document versions. My intention is to just use it to start a discussion.

It seems like the changes are pretty substantial and I'm not sure how we would go about turning the document versioning into a pluggable strategy. Thoughts @jeremydmiller?

This script can be used to migrate existing tables from uuid to bigint:

```sql
ALTER TABLE public.mt_doc_target
ALTER COLUMN mt_version DROP DEFAULT;

ALTER TABLE public.mt_doc_target
ALTER COLUMN mt_version TYPE bigint USING CAST(1 as bigint);

ALTER TABLE public.mt_doc_target
ALTER COLUMN mt_version SET DEFAULT 1;
```


mdissel:
My initial thought was, make it the version field type configurable and support for guid and incrementing integer values.

jeremydmiller:
@ejsmith What @mdissel said. I really, really don't want to force this change on every user and I was going to push for this even being document type by document type. 

To make it pluggable, you'd probably try to absorb the differences in types by "baking" the varying logic into `IDocumentStorage<T>`. For methods like `IDocumentSession.Store()`, use overloads. A big chunk of your effort here was changing tests that look for generated SQL, and we dodge all that effort by making this configurable so those tests wouldn't have to change.

mdissel:
I didn't know this feature, but apparently Postgres has a system column that can be used as a concurrency field, xmin
see https://github.com/npgsql/Npgsql.EntityFrameworkCore.PostgreSQL/issues/19
and https://www.postgresql.org/docs/current/static/ddl-system-columns.html

jeremydmiller:
Meh, would have been good to know about this before we did the versioning. Still could make it be yet another opt in versioning mode.

ejsmith:
Ok, couple questions:
- Would concurrency use a known set of implementations using a type enum like `TenancyStyle` or a strategy based pattern similar to `IIdGeneration`?
- Do you guys think it's worth the added complexity to the public API that would be required to make the versioning strategy pluggable? I'd like to have an incrementing version strategy, but I'm questioning if it's worth the complexity. This feels like something you just want to enable as an end user and then not think about it and definitely not have to configure it. I'm kind of leaning towards, if we don't want to change the existing behavior and the benefits of changing aren't worth forcing users to do a simple migration, then we should probably just leave it as is.
    - Benefits of incrementing integer version:
        - Can easily see how many times a document has been modified.
        - Less storage per row and in the app: uuid (16 bytes) vs bigint (8 bytes) Probably not a big deal considering the documents are stored as JSON anyway.

I know @jeremydmiller is big time against any breaking changes and the 2.0 ship has already sailed, so I'm leaning towards just dropping this unless we really think that pluggable version strategies would be beneficial.

Here is an initial set of API changes that I can think of:
- Change `IDocumentSession.Store` to use `object` data type for version.
- Change `IIdentityMap.Get` and `IIdentityMap.Store` to use `object` data type for versions.
- Change `VersionTracker.Store` and `VersionTracker.Version` to use `object` data type for versions.
- Change `DocumentMetadata.CurrentVersion` to `object` data type.
- Change `VersionArgument` to use `object` data type.
- Add ability to map version strategy and version field on `IDocumentMapping`
    - `_.Schema.For<Issue>().Version(i => i.Version).UseOptimisticConcurrency(true);`
    - Would we try and update the document value in the document before persisting it? Or let it be our of sync with the column value?
    - Maybe we would just recommend that you mark the version field on your model as `[JsonIgnore]`
    - Should optimistic concurrency be automatically enabled if you set the Version field? Seems reasonable to me.
    - When this field is mapped, it would automatically be used for `Store` operations as the current version. But how you would decide to ignore it and force overwrite?

jeremydmiller:
"I know @jeremydmiller is big time against any breaking changes" -- Jeremy is the person who's gonna get screamed at for causing so much churn to production databases that are probably working just fine as is. It's worth the work from me to make this configurable.

* I'd rather do overloads on `IDocumentStore` and maybe the internal `VersionTracker` rather than use `object`
* On `DocumentMetadata`, I think I'd take a page out of the event store "Guid or string identifier" option and just have two different props on `DocumentMetadata`, w/ maybe a new prop called "Revision" for the int.
* I'd make an `IntVersionArgument` instead of object and switch within `DocumentTable` based on that

* As for the last bullet points:

1. Yep
1. I say yes, because it'd make the mechanics quite a bit easier on the read side
1. I don't think that's going to be necessary or desirable IMO
1. I say yes
1. As of 2.0 you can create an `IDocumentSession` that turns off all concurrency checks explicitly so that you can do overwrites regardless of version

ejsmith:
Ugh, I should have used a numbered list instead of bullets. I'm trying to correlate your answers to my questions.

1. I think you are saying that yes, you do think that making the versioning strategy configurable is worth the added complexity (extra overloads and configuration options) in the client API and you want to move forward with this feature.
2. I don't see an answer to if the versioning strategy should be an open ended `IIdGeneration` style or closed (Guid and revision) `TenancyStyle` setting where there would just be the 2 options to pick from. It seems like it would be easier to implement the closed enum style and we would also know exactly which version parameter overloads are needed. Which are you thinking?
3. You want to use overloads for methods instead of using object where possible.
4. You want the optimistic concurrency option turned on automatically if the Version field on the doc is configured.
5. You want to try and keep the document values in sync with the version metadata if the version field is mapped on the document. This should be doable, but it might be tricky to rollback when the DB transaction fails.
6. I know at one point you mentioned that you were going to take a stab at this. Are you wanting to do this or would you like me to take a stab at it?

jeremydmiller:
@ejsmith:

1. Yes
1. I don't think it needs to be an open ended extension point and that's a bit more complexity
1. Yes
1. I say yes
1. I'm not seeing that as that big of a deal
1. Honestly, I'm not enthusiastic about any of this, but this is something I'd rather do since it hits so many internals

ejsmith:
Ok sounds good. You can close this then if you want.

jeremydmiller:
Just isn't gonna happen, sorry.

PR #841, opened 2017-08-06
Add proper handling of async in ApplyAsync method for ViewProjection

oskardudycz:
Currently ViewProjection even though it has `async Task IProjection.ApplyAsync(IDocumentSession session, EventPage page, CancellationToken token)` method it's not doing any async code there - it doesn't await for applyProjections so it's not possible to use async code in the persisting event projection method.

I know that there is AsyncDaemon, that helps with some of the async cases, but it's complext in usage and sometimes there is a need to use the code that has only async API (like in my commercial project, most of the code doesn't support synchronous access). Imho it would be good from the performance  and consistency perspective to allow to use async code in the `ProjectEvent` persist method.

In my PR, I changed the `Handler` definition to always return Task. For the non async code it just returns `Task.Completed` (by convention found in other Marten classes). That gives the unified usage for both synchronous and asynchronous projection handling. If the async code is used in synchronous `applyProjections` then the `.Wait` method is being used.

I added also the tests, that mimics the original tests. I'm not 100% sure that it's the best way to do doubled code with that. The other possibility is to add additional 3 events, that we could use in `PersistViewProjection` class and check in same scenarios. But I wasn't sure if that won't break the convention of using same events in documentation and in the tests.

### DIFF
--- a/src/Marten.Testing/Events/Projections/custom_async_transformation_of_events.cs
+++ b/src/Marten.Testing/Events/Projections/custom_async_transformation_of_events.cs
@@ -1,0 +1,335 @@
+﻿using System;
+using System.Collections.Generic;
+using Marten.Events.Projections;
+using Marten.Services;
+using Shouldly;
+using Xunit;
+using System.Threading.Tasks;
+
+namespace Marten.Testing.Events.Projections
+{
+    public class project_events_async_from_multiple_streams_into_view : DocumentSessionFixture<IdentityMap>
+    {
+        static readonly Guid streamId = Guid.NewGuid();
+        static readonly Guid streamId2 = Guid.NewGuid();
+
+        QuestStarted started = new QuestStarted { Id = streamId, Name = "Find the Orb" };
+        QuestStarted started2 = new QuestStarted { Id = streamId2, Name = "Find the Orb 2.0" };
+        MonsterQuestsAdded monsterQuestsAdded = new MonsterQuestsAdded { QuestIds = new List<Guid> { streamId, streamId2 }, Name = "Dragon" };
+        MonsterQuestsRemoved monsterQuestsRemoved = new MonsterQuestsRemoved { QuestIds = new List<Guid> { streamId, streamId2 }, Name = "Dragon" };
+        QuestEnded ended = new QuestEnded { Id = streamId, Name = "Find the Orb" };
+        MembersJoined joined = new MembersJoined { QuestId = streamId, Day = 2, Location = "Faldor's Farm", Members = new[] { "Garion", "Polgara", "Belgarath" } };
+        MonsterSlayed slayed1 = new MonsterSlayed { QuestId = streamId, Name = "Troll" };
+        MonsterSlayed slayed2 = new MonsterSlayed { QuestId = streamId, Name = "Dragon" };
+        MonsterDestroyed destroyed = new MonsterDestroyed { QuestId = streamId, Name = "Troll" };
+        MembersDeparted departed = new MembersDeparted { QuestId = streamId, Day = 5, Location = "Sendaria", Members = new[] { "Silk", "Barak" } };
+        MembersJoined joined2 = new MembersJoined { QuestId = streamId, Day = 5, Location = "Sendaria", Members = new[] { "Silk", "Barak" } };
+
+        [Fact]
+        public void from_configuration()
+        {
+            StoreOptions(_ =>
+            {
+                _.AutoCreateSchemaObjects = AutoCreate.All;
+                _.Events.InlineProjections.AggregateStreamsWith<QuestParty>();
+                _.Events.ProjectView<PersistedView, Guid>()
+                    .ProjectEventAsync<QuestStarted>((view, @event) => { view.Events.Add(@event); return Task.CompletedTask; })
+                    .ProjectEventAsync<MembersJoined>(e => e.QuestId, (view, @event) => { view.Events.Add(@event); return Task.CompletedTask; })
+                    .ProjectEventAsync<MonsterSlayed>(e => e.QuestId, (view, @event) => { view.Events.Add(@event); return Task.CompletedTask; })
+                    .DeleteEvent<QuestEnded>()
+                    .DeleteEvent<MembersDeparted>(e => e.QuestId)
+                    .DeleteEvent<MonsterDestroyed>((session, e) => session.Load<QuestParty>(e.QuestId).Id);
+            });
+
+            theSession.Events.StartStream<QuestParty>(streamId, started, joined);
+            theSession.SaveChanges();
+
+            theSession.Events.StartStream<Monster>(slayed1, slayed2);
+            theSession.SaveChanges();
+
+            theSession.Events.Append(streamId, joined2);
+            theSession.SaveChanges();
+
+            var document = theSession.Load<PersistedView>(streamId);
+            document.Events.Count.ShouldBe(5);
+            document.Events.ShouldHaveTheSameElementsAs(started, joined, slayed1, slayed2, joined2);
+
+            theSession.Events.Append(streamId, ended);
+            theSession.SaveChanges();
+            var nullDocument = theSession.Load<PersistedView>(streamId);
+            nullDocument.ShouldBeNull();
+
+            // Add document back to so we can delete it by selector
+            theSession.Events.Append(streamId, started);
+            theSession.SaveChanges();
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@ericgreenmix What do you think of this one?

@oskardudycz Oh man, we've gotta talk about this one first. It's a breaking API change, so it shouldn't technically go in until 3.0. We could finesse it by saying not many folks will be using ViewProjection yet.

oskardudycz:
@jeremydmiller well, imho the backward compatibility is kept. For the ProjectEvent it should work the same as it was before. Imho it's just making sure that ProjectAsync is really async. But yeah, I know what you meant in general. You wan't to be carefull and double check that it do no harm.

Maybe for the version 2 we could have less invasive change and add type like AsyncViewProjection that will contain suggested by me ProjectAsync methods?

I'm using ViewProjection for most of my read module Event Projections in my commercial project ;) But because it's not fully async then I had to use for my async methods that I use there sth. like `AsyncMethod().Result`, which is imho bad solution ;)

ericgreenmix:
@jeremydmiller @oskardudycz 

First, ignoring the code, I just want to add a word of caution on the need to have async aggregator methods. Ideally, these methods should just be taking the event and applying it to the model. If you are needing to make async requests to other systems, I would highly suggest you look at enriching your events to have the necessary data. Adding this data to your events also ensures that your model at a single point in time is always the same. If you were calling other systems during a rebuild, theoretically your data could change causing for unreliable state of projections after a rebuild.

As for the code, I would like to see this implemented without having to call `Wait` on `applyProjections`. You could do something like have two separate `Handler` delegates on the `EventHandler` and set the appropriate one based on if the user is calling the `Async` version or not. You might even be able to do away with having `ProjectEvent` and `ProjectEventAsync` be two separate methods and just set the appropriate `Handler` based on the return type of aggregator method provided.

oskardudycz:
@ericgreenmix I agree that we should avoid if possible using async code (especially those one that runs longer) in the ApplyProjection. But sometimes it's usefull and in fact it's already defined that we have `IProjection.ApplyAsync` in Marten, but ViewProjection implementation right now is not async at all.

I could add separate EventHandler types to show explicitly what type of synchronization are they for. But async handlers when running `applyProjectsions` would still need to have `.Wait()` method being called. If we won't call async handlers in applyProjections then it might lead to the developer confusion, because he should know on the ViewProjection definition if someone in the business code will be using `documentSession.SaveChanges` or  `documentSession.SaveChangesAsync `. Imho that would be odd when someone publish events and not all defined handlers would be called. That would lead to more serious, hard to find and debug bugs and it's more error prone.

jeremydmiller:
@oskardudycz If you need to support both sync and async, I think you'd be a lot better adding a quiet `return Task.CompletedTask;` around the sync code to be fully async rather than doing a `Task.Wait()` in the sync version. I think you're gonna need to enhance the ViewProjection stuff to allow users to register either sync or async transformation lambdas.

Putting it another way, we won't allow any `Task.Wait()` calls in the real code because of the risk of deadlocks in production.

oskardudycz:
I already added Task.CompletedTask for async applyProjections.

Ok, so what should happen when you defined async event handler and you're running synchronous applyProjections? Should such projection be ignored? If so - that could be hard to understand for the end user.

I understand that you wouldn't like to risk deadlocks, I'm also not happy with using wait. But what choice do we have when current api forces to have both sync and async code? Any ideas how I could fix it?

In general - why we have ApplyAsync if we don't want to use it asynchronously?

ericgreenmix:
@oskardudycz I see what you are saying. I don't think `Apply` vs. `ApplyAsync` should ignore any of the transformation methods, as that would be confusing and bad (as you said).

I don't have time at the moment to look deeper into the code, but there has to be a way to handle this without relying on `Task.Wait()` like Jeremy said. Since Marten supports net46, where Synchronization Contexts are still a thing, there will be the possibility of deadlocks when using it.

oskardudycz:
@ericgreenmix so when you find some spare time, I'd be grateful for recommendations.

I already was trying before sending pull request some alternatives for the .Wait() calls, but didn't find anything better than that. I will try again, get more deeper and find some other possibilities. 

Thanks Guys for a feedback!

oskardudycz:
Ok, I found 2 possibilities:
- https://stackoverflow.com/a/22629216 - Task.Run(AsynMethod).Wait - should be deadlock safe,
- https://stackoverflow.com/a/28307965 - using SynchronizationContext
What do you think?

jeremydmiller:
@oskardudycz It's just not necessary. Have it be an async when it runs in ViewProjection, but allow users to define sync or async delegates, just when it's a sync, wrap it in a different func that returns Task.CompletedTask. That's my vote over getting clever w/ Task

oskardudycz:
@jeremydmiller I think that we're talking about the different things. Like I stated above - I already used Task.CompletedTask in my pull request. Calling sync method asynchronously is easy and already handled as you described in PR. @ericgreenmix agreed that my question is valid. I'm asking what should happen when we're calling sync applyProjections and we have defined async Event Handler by ProjectToAsync. Imho we have then 2 choices:
- ignore them,
- runs them synchronously.

First one would be imho confusing and error prone. I decided to go in my PR with second option by using .Wait(). I agree that it can provide deadlock and it's wrong, so I suggested those 2 nondeadlocking ways.

Could you elaborate why it's not nessesary? Maybe I'm missing the point and not catching your idea properly.

Of course I'm not pushing to add that feature, this is your decision. I can only say that in my business possibility to have full async handling of ApplyProjection would be great relief.

oskardudycz:
@jeremydmiller @ericgreenmix I added `NoSynchronizationContextScope` because it's suggested also in other cases for nondeadlocking run of asynchronous code synchronously. See changes in https://github.com/JasperFx/marten/pull/841/commits/e5d7a7984eafb588a32d4e9b80d4aa707fe57f4d

Would be grateful for your comments and the final decision about go/no go for that PR ;)

jeremydmiller:
@oskardudycz I'd still prefer you make the sync transforms act like async instead of making the async transforms try to be sync.

oskardudycz:
@jeremydmiller huh... I understand that, but how to handle the case that I described in my previous comment and the question I asked?

jeremydmiller:
Ah, gotcha, I see. I'm not super wild about the async to sync, but let's just get this in.

oskardudycz:
@jeremydmiller yup, I agree that it's not the ideal solution. But imho keeping the idea that Marten provides both async and synch api forces us to provide such cases... The other possibility is to extend `EventHandler` to sth like:
`private class EventHandler
    {
        public Func<IDocumentSession, object, Guid, TId> IdSelector { get; }
        public Func<IDocumentSession, object, Guid, List<TId>> IdsSelector { get; }
        public Action<TView, object> Handler { get; }
        public Func<TView, object, Task> AsyncHandler { get; }
        public ProjectionEventType Type { get; set; }

        public EventHandler(
            Func<IDocumentSession, object, Guid, TId> idSelector,
            Func<IDocumentSession, object, Guid, List<TId>> idsSelector,
            Func<TView, object, Task> asyncHandler,
            ProjectionEventType type)
        {
            IdSelector = idSelector;
            IdsSelector = idsSelector;
            AsyncHandler = asyncHandler;
            Handler = 
                (view, data) => 
                {
                    using (Util.NoSynchronizationContextScope.Enter())
                    {
                        AsyncHandler(view, data).Wait();
                    }
                };
            Type = type;
        }

        public EventHandler(
            Func<IDocumentSession, object, Guid, TId> idSelector,
            Func<IDocumentSession, object, Guid, List<TId>> idsSelector,
            Action<TView, object> handler,
            ProjectionEventType type)
        {
            IdSelector = idSelector;
            IdsSelector = idsSelector;
            Handler = handler;
            AsyncHandler = (view, data) => { Handler(view, data); return Task.CompletedTask; };
            Type = type;
        }
    }`
And then call those `Handler` or `AsyncHandler` methods in `ProjectTo` and `ProjectToAsync`. That would removed the need to call `.Wait()` when someone is using only synchronous code. But the problem with async handlers when running `applyProjections` synchronously would stay. But tbt this should work more or less the same, but will add some code complexity. What do you think?

oskardudycz:
@jeremydmiller @ericgreenmix guys, what's the final decision? :) Is it ready to be merged, or is it not accepted and I should close merge request?

ericgreenmix:
I am good with this.

Issue #844, opened 2017-08-10
Linq parser error when soft deletes are configured and multiple where clauses are applied


jeffdoolittle:
This is a regression from 1.x to 2.x

This used to work in Marten 1.x:

```csharp
var result = await session
    .Query<MyDocument>()
    .Where(x => x.TenantId == request.TenantId)
    .Where(x => x.Id == request.Id)
    .SingleOrDefaultAsync();
```

But now only this works with 2.x:

```csharp
var result = await session
    .Query<MyDocument>()
    .Where(x => x.TenantId == request.TenantId && x.Id == request.Id)
    .SingleOrDefaultAsync();
```

The exception stack trace is:

```
System.NotSupportedException: Collection was of a fixed size.
   at System.SZArrayHelper.Add[T](T value)
   at Marten.Linq.WhereFragmentExtensions.Append(IWhereFragment fragment, IWhereFragment other)
   at Marten.Linq.WhereFragmentExtensions.Append(IWhereFragment fragment, IWhereFragment[] others)
   at Marten.Schema.DocumentMapping.FilterDocuments(QueryModel model, IWhereFragment query)
   at Marten.Linq.Model.LinqQuery`1.buildWhereFragment()
   at Marten.Linq.Model.LinqQuery`1..ctor(DocumentStore store, QueryModel model, IIncludeJoin[] joins, QueryStatistics stats)
   at Marten.Linq.MartenQueryable`1.executeAsync[TResult](Func`2 source, CancellationToken token)
   at Marten.Linq.MartenQueryable`1.ToListAsync[TResult](CancellationToken token)
   at Marten.QueryableExtensions.ToListAsync[T](IQueryable`1 queryable, CancellationToken token)
```

jeffdoolittle:
Testing revealed that the issue arises when `SoftDelete` is configured for the Document Type.

jeremydmiller:
Easy money, it's fixed and will be in 2.1 later today.

PR #845, opened 2017-08-10
failing test demonstrating GH-844

jeffdoolittle:
First test fails (with soft deletes configured). 

Second test passes (exact same code but without soft deletes configured).


jeremydmiller:
Got it, thanks for doing this

PR #859, opened 2017-08-19
Correcting a typo I noticed when setting up for development.

slashben:


jeremydmiller:
Grumble. That's gone back and forth a couple times on different platforms. Thanks for catching that!

PR #873, opened 2017-09-12
Patch 1

nehapuniani:
IMartenQueryable does not work with respect to ignoring the case

Issue #834, opened 2017-07-29
linq support for any and startswith


mdissel:
My model is:
```
public class Contact {
	List<string> Tags {get;set;}
}
```
and now trying to count all contacts that have a specific Tag that starts with X
`session.Query<Contact>().Count(x => x.Tags.Any(y => y.StartsWith("X"))`

Currently (v2 rc) I get a NotSupportedException. Any workarounds?

jeremydmiller:
Like I say every single time a Linq problem comes up, you almost have to explicitly code for every possible permutation of operators. I don't think that that usage could be done very easily if at all with the current JSONB operators.

As a workaround, you can add a read only property to your Contact document like:

```
public bool HasXTags
{
    get
    {
          return Tags.Any(y => y.StartsWith("X");

     }
}
```

then query by:

```
session.Query<Contact>().Count(x => x.HasXTags)
```

mdissel:
I tried to find a sql command for this, but didn't find a simple solution other then unnest the array. 
In my case i'm storing a tree structure like ["'/1/2/3/4'", "1/2/4"] and i want to find all the childs of "/1/" for example, so a simple property at the top level will not help in this case. A workaround for now is a OR combination of every child, but that's not very efficient with large number of combinations
`session.Query<Contact>().Count(x => x.Tags.Contains("/1/") || x.Tags.Contains("/1/2/") || etc...);`

jokokko:
@mdissel 
```cs
session.Query<Contact>(@"
SELECT d.data
FROM public.mt_doc_contact d,
    LATERAL jsonb_array_elements_text(d.data -> 'Tags') t
WHERE t.value LIKE 'X%'")
```

mdissel:
Thanks!!

jeffdoolittle:
@mdissel or @jokokko - want to turn that into an `IMethodCallParser` and submit a PR?

http://jasperfx.github.io/marten/documentation/documents/advanced/customizing_linq/

jokokko:
@jeffdoolittle I think I can take a stab at it. Might not be totally trivial though, because of the LATERAL...

jeffdoolittle:
Helpful place to start: 

https://github.com/JasperFx/marten/blob/8069bb0657c42672b5fcfaf600b0aae20d93936e/src/Marten/Linq/Parsing/IsOneOf.cs

jokokko:
Thanks. I hope I can manage. I have a pending PR for one IMethodCallParser already }:]

[...]

PR #875, opened 2017-09-14
Add ignore case constructor overload for StringContains

mysticmind:


mysticmind:
Please use this PR instead of #873

mysticmind:
@nehapuniani is using a sample query as below:
```
var searchTags = new List<string> { "Test1", "TEST2" };
var results = session.Query<Contact>().Where(x => x.Tags.Any(q => q.Name.Contains(searchTags[0], StringComparison.OrdinalIgnoreCase)));
```

When I try out the above, it actully results in `NotSupportedException` since for a collection, only `==` is supported within `Any()`. Hence the original fix provided by @nehapuniani won't work (this PR was raised on her behalf by me).

The solution is to fall to plain sql query as below:
```
session.Query<Contact>($"SELECT d.data FROM public.mt_doc_contact d,LATERAL jsonb_array_elements(d.data->'Tags') t WHERE t.value->>'Name' ILIKE '{searchTags[0]}'");
```

@nehapuniani If you have any further comments or thoughts, let me know. Otherwise I will close this PR by tomorrow.

nehapuniani:
ILike is not a good solution as it will drastically affect the performance. 
Does Jeremy or you have a better solution ?
Thanks, Neha

mysticmind:
You could use equals check using the SQL as below:
```
session.Query<Contact>($"SELECT d.data FROM public.mt_doc_contact d,LATERAL jsonb_array_elements(d.data->'Tags') t WHERE t.value->>'Name' = '{searchTags[0].ToLower()}'");
```

nehapuniani:
That's exactly what i am doing and this adds case sensitivity and that was the issue that i reported. Jeremy said overloading that method will solve the issue. We need better solution.
Thanks, Neha

mysticmind:
I am closing this PR since the applied changes don't solve the underlying problem at hand. Few proposals provided, not sure what else we could do here.

@nehapuniani you could raise a new issue and outline the details. A similar issue #834 is already there, you could add your comments to the existing issue as well.

Issue #604, opened 2016-11-01
Be able to use public Field's as parameters in compiled queries


jeremydmiller:
Not a super high priority, but still

jeremydmiller:
This is done by #876

PR #876, opened 2017-09-23
Compiled Queries - Adds support for fields in query filters

timbarker:
In the marten [docs for compiled queries](http://jasperfx.github.io/marten/documentation/documents/querying/compiled_queries/) it states that properties and fields can be used in the query filter.  This currently works fine for properties but when a field is used this exception is raised:
```
Exception has occurred: CLR/System.InvalidCastException
An unhandled exception of type 'System.InvalidCastException' occurred in Marten.dll: 'Unable to cast object of type 'System.Reflection.RtFieldInfo' to type 'System.Reflection.PropertyInfo'.'
   at Marten.Linq.Compiled.CompiledQueryMemberExpressionVisitor.VisitMember(MemberExpression node)
```
This is because the CompiledQueryMemberExpressionVisitor is expecting only properties. This PR adds support for fields as well as properties. I believe this is what is being referred to in issue #604

### DIFF
--- a/src/Marten.Testing/Linq/Compiled/compiled_query_Tests.cs
+++ b/src/Marten.Testing/Linq/Compiled/compiled_query_Tests.cs
@@ -60,6 +60,18 @@ namespace Marten.Testing.Linq.Compiled
             UserByUsername.Count.ShouldBe(1);
         }
 
+        [Fact]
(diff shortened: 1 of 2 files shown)


jeremydmiller:
Sweet, this'll be in either 2.3.1 or 2.4 tonight or tomorrow.

Issue #877, opened 2017-09-24
Foreign Key exception when object uses Document Hierarchy


slovely:
I've spotted an issue when an entity has a ForeignKey setup to an entity that defines SubClasses.  It's equally possible I am doing something wrong!

e.g.
Given this setup...
```
        public class Person
        {
            public int Id { get; set; }
        }

        public class Employee : Person
        {
        }

        public class Address
        {
            public int Id { get; set; }
            public int ParentId { get; set; }
        }

        // STORE OPTIONS
        _.Schema.For<Person>()
            .AddSubClass<Employee>();

        _.Schema.For<Address>()
            .ForeignKey<Person>(c => c.ParentId);
```
When executing this code....
```
        using (var session = theStore.OpenSession())
        {
           // Creating an employee with an address - the employee object should be
           // inserted *before* the address object.
           var employee = new Employee {Id = 222};
           session.Store(employee);
           var address = new Address {ParentId = 222, Id = 42};
           session.Store(address);
           session.SaveChanges();
       }
```
I get a FK exception raised when SaveChanges is called.
I've done a bit of debugging, and I think that the issue is in `UnitOfWork.GetTypeDependencies` function.  When the 'Address' mapping is examined, it see `Person` as a dependency, but not 'Employee', so it attempts to insert the address before the employee.

I've added an integration test for this and a possible fix, although it makes `GetTypeDependencies` more expensive, so it's possibly something that should be calculated when the store is setup.  It might be that a unit test would be better for this, although I don't think the dependency information is exposed from `UnitOfWork` at the moment.

Please let me know if you want/need more info, or anything changed in the commit.  More than happy to help if I can.

slovely:
Closing this one, as I have replaced with PR #878

PR #878, opened 2017-09-24
Test and possible fix for FK and sub-classes exception #877

slovely:
Comment copied from #877:

I've spotted an issue when an entity has a ForeignKey setup to an entity that defines SubClasses.  It's equally possible I am doing something wrong!

e.g.
Given this setup...
```
        public class Person
        {
            public int Id { get; set; }
        }

        public class Employee : Person
        {
        }

        public class Address
        {
            public int Id { get; set; }
            public int ParentId { get; set; }
        }

        // STORE OPTIONS
        _.Schema.For<Person>()
            .AddSubClass<Employee>();

        _.Schema.For<Address>()
            .ForeignKey<Person>(c => c.ParentId);
```
When executing this code....
```
        {
           // Creating an employee with an address - the employee object should be
           // inserted *before* the address object.
           var employee = new Employee {Id = 222};
           session.Store(employee);
           var address = new Address {ParentId = 222, Id = 42};
           session.Store(address);
           session.SaveChanges();
       }
```
I get a FK exception raised when SaveChanges is called.
I've done a bit of debugging, and I think that the issue is in `UnitOfWork.GetTypeDependencies` function.  When the 'Address' mapping is examined, it see `Person` as a dependency, but not 'Employee', so it attempts to insert the address before the employee.

I've added an integration test for this and a possible fix, although it makes `GetTypeDependencies` more expensive, so it's possibly something that should be calculated when the store is setup.  It might be that a unit test would be better for this, although I don't think the dependency information is exposed from `UnitOfWork` at the moment.

Please let me know if you want/need more info, or anything changed in the commit.  More than happy to help if I can.

### DIFF
--- a/src/Marten.Testing/Bugs/SubClassForeignKeyBugs.cs
+++ b/src/Marten.Testing/Bugs/SubClassForeignKeyBugs.cs
@@ -1,0 +1,51 @@
+﻿using Marten.Services;
+using Xunit;
+
+namespace Marten.Testing.Bugs
+{
+    public class SubClassForeignKeyBugs : IntegratedFixture
+    {
+        public class Person
+        {
+            public int Id { get; set; }
+        }
+
+        public class Employee : Person
+        {
+        }
+
+        public class Address
+        {
+            public int Id { get; set; }
+            public int ParentId { get; set; }
+        }
+        
+        public SubClassForeignKeyBugs()
+        {
+            StoreOptions(_ =>
+            {
+                _.Schema.For<Person>()
+                    .AddSubClass<Employee>();
+
+                _.Schema.For<Address>()
+                    .ForeignKey<Person>(c => c.ParentId);
+            });
+        }
+
+        [Fact]
+        public void ForeignKeysOnDerivedClassesShouldBeInsertedFirst()
+        {
+            using (var session = theStore.OpenSession())
+            {
+                // Creating an employee with an address - the employee object should be
+                // inserted *before* the address object.
+                var employee = new Employee {Id = 222};
+                session.Store(employee);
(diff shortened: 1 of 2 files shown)


slovely:
Firstly, apologies for creating both an issue (#877) and a PR about the same thing.  I'm closing the issue and moving the details into here.
Also, I've spotted a related issue to this when an entity with SubClasses itself has a FK to another entity.  An incorrect insert statement sequence is generated.  For example:

```
        public class Department
        {
            public int Id { get; set; }
        }

        public class Person
        {
            public int Id { get; set; }
            public int DepartmentId { get; set; }
        }

        public class Employee : Person
        {
        }

        // STORE OPTIONS
        _.Schema.For<Person>()
            .AddSubClass<Employee>()
            .ForeignKey<Department>(p => p.DepartmentId);

        // This code will fail
        {
            var department = new Department {Id = 1};
            session.Store(department);
        
            var employee = new Employee {Id = 2, DepartmentId = 1};
            session.Store(employee);
            // SaveChanges tries to insert the employee before the department
            session.SaveChanges();
        }
```
Again, the problem appears to be in `GetTypeDependencies` - this time because it casts the result of `_tenant.MappingFor(EMPLOYEE)` to DocumentMapping, but it is a `SubClassMapping` so doesn't pick up the fact it has an FK to Department.

I've added another commit with a test and possible fix for this.

jeremydmiller:
@slovely I'll get these in *some time* this week.

@jimgolfgti Thanks for looking through this

slovely:
Thanks guys!  No rush to get it in @jeremydmiller

Issue #796, opened 2017-06-13
Index Constraints owned by a PostgreSQL group role, not a user role


bradyclifford:
In our Postgres environment, every object is owned by a group role, not a user role.  This allows multiple user roles to modify schema when needed. We don't give permissions to a user role directly because it's not the same from one environment to the next.

In Marten when doing the schema check, the query it uses for getting all of the indices requires that the index is owned by a User role.
```JOIN pg_user AS U ON i.relowner = U.usesysid```

Marten is not able to find our new index constraint and it will throw an exception when we have auto create disabled in our environments.

**Comment by jeremydmiller** [comment]
@bradyclifford Can you just get around this by going AutoCreate.None in other environments? At this point, that never checks the schema itself. I'm hesitant to do any more customization of Marten to acomodate our database team.

**Comment by bradyclifford** [comment]
AutoCreate.None doesn't fix the problem to my recollection.  It has been awhile, so I will have to triage it again.

Andrew might have additional details.

**Comment by bradyclifford** [comment]
To summarize, there are 2 issues.
1. `AutoCreate.None` is still asserting schema
1. `pg_roles` should be used instead of `pg_user` when asserting; it will still include any LOGIN user roles

As an example:
```
SELECT
  R.rolname                AS role_name,
  ns.nspname               AS schema_name,
  pg_catalog.textin(pg_catalog.regclassout(idx.indrelid :: REGCLASS)) AS table_name,
  i.relname                AS index_name,
  pg_get_indexdef(i.oid) as ddl,
  idx.indisunique          AS is_unique,
  idx.indisprimary         AS is_primary,
  am.amname                AS index_type,
  idx.indkey,
       ARRAY(
           SELECT pg_get_indexdef(idx.indexrelid, k + 1, TRUE)
           FROM
             generate_subscripts(idx.indkey, 1) AS k
           ORDER BY k
       ) AS index_keys,
  (idx.indexprs IS NOT NULL) OR (idx.indkey::int[] @> array[0]) AS is_functional,
  idx.indpred IS NOT NULL AS is_partial
FROM pg_index AS idx
  JOIN pg_class AS i
    ON i.oid = idx.indexrelid
  JOIN pg_am AS am
    ON i.relam = am.oid
  JOIN pg_namespace AS NS ON i.relnamespace = NS.OID
  JOIN pg_roles AS R ON i.relowner = U.oid
[...]

PR #917, opened 2017-11-10
Search Schema by Role not User

bradyclifford:
When schema is created under a context other than a user, Marten
assertion fails.

- Change assertion querying from utilizing `pg_user` to instead `pg_roles`

Resolves Issue #796

### DIFF
--- a/src/Marten/Schema/DbObjects.cs
+++ b/src/Marten/Schema/DbObjects.cs
@@ -61,7 +61,7 @@ namespace Marten.Schema
         {
             var sql = @"
 SELECT
-  U.usename                AS user_name,
+  R.rolname                AS role_name,
   ns.nspname               AS schema_name,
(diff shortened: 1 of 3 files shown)


bradyclifford:
@jeremydmiller any luck reviewing this?  We are having assertions issues running the Marten.Assert command for health check purposes.

mysticmind:
I think it would help to keep the implementation configurable to support both user and role (with default behaviour as user). Just changing it to use role may break the current behaviour and may cause issues for others using Marten in production.

oskardudycz:
I fully agree with @mysticmind - we should avoid breaking changes when they're not needed. I think that we could add it if you @bradyclifford provide it as optional (eg. based on the switch) and leaving default behaviour as it is.

@bradyclifford could you rebase it with recent master and update the code with such switch?

jeremydmiller:
I don't think this one is ever going to happen. I'm going benevolent dictator today and closing this.

PR #919, opened 2017-11-13
fix the initial members

mburleigh:


jeremydmiller:
I'm actually a little bit ashamed here, got it;)

Issue #925, opened 2017-12-03
Patch -> SaveChanges -> Load: stale result


migajek:
When modifying document(s) via patch api, calling `SaveChanges` and then `Load` does not work as expected - returns stale result. See tests below.

I know one could manually `Eject` the document, but then again it'd be required to `Load` it first.

For me, this breaks the scenario of events projection - in reaction to one event, the documents are patched. Reacting to other events, documents are Loaded first and then some data is copied from one document to another. All of this happens in one session (managed by async daemon)

One fix that came to my mind is to eject all documents matching the Patch'es `IWhereFragment` when applying it. What do you think of this approach? I think I could try implementing it

tests: 
```
        [Fact]
        public async Task Patch()
        {
            var id = Guid.NewGuid();
            using (var sess = theStore.OpenSession())
            {
                sess.Store(new Model() { Id = id, Name = "foo" });
                sess.Patch<Model>(id).Set(x => x.Name, "bar");
                await sess.SaveChangesAsync();                
                sess.Query<Model>().Where(x => x.Id == id).Select(x => x.Name).Single().ShouldBe("bar");
                sess.Load<Model>(id).Name.ShouldBe("bar"); // fails. Name is null.
            }
        }

        [Fact] // this one passes
        public async Task LoadAndChange()
        {
            var id = Guid.NewGuid();
            using (var sess = theStore.OpenSession())
            {
                sess.Store(new Model(){Id = id, Name = "foo"});
                sess.Load<Model>(id).Name = "bar";
                await sess.SaveChangesAsync();                
                sess.Query<Model>().Where(x => x.Id == id).Select(x => x.Name).Single().ShouldBe("bar");
                sess.Load<Model>(id).Name.ShouldBe("bar");
            }
        }
```

jeremydmiller:
Ugh, that's going to be annoying. If there's a patch, I'd vote to just eject all models of that type. Might end up being cheaper at runtime than it would be to do it selectively

migajek:
ok, this sounds to be even easier. I'll try a PR

jeremydmiller:
Fixed by #926

PR #926, opened 2017-12-03
after patching, eject all documents of patched types

migajek:
GH-925

### DIFF
--- a/src/Marten/IDocumentSession.cs
+++ b/src/Marten/IDocumentSession.cs
@@ -207,6 +207,13 @@ namespace Marten
         /// <typeparam name="T"></typeparam>
         /// <param name="document"></param>
         void Eject<T>(T document);
+
+        /// <summary>
+        /// Completely remove all the documents of given type from this session's unit of work tracking and identity map caching
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="type"></param>
+        void EjectAllOfType(Type type);
     }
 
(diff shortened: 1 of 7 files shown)


src/Marten/IDocumentSession.cs
```diff
@@ -201,12 +201,19 @@ public interface IDocumentSession : IQuerySession
...
+        /// <typeparam name="T"></typeparam>
         /// <param name="document"></param>
         void Eject<T>(T document);
+
+        /// <summary>
+        /// Completely remove all the documents of given type from this session's unit of work tracking and identity map caching
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="type"></param>
+        void EjectAllOfType(Type type);
```
jeremydmiller: That bumps this to 2.6.0 instead of 2.5.1. No big deal. Makes tracking a little easier

jeremydmiller:
Looks good, and sorry for the delay getting this in. It'll have to be in 2.6.0. How soon do you need this? I know there's some outstanding work that hasn't come in yet I'd like to wait for if I could.

migajek:
Great, thanks for merging :) I'm not in a hurry, I've already setup appveyor builds of my fork and I'm using it's custom nuget feed so I'm fine :)

It'd be great if you could review my second PR (#927) as I'm more afraid of it's concept not fitting into Marten, in such case I cannot build my app around it as I want to eventually switch to the "main" Marten and stop using my fork :)

PR #927, opened 2017-12-03
Multi document type projection

migajek:
This change makes it possible for the Async Daemon to handle projections which are managing several document types (read models). 

An example use case included in tests

### DIFF
--- a/src/Marten/Events/Projections/IProjection.cs
+++ b/src/Marten/Events/Projections/IProjection.cs
@@ -22,5 +22,14 @@ namespace Marten.Events.Projections
         {
             return (projection as IDocumentProjection)?.Produces ?? projection.GetType();
         }
+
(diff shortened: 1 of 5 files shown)


src/Marten/Events/Projections/IProjection.cs
```diff
@@ -22,5 +22,14 @@ public static Type ProjectedType(this IProjection projection)
         {
             return (projection as IDocumentProjection)?.Produces ?? projection.GetType();
         }
+
+        public static Type[] ProjectedTypes(this IProjection projection)
```
jeremydmiller: Clever. Avoids the breaking API change.

jeremydmiller:
@migajek Looks good, and I'm pulling this in now, but could I talk you into another PR adding a little bit to the docs on multi-document projections?

Issue #932, opened 2017-12-18
LINQ expression not parsed correctly when querying inherited class with at least one OR in the query


VladanStojanovic:
The final query has `**or** mt_doc_type = 'INHERITED_CLASS'` instead of `**and** mt_doc_type = 'INHERITED_CLASS'` appended in where fragment, making the query ignore all conditions and just return all the data of inherited type.

Test class (based on existing tests in Marten.Testing/Bugs) is attached.

[Bug_932_or_query_in_inherited_class.zip](https://github.com/JasperFx/marten/files/1569706/Bug_932_or_query_in_inherited_class.zip)

jeremydmiller:
@VladanStojanovic Instead of zip files, can you do a PR with the failing tests? That's a whole lot easier on us to consume and fix.

VladanStojanovic:
@jeremydmiller Yep, doing it now. Thanks for the quick response.

VladanStojanovic:
@jeremydmiller Added a PR. Will try debugging the tests and see if I can figure out anything else that can be helpful.

VladanStojanovic:
The problem seems to stem from the way user-defined query is combined with doc type `where` fragment, always using the user-defined query's operator to combine them. User-defined query in `SubClassMapping.FilterDocuments` should be wrapped in a query utilizing the `and` operator before Appending the doc type fragment. But I don't know if it would break some other functionality.

A workaround is to always define your query with a clause that always evaluates to `true`, like so:

> session.Query<Doc>().Where(x => **true &&** (x.Id == 1 || x.Id == 2)).Count();

jeremydmiller:
@kYann did a fix for this one. Will be in Marten 2.8.0.

PR #933, opened 2017-12-18
Tests for bug #932

VladanStojanovic:
Tests are showing how a simple query that includes a logical disjunction works correctly with the base class and generates a wrong query for the inherited one.

Bug is described in more detail in issue #932 .


jeremydmiller:
@VladanStojanovic Ugh, I dropped the ball on this one. I do have the fix for the original issue in from @kYann.

Issue #939, opened 2018-01-04
Ability to pass in instance of Aggregate when applying events for a Stream


Sacrelicious:
We need to supply some external information to the Aggregate we are building to support Applying certain events, as such we need to be able to pass in an instantiated instance of the Aggregate type, instead of just relying on `new T()` every time we call `AggregateStream<T>...`

jeremydmiller:
Closed by https://github.com/JasperFx/marten/commit/502793528001c9144e3c6c4ac2d3c56c9c0aa936

PR #940, opened 2018-01-04
Ability to use instantiated Aggregate object when applying events for stream #939

Sacrelicious:


jeremydmiller:
@Sacrelicious Ugh, sorry for being so slow getting back to you. The code looks fine, except -- and I know this blows -- I need you to make overloads for those additional methods where you pass in state instead of the optional argument. I know it doesn't change your code, but it's a breaking binary change and would require a bump to 3.0 I don't wanna do anytime soon.

And if you don't mind, would you add a sample of doing this in the docs on aggregation?

If you don't have time to get to this soon, I can take care of it later, but it'll be a bit.

PR #945, opened 2018-01-16
Proposal to add full text document searching using PostgreSQL 10 features

CodingGorilla:
I've added a basic implementation to allow full text searching using the new features in PGSQL 10.  My big question before merging this pull request is whether I handled the execution of the query in the proper (best?) manner.  Also, what kind of tests are needed for this, if any?


oskardudycz:
@jeremydmiller Postgresql seems to be supported both in Amazon and Azure:
- https://aws.amazon.com/about-aws/whats-new/2018/02/postgresql-10-now-supported-in-amazon-rds/ 
- https://azure.microsoft.com/en-us/updates/postgresql-10-on-azure-database-for-postgresql/
Seems to me that we're not blocked anymore

src/Marten/Storage/StorageFeatures.cs
```diff
@@ -1,245 +1,245 @@
-using System;
```
oskardudycz: @CodingGorilla was there any change to this file? Is it just some weird EOL formatting issue?

CodingGorilla:
Yea, I haven't touched this in a while, but I'll take a whack at it this weekend.

src/Marten/Storage/StorageFeatures.cs
CodingGorilla: This might just have been my R# doing this on it's own as I was making my way through the code.  I think I will just try to recreate these changes on top of master, rather than rebasing.  That is probably easier than rebasing and it will clear up issues like this where I honestly don't remember what I changed.

CodingGorilla: @oskardudycz Yea, originally I didn't do unit tests because I was looking for some guidance as to whether I was doing this "the best way" (as asked in my original comment).  But if this is good, then I'll go ahead and add the tests.

oskardudycz: For me it looks good. I like that it’s not breaking change and extending what we have. But it would be nice to see what others think. I’ll try to go during the weekend deeper into this feature, but so far so good.

mysticmind:
I ran through the changes and it looks good to me. We should all be set once @CodingGorilla does a rebase and adds unit tests.

src/Marten/Schema/FullTextIndex.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            _table = parent.Table;
+            _config = config;
+        }
+
+        public string IndexName
+        {
+            get
+            {
+                if(_indexName.IsNotEmpty())
```
oskardudycz: What I'd suggest is to move this logic of checking and parsing to setter method. Setter method will be called less, so that will reduce the number of potential allocations

CodingGorilla:
Closing this PR in favor of #1098 .

PR #981, opened 2018-03-03
Add documentation for AnyTenant and TenantIsOneOf

mysticmind:


cocowalla:
Hah, I just came across these 2 today with intellisense, so thanks for this 👍

PR #987, opened 2018-03-15
UpdateBatchSize with delete

teamdynamiq:
Seems like batching is not taken into consideration when a delete is used in the same unit of work.


jeremydmiller:
Got it in. Will be in 2.7.1 later today when everything runs and the nuget is published

PR #989, opened 2018-03-21
make LazyLoadedProjection play nicer with async daemon

migajek:
before this change, the async daemon is not able to clear the storage before replaying all events.

it looks for storage table for the projection itself, not taking into consideration that the projection might not be an aggregate one. 

note: the test included proves it to some point only. 
before the change (when the LazyLoadedProjection didn't implemented `IDocumentsProjection`), the test runned forever instead of failing.

There seem to be some other issue making the `Daemon.RebuildAll` Task to never complete (instead of fail) when there's an exception in `DocumentMapping.Validate`. But this is kind of an async hell so I didn't manage to dig into it. the only way you can tell there was a problem is by intercepting async daemon messages (there'll be a message with a validation exception)

### DIFF
--- a/src/Marten.Testing/Events/Projections/Async/MultidocumentProjectionTests.cs
+++ b/src/Marten.Testing/Events/Projections/Async/MultidocumentProjectionTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


oskardudycz:
My few cents, as author of  `LazyLoadedProjection`:
I haven't used it with the `AsyncDaemon`, only with `ViewProjection` so I din't thought about such usage. Initially I was thinking to add `IDocumentProjection` initally, but I stayed with the more basic interface. I don't have anything against addint `IDocumentProjection` to it. There is a rare chance that someone would want to use it with different type of projection.

migajek:
perfect, thank you!

Issue #997, opened 2018-04-13
Query not working when using OR with document hierarchies


kYann:
Hello !

When doing a linq query with OR, then all the documents of the table are returned.

```
var query = session.Query<MegaUser>()
                       .Where(_ => _.DisplayName == "Yann" || _.DisplayName == "Robin");
```

When outputing the sql, I can see that the OR operator is used on the type field, and this is why all the rows are returned

`select d.data, d.id, d.mt_doc_type, d.mt_version from public.mt_doc_user as d where (d.data ->> 'DisplayName' = :arg0 or d.data ->> 'DisplayName' = :arg1 or (d.mt_doc_type = 'mega_user'))`

See full demo here https://gist.github.com/kYann/9fe0fd4e860049552439984789e35e8f

Thanks !

kYann:
Duplicate #932

PR #998, opened 2018-04-13
fix issue #997 by forcing and

kYann:
#997 Fix


jeremydmiller:
@kYann I'm gonna get this right now, but add a regression test. Next time though, can you add a test demonstrating the problem and locking down the problem?

jeremydmiller:
Got it in with a rebase.

Issue #960, opened 2018-02-16
DROP INDEX CONCURRENTLY must be first action in transaction


NicoKno:
I get an exception with the following code and I can't find the actual problem.
The code is running against a docker postgres:latest container which should be a Postgres 10.2 DB.

``` csharp
public async Task<int> CountAsync(string userId, string modulePackageId, DateTime bookingDate, DateTime expiryDate)
{
            var result = await _session.Query<TransactionEvent>().Where(transaction => transaction.UserId.Equals(userId, StringComparison.Ordinal) &&
                                                                           transaction.PackageId.Equals(modulePackageId, StringComparison.Ordinal) &&
                                                                           transaction.Timestamp >= bookingDate &&
                                                                           transaction.Timestamp < expiryDate &&
                                                                           transaction.BillingExecuted).CountAsync();

            return result;
}
```

The exception that is thrown:
```
0A000: DROP INDEX CONCURRENTLY must be first action in transaction) ---> Marten.MartenCommandException: Marten Command Failure:
-- Marten.Storage.DocumentTable
drop index concurrently if exists public.mt_doc_transactionevent_idx_timestamp;


-- Marten.Storage.DocumentTable
drop index concurrently if exists public.mt_doc_transactionevent_idx_user_id;


-- Marten.Storage.DocumentTable
drop index concurrently if exists public.mt_doc_transactionevent_idx_package_id;


-- Marten.Storage.DocumentTable
drop index concurrently if exists public.mt_doc_transactionevent_idx_contingent_id;

0A000: DROP INDEX CONCURRENTLY must be first action in transaction ---> Npgsql.PostgresException: 0A000: DROP INDEX CONCURRENTLY must be first action in transaction
   at Npgsql.NpgsqlConnector.<DoReadMessage>d__148.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at System.Runtime.CompilerServices.ValueTaskAwaiter`1.GetResult()
   at Npgsql.NpgsqlConnector.<ReadMessage>d__147.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1006, opened 2018-04-29
Fix for bug #960

Miggleness:
Includes fix and test for bug #960


jeremydmiller:
Got it. Doofus me introduced some build failures from a "fix" that I need to address before pulling this in. Might take a week. Will have this in 2.8.

PR #1021, opened 2018-05-22
Factorize EventMappingFor related duplicated logic in event selectors

smoothdeveloper:
This factorizes the event mapping resolution which was duplicated 4 times in the code with two variants.

Slight drawback is that we pull dotnettypename from reader upfront.


jeremydmiller:
Yeah, I'm sorry. I think this just isn't important enough to bring in. I'm closing this for the sake of housekeeping

PR #1024, opened 2018-06-01
Fix for netcore 2.1 Linq support

jarroda:
This is the best we could come up with for a fix for this.  Reflection will have to try to get the overloads for `String.Contains` at runtime, and what it gets will vary depending on whether its running on 2.0 or 2.1.


jarroda:
Looks like we're waiting on 2.1 support on Appveyor https://github.com/appveyor/ci/issues/2373
Later today perhaps?

ericgreenmix:
@jeremydmiller AppVeyor is now supporting .net 2.1 as of yesterday. Can we get the build re-ran?

jeremydmiller:
@ericgreenmix @jarroda  On its way.

PR #1031, opened 2018-06-14
BulkInsertion optimization

oakio:
1. BulkInsert method now accept not only arrays, but also lists and hashsets.
2. Reduce GC memory pressure on batch gathering

### DIFF
--- a/src/Marten/DocumentStore.cs
+++ b/src/Marten/DocumentStore.cs
@@ -134,7 +134,7 @@ namespace Marten
         public IDocumentSchema Schema { get; }
         public AdvancedOptions Advanced { get; }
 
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@oakio Sorry I've been so slow on this one. Looks good to me, but...

It's nit-picky, but this is a breaking public API change so it would need to either be overloads that would clutter up the API or we start working on a 3.0 and roll this into that. I'm using this PR and a couple other breaking changes to start a 3.0 release, so look for this to be into a 3.0.0-alpha-1 release sometime next week.

jeremydmiller:
My bad, I didn't read the code closely enough. I'm pulling this in shortly then.

PR #1037, opened 2018-07-05
Added PLV8 extension creation option

oakio:
Allow to automatically create PLV8 extension when new database created


jeremydmiller:
Got it, thank you. I'll try to have a 2.9 out later this week with this and some other things

PR #1040, opened 2018-07-10
Added multi-property indexing support (close #733)

### LINKED ISSUE #733: Multi-Property indexing support
Implement support for multi-property indexing. 

Proposed API:

```csharp
    _.Schema
        .For<User>()
        .Index(x =>
        {
            x.AddColumn(prop => prop.FirstName);
            x.AddColumn(prop => prop.LastName);

            x.IndexName = "first_last_name_index";
        });
```

This can work along-side the existing index API for single property, however the existing api will call into the new api as they are both the same, except that the new api supports multiple properties.

oakio:
Close https://github.com/JasperFx/marten/issues/733

### DIFF
--- a/src/Marten.Testing/Acceptance/computed_indexes.cs
+++ b/src/Marten.Testing/Acceptance/computed_indexes.cs
@@ -1,4 +1,7 @@
-﻿using System.Linq;
+using System;
+using System.Collections.Concurrent;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@oakio Sweet! Could I ask you to add some content to the docs page on calculated indexes with a sample? Could even just be embedding the code for your unit test.

PR #1045, opened 2018-07-14
Wip update to Npgsql 4.0

oskardudycz:
Follow up of https://github.com/JasperFx/marten/pull/1033 by @jacobpovar

See more about the reason of such changes in:
- https://github.com/npgsql/npgsql/issues/1483
- https://github.com/npgsql/Npgsql.EntityFrameworkCore.PostgreSQL/issues/303
- http://www.npgsql.org/doc/types/datetime.html
- http://www.npgsql.org/efcore/mapping/enum.html
- http://www.npgsql.org/doc/types/enums_and_composites.html


src/Marten/Marten.csproj
```diff
@@ -27,9 +31,10 @@
     <EmbeddedResource Include="Schema\SchemaObjects.sql" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="Npgsql" Version="4.1.0-ci.1131" />
```
jacobpovar: now when npgsql https://www.nuget.org/packages/Npgsql/4.0.2 with jsonb fix is released is this pre-release package necessary ?

oskardudycz: I think so, but I'll double check that

oskardudycz: @jacobpovar You're right `4.0.2` is enough.

jacobpovar: great :)

src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
```diff
@@ -161,7 +154,17 @@ private string buildSubQuery(SubQueryExpression subQuery, CommandBuilder command
             {
                 throwNotSupportedContains();
             }
-            var fromParam = command.AddParameter(from.Value);
+
+            var values = new List<string>();
```
jeremydmiller: Why are we having to do this? Is this something that got broken by npgsql 4?

src/Marten/Linq/QueryHandlers/CachedQueryHandler.cs
```diff
@@ -26,13 +24,20 @@ public CachedQueryHandler(object model, NpgsqlCommand template, IQueryHandler<T>
...
+        public Type SourceType => _handler.SourceType;
+
         public void ConfigureCommand(CommandBuilder builder)
         {
             var sql = _template.CommandText;
             for (var i = 0; i < _setters.Length && i < _template.Parameters.Count; i++)
             {
                 var param = _setters[i].AddParameter(_model, builder);
+
+                if (param.Value is Enum)
```
jeremydmiller: Why is this here? Marten has configurable enum storage of either "as strings" or "as int". This would possibly break here.

src/Marten/Schema/BulkLoading/BulkLoader.cs
```diff
@@ -121,6 +121,8 @@ private void load(ITenant tenant, ISerializer serializer, NpgsqlConnection conn,
                     _transferData(document, _mapping.AliasFor(document.GetType()), serializer, writer, textWriter, tenant.TenantId);
                     textWriter.Clear();
                 }
+                
+                writer.Complete();
```
jeremydmiller: This is new in npgsql4?

src/Marten/Schema/DuplicatedField.cs
```diff
@@ -38,9 +37,19 @@ public DuplicatedField(EnumStorage enumStorage, MemberInfo[] memberPath) : base(
                 PgType = "varchar";
             }
             else if (MemberType.IsDateTime())
+            {
```
jeremydmiller: Are all the Datetime/DatetimeOffset query tests passing?

src/Marten/Services/SprocCall.cs
```diff
@@ -90,6 +88,9 @@ public SprocCall JsonBodies(string argName, ArraySegment<char>[] bodies)
 
         public SprocCall Param(string argName, object value, NpgsqlDbType dbType)
         {
+            if ((value is Enum || value is Guid) && dbType == NpgsqlDbType.Varchar)
```
jeremydmiller: Same thing here about the enum configuration on the serializer

src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
oskardudycz: It seems so, we're passing the parameter as array of text (see [here](https://github.com/JasperFx/marten/pull/1045/files/e56226f5ff844a03a847ca0582459f0812161efd#diff-f98a5f4de4c572b88fea044c7ef852a8R168) ). I believe that previously it was the automated conversion of parameter, but right now it's not being performed. It throws exception that TextHandler cannot handle array of guids (or something like that). In general it's because Array of String != Array of Guids.

src/Marten/Linq/QueryHandlers/CachedQueryHandler.cs
oskardudycz: Yup, it's ugly hack. Without this casting it will throw that it doesn't know the Enum type. I need to revisit that but I believe that here it's only for querying and there would be some automated translation. I need to check it again if there is a chance to do it better.

src/Marten/Schema/BulkLoading/BulkLoader.cs
oskardudycz: I believe so. @jacobpovar ?

src/Marten/Schema/DuplicatedField.cs
oskardudycz: Yes.

src/Marten/Services/SprocCall.cs
oskardudycz: Here it's the same case as I described with Array of Guids. It seems that there is no more automatic conversion here. They provided some changes to automatically selecting the handlers, but I don't know precisely, because it's a lack of documentation.

src/Marten/Schema/BulkLoading/BulkLoader.cs
jacobpovar: Correct. 

From [npgsql 4 breaking changes](http://www.npgsql.org/doc/release-notes/4.0.html):

> You must now call NpgsqlBinaryImporter.Complete() to save your imported data; not doing so will roll the operation back. 

> 

>

jeremydmiller:
Hey everyone, this is now in the official "3.0" branch.

oskardudycz:
🎉 🎉 🎉

jacobpovar:
awesome!

Issue #1018, opened 2018-05-17
Unique Index Schema Update Assert Failure


afmuller:
Hi

See below program.

```
using Marten;
using Marten.Schema;
using System;

namespace Marten1
{
    public class Invoice
    {
        public Guid Id { get; set; }
        public Guid EntityId { get; set; }
        public string RecordId { get; set; }
        public Guid Version { get; set; }
        public string Principal { get; set; }
        public string DocType { get; set; }
    }

    public class MartenRegistry : Marten.MartenRegistry
    {
        public MartenRegistry()
        {
            For<Invoice>()
                .Duplicate(x => x.EntityId)
                .Duplicate(x => x.RecordId)
                .UniqueIndex(UniqueIndexType.DuplicatedField, x => x.EntityId, x => x.RecordId);
        }
    }

    class Program
    {
        private static IDocumentStore createDocumentStore()
        {
            var store = DocumentStore.For(_ =>
            {
                _.Connection("host=localhost;database=marten_test;password=velocity;username=velocity");
                _.Schema.Include<MartenRegistry>();
                _.PLV8Enabled = false;
                _.AutoCreateSchemaObjects = AutoCreate.All;
            });

            return store;
        }

        static void Main(string[] args)
        {
            var store = createDocumentStore();
            store.Schema.ApplyAllConfiguredChangesToDatabase();
            try
            {
                store.Schema.AssertDatabaseMatchesConfiguration();
            }
            catch (Exception ex)
            {
                Console.WriteLine(ex.Message);
            }
            Console.ReadLine();
        }
    }
}

```

Even though the database (and the unique index) is successfully created (because of ApplyAllConfiguredChangesToDatabase) the AssertDatabaseMatchesConfiguration always fails with this error message:
```
Configuration to Schema Validation Failed! These changes detected:

-- Marten.Storage.DocumentTable
drop index public.mt_doc_invoice_idx_entity_id_record_id;
CREATE UNIQUE INDEX mt_doc_invoice_idx_entity_id_record_id ON public.mt_doc_invoice ("entity_id", "record_id");;

```
The schema migration does not appear to be detecting that the unique index already exists. 

I am running Marten 2.8.0.

jacobpovar:
[...]

PR #1046, opened 2018-07-17
Fix GH-1018

mysticmind:
- Fix bug GH-1018 to remove extra space between index keys from actual DDL statament for multi key indices
in Matches function which results in mismatch
- Add unit test


jeremydmiller:
Oh sweet, thanks!

PR #1053, opened 2018-07-22
add customization of event projection key name

jacobpovar:
This PR adds customization of key name for records in mt_event_progression table used by Deamon to track progress.  Full projection class typename is still used by default.


jeremydmiller:
@jacobpovar Sorry for the long delay on this one. I picked up your commit w/ a cherry pick, so the PR itself won't get merged.

jacobpovar:
that's ok. thanks!

PR #1055, opened 2018-07-24
Only check table indices against the configured schema. This

jokokko:
* Makes `IDbObjects.AllIndexes` implementation respect the documented behavior (get indices for the current tenant)
* Allows indice check not to fail in PG with `XX000: could not open relation with OID ..` when changes happen in unrelated schemas (e.g. schema is dropped so that an OID no longer exists when `pg_get_indexdef` executes)
  * Concurrent test scenarios where schemas are treated as "disposable", e.g. only created for the duration of tests, won't sporadically fail in schema checks & patch generation


jeremydmiller:
Got it. I think we're gonna have to have a 2.10.0 pretty soon that'll have this one

PR #1056, opened 2018-07-24
Fix #1014. Default mode for the store now AutoCreate.CreateOrUpdate. …

### LINKED ISSUE #1014: Data mysteriously disappeared
Over the weekend I've migrated our database from RavenDb to Marven and the overall process was extremely smooth (great job!). But today there was a very frightening accident. So I've been adding data to the database using BulkInsert (a couple of thousand documents per second) and it everything was fine. But then I tried to read data from the same database and two things happened:
1. The program saving data using BulkInsert stoped working: every BulkInsert resulted in `System.InvalidOperationException: This NpgsqlTransaction has completed; it is no longer usable` It seems like it could be connected to pooling, so I disabled it and it seems to work, at least for now
2. All the data from the table I've been reading from, disappeared. Taking into account the previous error, it looks like for all BulkInserts there was a global transaction that was rolled back. I find it very unlikely, just wanted to share this hoping may be more experienced Marten users could share their insights.

jokokko:
…Docs adjusted accordingly.

### DIFF
--- a/documentation/documentation/documents/index.md
+++ b/documentation/documentation/documents/index.md
@@ -14,9 +14,9 @@ with Marten can be as simple as opening a `DocumentStore` to your new Postgresql
 The code sample above sets up document storage against the Postgresql at the connection string you supplied. In this "quickstart" configuration,
 Marten will build new database tables and functions for each new document type as needed **if those database schema objects do not already exist**.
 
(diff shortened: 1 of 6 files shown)


jeremydmiller:
I'm going to pull this one into the 3.0 branch, hopefully in the next couple days. Thanks for doing this one!

oskardudycz:
Imho this would be nice to get into 3.0 finally :)

Issue #1042, opened 2018-07-12
DROP INDEX CONCURRENTLY invalid within a transaction block on Postgres 10


droach282:
Hello,

We're having a rough time running the update scripts created by marten command line tooling.  If there are any indexes being created in the scripts, they universally fail with the following message:

`psql:test.sql:82: ERROR:  DROP INDEX CONCURRENTLY cannot be executed from a function or multi-command string`

In doing some research on this, I found the following on [PostgreSQL's documentation](https://www.postgresql.org/docs/current/static/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY):

> a regular CREATE INDEX command can be performed within a transaction block, but CREATE INDEX CONCURRENTLY cannot.

In looking at the scripts, all of the offending CREATE INDEX CONCURRENTLY statements appear to be contained within transaction blocks.

mysticmind:
@droach282 could you please provide a repro unit test for this issue?

droach282:
Sure, here are the steps that I've used to reproduce.

1. Create a new database in Postgres 10 for use with Marten, with plv8 installed.
2. Declare indexes on some document types with "IsConcurrent = true" enabled.
3. Create a command line app with the Marten.CommandLine library, and run it with the "patch" command.
4. Attempt to run the generated patch.

In the generated script, you'll see that there is a concurrent index declaration within a transaction block, which Postgres 10 chokes on with the error listed above.

The "apply" command seems to work without incident.

droach282:
Ok, we can certainly do that at this point, but that makes me nervous going
forward. A major reason that were running from RavenDB is how expensive
index creation is, and the idea of locking a table in Postgres to create an
index seems like it would be just as bad. Am I right to be worried?

Miggleness:
@droach282 a viable workaround is to continue using the `IsConcurrent` flag and modify the resulting patch file to remove the concurrent index creation.

mysticmind:
@droach282 I have added an enhancement via PR #1058. With this by default system creates non-transactional script for patch and dump. If you pass flag `-t` or `--transactional-script` then it creates a transactional script. For your case with creating concurrent index, the non-transactional script should work.

I will wait for @jeremydmiller to review and confirm the changes.

mysticmind:
@jeremydmiller Can we close this against PR #1058?

jeremydmiller:
@mysticmind Sorry I was slow on this one. Yes. Closed by #1058

PR #1058, opened 2018-07-31
Add option to create transactional or non transactional script for patch and dump

mysticmind:
By default Marten always creates transactional script for patch and dump. When user choose to add indexes with `IsConcurrency=True`, create or drop index concurrently with a transaction block results in a script error as raised in #1042. 

This PR adds functionality to create transactional or non-transactional script using flag `[-t, --transactional-script]` passed to patch or dump. This way, user can choose to wrap a transaction or not.

With introduction of these changes, by default patch and dump will produce non-transactional script. Only with the flag passed, it produces transactional script. This changes the current default behaviour.


mysticmind:
@jeremydmiller Please note that the TravisCI failure is unrelated to this PR.

jeremydmiller:
@mysticmind Gotcha.

PR #1059, opened 2018-08-01
Add space below previous and next links

mderriey:
Hi 👋,

I ❤ the extensive docs but I find it hard to see and click the previous and next links at the bottom of the pages.
This PR adds some space below them; I picked 21px because that's the margin defined by the `<hr>` just above the `<nav>`.

### Before

before-adding-space-below

### After

after-adding-space-below


jeremydmiller:
Cool, thank you. I'll do a refresh of the docs before the week is out.

Issue #1064, opened 2018-08-03
Make dotnet build work on non-Windows platform (*nix, MacOS)


Miggleness:
As of the moment, running `dotnet build` fails on non-Windows platforms due to net46 being among the default target frameworks.  

Small nuance, but nice to have instead of having to change the target frameworks manually.

jeremydmiller:
Fixed by #1065

PR #1065, opened 2018-08-03
Updated csprojs under Marten.sln for solution to build successfully o…

Miggleness:
Updated csprojs under Marten.sln for solution to build successfully on non-Windows platform with use of targeting package. Reference to #1064 

Working on MacOS. Not tested on *nix.  I can proceed to apply the change to Marten 3.0 and other solution files if @jeremydmiller is OK with the change.


jeremydmiller:
@Miggleness Cool, thank you. Don't worry about the 3.0 branch, I'll rebase it on top of master some time pretty soon when I cut 2.10.

PR #1078, opened 2018-08-29
using Streams during an inline projection overrides event ids with Guid for string based events

david-driscoll:
This is a very subtle bug that I'm not 100% sure the best way to add a proper unit test for. 

Basically if you use `EventPage.Streams` during an inline projection, when using `StreamIdentity.AsString` the underlying events get updated to have both an `StreamId` Guid and a `StreamKey` which causes some strange issues later down the road.

I should have more details in a day or two, when I can get back to the problem and try to get a good repro around it.


david-driscoll:
Going to close for now, I might have missed something, I will reopen if my local testing pops up the specific bug again.

Issue #1081, opened 2018-09-07
Marten does not work when using Npgsql 4.0.2


jinhong-:
If ever you install Npgsql explicitly to version 4.0.2, you will be presented with the following error:
`Exception thrown: 'System.NullReferenceException' in Marten.dll: 'Object reference not set to an instance of an object.'
   at Marten.Util.TypeMappings.ToDbType(Type type)
   at Marten.Schema.DuplicatedField..ctor(EnumStorage enumStorage, MemberInfo[] memberPath)
   at Marten.Schema.DocumentMapping.DuplicateField(MemberInfo[] members, String pgType, String columnName)
   at Marten.Events.EventQueryMapping..ctor(StoreOptions storeOptions)
   at Marten.Storage.StorageFeatures.PostProcessConfiguration()
   at Marten.DocumentStore..ctor(StoreOptions options)
   at Marten.DocumentStore.For(Action`1 configure)

Probably got to do with this line:
https://github.com/JasperFx/marten/blob/eb079897138c62d5d765323769e000d27d9c1667/src/Marten/Util/TypeMappings.cs#L33

cocowalla:
If you want Npgsql 4.x, you need to use Marten 3.x.

If you mean you're using Marten 2.x and not explicitly referencing Npgsql 4.x, but another dependency _is_, that's not a Marten-related issue. You could try explicitly adding a reference to Npgsql 3.x, but that's not going to work if the package causing your problem requires Npgsql 4+.

jinhong-:
I understand that for 2.x only npgsql 3.x. My project isn't referencing a dependency that requires npgsql 4.x. I am not saying it is a marten related issue. What i am saying is that as an improvement, it would be good to configure the Marten so that it prevents npgsql 4.x from getting installed when it is installed. I will submit a PR

cocowalla:
OK, I misunderstood - adding some kind of maximum version Nuget constraint seems like a good idea, to me at least :) (I'm not the maintainer, just trying to help out)

jeremydmiller:
We know, and it's been a massive PITA for everyone. Your PR went into Marten 2.10.1, and I'm getting 3.0 ready right now

PR #1082, opened 2018-09-07
Restrict Npgsql version >=3.2.5 and <4.0

jinhong-:
This would allow nuget package that is published to limit the version of npgsql, this way we can enforce the version during package installation of Marten without waiting for runtime errors to occur
Related to #1081


oskardudycz:
Looks good to me. @jeremydmiller I think that merging this change would solve common issue that is visible on the gitter and lot of confusion. Especially after we finish with the final version 3.0.

Issue #930, opened 2017-12-15
AsPageList could be done with one database query


karelg:
AsPageList is executing two database queries, but can actually be implemented to use only one using the [Stats Linq operator](http://jasperfx.github.io/marten/documentation/documents/querying/paging/).

mysticmind:
@karelg I don't think you could pull it off with one query since `Stats` will provide the total records for the specific query in question. For fetching paged records, we will need to find out the total records to compute the number of pages based on the page size and then fetch the records for a specific page.

karelg:
@mysticmind I don't know if I'm missing something here, but I actually don't see any reason, why we can not fetch the page before computing the number of pages? _PageNumber_ and _pageSize_ should be sufficient input for that?

mysticmind:
@karelg Probably I should have detailed this in my earlier comment, `AsPagedList` does a little more to provide you details such as PageCount, HasPreviousPage, HasNextPage, IsFirstPage, IsLastPage etc which requires 2 calls. This nicely dovetails with the requirements of  a navigation type interface and also can validate if a page number is valid.

Agree with you that if you are only interested in getting records for a page for a given page number and page size, you can directly use a LINQ query using `Skip` and `Take`, pull it off with a single query. You don't need to use `AsPagedList`.

karelg:
@mysticmind I misunderstood from current [documentation](http://jasperfx.github.io/marten/documentation/documents/querying/paging/) that `Stats` was actually implemented as `IQueryable` extension method, but actually it's implemented as `MartenQueryable` method, so it's not currently possible to access this in `AsPagedList`, which operates over `IQueryable`. So I think, currently there is no simple way to make `AsPagedList` work with only a single round trip. Sorry for confusion.

mysticmind:
@karelg Thanks for your reply. Can you please go ahead and close this issue?

karelg:
@mysticmind Thanks! I will close this issue, given there is no trivial way for now to make 'AsPagedList' work with single round trip to database! But we can always easily implement a custom optimized paging solution for our applications using the 'Stats' method if performance is an issue.

mysticmind:
@karelg FYI, I have refactored the AsPagedList implementation to use QueryStatistics which will result in a single round trip to db.

PR #1085, opened 2018-09-11
Refactor and simplify the PagedList implementation

mysticmind:
- Refactor and simplify the PagedList implementation to use
QueryStatistics to use single db query GH-930
- BREAKING CHANGE: AsPagedList and AsPagedListAsync IQueryable<T> is
changed to IMartenQueryable<T>
- BREAKING CHANGE: some of the properties in IPagedList<T> has change been
changed from int to long
- Update unit tests


jeremydmiller:
@mysticmind If this is a breaking change, can you apply this to the 3.0 branch?

mysticmind:
I just relooked at the changes which I have marked as "breaking change":
- AsPagedList and AsPagedListAsync extension method signature with first parameter changed from IQueryable to IMartenQueryable. Even though the method signature have changed, it is still compatible and won't break existing users.
- IPagedList interface change with properties changed from int to long also won't break existing end users.

Given the above details, I think to have incorrectly commented as "breaking change". Can we merge into master itself?

karelg:
@mysticmind Unfortunately signature change will break things, if we are filtering and ordering results with LINQ before we are calling _AsPagedList_. I think, we should not change the signature, instead we could just check inside _PageList_, if _IQueryable_ is an instance of _IMartenQueryable_ and use the _QueryStatistics_ from it for more efficient query and just fall back to old implementation, if it is not. What do you think?

mysticmind:
@karelg good idea, will change the implementation accordingly.

mysticmind:
@karelg I had a look at it, doing a fallback will be problematic since the IPagedList also has some signature changes. In theory these are still breaking changes since the signature have changed. I think it is better to be pushed to 3.0 branch as suggested by @jeremydmiller

mysticmind:
Based on my prior comments, closing this. I will send a fresh PR for 3.0 branch

PR #1088, opened 2018-09-14
Refactor and simplify the PagedList implementation

mysticmind:
- Refactor and simplify the PagedList implementation to use
QueryStatistics to use single db query GH-930
- BREAKING CHANGE: AsPagedList and AsPagedListAsync IQueryable<T> is
changed to IMartenQueryable
- BREAKING CHANGE: some of the properties in IPagedList<T> has change been
changed from int to long
- Update unit tests


mysticmind:
@jeremydmiller I have resubmitted the PR to 3.0 branch.

kasparkallas:
I believe the more correct name for the method is "ToPagedList" instead of "AsPagedList" because it executes the queryable and transfers the result to a new collection. Using "AsSomething" is more suitable for methods that keep the original object reference and cast it to a different shape (like "AsQueryable" and "AsEnumerable"), as far as I know.

mysticmind:
I think it is even more unambiguous  if we simply call it as `PagedList`

mysticmind:
@kasparkallas I went ahead and changed to `ToPagedList` and `ToPagedListAsync` as per your suggestion.

kasparkallas:
@mysticmind Thank you so much!

At the moment the drawback with the new implementation is that you can not have additional LINQ filtering after calling the paging method:

```
_documentSession.Query<Target>()
    .ToPagedList(pageNumber, pageSize).Where(t => t.Number > 50); 
// Does filtering in memory.

_documentSession.Query<Target>().Where(t => t.Number > 50)
    .ToPagedList(pageNumber, pageSize); 
// Does not work because the paging method only works with IMartenQueryable at the moment 
// because .Where() returns IQueryable.
```

From talking to @karelg I believe that it should be possible to cast IQueryable to IMartenQueryable. If that works then that would be the cleanest and most powerful api I believe:
```
_documentSession.Query<Target>()
    .Where(t => t.Number > 50).ToPagedList(pageNumber, pageSize); 
// This work as long as the queryable starts off of Marten.
```

mysticmind:
@kasparkallas yeah, you are right. Let me dig a little more on this.

mysticmind:
@jeremydmiller I believe this PR is now *ready* for your review and pull into the 3.0 branch.

PR #1095, opened 2018-09-27
Add Bullseye based .NET Core build utility tool and misc changes

mysticmind:
- Add Bullseye based .NET Core build utility project (rake equivalent) and helper build
files (.cmd,.sh and .ps1)
- Move docs.csproj to tools\stdocs
- Update rake tasks for stdocs to point to the new path for docs.csproj
- Update StoryTeller to use v5.2.0
- Update rake tasks for storyteller and open_st commandline to use StoryTeller v5.2.0
- Fix to check not null before calling dispose on _daemonHelper on Marten.StoryTeller


oskardudycz:
@mysticmind will go through that today evening or tomorrow morning

mysticmind:
@oskardudycz sure, thanks!

oskardudycz:
For me it looks good and clean, although I'm not neither a Rake specialist nor Bullseye.

@mysticmind besides the build cleanup what's the main advantage? Easier build scripts mantainence?

mysticmind:
@oskardudycz See my comments below:

- Currently we use the Rake task `ci` task for running in CI environments i.e. both Appveyor and TravisCI. Also we use several of the other tasks such as `docs` to run docs locally, run `pack` to create the nuget packages, run StoryTeller etc.
- Rake is a Ruby based build tool. Anyone using it will require to install Ruby and use `gem` to install Rake. Even though it is a one time thing, many .NET developers are not that familiar and it adds a little learning curve to prepare their development environment. Also if you look at the TravisCI config, I had to specifically add few lines to install Ruby and Rake to use the rake `ci` task
- The new Bullseye based build tool is .NET Core based and does not require any other development dependencies to be installed. It has got all the required tasks which are available in the Rake build.
- Note that you can easily work with Marten without the a build tool. Especially for maintainers,  build tool eases the process of reducing the number of keystrokes required to run a task and also don't need to remember to type down the commands with all parameters.

In summary, the idea is to replace the Rake with the new build tool and transition to use the new one for CI environments which is a next step.

oskardudycz:
@mysticmind there were times in my career that I was Ruby on Rails developer, long time... ;)

I like the idea, to be true, for me this Rake configuration was always something that I skipped. I usually prefer not to mess my local workstation with too many dependencies and configration stuf. I wasn't using Bullseye yet, but it seems that I definitely should finally try.

I'm approving that. Would love to see what @jeremydmiller thinks about that.

oskardudycz:
For me it's fine 👍

adamralph:
@mysticmind out of interest, why did you remove `set -euo pipefail` from `build.sh`?

mysticmind:
@adamralph I was doing some tests using the build.sh file in a Ubuntu docker container and the file was not running at all. I had commented out that line for troubleshooting. Apparently I figured out that it had the CRLF line feed rather than LF which was causing issues. I will reinstate it back.

mysticmind:
Got a go ahead from @jeremydmiller as well, went ahead merged this PR

PR #1098, opened 2018-10-01
Reimplementing full text search on top of the latest code

CodingGorilla:
This adds an implementation of full text searching on top of the latest Marten code base.  Replacing #945 .

### DIFF
--- a/src/Marten.Testing/Acceptance/full_text_index.cs
+++ b/src/Marten.Testing/Acceptance/full_text_index.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using Marten.Schema;
 using Marten.Testing.Documents;
 using Shouldly;
 using Xunit;
@@ -20,8 +21,15 @@ namespace Marten.Testing.Acceptance
                                               _.Schema.For<User>().FullTextIndex();
                                           });
 
-            using(var session = store.QuerySession())
+            using(var session = store.OpenSession())
             {
+                session.Store(new User { FirstName = "Jeremy", LastName = "Miller", UserName = "jmiller"});
+                session.Store(new User { FirstName = "Lindsey", LastName = "Miller", UserName = "lmiller"});
+                session.Store(new User { FirstName = "Max", LastName = "Miller", UserName = "mmiller"});
+                session.Store(new User { FirstName = "Frank", LastName = "Zombo", UserName = "fzombo"});
+                session.Store(new User { FirstName = "Somebody", LastName = "Somewher", UserName = "somebody"});
+                session.SaveChanges();
+                
                 var somebody = session.Search<User>("somebody");
             }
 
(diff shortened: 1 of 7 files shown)


CodingGorilla:
@oskardudycz I'm looking for ways to unit test the new search methods on the `QuerySession`, do you have any recommendations or examples that I can base the tests off of?

mysticmind:
@CodingGorilla Our CI configuration on both Appveyor and TravisCI are running on PostgreSQL 9.5, we have to bump that to use PG10. I will get this sorted for you.

CodingGorilla:
@mysticmind Thanks, but no rush, I'm still working on this and adding unit tests.  At this point, my unit tests don't work locally either; they never seem to actually create the index and I'm trying to figure out why.

oskardudycz:
@CodingGorilla regarding unit tests, I think that testing QuerySession won't bring much value. For that I would recommend integration tests.

What can be unit tested are public members:
- different cases for `IndexName` generation (so all that ifs in getter),
- DDL generation - which you already tested,
- Matching Indexes (eg. if Index was already defined for Document)

Btw. regarding [my comment](https://github.com/JasperFx/marten/pull/945#discussion_r221584360) from old PR,  about moving logic from getter to setter. I think that it's doable if we set default value for `_indexName`. Of course I won't be pushing for that if it's the same logic as currently existing for `ComputedIndex`, but that's something that I noted as potential prace for refactoring.

CodingGorilla:
@oskardudycz I'm not quite sure the best way to check for matching indexes, can you point to any other examples of tests for that?

oskardudycz:
@CodingGorilla I think that we don't have such sample yet. I'll try to prepare tommorow something for you in separate PR removing also some technical debt around that ;) It seems that PRs are motivating for both sides ;)

CodingGorilla:
It seems like that's not the responsibility of the index itself, wouldn't that be handled by the `DocumentMapping` since it owns the collection of indexes?

oskardudycz:
Well, yes and now. In general I believe it's good to have public API tested (in this case public members). Iimho it would be nice to have dedicated unit tests for the FullTextIndex by itself.  
But!
If you find it easier through `DocumentMapping` - I'm fine. 
The main logic of comparing table schema is happening in TableDelta. So I'm also fine of placing those test in [TableDeltaTests](https://github.com/JasperFx/marten/blob/master/src/Marten.Testing/Storage/TableDeltaTests.cs).

oskardudycz:
@CodingGorilla as I see the `Matches` contains business logic, so imho it's worth to test it. Especially that as I said, it can be compared to any of the other indexes. Cases that we'd need to imho check at least:
- comparing with other full text index - checking if it returns true,
- comparing to other index definition that shouldn't match - so checking for false scenario.

Imho this is the good place for unit testing, because they would be easy to write and give business value. 
This part is imho really important to test, because we wouldn't like to have scenario, where someone adds full text index and get messy with other indexes that he had already defined.

But as I said above. If you find better way for testing those cases. I'm fine with that.

For sure the most important tests are:
- integration tests checking if the full text search queries work,
- proper database query definition.

CodingGorilla:
Ok, you want to test the `IIndexDefinition.Matches()`, that makes sense, I'm just not sure how to construct the bits that are required in order to test that method specifically.  

Realistically there should only ever be one full text index per document type, there's never a reason to have more than one that I can think of.  I considered adding some logic checking the `DocumentMapping`, but the question was: what to do if someone tries to add another?  I didn't like the idea of throwing an exception, and in reality it doesn't _hurt_ to have more than one, it's just a waste.

oskardudycz:
@CodingGorilla I think that it'd be nice to force someone to keep only one full text index, so Imho throwing some meaningfull exception is nothing bad, although currently this is happening silently, so maybe it's better to keep convention. But that's a valid point that we should reconcider globally for indexes.

What I'm concerned more is that we need to make sure that when user updates config of the existing FullTextIndex then that:
- new one won't be added,
- index definition will be updated

Would be great to add integration test for that.

mysticmind:
@CodingGorilla could you please rebase this PR with Master? PG10 is available in CI env.

CodingGorilla:
@mysticmind I will try to get it done this weekend.

CodingGorilla:
@mysticmind Done.

PR #1101, opened 2018-10-03
refactor martenbuild.cs a little

adamralph:


adamralph:
It seems the CI gods are on my side today. Both builds green first time! :feelsgood:

martenbuild.cs
```diff
@@ -8,46 +8,33 @@ namespace martenbuild
...
             Target("connection", () =>
             {
-                File.WriteAllText("src/Marten.Testing/connection.txt", connection);
+                File.WriteAllText("src/Marten.Testing/connection.txt", Environment.GetEnvironmentVariable("connection"));
             });
 
             Target("install", () =>
             {
-                if (pid == PlatformID.Unix || pid == PlatformID.MacOSX)
+                if (platformID == PlatformID.Unix || platformID == PlatformID.MacOSX)
```
mysticmind: Another refactoring area I am observing is checking for the platform, have a helper method as below
```csharp
private static bool IsWindowsPlatform()
{
    var platformID = Environment.OSVersion.Platform;
    
    if (platformID == PlatformID.Unix || platformID == PlatformID.MacOSX)
    {
	return false;
    }

    return true;
}

// change the appropriate code as below to simplify it
IsWindowsPlatform ? Run("cmd.exe", "/c npm run test") : Run("npm", "run test");
```

mysticmind:
@adamralph Thanks for the refactoring changes!

martenbuild.cs
mysticmind: @adamralph I will take care of applying the above.

adamralph: @mysticmind given that the choice between `Run("npm", "...");` and `Run("cmd.exe", "/c npm ...");` is made in two places (`install` and `mocha` targets) I would consider creating a `RunNpm(string args)` method to wrap that decision.

Issue #1105, opened 2018-10-05
Supported child collection formats when passed an array value to a non-array property


Nysosis:
Marten Version: 3.0.1
Framework: DotNetCore 2.1
Postgres Version: PostgreSQL 9.6.1, compiled by Visual C++ build 1800, 64-bit

There seems to be a discrepancy on the way that collections are being serialised into the JSON document.

Given this class:
```
	public class MartenData {
		public int Id { get; set; }

		public string[] Array { get; set; }
		public IEnumerable<string> Enumerable { get; set; }
		public IEnumerable<string> IListAsArray { get; set; }
		public IEnumerable<string> IListAsList { get; set; }
		public List<string> List { get; set; }
	}
```

and this instance:
```
		public async Task Create() {
			this._Session.Store(new MartenData {
				Id = 1,
				Array = new[] { "a", "b" },
				Enumerable = new[] { "a", "b" },
				IListAsArray = new[] { "a", "b" },
				IListAsList = new List<string> { "a", "b" },
				List = new List<string> { "a", "b" }
			});

			await this._Session.SaveChangesAsync();
		}
```

The value saved into the database is:
```
{
    "IListAsList": [
        "a",
        "b"
    ],
    "List": [
        "a",
        "b"
    ],
    "IListAsArray": {
        "$type": "System.String[], System.Private.CoreLib",
        "$values": [
            "a",
            "b"
        ]
    },
    "Array": [
        "a",
        "b"
    ],
    "Enumerable": {
        "$type": "System.String[], System.Private.CoreLib",
        "$values": [
            "a",
            "b"
        ]
    },
    "Id": 1
}
```

Note on `IListAsArray`, and `Enumerable` have extended data about the type, this is resulting on `where` statements on these child properties to not work.

oskardudycz:
You can try to setup TypeNameHandling settings of Json serializer to None https://www.newtonsoft.com/json/help/html/SerializeTypeNameHandling.htm 

I have

oskardudycz:
@Nysosis I added tests for your case https://github.com/JasperFx/marten/pull/1107/files

Could you compare that they reflect your query and your json serializer configuration?

oskardudycz:
@Nysosis Ok, my tests were false positive. Now they're failing. I'll look on that.

oskardudycz:
[...]

PR #1107, opened 2018-10-06
Added Tests for Query with inner query

oskardudycz:
Added tests for issue https://github.com/JasperFx/marten/issues/1105


src/Marten.Testing/Linq/query_with_inner_query.cs
```diff
@@ -0,0 +1,119 @@
...
+                results.All(e => e.Enumerable.Contains(SearchPhrase)).ShouldBeTrue();
+            }
+        }
+
+        private void ConfigureStoreToHaveJsonTypeNameHandlingNone()
+        {
+            var serializer = new JsonNetSerializer();
+            serializer.Customize(s =>
+            {
+                s.TypeNameHandling = Newtonsoft.Json.TypeNameHandling.None;
```
mysticmind: I believe the same can be applied as an attribute on the IEnumerable properties on the class to achieve the same rather than customizing the serializer at the store level. Thoughts?

oskardudycz: @mysticmind good catch, I'll add next case checking that

robert-skarzycki:
@oskardudycz, what about one more level down, i.e. having creation of test data looking like this:
```
 public static TypeWithInnerCollections Create(params string[] array)
        {
            return new TypeWithInnerCollections()
            {
                Id = Guid.NewGuid(),
                Array = array,
                Flatten = array.Aggregate((i, j) => i + j),
                Enumerable = array.AsEnumerable(),
                IEnumerableFromArray = array,
                IEnumerbaleFromList = array.ToList(),
                List = array.ToList(),
                IList = array.ToList(),
                IReadonlyCollection = array.ToList(),
                IReadonlyCollectionOfInnerClasses = new List<TypeWithInnerCollections>
                    {
                        new TypeWithInnerCollections()
                        {
                            Id = Guid.NewGuid(),
                            Array = array,
                            Flatten = array.Aggregate((i, j) => i + j),
                            Enumerable = array.AsEnumerable(),
                            IEnumerableFromArray = array,
                            IEnumerbaleFromList = array.ToList(),
                            List = array.ToList(),
                            IList = array.ToList(),
                            IReadonlyCollection = array.ToList(),
                            IReadonlyCollectionOfInnerClasses = new List<TypeWithInnerCollections>
                            {
                                new TypeWithInnerCollections()
                                {
                                    Id = Guid.NewGuid(),
                                    Array = array,
                                    Flatten = array.Aggregate((i, j) => i + j),
                                    Enumerable = array.AsEnumerable(),
                                    IEnumerableFromArray = array,
                                    IEnumerbaleFromList = array.ToList(),
                                    List = array.ToList(),
                                    IList = array.ToList(),
                                    IReadonlyCollection = array.ToList(),
                                }
                            }
                        }
                    }
            };
```
?

Then having a query over this extra-level leaves would fail:
```
await query.Query<TypeWithInnerCollections>()
                .Where(x => x
                .IReadonlyCollectionOfInnerClasses
                .Any(y => y.IReadonlyCollectionOfInnerClasses
                .Where(e => e.Flatten == "onetwo")
                .Any())
                || x
                .IReadonlyCollectionOfInnerClasses
                .Any(y => y.IReadonlyCollectionOfInnerClasses
                .Where(e => e.Flatten == "twothree").Any())
                )
                .ToListAsync();
```
with exception stating that only `Contains` is allowed inside `Any`.

P.S. That's what I meant by asking about subqueries on Gitter: https://gitter.im/JasperFx/marten?at=5c66d8b64bafd97ca3f8f457

oskardudycz:
@robert-skarzycki I agree, current support for InnerCollection queries is far from being ideal. I'll try to revisit that after I finish my work on NodaTime support. But the fix won't be quick, probably no sooner than in version 3.6...
For now as a workaround I would suggest to use getters with flattened strucures. It's workaround as this will make data denormalized but it would work.

mysticmind:
@oskardudycz I have added `CollectionToArrayJsonConverter<T>` under `Marten.Util` and updated the unit tests to apply attribute on collection properties `[JsonConverter(typeof(Marten.Util.CollectionToArrayJsonConverter<TypeWithInnerCollections>))]`. All tests are green.

With this fix, we can close #1105 and #1192. We will also need to add some documentation on this.

oskardudycz:
@mysticmind I made few additional improvements:
- Refactored CollectionToArrayJsonConverter to not need generic parameter,
- added JsonNetContractResolver to allow global setting of storing collections as arrays

Could you have a look and share your thoughts? 😄

mysticmind:
@oskardudycz Looks like the tests are failing with some recent changes, can you please have a look?

oskardudycz:
@mysticmind should be good now. Too long document class name ;)

src/Marten/Util/CollectionToArrayJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+using Newtonsoft.Json;
+
+namespace Marten.Util
+{
+    /// <summary>
+    /// Serialize collection type property to JSON array using a custom Newtonsoft.Json JsonConverter
+    /// Note that without using custom `JsonConverter`, `Newtonsoft.Json` stores it as $type and $value.
+    /// Or you may need to resort to `Newtonsoft.Json.TypeNameHandling.None` which has its own side-effects
+    /// </summary>
+    /// <typeparam name="T">Type of the collection</typeparam>
```
mysticmind: Remove this document comment, not applicable any more

src/Marten/Util/CollectionToArrayJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class CollectionToArrayJsonConverter : JsonConverter
```
mysticmind: I think it would make sense to rename this class to `JsonNetCollectionToArrayJsonConverter` and move it under `Marten.Services.Json` namespace

src/Marten/Services/JsonNetSerializer.cs
```diff
@@ -119,22 +123,27 @@ public Casing Casing
...
-                    _serializer.ContractResolver = new DefaultContractResolver { NamingStrategy = new SnakeCaseNamingStrategy() };
-                    _clean.ContractResolver = new DefaultContractResolver { NamingStrategy = new SnakeCaseNamingStrategy { ProcessDictionaryKeys = true, OverrideSpecifiedNames = true } };
-                }
+                _serializer.ContractResolver = new JsonNetContractResolver(_casing, CollectionStorage);
+                _clean.ContractResolver = new JsonNetContractResolver(_casing, CollectionStorage);
+            }
+        }
+
+        /// <summary>
+        /// Specify whether properties in the JSON document should use Camel or Pascal casing.
```
mysticmind: Looks like a copy & paste glitch, fix the xml comment ;-)

oskardudycz:
@mysticmind comments applied ;)

PR #1110, opened 2018-10-09
 fixes #1109 async batch updates sent fifo

### LINKED ISSUE #1109: SaveChangesAsync sends batchCommands in reverse order
If first doing a DeleteWhere followed by upserting more documents than UpdateBatchSize value, at least one of the upserts will be clobbered when using SaveChangesAsync.

Basically, this is the issue reported and fixed in #987 but async.

gregfiske:
The PR is off of master (3.0) but I _really_ would like this fixed for 2.10.  What would make that easier to do - should I make a PR off of that branch?  I'd rather not go through regression testing for npgsql 4 on my app at the moment if I don't have to.

### DIFF
--- a/src/Marten.Testing/Util/update_batch_Tests.cs
+++ b/src/Marten.Testing/Util/update_batch_Tests.cs
@@ -71,5 +71,23 @@ namespace Marten.Testing.Util
                 session.Query<Target>().Count().ShouldBe(100);
             }
         }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


mysticmind:
@gregfiske I think we can possibly cherry pick the changes and apply it on 2.0 branch. Let me check on it. so eventually you are looking at a bug fix release on 2.x i.e. v2.10.2

mysticmind:
@gregfiske v2.10.2 is available Nuget

Issue #1113, opened 2018-10-10
Docs: Add documentation for logging and timing queries related questions


mysticmind:
Cover documentation in  section http://jasperfx.github.io/marten/documentation/troubleshoot/

mysticmind:
Closed via #1115

PR #1115, opened 2018-10-11
Add mention of Npgsql logging and perf counters under FAQ & troublesh…

jokokko:
…oot. #1113.


jokokko:
For hygiene reasons, I think I won't merge my own PRs.

PR #1116, opened 2018-10-11
Eliminate usage of reflection in TypeMappings - fixes 1049

### LINKED ISSUE #1049: Eliminate usage of reflection in TypeMappings
Bits of code we need in Npgsql are inexplicably private, so we've always used Reflection to get at it. That's slow, and their code has been buggy in npgsql 4, so let's just write out own mappings from .Net type to NpgsqlDbType

mysticmind:
- Eliminate usage of reflection in TypeMappings by using INpgsqlTypeMapper.Mappings - fixes #1049
- Update logic of PgTypes to use type mappings from INpgsqlTypeMapper.Mappings

Note: With the above fix, we don't need any change from Npgsql folks. @oskardudycz you can review and close the issue raised in Npgsql repo.

### DIFF
--- a/src/Marten/Util/TypeMappings.cs
+++ b/src/Marten/Util/TypeMappings.cs
@@ -16,7 +16,7 @@ namespace Marten.Util
 
         static TypeMappings()
         {
-            // Create the CLR type type to NpgsqlDbType and PgTypeName mapping from exposed INpgsqlTypeMapper.Mappings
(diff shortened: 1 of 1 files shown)


jeremydmiller:
I'm pulling this one in. Looks perfectly fine to me

oskardudycz:
@mysticmind great job!

PR #1124, opened 2018-10-18
Implement RetryPolicy extension point feature to handle transient errors - fixes #1119

### LINKED ISSUE #1119: Retry when transient errors occur
Is there anything built in to Marten to enable automatic retries when transient errors occur, or some kind of extensibility point where it could be plugged in?

mysticmind:
- Add `IRetryPolicy` and `NulloRetryPolicy`
- Implemement RetryPolicy to configure the same via `StoreOptions` and wrap the Npgsql calls in `ManagedConnection` using retry policy - fixes #1119
- Update `ManagedConnection` unit tests to retrofit retry policy changes.

### DIFF
--- a/src/Marten.Testing/Services/ManagedConnectionTests.cs
+++ b/src/Marten.Testing/Services/ManagedConnectionTests.cs
@@ -12,7 +12,7 @@ namespace Marten.Testing.Services
         [Fact]
         public async Task increments_the_request_count()
         {
-            using (var connection = new ManagedConnection(new ConnectionSource()))
+            using (var connection = new ManagedConnection(new ConnectionSource(), new NulloRetryPolicy()))
             {
                 connection.RequestCount.ShouldBe(0);
 
@@ -56,7 +56,7 @@ namespace Marten.Testing.Services
         {
             var ex = new DivideByZeroException();
             var logger = new RecordingLogger();
-            using (var connection = new ManagedConnection(new ConnectionSource()) {Logger = logger})
+            using (var connection = new ManagedConnection(new ConnectionSource(), new NulloRetryPolicy()) {Logger = logger})
             {
                 Exception<DivideByZeroException>.ShouldBeThrownBy(() =>
                 {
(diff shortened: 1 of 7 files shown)


mysticmind:
@jokokko I don't think there is any binary compatibility change done. Can you please elaborate what you think as a binary compatibility change?

jokokko:
> 
> 
> @jokokko I don't think there is any binary compatibility change done. Can you please elaborate what you think as a binary compatibility change?

Changing the existing CTORs, rather than introducing new ones.

jeremydmiller:
@mysticmind @jokokko My opinion is that `ManagedConnection` is an implementation detail even through it's public. I think this is an additive change in terms of binary compatibility and a minor release point for SemVer.

jeremydmiller:
To @jokokko 's point, I think you get around this by just adding a ctor signature w/o the IRetryPolicy argument

jokokko:
> @mysticmind @jokokko My opinion is that `ManagedConnection` is an implementation detail even through it's public. I think this is an additive change in terms of binary compatibility and a minor release point for SemVer.

It's an additive change and can thus be kept binary compatible by introducing new CTORs (not amending current CTORs with default arguments). Right? Again, just my humble opinion, but having been burned by these kind of changes, I like to take a stance on them.

mysticmind:
okay, I will update and add new CTORs.

mysticmind:
@jokokko Please check the latest change and let me know if it satisfies binary compatibility.

jokokko:
> 
> 
> @jokokko Please check the latest change and let me know if it satisfies binary compatibility.
> 
> image
> Also see above, Rider clearly tells me that the existing CTORs are not being used in any code path.

Looks good signature-wise. Overall, having binary backwards compatibility checks for the public surface of Marten would be a really neat addition.

mysticmind:
@jeremydmiller we can plan to merge this PR once your review and give a go ahead.

PR #1125, opened 2018-10-19
Added IsSubsetOf linq support

oakio:


oakio:
Independent from my changes tests were failed. 
What else is strange that different CI-systems failed in different ways. It possible rerun CI?

mysticmind:
@oakio I have rerun the TravisCI build. Since everything is integration tests, there are few random blinking tests which we will need to check.

oakio:
I agree with an idea not to add extra extensions methods. I can leave only `ISet<>.IsSubsetOf` method linq support.

PR #1130, opened 2018-11-02
Replaced "sequential Guid algorithm" link

Maximusya:
The former link was to a no longer existing Wikipedia page (and the one you got redirected to contains a vague hint to where to read about the algorithm itself).
Marten's CombGuidIdGeneration.cs contains all the info (references and the implementation)


jeremydmiller:
got it, thanks

PR #1133, opened 2018-11-08
Allow char buffer pooling for JsonNet serializer

oakio:
Reducing GC pressure for json serialization/deserialization using character buffer pool.


oakio:
@oskardudycz 
1. Added description.
2. `JsonNetArrayPool` just wrapper for standard `ArrayPool` from `System.Buffers`, so I think there are no need extra tests.

PR #1137, opened 2018-11-15
Add a test for building a deep expression for an enum

amferguson:
For what it's worth, both versions of the test fail against 3.1, but only the `EnumStorage.AsString` does in 2.10, but it looks like the reason for the difference is because in 2.x, in the `ToExpression` method, it only tried the extra `Expression.Call` statement when `EnumStorage` was set as a string.


oskardudycz:
@amferguson I'm continuing my investigations in follow up PR: https://github.com/JasperFx/marten/pull/1139

mysticmind:
#1139 adds the unit tests required. Closing this.

PR #1139, opened 2018-11-17
WIP: Fix enum querying

oskardudycz:
Follow up of https://github.com/JasperFx/marten/pull/1137. Fixed test case but added other failing integration test

### DIFF
--- a/src/Marten/Linq/ContainmentWhereFragment.cs
+++ b/src/Marten/Linq/ContainmentWhereFragment.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using Baseline;
+using Marten.Schema;
 using Marten.Util;
 using Npgsql;
 using NpgsqlTypes;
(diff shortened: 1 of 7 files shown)


jeremydmiller:
@oskardudycz et al,  Hey all, I'm going to commit to having this done by the end of tomorrow.

mysticmind:
@oskardudycz Please have a look my code changes which fixes the issue at hand.

src/Marten/Linq/ContainmentWhereFragment.cs
```diff
@@ -98,5 +101,18 @@ public static IWhereFragment SimpleArrayContains(MemberInfo[] members, ISerializ
 
             throw new NotSupportedException();
         }
+
+        private static object GetMemberValue(MemberInfo member, object expressionValue, EnumStorage enumStorage)
+        {
+            var value = expressionValue;
+
+            if (member.MemberType is MemberTypes.Property && ((PropertyInfo)member).PropertyType.IsEnum
```
jeremydmiller: @mysticmind There should be an extension method in the codebase called `GetMemberType()` that will deal w/ the PropertyInfo vs. FieldInfo differences here. You'll want to use that so that this fix applies just as well to public Fields

jeremydmiller:
@oskardudycz @mysticmind Ah, guys, I really did want to try to get to this one. @mysticmind, I added one small comment that would generalize your fix to apply to Fields as well w/ a helper method in Marten for exactly this kind of thing.

src/Marten/Linq/ContainmentWhereFragment.cs
mysticmind: @jeremydmiller sure, will update the logic using `GetMemberType()`

mysticmind:
@jeremydmiller if you are think the changes are good now, we can merge the PR.

Issue #1141, opened 2018-11-23
Unused Jointype parameter in Join with dictionary


unusual-thoughts:
I believe https://github.com/JasperFx/marten/pull/507 missed the case of Joins with a Dictionary.

`JoinType` parameter is dropped here:
https://github.com/JasperFx/marten/blob/eb079897138c62d5d765323769e000d27d9c1667/src/Marten/Linq/MartenQueryable.cs#L110

INNER JOINS are always emitted when including with a dictionary

unusual-thoughts:
https://github.com/JasperFx/marten/pull/1142

PR #1142, opened 2018-11-23
Use JoinType parameter in Join with dictionary

unusual-thoughts:
Fix https://github.com/JasperFx/marten/issues/1141


oskardudycz:
Looks good. Thanks for this fix!+++

PR #1158, opened 2018-12-19
flow through enumstorage to base field

### LINKED ISSUE #1157: EnumStorage isn't taken into account when Duplicated Field update SQL is generated
Even though the `enumStorage` from the `StoreOptions` is passed into the` DuplicatedField` constructor (`public DuplicatedField(EnumStorage enumStorage, MemberInfo[] memberPath) : base(enumStorage, memberPath)`), that value isn't set in the `Field`'s `_enumStorage` readonly property in the base constructor.  This leads to always generating int-casting update fragments, due to the following code in `DuplicatedField`: 
```csharp
public string UpdateSqlFragment()
{
    var jsonField = new JsonLocatorField("d.data", _enumStorage, Casing.Default, Members);
    // HOKEY, but I'm letting it pass for now.
    var sqlLocator = jsonField.SqlLocator.Replace("d.", "");
   return $"{ColumnName} = {sqlLocator}";
}
```

I can submit a PR to add the `_enumStorage = enumStorage` to the Field constructor, but I am unsure about testing requirements here.

baronfel:
Fixes #1157 by flowing through the enumStorage into the readonly field.

I'm not sure how/where appropriate testing should be applied.  It seems pretty trivial to assert that the getter is changed, should I test the difference in `UpdateSqlFragment`?


baronfel:
Any pointers here as to the best way to set up a test to verify this behavior/polish this up for a merge?

oskardudycz:
@baronfel thank you for the PR. Sorry for long answer. I'll review it today.

oskardudycz:
@baronfel changes looks good. I'd suggest to werify at least if the `EnumStorage` is properly set for both of the `EnumStorage` values.
You could extend [this test](https://github.com/JasperFx/marten/blob/master/src/Marten.Testing/Schema/DuplicatedFieldTests.cs#L53) adding check there and `Theory` data simmilar to the [tests for `DuplicatedFieldEnumStorage](https://github.com/JasperFx/marten/blob/master/src/Marten.Testing/StoreOptionsTests.cs#L338). 

If you be so kind to add such scenario - that would be great

baronfel:
Thanks for the pointers @oskardudycz, I'll try to send that in today.

oskardudycz:
@baronfel cool! I'll merge it asap it's ready

baronfel:
Would you prefer using reflection to check that the field is set, or checking the output of the update sql to ensure that the type-cast is inserted where appropriate?

baronfel:
I went ahead with the 'test the generated update sql for the typecast' approach.

oskardudycz:
@baronfel great. I also prefer this solution 👍

PR #1160, opened 2018-12-23
Added functionality for apply to be called on base class

### LINKED ISSUE #1159: Projections Apply Event is not called for base class, when Apply for child class is not present.
When an aggregation has an Apply event for a base class, but no Apply event for the Event class itself, the base Apply is not called.

malscent:
if apply method for event missing. Fixes #1159


malscent:
While a fun exercise, ultimately, this was decided that the best course of action was to create a custom IProjection and custom IAggregator.

PR #1162, opened 2019-01-03
Fix duplicate field null scenarios

### LINKED ISSUE #1155: Exceptions thrown for various Duplicate field scenarios involving nulls
I've found a couple of scenarios where exceptions are being thrown when storing documents due to null values for Duplicate fields.
```csharp
[Fact]
public void when_field_is_null_due_to_nesting()
{
    StoreOptions(_ => _.Schema.For<Target>().Duplicate(t => t.Inner.Number));

    {
        session.Store(new Target
        {
            Number = 1,
            Inner = null
        });

        session.SaveChanges();
    }
    
    {
        session.Query<Target>().Where(x => x.Number == 1)
            .ToArray()
            .Select(x => x.Number)
            .ShouldHaveTheSameElementsAs(1);
    }
}
```
Throws a NullReferenceException.

```csharp
[Fact]
public void when_string_enum_is_null_due_to_nullable_type()
{
    StoreOptions(_ =>
    {
        _.Serializer(new JsonNetSerializer{EnumStorage = EnumStorage.AsString});
        _.Schema.For<Target>().Duplicate(t => t.NullableColor);
    });

    {
        session.Store(new Target
        {
            Number = 1,
            NullableColor = null
        });

        session.SaveChanges();
    }
    
    {
        session.Query<Target>().Where(x => x.Number == 1)
            .ToArray()
            .Select(x => x.Number)
            .ShouldHaveTheSameElementsAs(1);
    }
}
```
Throws ArgumentNullException when calling `System.RuntimeType.GetEnumName()` internally.
N.B. I added a `NullableColor` property of type `Colours?` to the Target class for this one.

I'd be happy to contribute a PR if you accept them and these scenarios are valid use cases.

richardoliver-umusic:
Fixes #1155 scenarios where duplicate fields or their ancestors can be null.
When this is the case the default value will be stored to the duplicate field.

I've extended the expression generation logic to add null checks when getting values for duplicate fields and compiled queries. 

Excerpt from comments below:
```
            // Builds expression to retrieve value including enum conversion and null checks:
            // Simple property/field                 target => target.Property
            // Enum conversion to int                target => Convert.ToInt32(target.EnumProperty)
            // Enum conversion to string             target => Enum.GetName(type, target.EnumProperty)
            // Nested property/field null checks     target => target.Inner != null ? target.Inner.Property : default()
```

I've added tests scenarios specific the the use cases I came across the issue in Bug_1155_null_duplicate_fields.cs. I'm not sure if you'd prefer these to be move into a more suitable location? The code path is also used by compiled queries, all the existing tests pass but I haven't added any additional tests for them.

Let me know what else you'd like adding/changing/thinking about.

### DIFF
--- a/src/Marten.Testing/Bugs/Bug_1155_null_duplicate_fields.cs
+++ b/src/Marten.Testing/Bugs/Bug_1155_null_duplicate_fields.cs
@@ -1,0 +1,198 @@
+﻿using System.Linq;
+using Marten.Services;
+using Xunit;
+
+namespace Marten.Testing.Bugs
+{
+    public class Bug_1155_null_duplicate_fields : IntegratedFixture
+    {
+        [Fact]
+        public void when_enum_is_null_due_to_nullable_type()
+        {
+            StoreOptions(_ =>
+            {
+                _.Serializer(new JsonNetSerializer {EnumStorage = EnumStorage.AsInteger});
+                _.Schema.For<Target>().Duplicate(t => t.NullableColor);
+            });
+
+            using (var session = theStore.OpenSession())
+            {
+                session.Store(new Target
+                {
+                    Number = 1,
+                    NullableColor = null
+                });
+
+                session.SaveChanges();
+            }
+
+            using (var session = theStore.OpenSession())
+            {
+                session.Query<Target>().Where(x => x.Number == 1)
+                    .ToArray()
+                    .Select(x => x.Number)
+                    .ShouldHaveTheSameElementsAs(1);
+            }
+        }
+
+        [Fact]
+        public void when_enum_is_null_due_to_nesting()
+        {
+            StoreOptions(_ =>
+            {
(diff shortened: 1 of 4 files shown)


src/Marten.Testing/Bugs/Bug_1155_null_duplicate_fields.cs
```diff
@@ -0,0 +1,198 @@
...
+                session.Store(new Target
+                {
+                    Number = 1,
+                    NullableColor = null
+                });
+
+                session.SaveChanges();
+            }
+
+            using (var session = theStore.OpenSession())
```
oskardudycz: As all tests have the same Asert logic - imho it would be good to move it to the private method and reuse it.

richardoliver-umusic:
Thanks all!

PR #1154, opened 2018-12-17
Supporting NodaTime

oskardudycz:
See more here: https://github.com/JasperFx/marten/issues/1027

It's still work in progress, but I wanted to get asap initial thoughts about code structure (plugin inside main repo etc.), gather code suggestion etc.

@jeremydmiller @mysticmind @jokokko what do you think?

p.s. changes are a bit polluted with `Shouldly` usage. I removed `ShouldBeTrue` and `ShouldBeFalse` from our custom implemnentation, as it's right now supported by `Shouldly`

PR #1164, opened 2019-01-08
Respect custom mappings in NPGSQL's GlobalTypeMapper

eouw0o83hf:
`Marten.Util.TypeMappings` pre-caches Type mappings in a static constructor, using the Npgsql registry. If client code depends on any custom mappings which are registered with Npgsql, they will not be picked up, since static constructors execute very early. (There may be a way to sneak a value into Npgsql before Marten gets initialized, but I was unable to find a way to make it happen.)

This change (or one like it) would allow those type mappings to be re-executed after any other custom mappings are added.

Thoughts?

### DIFF
--- a/src/Marten.Testing/TypeMappingsTests.cs
+++ b/src/Marten.Testing/TypeMappingsTests.cs
@@ -34,6 +34,14 @@ namespace Marten.Testing
             ShouldThrowExtensions.ShouldThrow<Exception>(() => TypeMappings.ToDbType(typeof(UnmappedTarget)));
         }
 
+
+        [Fact]
+        public void execute_get_pg_type_default_mappings_resolve()
+        {
+            TypeMappings.GetPgType(typeof(long), EnumStorage.AsString).ShouldBe("bigint");
+            TypeMappings.GetPgType(typeof(DateTime), EnumStorage.AsString).ShouldBe("timestamp without time zone");
+        }
+
         [Fact]
         public void execute_get_pg_type_custom_mappings_resolve_or_default_to_jsonb()
         {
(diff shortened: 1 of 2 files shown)


mysticmind:
@jeremydmiller or @cocowalla  could you please review and weigh in on this PR?

oskardudycz:
@eouw0o83hf I'm fine with those changes, but that won't change much, as we had still lot of hardcoded code related to types and querying (especially related to Date types). 

I'm currently working on simmilar stuff, so making Marten extensible to eg. support `NodaTime` library https://github.com/JasperFx/marten/pull/1154. It's not easy.

But as I said, I like this addition, it's not enough to make Marten extensible, but it's step forward. If we conjoin our changes later then it might make plugin creation in this area easier.

jeremydmiller:
My 2 cents worth:

I don't mind taking this in as is, but it's an ugly workaround that'll get by many users. What if we were to make *our* copy of the TypeMapping lazy evaluated per type? Assuming that you get your custom mappings in first before trying to do any kind of Marten operation, you should be fine.

oskardudycz:
@eouw0o83hf I confirm, it shouldn't make any collision with my changes. I think that the can go well together.

eouw0o83hf:
Fantastic. Lazy resolution code and tests are in.

oskardudycz:
Looks good to me.

oskardudycz:
I think that if @jeremydmiller will agree on that then we can merge it.

jeremydmiller:
@mysticmind My only comment is that this is probably a perfect place to use the ImTools HashMap

mysticmind:
@jeremydmiller FYI, PR #1183 changes ConcurrentDictionary to ImTools HashMap.

PR #1167, opened 2019-01-13
Full text improvements

oskardudycz:
# Description
Currently it's Full Text search allow to define only one index per document. For most of the simple cases it's enough, but it might be not efficient to index whole `data`, when only one of few fields should be filtered. It's also not enough if we'd have to support more than language and filter only texts in specific language.

## Type of change

- added possibility to define custom data config (via config parameter of AddFullTextIndex method or via selected document members),
- changed index name to contain regional config - thus allow to configure more than one FTS for different languages,
- added full text index attribute to make easier fts fields definiton,
- added support for Conjoined Tenancy.

Also added possibility to define custom index name in AddUniqueIndex method with members expression.

 [x] New feature (non-breaking change)

# How Has This Been Tested?

Added new [tests to already existing for FTS](https://github.com/JasperFx/marten/pull/1167/files#diff-3cd0582ff65551d6588c90142e70e847)

Manually checked also execution plans for:

1. Two FTS indexes for different languages and single members selected
- Table definition
```
CREATE TABLE public.mt_doc_user
(
    id uuid NOT NULL,
    data jsonb NOT NULL,
    mt_last_modified timestamp with time zone DEFAULT transaction_timestamp(),
    mt_version uuid NOT NULL DEFAULT (md5(((random())::text || (clock_timestamp())::text)))::uuid,
    mt_dotnet_type character varying COLLATE pg_catalog."default",
    CONSTRAINT pk_mt_doc_user PRIMARY KEY (id)
)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

CREATE INDEX mt_doc_user_french_idx_fts
    ON public.mt_doc_user USING gin
    (to_tsvector('french'::regconfig, data ->> 'LastName'::text))
    TABLESPACE pg_default;

CREATE INDEX mt_doc_user_italian_idx_fts
    ON public.mt_doc_user USING gin
    (to_tsvector('italian'::regconfig, data ->> 'FirstName'::text))
    TABLESPACE pg_default;
```
- query plan:

obraz

2. Single FTS index for default language and two members selected
- Table definition
```
CREATE TABLE public.mt_doc_user
(
    id uuid NOT NULL,
    data jsonb NOT NULL,
    mt_last_modified timestamp with time zone DEFAULT transaction_timestamp(),
    mt_version uuid NOT NULL DEFAULT (md5(((random())::text || (clock_timestamp())::text)))::uuid,
    mt_dotnet_type character varying COLLATE pg_catalog."default",
    CONSTRAINT pk_mt_doc_user PRIMARY KEY (id)
)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

CREATE INDEX mt_doc_user_english_idx_fts
    ON public.mt_doc_user USING gin
    (to_tsvector('english'::regconfig, ((data ->> 'FirstName'::text) || ' '::text) || (data ->> 'LastName'::text)))
    TABLESPACE pg_default;
```
- query plan:

obraz

@CodingGorilla: I would love to see your point of view about those changes as you're inital author of this feature.

I plan to also update documentation, when we have agreement on the API.

### DIFF
--- a/documentation/documentation/documents/querying/linq.md
+++ b/documentation/documentation/documents/querying/linq.md
@@ -213,7 +213,7 @@ Use `TenantIsOneOf` to query on a selected list of tenants.
 
 ## Text Search
 
-Postgres constains built in [Text Search functions](https://www.postgresql.org/docs/10/textsearch-controls.html). They enable the possibility to do more sophisticated searching through text fields. Marten gives possibility to define <[linkto:documentation/documents/configuration/full_text;title=Full Text Indexes]> and perform queries on them.
+Postgres contains built in [Text Search functions](https://www.postgresql.org/docs/10/textsearch-controls.html). They enable the possibility to do more sophisticated searching through text fields. Marten gives possibility to define <[linkto:documentation/documents/configuration/full_text;title=Full Text Indexes]> and perform queries on them.
 Currently three types of full Text Search functions are supported:
 
 * regular Search (to_tsquery)
@@ -232,7 +232,7 @@ All types of Text Searches can be combined with other Linq queries
 
 <[sample:text_search_combined_with_other_query_sample]>
 
-They alow also to specify language (regConfig) of the text search query (by default `english` is being used)
+They allow also to specify language (regConfig) of the text search query (by default `english` is being used)
 
 <[sample:text_search_with_non_default_regConfig_sample]>
 
--- a/documentation/documentation/documents/querying/linq.md
+++ b/documentation/documentation/documents/querying/linq.md
@@ -211,6 +211,32 @@ Query data from all tenants using `AnyTenant` method.
 Use `TenantIsOneOf` to query on a selected list of tenants.
 <[sample:tenant_is_one_of]>
 
+## Text Search
+
+Postgres constains built in [Text Search functions](https://www.postgresql.org/docs/10/textsearch-controls.html). They enable the possibility to do more sophisticated searching through text fields. Marten gives possibility to define <[linkto:documentation/documents/configuration/full_text;title=Full Text Indexes]> and perform queries on them.
+Currently three types of full Text Search functions are supported:
+
+* regular Search (to_tsquery)
+
+<[sample:search_in_query_sample]>
+
+* plain text Search (plainto_tsquery)
+
+<[sample:plain_search_in_query_sample]>
+
+* phrase Search (phraseto_tsquery)
+
+<[sample:phrase_search_in_query_sample]>
+
+All types of Text Searches can be combined with other Linq queries
+
+<[sample:text_search_combined_with_other_query_sample]>
+
+They alow also to specify language (regConfig) of the text search query (by default `english` is being used)
+
+<[sample:text_search_with_non_default_regConfig_sample]>
+
+
 ## Supported Types
 
 At this point, Marten's Linq support has been tested against these .Net types:
--- a/src/Marten/Linq/Parsing/FullTextSearchMethodCallParser.cs
+++ b/src/Marten/Linq/Parsing/FullTextSearchMethodCallParser.cs
@@ -38,7 +38,7 @@ namespace Marten.Linq.Parsing
                 throw new ArgumentException("Search Term needs to be string", "searchTerm");
 
             if (expression.Arguments.Count > 2 && expression.Arguments[2].Type != typeof(string))
-                throw new ArgumentException("Reg config needs to be string", "searchTerm");
+                throw new ArgumentException("Reg config needs to be string", "regConfig");
 
             var searchTerm = (string)expression.Arguments[1].Value();
 
--- a/src/Marten/Linq/Parsing/FullTextSearchMethodCallParser.cs
+++ b/src/Marten/Linq/Parsing/FullTextSearchMethodCallParser.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Linq.Expressions;
+using Marten.Linq.WhereFragments;
+using Marten.Schema;
+
+namespace Marten.Linq.Parsing
+{
+    public abstract class FullTextSearchMethodCallParser : IMethodCallParser
+    {
+        private readonly string methodName;
+        private readonly string searchFunction;
+
+        protected FullTextSearchMethodCallParser(string methodName, string searchFunction)
+        {
+            this.methodName = methodName;
+            this.searchFunction = searchFunction;
+        }
+
+        public bool Matches(MethodCallExpression expression)
+        {
+            return expression.Method.Name == methodName
+                   && expression.Method.DeclaringType == typeof(LinqExtensions);
+        }
+
+        public IWhereFragment Parse(IQueryableDocument mapping, ISerializer serializer, MethodCallExpression expression)
+        {
+            if (expression.Arguments.Count < 2 || expression.Arguments[1].Value() == null)
+                throw new ArgumentException("Search Term needs to be provided", "searchTerm");
+
+            if (expression.Arguments[1].Type != typeof(string))
+                throw new ArgumentException("Search Term needs to be string", "searchTerm");
+
+            if (expression.Arguments.Count > 2 && expression.Arguments[2].Type != typeof(string))
+                throw new ArgumentException("Reg config needs to be string", "searchTerm");
+
+            var searchTerm = (string)expression.Arguments[1].Value();
+
+            var regConfig = expression.Arguments.Count > 2 ?
+                expression.Arguments[2].Value() as string : FullTextIndex.DefaultRegConfig;
+
+            return new FullTextWhereFragment(
+                mapping as DocumentMapping,
+                searchFunction,
+                searchTerm,
+                regConfig);
+        }
+    }
+}
--- a/src/Marten/Linq/Parsing/FullTextSearchMethodCallParser.cs
+++ b/src/Marten/Linq/Parsing/FullTextSearchMethodCallParser.cs
@@ -5,12 +5,19 @@ using Marten.Schema;
 
 namespace Marten.Linq.Parsing
 {
+    public enum FullTextSearchFunction
+    {
+        to_tsquery,
+        plainto_tsquery,
+        phraseto_tsquery
+    }
+
     public abstract class FullTextSearchMethodCallParser : IMethodCallParser
     {
         private readonly string methodName;
-        private readonly string searchFunction;
+        private readonly FullTextSearchFunction searchFunction;
 
-        protected FullTextSearchMethodCallParser(string methodName, string searchFunction)
+        protected FullTextSearchMethodCallParser(string methodName, FullTextSearchFunction searchFunction)
         {
             this.methodName = methodName;
             this.searchFunction = searchFunction;
--- a/src/Marten/Linq/WhereFragments/FullTextWhereFragment.cs
+++ b/src/Marten/Linq/WhereFragments/FullTextWhereFragment.cs
@@ -1,0 +1,33 @@
+﻿using System.Linq;
+using Marten.Schema;
+
+namespace Marten.Linq.WhereFragments
+{
+    internal class FullTextWhereFragment : WhereFragment
+    {
+        public FullTextWhereFragment(DocumentMapping mapping, string searchTerm, string regConfig, string searchFunction)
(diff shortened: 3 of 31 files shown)


CodingGorilla:
It all looks fine to me.  I am by no means an expert on PostgresSQL or full text indexing in it.  I wrote the original feature as a bit of a learning experience with PGSQL.  These changes actually look pretty minimal, so far.

oskardudycz:
@CodingGorilla thank you for your feedback 👍 Yes, it's just step forward, I'm also not such expert about FTS on Postgres. So for sure it's just the first step, not the last one.

Looking at the query plans it seems that it's using the defined indexes, so from this stand point - it looks good. For sure we'd need some more investigation on performance, maybe some benchmarks etc.

oskardudycz:
Guys, I added first round of samples and documentation as @jokokko and @CodingGorilla were fine with initial concept. I'll try to finalize rest of documentation and testing tomorrow or on wednesday.

oskardudycz:
@jeremydmiller @CodingGorilla @mysticmind @jokokko I added possibilty to extend `MartenQueryable` with `WhereFragment` and used it in `FullTextSearch`. See more in this commit: https://github.com/JasperFx/marten/pull/1167/commits/8a79740d8faf2e403acaa25891cd20db6b2ad7c3

This makes possible to join FullTextSearch mechanism together with queryable. It'll make also possible to use it even with any other custom where.

I know that this change might be controvercial, but imho it'll make in future easy extensible plugging in other such queries. It will also make Marten users to even write their own custom wheres without needing to extend Marten core logic.

I'd be grateful for the feedback!

CodingGorilla:
@oskardudycz Can you show an example of how you expect this to be used?  It seems like there's likely some confusion as to how this might end up working that could cause queries to become 'garbeled' (for lack of a better way to explain it).

oskardudycz:
@CodingGorilla I added also sample in code for the usage of conjoined FullText search with queryable. I added also FullTextSearch methods to QueryableExtensions to allow usage on IQueryable<T> interface so:

```
var travelBlogPosts = session.Query<BlogPost>()
                    .Where(blogPost => blogPost.Category == "Travel")
                    .Search("Lorem ipsum")
                    .ToList()
```

CodingGorilla:
Ok, never mind my original question, my brain was still stuck in relational query mode. :)  But, while this looks like it would work, it seems like it kind of defeats the purpose of a full text search doesn't it?  You're doing a FTS inside one or more fields?

oskardudycz:
@CodingGorilla well, with this improvements you have now just more options than you had. You can do search as was originally designed - so on all contents of data field, but if you'd like to you can select to index only one (or more) fields from data - so reuse documents.

It also gives possibility to define more thant one FTS index on the document - so support eg. translations. You can have document with different versions for different languages and search with exact language (the Blog Post or Article could be good example of that).

Also filtering out some records (so joining FTS searches with query) imho is one of the useful features, even from performance perspective.

If you'll be doing fts then it will try to find and reuse defined index or filter whole data (see the logic for that [here](https://github.com/JasperFx/marten/pull/1167/files#diff-6019345c44b68252abc72ff417a6f4f0R280)). Query plans confirms that.

Of course normally I'd suggest to have specific document designed for FTS, but purpose of this PR is just to give users more options and flexibility.

CodingGorilla:
I'm not always a big fan of giving people more ways to accomplish the same thing.  It adds complexity and makes troubleshooting and support very difficult at times.  But if this is a feature that you find useful, then I say "Make it so!". 😆

oskardudycz:
@CodingGorilla yes, I know what you're saying. That's why I wrote that it's a bit controvercial ;) I already got such feature request from my colleagues. I also personally see the value on that, but I'm also a bit afraid that people my overuse this stuff. I'll try to put some guidance/best practices in the docs.

@CodingGorilla Thank you a lot for a good constructive feedback! And what's more - I learned new word - `garbeled` 😆

@jeremydmiller @mysticmind @jokokko what are your thoughts?

cocowalla:
> Also filtering out some records (so joining FTS searches with query) imho is one of the useful features, even from performance perspective

As a (happy!) user of Marten, I would find this really useful - I would think it's a really common scenario too.

oskardudycz:
Thanks @cocowalla, appreciate the feedback 👌

jeremydmiller:
@oskardudycz Oh man, I should have been paying more attention. I want to push you toward a little different implementation in the Linq mechanics. I think that explicitly adding `IWhereFragment` objects to `IMartenQueryable` is polluting the internal APIs and forcing you to make a lot of new object allocations from the before and adding complexity I don't think we want.

What I would suggest is using the `IMethodCallParser` mechanism instead, and let the normal `IMartenQueryable` mechanics do its thing. So instead of these methods directly on `IMartenQ
ueryable` adding WhereFragments directly:

```
        /// <summary>
        /// Performs a full text search against <typeparamref name="TDoc"/>
        /// </summary>
        /// <param name="searchTerm">The text to search for.  May contain lexeme patterns used by PostgreSQL for full text searching</param>
        /// <param name="regConfig">The dictionary config passed to the 'to_tsquery' function, must match the config parameter used by <seealso cref="DocumentMapping.AddFullTextIndex(string)"/></param>
        /// <remarks>
        /// See: https://www.postgresql.org/docs/10/static/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES
        /// </remarks>
        IMartenQueryable<T> Search(string searchTerm, string regConfig = FullTextIndex.DefaultRegConfig);

        /// <summary>
        /// Performs a full text search against <typeparamref name="TDoc"/> using the 'plainto_tsquery' search function
        /// </summary>
        /// <param name="queryText">The text to search for.  May contain lexeme patterns used by PostgreSQL for full text searching</param>
        /// <param name="regConfig">The dictionary config passed to the 'to_tsquery' function, must match the config parameter used by <seealso cref="DocumentMapping.AddFullTextIndex(string)"/></param>
        /// <remarks>
        /// See: https://www.postgresql.org/docs/10/static/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES
        /// </remarks>
        IMartenQueryable<T> PlainTextSearch(string searchTerm, string regConfig = FullTextIndex.DefaultRegConfig);

        /// <summary>
        /// Performs a full text search against <typeparamref name="TDoc"/> using the 'phraseto_tsquery' search function
        /// </summary>
        /// <param name="queryText">The text to search for.  May contain lexeme patterns used by PostgreSQL for full text searching</param>
        /// <param name="regConfig">The dictionary config passed to the 'to_tsquery' function, must match the config parameter used by <seealso cref="DocumentMapping.AddFullTextIndex(string)"/></param>
        /// <remarks>
        /// See: https://www.postgresql.org/docs/10/static/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES
        /// </remarks>
        IMartenQueryable<T> PhraseSearch(string searchTerm, string regConfig = FullTextIndex.DefaultRegConfig);
```

Let's make these strictly extension methods like:

```
public static class FullTextExtensions
{
    public static bool PhraseSearch(this T entity, string searchTerm, string regConfig = FullTextIndex.DefaultRegConfig)
    {
        return true; // doesn't really matter here
     }
}
```

and a corresponding `IMethodCallParser` for each new method that returns the same IWhereFragment you've already built.

If you want, you can come back with this too in the same extension method class:

```
// just syntactical sugar
public static IMartenQueryable<T> PhraseSearch(this IMartenQueryable<T> queryable, args)
{
    
    return queryable.Where(x => x.PhraseSearch(args));'
}
```

By relying on the IMethodCallParser, you're making fewer changes to the internals of MartenQueryable and doing fewer object allocations

oskardudycz:
Ok, it's fixed. I'll polish the changes (tests, documentation) and will try to finish that in the next 2-3 days.

oskardudycz:
Ok, I added documentation, more tests. It's ready for final review! :)
@jeremydmiller @mysticmind @CodingGorilla @jokokko

mysticmind:
@oskardudycz I had merged the PR #1168 with few minor tweak, this is resulting in a conflict in this PR for README.md, I will take a look to fix it.

oskardudycz:
@mysticmind sure, but if you'd like I can do rebase by myself.

oskardudycz:
@mysticmind I did additonal rebase to force Github to not scream about conflicts ;)

src/Marten/Linq/WhereFragments/FullTextWhereFragment.cs
```diff
@@ -0,0 +1,33 @@
...
+        public FullTextWhereFragment(DocumentMapping mapping, string searchFunction, string searchTerm, string regConfig = FullTextIndex.DefaultRegConfig)
+            : base(GetFilter(mapping, searchFunction, searchTerm, regConfig))
+        {
+        }
+
+        private static string GetFilter(DocumentMapping mapping, string searchFunction, string searchTerm, string regConfig)
+        {
+            var dataConfig = GetDataConfig(mapping, regConfig);
+
+            return $"to_tsvector('{regConfig}', {dataConfig}) @@ {searchFunction}('{regConfig}', '{searchTerm}')";
```
mdissel: Sql injection possible?

jokokko: Good catch (and shame on me for being too hasty)!. It is. `x.Search(@"somefilter ') ; drop table mt_doc_blogpost ; select ('"))`.

oskardudycz: Yeah, I forgot about that. I was supposed to fix that issue https://github.com/JasperFx/marten/issues/1170. I'll work on that

oskardudycz: @mdissel @jokokko  @RobSchoenaker Fixed in https://github.com/JasperFx/marten/pull/1167/commits/b31120cbd2b91844f3905bed69be998a6bd5908a

src/Marten/Linq/Parsing/FullTextSearchMethodCallParser.cs
```diff
@@ -0,0 +1,55 @@
...
+        public IWhereFragment Parse(IQueryableDocument mapping, ISerializer serializer, MethodCallExpression expression)
+        {
+            if (expression.Arguments.Count < 2 || expression.Arguments[1].Value() == null)
+                throw new ArgumentException("Search Term needs to be provided", "searchTerm");
+
+            if (expression.Arguments[1].Type != typeof(string))
+                throw new ArgumentException("Search Term needs to be string", "searchTerm");
+
+            if (expression.Arguments.Count > 2 && expression.Arguments[2].Type != typeof(string))
+                throw new ArgumentException("Reg config needs to be string", "searchTerm");
```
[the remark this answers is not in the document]
oskardudycz: Fixed.

documentation/documentation/documents/querying/linq.md
```diff
@@ -211,6 +211,32 @@ Query data from all tenants using `AnyTenant` method.
...
+
+* phrase Search (phraseto_tsquery)
+
+<[sample:phrase_search_in_query_sample]>
+
+All types of Text Searches can be combined with other Linq queries
+
+<[sample:text_search_combined_with_other_query_sample]>
+
+They alow also to specify language (regConfig) of the text search query (by default `english` is being used)
```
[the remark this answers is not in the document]
oskardudycz: Fixed.

documentation/documentation/documents/querying/linq.md
```diff
@@ -211,6 +211,32 @@ Query data from all tenants using `AnyTenant` method.
 Use `TenantIsOneOf` to query on a selected list of tenants.
 <[sample:tenant_is_one_of]>
 
+## Text Search
+
+Postgres constains built in [Text Search functions](https://www.postgresql.org/docs/10/textsearch-controls.html). They enable the possibility to do more sophisticated searching through text fields. Marten gives possibility to define <[linkto:documentation/documents/configuration/full_text;title=Full Text Indexes]> and perform queries on them.
```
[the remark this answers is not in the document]
oskardudycz: Fixed.

oskardudycz:
@mysticmind I added asserts for some of the sample methods in tests. Removed `[Fact]` attribute from the sample methods that are covered in other tests.

All comments addressed.

PR #1168, opened 2019-01-21
Update run dependency docs and add link to automated docker solution

eouw0o83hf:
`README.md` contains a link for getting started with PLV8 on windows, but nothing for non-windows systems. I wrote an [automated Docker solution](https://github.com/eouw0o83hf/marten-local-db) to get testing up and running with one script call, and wanted to share it with other prospective contributors.

If it would be more useful to just add that container system to this repository, I'm glad to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,13 +17,25 @@ Marten also provides .NET developers with an ACID-compliant event store with use
 
 Before getting started you will need the following in your environment:
 
-* Access to a PostgreSQL **9.5+** database.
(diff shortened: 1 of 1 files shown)


colin-anderson:
@eouw0o83hf If you want to use a newer version of PLV8, or you want to use Postgres 10.6 or 11.1, we have images available [here](https://hub.docker.com/r/ionx/postgres-plv8) you could use in your `Dockerfile`.

PR #1171, opened 2019-01-23
Testing the idea of switching some of the ConcurrentDictionary usages…

jokokko:
… to ImHashMap from https://github.com/dadhi/ImTools

Some benchmarks to (in)validate the justification for such changes (also locally benched ConcurrentDictionary against ImHashMap when lookup was by Guid, in which case CD performed better [with default Guid.GetHashCode]).
Lets see if this goes anywhere.
Further remarks: if ImTools is used with Cache in IdentityMap, it cannot be made internal since Cache is a protected member within a public type. In that sense, switching the cache type is a breaking change should anybody have their own implementation in place (though we don't allow pluggin it into Marten anywhere).
Lastly, upgraded BenchmarkDotNet, fixed existing benchmarks for compilation to pass.

### DIFF
--- a/src/Marten/Events/EventGraph.cs
+++ b/src/Marten/Events/EventGraph.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -8,6 +7,7 @@ using Baseline;
 using Marten.Events.Projections;
 using Marten.Schema;
 using Marten.Storage;
+using Marten.Util;
 
 namespace Marten.Events
 {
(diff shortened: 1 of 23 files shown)


cocowalla:
Few questions spring to mind:

1. Would we be likely to see benefits in real world use?
2. Are `ConcurrentDictionary`s in the hot path?
3. Is it worth swapping out something so battle-tested for a relative unknown?

Apologies if these come across as negative, it's not my intention (I'm something of a perf-seeking demon myself) :smile:

jokokko:
> 
> 
> Few questions spring to mind:
> 
>     1. Would we be likely to see benefits in real world use?
> 
>     2. Are `ConcurrentDictionary`s in the hot path?
> 
>     3. Is it worth swapping out something so battle-tested for a relative unknown?
> 
> 
> Apologies if these come across as negative, it's not my intention (I'm something of a perf-seeking demon myself) 😄

All good questions... This potential change was proposed & discussed in Gitter (based on the use of ImHashMap in Lamar), so I threw in a bit of code. The benches provided all were in favor of ImHashMap, but that's far from a rigorous proof.

jeremydmiller:
@cocowalla I'm using ImTools in other libraries (Lamar) with no issues, and it's at the core of DryIoC.

oskardudycz:
@jokokko see also changes that came with https://github.com/JasperFx/marten/pull/1164

mysticmind:
@jokokko and @oskardudycz  I have fixed the code indenting issues. Once the CI tests are green, will proceed to merge this PR. After the merging the PR, we can apply a change to #1164 related changes to use ImTools HashMap.

oskardudycz:
Cool! Thanks @mysticmind! 👍

Issue #1173, opened 2019-01-25
Patch API does not respect "TypeNameHandling" when serializing nested objects


psmith6:
When setting custom serializing settings below, the patch api calls fail to set the `$type` property.
```
var serializer = new JsonNetSerializer();
serializer.Customize(config =>
{
    config.TypeNameHandling = Newtonsoft.Json.TypeNameHandling.Objects;
});
```
I've noticed that TypeNameHandling is hardcode to `TypeNameHandling.None` in JsonNetSerializer seen below. When the patch dictionary gets serialized with this setting, it will prevent the inner object from storing its type information.

https://github.com/JasperFx/marten/blob/eb079897138c62d5d765323769e000d27d9c1667/src/Marten/Services/JsonNetSerializer.cs#L33-L39

jeremydmiller:
@psmith6 I don't think Patch is every going to play nicely with any TypeNameHandling strategy other than None. The Patch functionality actually uses Javascript at the moment, and knows absolutely nothing about Newtonsoft functionality

cocowalla:
I hit this problem a few months back too, and am getting back round to looking at it now. Only workaround I can see right now is not to use the patch API for interface types.

@jeremydmiller Marten already supports [querying for raw JSON](http://jasperfx.github.io/marten/documentation/documents/querying/query_json/) - what are your thoughts on extending the Patch API to support setting raw JSON? This would allow full control over serialisation, and might also be useful in cases where you already have JSON and want to avoid an extra serialisation/deserialisation?

oskardudycz:
@cocowalla That sound reasonable for me. It's similar case as for giving users ability to pass custom sql for orderby. That gives user possibility to do customisation even if we don't support this case yet in Marten.

cocowalla:
For info, I tried out the `JsonConverter` option, and it does work as expected. It works for my use case, but obviously it's not a great solution if you have a lot of types to create a converter for.

jeremydmiller:
@cocowalla Getting back into things this week.

"@jeremydmiller Marten already supports querying for raw JSON - what are your thoughts on extending the Patch API to support setting raw JSON?" -- that's basically what it does today, or am I missing something?

jeremydmiller:
Some new thoughts on this one today:

* It looks like it's possible to **not** use the "clean" serializer when appending items
* The patching is going to have to be smart enough to "know" when an item is being appended that would force Newtonsoft to use the $type, or "know" that the type handling is set to do so.

[...]

PR #1174, opened 2019-01-25
Fix patch api to support TypeNameHandling.Object

psmith6:
This is a quick hack to demonstrate the issue and provide a possible solution to https://github.com/JasperFx/marten/issues/1173.

### DIFF
--- a/src/Marten/Patching/PatchOperation.cs
+++ b/src/Marten/Patching/PatchOperation.cs
@@ -30,10 +30,26 @@ namespace Marten.Patching
         }
 
         // TODO -- come back and do this with a single command!
+        private const string VALUE_LOOKUP = "___VALUE___";
         public void ConfigureCommand(CommandBuilder builder)
         {
-            var patchJson = _serializer.ToCleanJson(_patch);
-            var patchParam = builder.AddJsonParameter(patchJson);
+            var patchParam = builder.AddJsonParameter(_serializer.ToCleanJson(_patch));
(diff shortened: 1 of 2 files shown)


src/Marten/Patching/PatchOperation.cs
```diff
@@ -30,10 +30,21 @@ public class PatchOperation : IStorageOperation, NoDataReturnedCall
         }
 
         // TODO -- come back and do this with a single command!
+        private const string VALUE_LOOKUP = "___VALUE___";
         public void ConfigureCommand(CommandBuilder builder)
         {
-            var patchJson = _serializer.ToCleanJson(_patch);
-            var patchParam = builder.AddJsonParameter(patchJson);
+            var value = _serializer.ToJson(_patch["value"]);
```
jeremydmiller: Ugh, ugh, ugh. I don't have any other recommendations, but this is making it way more inefficient. Let's let this go for now, but maybe if we ever get around to the Linq overhaul, we can do something with the `IField` model so that it "knows" whether or not it needs to do the heavyweight serialization state here or can get away with something lighter like just the text.

jeremydmiller:
Jeremy says "Ship it!", but we might wanna revisit this in 4.0 with an eye toward performance

oskardudycz:
@jeremydmiller I created follow up issue https://github.com/JasperFx/marten/issues/1182 for Milestone 4.0

PR #1175, opened 2019-01-28
On the ViewProjection, if the Id selector returns the default of the projection type Id, then do not add that viewId as a new projection.

ericgreenmix:
This prevents default type ids from creating new erroneous documents.

### DIFF
--- a/src/Marten.Testing/Events/Projections/custom_transformation_of_events.cs
+++ b/src/Marten.Testing/Events/Projections/custom_transformation_of_events.cs
@@ -361,20 +361,44 @@ namespace Marten.Testing.Events.Projections
             theSession.Events.StartStream<QuestParty>(streamId, started);
             theSession.SaveChanges();
 
-            var docoument = await theSession.LoadAsync<PersistedView>(streamId);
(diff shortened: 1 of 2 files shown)


ericgreenmix:
@oskardudycz I don't really see this as a big enough breaking change to delay until version 4. Yes, if somebody was abusing the `ProjectEvent` id selector and intentionally feeding it in the default type, then it could cause their code to stop creating new documents with a random generated Id. However, this is not the intended use of the id selector. 

This code change is really there to prevent potential bugs when selecting the property(s) to be used as the Id of the projected document when the selector results in a `default(TypeId)` type. I.E., a nullable Guid from the event gets set as the Id.

oskardudycz:
Yes, I understand the intent for this change, and I agree with that. It's imho good change, but in general we should avoid breaking changes if it's possible. Imho this place might have been "abused" by the users, so they might be surprised.

I'll let @jeremydmiller decide. If he also agrees that it's breaking, then I'd suggest to add some flag as I suggested in previous comments, then I'm totally positive on merging it asap.

PR #1202, opened 2019-02-25
Adding the ability to create a foreign key from a marten document to non-marten table

CodingGorilla:
This is obviously not complete, but I wanted to get this request started to get some feedback on the idea and the implementaion.  The first big thing is that I made the `ForeignKeyDefinition` an abstract class, I think this is a breaking change since the `DocumentMapping.ForeignKeys` property is public so people could potentially create their own `ForeignKeyDefinition`s and add them manually.  I considered keeping the class non-abstract and instead making the `ReferenceType` and `ToDDL()` virtual, but I'm not a big fan of virtual members.

### DIFF
--- a/src/Marten.Testing/Schema/ForeignKeyDefinitionTests.cs
+++ b/src/Marten.Testing/Schema/ForeignKeyDefinitionTests.cs
@@ -14,7 +14,7 @@ namespace Marten.Testing.Schema
         [Fact]
         public void default_key_name()
         {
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@CodingGorilla @oskardudycz Meh, to avoid breaking the name and to keep it a bit simpler, could we add additional constructor functions to `ForeignKeyDefinition` for the FK to another table? And internally use a Func<DbObjectName> member that can determine the linked table/column name? 

It's nit-picky, but it avoids changing or breaking the existing API, and ForeignKeyDefinition could easily be used as a public API.

CodingGorilla:
@jeremydmiller @oskardudycz Ok, I will close this pull request and create a new one with that approach.

PR #1204, opened 2019-02-28
Add missing semicolons when dropping system functions

richardoliver-umusic:
Hi all, we came across a minor issue when generating a patch rollback script for a new marten database, where the drop function statements for a few of the system functions didn't have semicolons following them. The affecting functions are mt_immutable_timestamp, mt_immutable_timestamptz, and mt_append_event. This PR fixes them and adds a test for it.

The whitespace changes in AppendEventFunction.cs are replacing all tabs with spaces as the file was mixed. Let me know if you'd rather not have the whitespace changes included and i'll remove them.

As always let me know if anything else is required.

I found a couple other issues related to patch rollback scripts that i'll create issues for as they likely require a bit more discussion than this.

### DIFF
--- a/src/Marten.Testing/Schema/creating_a_full_patch.cs
+++ b/src/Marten.Testing/Schema/creating_a_full_patch.cs
@@ -1,5 +1,6 @@
 ﻿using Marten.Testing.Documents;
 using Xunit;
+using IssueAssigned = Marten.Testing.Events.IssueAssigned;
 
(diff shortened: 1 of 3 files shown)


mysticmind:
@richardoliver-umusic  Raised a separate issue #1216 pertaining to mixed tab spaces.

richardoliver-umusic:
@mysticmind No problem, whitespace changes removed.

Issue #996, opened 2018-04-09
Consider using "tombstone" placeholders in the event table


jeremydmiller:
Use “tombstone” event placeholders in place of failures. This might make the async daemon **way** easier.

jeffdoolittle:
Can you provide more details on the nature of the problem and the proposed solution?

jeremydmiller:
I'm folding this into #1584

jeremydmiller:
I did a terrible job of this in #1584 and I'm reopening this separately today to redo it. The tombstones have to be interleaved w/ the events for this to matter.

PR #1208, opened 2019-03-04
Add scenario (docs) for copying & transforming event streams (for versioning…

jokokko:
…, data modification)

### DIFF
--- a/src/Marten.Testing/ScenarioCopyAndReplaceStream.cs
+++ b/src/Marten.Testing/ScenarioCopyAndReplaceStream.cs
@@ -1,0 +1,107 @@
+﻿using System.Linq;
+using Marten.Events;
+using Marten.Services;
+using Marten.Testing.Events;
+using Marten.Testing.Events.Projections;
+using Xunit;
+
+namespace Marten.Testing
+{
+	public class ScenarioCopyAndReplaceStream : DocumentSessionFixture<NulloIdentityMap>
+	{
+		public ScenarioCopyAndReplaceStream()
+		{
+			StoreOptions(options =>
+			{
+				options.Events.StreamIdentity = StreamIdentity.AsString;
+			});
+		}
+
+		[Fact]
+		public void SampleCopyAndTransformStream()
+		{
+			// SAMPLE: scenario-copyandtransformstream-setup
+			var started = new QuestStarted { Name = "Find the Orb" };
+			var joined = new MembersJoined { Day = 2, Location = "Faldor's Farm", Members = new[] { "Garion", "Polgara", "Belgarath" } };
+			var slayed1 = new MonsterSlayed { Name = "Troll" };
+			var slayed2 = new MonsterSlayed { Name = "Dragon" };
+			
+			using (var session = theStore.OpenSession())
+			{
+				session.Events.StartStream<Quest>(started.Name,started, joined, slayed1, slayed2);
+				session.SaveChanges();
+			}
+			// ENDSAMPLE
+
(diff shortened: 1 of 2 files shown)


src/Marten.Testing/ScenarioCopyAndReplaceStream.cs
```diff
@@ -0,0 +1,107 @@
...
+	{
+		public ScenarioCopyAndReplaceStream()
+		{
+			StoreOptions(options =>
+			{
+				options.Events.StreamIdentity = StreamIdentity.AsString;
+			});
+		}
+
+		[Fact]
```
mysticmind: If this a sample for docs, can we remove/comment the `[Fact]` so that it does not get run as a unit test.

jokokko: @mysticmind: Certainly can be removed. Comes with the downside that should there be behavioural breaking changes (in this case something in the execution path leading to an exception, which is not likely :), the sample breaks & goes unnoticed.

mysticmind: @jokokko okay, make sense :-)

mysticmind:
@jokokko quick comment, could you please fix the tab spacing in code changes to use 4 tab spaces?

oskardudycz:
@jokokko this sample is interesting and something that we don't have covered. I was thinking about for quite some time. Is this something that you applied also practically in your project?

I have few concerns that we'd need to diccuss.

From my practice, my common scenario for working with event streams was:
1. Create Stream, publish some events.
2. Create projection (inline, or custom) that is updated during publishing events from streams.

Then while working on aggregates and getting actual aggregate stream - use online aggegation so `AggregateStream` method. (so this is how I was working with write model).

While doing queries I was using projection. So Inline projections, as they are snapshots of the write model state - I was using for the write model validation, etc. Custom projections - for the read model.

So having those end usage scenarios, let's consider how this sample would help, or what might be missing.

1. If we create copy of the stream, with new Id then we will get new record in our read model. How will we know that it's just a new "proper" item? How will we distinguish that we should not use old one? Imho we should publish "delete event" for the old stream. Having that for custom projection we could use `DeleteEvent` to remove the old record. For inline projection we don't have yet automated mechanism for deleting projection item, but it can be tricked (let's not get into details right now how...). So Delete Event should `more or less` should solve the case for read model.

2. Write model. This is the harder part. Currently we don't have the way to mark the stream as "deleted", or "obsolete". At least none built in way. We could think about some ["tombstoning"](https://github.com/JasperFx/marten/issues/996) but still, no out of the box solution. But having "delete event" suggested in the previous point we could implement custom "soft delete".

So to sum up, imho we should add to this sample also publishing delete event for the old stream. Having that it would make possible to remove the record from projection and also mark the stream as "deleted"/"inactive", so make soft delete with custom logic.

Imho this will make this sample more complete and more helpful for the end usage. Also maybe we could show mor real life sample - like bank account, rather than slaying monsters? :)

@jokokko what are your thoughts?

mysticmind:
@jokokko and @oskardudycz are we planning to add further docs or we are good with this PR and add any docs further as another PR?

oskardudycz:
@mysticmind that's more question to @jokokko. From my perspective it would be nice to extend this scenario with handling DeleteEvent.

mysticmind:
@oskardudycz shall we go ahead to merge the current changes and raise a separate issue for your comments?

mysticmind:
@oskardudycz since you merged this, I have set the milestone for this back to 3.5.0

Issue #1108, opened 2018-10-08
Unique index per tenant


Nysosis:
Currently it doesn't seem possible to configure unique indices to take multi-tenancy into consideration.

Would it make sense to extend the `UniqueIndexType` enum to have a `ComputedPerTenant` option, which if set on a property will include the `tenant_id` column onto the index definition when building the table?

oskardudycz:
@Nysosis sorry for long answer. I'll review such possibility and get back with a feedback.

oskardudycz:
@Nysosis it's implemented in https://github.com/JasperFx/marten/pull/1212. It'll be available in Marten 3.5.0

PR #1212, opened 2019-03-08
Basic support for Unique index per tenant #1108

jaymcguinness:
My first pull request for Marten so do let me know if I've missed anything.

This should provide a basic level of support for users who need unique index support across multiple tenants - see #1108 

I've added in test for the simplest case - the test suite continues to pass for me locally.

This should work for simple cases using ComputedIndex - I've yet to test for DuplicatedFields - so I'm not sure this will work just yet.


jaymcguinness:
Will fix whitespace on PR

src/Marten/Schema/UniqueIndexType.cs
```diff
@@ -3,6 +3,7 @@
     public enum UniqueIndexType
     {
         DuplicatedField,
-        Computed
+        Computed,
+        Tenanted
```
oskardudycz: I think that this is redundant. I don't see much value on that, especially that. Unique Index might be both computed and tenanted.

oskardudycz:
@jaymcguinness it would be also great if you could also update docs.

oskardudycz:
@jaymcguinness it would be also great if you could also update docs.

jaymcguinness:
@oskardudycz I agree about the Tenanted value on UniqueIndexType not being quite right, but we do need something to differentiate between unique indices which are global vs. the tenant.

Marten as it currently stands would allow for a unique index on a table across all tenants, so I need to be careful not to introduce a breaking change.  I also think there are valid use cases where you might still want a unique index to work across all tenants not just one - eg. am thinking a SaaS feature which allows tenanted users to  set multiple domain names - these things need to be unique in the real world so that tenant resolution would work via domain name etc.  

Therefore I'd propose that I do this:

* remove the Tenanted value from the UniqueIndexType enum
* add overloads to fluent interface/attributes where we set UniqueIndex to take a TenancyStyle property - we'd set the default to work across all tenants which means it won't break for existing codebases and users can opt-in to the unique index working at tenant level at their discretion.
* I'll write more tests to cover these scenarios and get it working for DuplicatedField too otherwise it'll be half-baked.

Thoughts?

ps. More than happy to update the docs once we are all happy with the approach.

oskardudycz:
@jaymcguinness points that you suggest are 100% aligned with what I was thinking about 👍

oskardudycz:
@jaymcguinness will you be able to finish the PR in the next few days? We'd like to get it to `v3.5.0`. If you're not able then I can try to merge it and add missing stuff.

jaymcguinness:
Should be sorted by end of week.

oskardudycz:
Thank you! If you need some help feel free to write here on on gitter 👍

src/Marten.Testing/Schema/UniqueIndexMultiTenantTests.cs
```diff
@@ -0,0 +1,193 @@
...
+        private class Project2 : Project { } //used for duplicatedfield index tests
+
+        //used for attributes index tests
+        private class Project3 : Project
+        {
+            [UniqueIndex(IsScopedPerTenant = true, IndexType = UniqueIndexType.Computed)]
+            public string Code { get; set; }
+        } 
+
+        //private class BadProject : Project { } //used for bad index tests
```
oskardudycz: Could you remove commented out code if it's not being used?

src/Marten.Testing/Schema/UniqueIndexMultiTenantTests.cs
```diff
@@ -0,0 +1,193 @@
...
+
+            public Guid Id { get; set; }
+
+            public string Name { get; set; }
+        }
+
+        private class Project2 : Project { } //used for duplicatedfield index tests
+
+        //used for attributes index tests
+        private class Project3 : Project
```
oskardudycz: Could you change name of this class from `Project3` to more descriptive - eg. `ProjectWithUniqueCodePerTenant`

src/Marten/MartenRegistry.cs
```diff
@@ -224,6 +224,22 @@ public DocumentMappingExpression<T> UniqueIndex(UniqueIndexType indexType, strin
...
+            /// <summary>
+            /// Creates a unique index on this data member within the JSON data storage
+            /// </summary>
+            /// <param name="indexType">Type of the index</param>
+            /// <param name="indexTenancyStyle">Style of tenancy</param>
+            /// <param name="indexName">Name of the index</param>
+            /// <param name="isScopedPerTenant">Whether the unique index applies on a per tenant basis</param>
+            /// <param name="expressions"></param>
+            /// <returns></returns>
+            public DocumentMappingExpression<T> UniqueIndex(UniqueIndexType indexType, string indexName, bool isScopedPerTenant=false, params Expression<Func<T, object>>[] expressions)
```
oskardudycz: Formatting:
```suggestion
            public DocumentMappingExpression<T> UniqueIndex(UniqueIndexType indexType, string indexName, bool isScopedPerTenant = false, params Expression<Func<T, object>>[] expressions)
```

src/Marten.Testing/Schema/UniqueIndexMultiTenantTests.cs
```diff
@@ -0,0 +1,193 @@
...
+using Shouldly;
+using Xunit;
+
+namespace Marten.Testing.Schema
+{
+
+    public class UniqueIndexMultiTenantTests
+    {
+        public const string UniqueSqlState = "23505";
+
```
oskardudycz: this empty line is redundant

src/Marten.Testing/Schema/UniqueIndexMultiTenantTests.cs
```diff
@@ -0,0 +1,193 @@
+﻿using System;
+using Marten.Schema;
+using Marten.Storage;
+using Npgsql;
+using Shouldly;
+using Xunit;
+
+namespace Marten.Testing.Schema
+{
+
```
oskardudycz: this empty line is redundant

src/Marten.Testing/Acceptance/unique_indexes.cs
```diff
@@ -112,5 +112,22 @@ public void example_using_a_multiple_properties_duplicate_field_unique_index()
...
+        [Fact]
+        public void example_using_a_per_tenant_scoped_unique_index()
+        {
+            // SAMPLE: per-tenant-unique-index
+            var store = DocumentStore.For(_ =>
+            {
+                _.Connection(ConnectionSource.ConnectionString);
+
+                // This creates a duplicated field unique index on firstname, lastname and tenant_id
+                _.Schema.For<User>().MultiTenanted().UniqueIndex(UniqueIndexType.DuplicatedField,"index_name",true, x =>  x.FirstName, x=>x.LastName);
```
oskardudycz:
```suggestion
                _.Schema.For<User>().MultiTenanted().UniqueIndex(UniqueIndexType.DuplicatedField,"index_name",true, x =>  x.FirstName, x => x.LastName);
```

src/Marten.Testing/Acceptance/unique_indexes.cs
```diff
@@ -112,5 +112,22 @@ public void example_using_a_multiple_properties_duplicate_field_unique_index()
...
+
+                // This creates a computed unique index on client name and tenant_id
+                _.Schema.For<Client>().MultiTenanted().UniqueIndex(UniqueIndexType.Computed, "index_name", true, x => x.Name);
```
oskardudycz:
```suggestion
                _.Schema.For<Client>().MultiTenanted().UniqueIndex(UniqueIndexType.Computed, isScopedPerTenant: "index_name", true, x => x.Name);
```

oskardudycz:
@jaymcguinness looks good 👍 could you also rebase your changes with latest code?

oskardudycz:
Updated, rebased and merge with: https://github.com/JasperFx/marten/pull/1232#merged-event

Issue #1198, opened 2019-02-21
Allow Duplicated fields to be marked as NOT NULL


cocowalla:
Marten allows duplicating fields from JSON as relational fields. I came across cases where duplicated fields were being marked as nullable, even though they were for non-nullable .NET types.

I guess this was happening because the properties in question belong to a nested type, so in *theory* could be null (even though they never would be in my case).

I mentioned this in gitter, and Jeremy suggested a PR, so this issue is to track it.

I'll have a stab at this one when I get time

cocowalla:
As a workaround, this can be done by using the `pgType` argument when calling `DocumentMappingExpression.Duplicate(...)`, for example:

```
For<Event>()
  .Duplicate(x => x.Message.Timestamp, "timestamp with time zone NOT NULL");
```

While it would be relatively straightforward to add support for declaratively adding the `NOT NULL` constraint, it does involve touching quite a lot of code, and would break binary compatibility without a slew of overloads being added.

So instead of adding this feature, I've made a small change to one of the examples in the docs, to demonstrate that the `pgType` argument can be used to add constraints to duplicated fields.

cocowalla:
I take it back, the workaround doesn't work, as it causes syntax errors in the generated DDL for the upsert/insert/update scripts

PR #1214, opened 2019-03-10
 Demo adding constraint to duplicated field (#1198)

cocowalla:
#1198


cocowalla:
Dammit, I pulled the trigger on this one a little early - seems it doesn't play nicely with the upsert functions...

Issue #1199, opened 2019-02-21
Allow specifying index order for Duplicated fields


cocowalla:
Marten allows duplicating fields from JSON as relational fields, and creating an index for them.

I have a case where I want to duplicate a `timestamptz` field, and have it sorted in descending order, which Marten doesn't yet support (confirmed in gitter).

This issue is to track adding support for specifying the index order for Duplicated fields

I'll have a stab at this one when I get time, along with the similar #1198

barryhagan:
It will be nice to have a declarative way to set this, but you can alter the index expression right now:

`For<T>().Duplicate(a => a.Timestamp, configure: idx => { idx.Expression = "? DESC"; });`

cocowalla:
Created PR #1213

PR #1218, opened 2019-03-11
Example showing customising the sort order of a duplicated index

cocowalla:
Example for changes made in #1199 / #1213


cocowalla:
@oskardudycz I can't add it _exactly_ there, as index sort order is only applicable for B-tree indexes - I can add a new `_.Schema.For<User>` block to demo it tho... but... I realised I kind of forgot about computed indexes in my previous PR 🙄

I've added a commit to this PR to support for changing sort order with `ComputedIndex`.

Issue #1220, opened 2019-03-13
IsOneOf for array using include fails 


KevinRecuerda:
This code fails : `column reference "data" is ambiguous`
```csharp
var requestedTags = new[] {"DIY", "TAG"};
var users = new List<User>();
var issues = query.Query<Issue>()
                  .Include(x => x.AssigneeId, users)
                  .Where(x => x.Tags.Any(t => requestedTags.Contains(t)))
                  .ToList();
```

PR #1221, opened 2019-03-13
IsOneOf for array using include fails

KevinRecuerda:
#1220


jeremydmiller:
Doh! I know who'd the guilty party on that one. When and if we pull off the big ol' Linq restructuring, I'm gonna say we kill off all the "d." prefix business. That's a vestigal tail from very, very early experiments.

PR #1227, opened 2019-03-15
Adjust Upsert, Insert and Update Functions for Duplicated Fields with User Defined Type

### LINKED ISSUE #1225: Workflow for Custom Types
I defined a duplicate field with a custom type according to [this doc](http://jasperfx.github.io/marten/documentation/documents/customizing/duplicated_fields/):
`For<Structure>().Duplicate(x => x.Molecule, pgType: "molecule('sample')");`

The respective `mt_doc_structure` table definition looks good (it created a duplicate field `molecule` with type `molecule`).

You can insert a value to the molecule-type field via a text like this:
`INSERT INTO test (molecule) VALUES ('c1ccccc1');`

But when I try to insert via the `DocumentStore` I get following error:
> Marten Command Failure:
> select structure.mt_upsert_structure(arg_molecule := :arg0, doc := :arg1, docDotNetType := :arg2, docId := :arg3, docVersion := :arg4)
> 42883: function structure.mt_upsert_structure(arg_molecule => text, doc => jsonb, docdotnettype => character varying, docid => uuid, docversion => uuid) does not exist

The problem is that the definition of the `mt_upsert_structure` is this:
`mt_upsert_structure(arg_molecule molecule, doc  jsonb, docdotnettype  character varying, docid  uuid, docversion  uuid)`

So the `mt_upsert_structure` in the db expects the `arg_molecule` as type `molecule` but is called with `arg_molecule` as type `text`.

Is there a way to customize this for cases with custom type duplicate fields without equivalent C# type?

dapalmi:
Closes #1225

### DIFF
--- a/src/Marten.Testing/Bugs/Bug_1225_duplicated_fields_with_user_defined_type.cs
+++ b/src/Marten.Testing/Bugs/Bug_1225_duplicated_fields_with_user_defined_type.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using Marten.Schema;
+using Npgsql;
 using Shouldly;
 using Xunit;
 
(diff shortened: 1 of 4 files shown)


mysticmind:
@dapalmi Around 3-4 unit tests are failing due to the PR changes which looks to be breaking changes. Please provide details on the use case/at least a unit test which provides details on what you are trying to accomplish with this change.

Besides, any breaking change will need to be considered for 4.0 milestone.

dapalmi:
@mysticmind Thanks for your response. You can find details about the problem I have in the issue #1225 
I saw the failing unit tests, but wanted to wait for a feedback from the Marten pros, before I continue with this. I made this PR to help you understand where I think the issue is. I don't think this supposed to be a breaking change, I rather think I didn't provide a proper solution to that problem ;)
Maybe you can take a look at my issue and tell me what you think. Thanks

dapalmi:
@mysticmind I had another idea. Could you please check the changes and the unit test. But to be honest, I am still not sure if I understood the problem I have completely and if this is the right way to solve it.

mysticmind:
@dapalmi The approach which you are trying may not be the right way to fix this. Let me take a look at it and get back to you.

dapalmi:
@mysticmind Any suggestions on how to fix this a better way?

mysticmind:
@dapalmi I didn't get a chance to look at this, will check in the next couple of days and get back to you

mysticmind:
@dapalmi I had a look at the issue in detail and my observations are below:
- Marten is fancy library built on top of Npgsql so for any Postgres type/custom types (i.e.molecule), there has to be an equivalent `NpgsqlDbType` for things to work.
- As per my tests, upsert function is created properly. But problem occurs for the arg which is passed  to function for the duplicated field which has `NpgsqlDbType` as `text` (since there is no mapping available for `molecule`, Marten defaults to NpgsqldbType`  as `text`)
- Looks like you will need to implement a type handler as an Npgsql plugin.  Npgsql provides plugin for PostGIS, GeoJSON etc. you can check the plugin implementation in Npgsql source https://github.com/npgsql/npgsql/tree/dev/src.

In summary, this is not something which needs a fix in Marten. Please run through the above and let me know your thoughts/comments.

dapalmi:
@mysticmind Thanks. Unfortunately the error occurs before the db is involved. It would be great of Marten would default to text like you said. But it does not. Marten creates a upsert function which it later cannot find anymore. Please see my error description in #1225

mysticmind:
Below is the error which you posted:
```
Marten Command Failure:
select structure.mt_upsert_structure(arg_molecule := :arg0, doc := :arg1, docDotNetType := :arg2, docId := :arg3, docVersion := :arg4)
42883: function structure.mt_upsert_structure(arg_molecule => text, doc => jsonb, docdotnettype => character varying, docid => uuid, docversion => uuid) does not exist
```

Postgres resolve a function based on the function signature, note that Marten has created a function with signature `structure.mt_upsert_structure(molecule, jsonb, uuid, uuid)`, you can check the Postgres database to confirm the function exists. 

Argument `arg_molecule` has `NpgsqlDbType` as `text`. Since there is no `NpgsqlDbType`  mapping for Postgres custom type `molecule`, `NpgsqlDbType` has been resolved based on the dotnet type of the property in this case.  

So the function call results in a varying function signature `structure.mt_upsert_structure(text, jsonb, uuid, uuid)`. Postgres is not able to find a function with this signature hence returns a `42883: function does not exist` error.

For resolving custom postgres types, my previous [comment](https://github.com/JasperFx/marten/pull/1227#issuecomment-482624392) holds good.

If you want to create a duplicate field as text, don't set `pgType`, it will get create a column appropriately and things should work fine.

mysticmind:
@jeremydmiller, @oskardudycz, @jokokko Please chime in and let me know your thoughts/comments. We definitely need to document usage and limitations of using custom Postgres types within Marten.

dapalmi:
@mysticmind Thanks for your input and explanation. I did some more research and tests and found a workaround for this problem.
I just need to manually create a cast like this:
`CREATE CAST (cust_type AS text) WITH INOUT AS IMPLICIT;`
Maybe you want to add that to your documentation.
Which brings me to a question I have:
How do I sneak in custom SQL code like the CREATE CAST to be executed with AutoCreate.CreateOrUpdate on startup in Development and add it to the SQL file with WritePatch() in Production?
Thanks

mysticmind:
@dapalmi some pointers to get you going with creating the cast using Marten schema features extensions:
- check the documentation http://jasperfx.github.io/marten/documentation/documents/advanced/extensions/ 
- Check the `ISchemaObject` implementation for [Sequence.cs](https://github.com/JasperFx/marten/blob/eb079897138c62d5d765323769e000d27d9c1667/src/Marten/Storage/Sequence.cs) .

dapalmi:
@mysticmind Thanks. But the documentation is very cryptic about this. I tried following but nothing happens:
```
public class StructureStorage : FeatureSchemaBase
    {
        public StructureStorage(StoreOptions options) : base("structure", options)
        {
        }

        protected override IEnumerable<ISchemaObject> schemaObjects()
        {
            yield return new MoleculeCast();
            yield return new LoadFileFunction(Options.Events);
        }
    }

    public class MoleculeCast : ISchemaObject
    {
        public DbObjectName Identifier { get; }

        public IEnumerable<DbObjectName> AllNames()
        {
            yield return Identifier;
        }

        public void ConfigureQueryCommand(CommandBuilder builder)
        {

        }

        public SchemaPatchDifference CreatePatch(DbDataReader reader, SchemaPatch patch, AutoCreate autoCreate)
        {
            if (!reader.Read() || reader.GetInt32(0) == 0)
            {
                Write(patch.Rules, patch.UpWriter);
                return SchemaPatchDifference.Create;
            }

            return SchemaPatchDifference.None;
        }

        public void Write(DdlRules rules, StringWriter writer)
        {
            writer.WriteLine($"CREATE CAST (molecule AS text) WITH INOUT AS IMPLICIT;");
        }

        public void WriteDropStatement(DdlRules rules, StringWriter writer)
        {
            writer.WriteLine($"DROP CAST IF EXISTS (molecule AS text);");
        }
    }

public class LoadFileFunction : Function
    {
        private readonly EventGraph _events;

        public LoadFileFunction(EventGraph events) : base(new DbObjectName(events.DatabaseSchemaName, "mt_load_file"))
        {
            _events = events;
        }

        public override void Write(DdlRules rules, StringWriter writer)
        {
            var databaseSchema = _events.DatabaseSchemaName;

            writer.WriteLine($@"
                CREATE OR REPLACE FUNCTION {databaseSchema}.{Identifier}(text) RETURNS text AS '
                #!/bin/bash 
                curl $1 2>/dev/null 
                ' LANGUAGE plsh;
                ");

        }

        protected override string toDropSql()
        {
            return $"drop function if exists {Identifier}(text)";
        }
    }

            var store = DocumentStore.For(_ =>
            {
...
                _.Storage.Add<StructureStorage>();
            });
```

jeremydmiller:
@dapalmi @mysticmind What's left at this point to finish this off? Do you need more help with the ISchemaObject thing? You need something in your implementation that can check the database for the existence of your custom type. And I'd definitely ask for some new docs on this workflow.

My only other comment is to streamline the `DuplicatedField` class. I don't think we need all the extra properties, just PostgresType. The Expression code is so nasty as it is that I'd like to keep that as simple as it can be.

dapalmi:
@jeremydmiller I understand the need to keep things simple. After more consideration I think the best solution is to create the respective casts for custom defined types in the database together with the custom defined casts. So Marten does not need to worry about that.

mysticmind:
@jeremydmiller As indicated in my earlier comment, I will plan to add some documentation around usage, workflow and limitations around custom types.

PR #1231, opened 2019-03-21
added type mappings for serial and text/varchar arrays

egiraffe:
I was working on a project and encountered the error "Marten cannot derive updates for objects" when creating a document store for tables with a "serial" column and text array columns (varchar[], text[], etc). I think the issue is related to TypeMappings.cs and can be resolved with this PR.  

Path to repo to reproduce error: https://github.com/egiraffe/Marten-Serial
in the repo, everything runs correctly on create, but creating the store after the initial run fails.


oskardudycz:
@egiraffe will you find some time soon to add tests so we could close this PR?

mysticmind:
@egiraffe I restarted the Appveyor build which was failing due to some blinking unit tests. It is all green now.

PR #1238, opened 2019-03-27
Reverse delete operation ordering. Fixes #1229

### LINKED ISSUE #1229: Delete ordering issue
When trying to delete multiple different types of documents that have a FK relationship the deletes are ordered incorrectly.

I've not fully tracked down but believe it's because the operations are sorted topologically but for deletes that is the incorrect order. For now I've just got a failing test to check if my assumptions about it working are incorrect:

````c#
public class Bug_1229_delete_ordering: IntegratedFixture
{
    [Fact]
    public void delete_ordering_with_fk_relationships()
    {
        StoreOptions(_ =>
        {
            _.Schema.For<User>()
                .IdStrategy(new NoOpIdGeneration());

            _.Schema.For<Issue>()
                .IdStrategy(new NoOpIdGeneration())
                .ForeignKey<User>(u => u.AssigneeId);
        });

        var user = new User();
        var issue = new Issue { AssigneeId = user.Id };

        using (var query = theStore.OpenSession("Bug_1229"))
        {
            query.Store(user);
            query.Store(issue);

            query.SaveChanges();
        }

        using (var query = theStore.OpenSession("Bug_1229"))
        {
            query.Delete(issue);
            query.Delete(user);

            query.SaveChanges();
        }
    }
}
````

### PR DESCRIPTION
**Pull request by barclayadam**
A fix for issue #1229 to reverse the order of delete operations to enable deleting multiple documents in one batch when a FK relationship exists.

Instead of relying purely on a topological sort of document types, I ordered the list of operations as a whole, continuing to use the topological sort of types.

A potential performance optimisation would be to calculate the topological order of all document types mapped once when creating the store / options but wanted to concentrate on the correctness in this PR

### DIFF
--- a/src/Marten/Services/UnitOfWork.cs
+++ b/src/Marten/Services/UnitOfWork.cs
@@ -8,6 +8,7 @@ using Baseline;
 using Marten.Events;
 using Marten.Patching;
 using Marten.Schema;
+using Marten.Services.Deletes;
 using Marten.Storage;
 using Marten.Util;
 
@@ -80,7 +81,7 @@ namespace Marten.Services
         }
 
         public IEnumerable<PatchOperation> Patches()
-        {            
+        {
             return _operations.Value.Enumerate().SelectMany(x => x.Value).OfType<PatchOperation>();
         }
 
@@ -198,7 +199,7 @@ namespace Marten.Services
         {
             var changes = buildChangeSet(batch);
 
-            await batch.ExecuteAsync(token).ConfigureAwait(false);            
+            await batch.ExecuteAsync(token).ConfigureAwait(false);
 
             ClearChanges(changes.Changes);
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by barclayadam** [comment]
This build has 2 failing tests, but I could not get those to work locally before my changes either. Is this a known issue or am I missing something?

Would love to get this progressed if possible

**Comment by oskardudycz** [comment]
@barclayadam I'll verify that tomorrow.

File: src/Marten/Services/UnitOfWork.cs [hunk 1]
```diff
@@ -342,7 +336,73 @@ public void Eject<T>(T document)
...
+                // Will loop through up the inheritance chain until reaches the end or the index is found, used
+                // to handle inheritance as topologically sorted array may not have the subclasses listed
+                var documentType = x.DocumentType;
+                var index = 0;
+
+                do
+                {
+                    index = _topologicallyOrderedTypes.IndexOf(documentType);
+                    documentType = documentType.BaseType;
+                } while (index == -1 && documentType != null);
```
oskardudycz: Are you sure that we won't ever end up in infinite loop?

src/Marten/Services/UnitOfWork.cs
```diff
@@ -342,7 +336,73 @@ public void Eject<T>(T document)
...
+        {
+            private readonly Type[] _topologicallyOrderedTypes;
+
+            public StorageOperationComparer(Type[] topologicallyOrderedTypes)
+            {
+                _topologicallyOrderedTypes = topologicallyOrderedTypes;
+            }
+
+            public int Compare(IStorageOperation x, IStorageOperation y)
+            {
```
oskardudycz: Might be worth check Reference equals (for performance reasons).

src/Marten/Services/UnitOfWork.cs
```diff
@@ -342,7 +336,73 @@ public void Eject<T>(T document)
...
+                if (x == y)
+                {
+                    return 0;
+                }
+
+                if (x?.DocumentType == null || y?.DocumentType == null)
+                {
+                    return 0;
```
oskardudycz: Are you sure that if one of the storage operations are `null` then we should return `0`?

src/Marten/Services/UnitOfWork.cs
barclayadam: This will break out of the loop once `documentType.BaseType` becomes null, which happens once you hit `object` (goes up the type hierarchy of `documentType`).

src/Marten/Services/UnitOfWork.cs
barclayadam: Done

src/Marten/Services/UnitOfWork.cs
barclayadam: To be honest, I was not sure what to do here. I choose 0 to try and maintain relative order of operations in this case, as without knowing what they actually are the order it _should_ be feels undefined

barclayadam:
@oskardudycz Thanks for the review. I've updated the branch with the `ReferenceEquals` check and commented on your other points

jeremydmiller:
@oskardudycz The broken tests are valid. I think it's coming from some old bug fixes and edge cases where you get a mix of partial operations, delete by where clause, and other things.

I'm absolutely booked up today, but I could maybe try to take this one to the end tomorrow.

I'd like to get @barclayadam 's idea to memoize the topological sort of the type dependencies too

jeremydmiller:
No luck yet. The sorting algorithm isn't quite right, so there's still a bit more to do.

jeremydmiller:
@barclayadam @oskardudycz I added some minor tweaks to finesse this around the broken tests and soften the impact when it's not necessary. I pushed this to master directly w/ a rebase first (blame me and my inadequate Git skills). Thank you for the fix.

PR #1240, opened 2019-04-01
FTS index comparison must take into account automatic cast to text. N…

jokokko:
…ow the indices get recreated every time (with AutoCreate.All).

E.g. `to_tsvector('english'::regconfig, (data ->> 'name'::text)))` is searched for `((data ->> 'Name'))`. Make comparison on lower
invariant values & trim parenthesis.


jokokko:
Uh, some additional noise in the PR that did not see in my difftool...

Issue #1242, opened 2019-04-03
Event Store broken with NodaTime


lngr:
It seems the Event Store is broken when using NodaTime:

```
System.InvalidCastException: Can't cast database type timestamp with time zone to DateTime
   at Npgsql.NpgsqlDefaultDataReader.GetFieldValue[T](Int32 column) in C:\projects\npgsql\src\Npgsql\NpgsqlDefaultDataReader.cs:line 104
   at Marten.Events.StreamStateByIdHandler`1.Resolve(DbDataReader reader, IIdentityMap map, QueryStatistics stats)
   at Marten.Services.CommandRunnerExtensions.<>c__DisplayClass2_0`1.<Fetch>b__0(NpgsqlCommand c)
   at Marten.Services.ManagedConnection.Execute[T](NpgsqlCommand cmd, Func`2 func)
```

The call was `session.Events.FetchStreamState(streamId);`

This is  Marten 3.5.0 with Marten.NodaTime 1.0.0.

Looking at the source code, the reason seems to be several casts to `DateTime`, where `DateTimeOffset` would (maybe?) be a more appropriate option.

https://github.com/JasperFx/marten/blob/master/src/Marten/Events/StreamStateByGuidHandler.cs#L81

Unfortunately, `DateTime` is used at several places through the public interface of the event store (including `IEventStore`), so I'm not sure if this can easily be fixed...

oskardudycz:
@lngr I'll check it asap. I'll keep you updated.

lngr:
I did some tests. Changing `DateTime` to `DateTimeOffset` in the file above is not sufficient. The `GetFieldValue` requires a NodaTime type. 

I think the calls to `Resolve` and `ResolveAsync` require a conversion strategy that is replaced by `UseNodaTime()`.

(In `StreamStateByGuidHandler`, and probably also `StringIdentifiedEventSelector` and other places where `GetFieldValue<DateTime>` / `GetFieldValue<DateTimeOffset>` is used.

oskardudycz:
Thanks for investigation. I think that it might end up with need to extend that methods to be able to inject other date types (in this case taken from NodaTime).

tetious:
With the current draft PR, the EventStore async daemon is failing when the store is set to UseNodaTime, with the following:

```
System.NotSupportedException: The CLR type System.TimeSpan isn't natively supported by Npgsql or your PostgreSQL. To use it with a PostgreSQL composite you n
[...]

PR #1246, opened 2019-04-10
#1242 - fixes for EventStore NodaTime support

oskardudycz:


oskardudycz:
@jeremydmiller @mysticmind @jokokko could you look on this workaround solution? It should work as for now. It's not ideal, as was discussed in https://github.com/JasperFx/marten/issues/1242, but it works.

If you're fine with that - I'll add few tests more.

I have plan to provide PoC of Matrix tests for NodaTime. I have idea in my mind. I will prepare PoC in separate PR.

jeremydmiller:
@oskardudycz I'm fine with the approach as is, but it might be a little tighter if instead the new abstraction was how you read from a NpgdsqlDbReader or wrote to an NpgsqlParameter. I think you could make the non-Noda default be just a bit tighter.

oskardudycz:
@jeremydmiller thank you for having a look. I was also thinking about making it closer to Npgsql classes, but didn't come up with exact working solution. I'll try to rethink that. 
I agree that this is rather hotfix than real solution.

PR #1259, opened 2019-04-30
Feature/1258 cannot derive updates for user defined types

### LINKED ISSUE #1258: User Defined Types: AutoCreateSchemaObjects thinks it needs to update the schema objects
When using a user defined type for a `DuplicatedField` and the `AutoCreateSchemaObjects `option is not `AutoCreate.None`, then Marten thinks it has to update the schema objects every time the `DocumentStore` is created.

Additionally, when the option is the default value `AutoCreate.CreateOrUpdate`, following exception is thrown:

> System.InvalidOperationException : Marten cannot derive updates for objects public.mt_doc_userwithcustomtype

The reason is that the sql command to get the actual table information checks to field `data_type` of the `information_schema.columns` table. But this field contains `USER-DEFINED` when the column is a user-defined type. And when trying to compare with the expected table information there is a difference with the `Type` values.

My suggestion is to use the field `udt_name` of the `information_schema.columns` table when the `data_type` column contains `USER-DEFINED`.

I'll create a PR for this.

dapalmi:
Closes #1258

### DIFF
--- a/src/Marten.Testing/Bugs/Bug_1258_cannot_derive_updates_for_objects.cs
+++ b/src/Marten.Testing/Bugs/Bug_1258_cannot_derive_updates_for_objects.cs
@@ -36,8 +36,8 @@ public class Bug_1258_cannot_derive_updates_for_objects : IntegratedFixture
             using (var session = theStore.OpenSession())
             {
                 var cmd = new NpgsqlCommand(@"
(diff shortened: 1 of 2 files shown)


jeremydmiller:
I've reviewed this and it looks good to me. I'm pulling this in.

Issue #1193, opened 2019-02-15
Linq Count() - predicate omitted


robert-skarzycki:
Having that model (https://gist.github.com/robert-skarzycki/c0c26f32e0000c17faabc6c5debde55c) and data inserting code (https://gist.github.com/robert-skarzycki/aa92e11a27fdb33ea8f89e59ea90329d), this query returns 0 elements:
```
var res = session
                .Query<Root>()
                .Where(r => r.ChildsLevel1.Count(c1 => c1.Name == "child-1.1") == 1);
```
However, it should return 1 element ("root-1" contains exactly one child on level 1 with name "child-1.1").
I see that generated command (extract with debugger) looks like this:
```
select d.data, d.id, d.mt_version from public.mt_doc_root as d where jsonb_array_length(CAST(d.data ->> 'ChildsLevel1' as jsonb)) = :arg0
```
and the `:arg0` equals to `1`. That suggests me that the predicate inside `Count` is just omitted. I expect either having it working properly or to see exception stating such a thing is not allowed.

oskardudycz:
Thank you for creating this issue. As I stated in [my comment in other issue](https://github.com/JasperFx/marten/pull/1107#issuecomment-464845482) I'll try to work on that asap, but I can't put specific date where the fix will be delivered.

oskardudycz:
This case is still not fixed.
@jeremydmiller I added failing tests for this scenario: https://github.com/JasperFx/marten/pull/1263. Is this something that you plan to tackle in your Linq overhaul part 1?

PR #1263, opened 2019-05-11
Added tests for #1193 - Linq Count() - predicate omitted 

oskardudycz:


jeremydmiller:
@oskardudycz I'm pulling this in locally right now. And I've got the fix. Marten is moving again!

PR #1264, opened 2019-05-11
Fixes #1148 - Query with entity's own property comparison not working as expected

### LINKED ISSUE #1148: Query with entity's own property comparison not working as expected
When comparing entity's own properties in a query results in a query where properties are compared vice-versa.

Example: `session.Query<TestEntity>().Where(e => e.Second > e.First)`

Expected result: `select d.data, d.id, d.mt_version from public.mt_doc_testentity as d where CAST(d.data ->> 'Second' as integer) > CAST(d.data ->> 'First' as integer)`

Actual result: `select d.data, d.id, d.mt_version from public.mt_doc_testentity as d where CAST(d.data ->> 'First' as integer) > CAST(d.data ->> 'Second' as integer)`

Simple repro code here: https://gist.github.com/kasparurban/1e816c190d8f03abbf364ff4eb8f0e6f

oskardudycz:
Fixes #1148


src/Marten/Linq/Parsing/SimpleBinaryComparisonExpressionParser.cs
```diff
@@ -34,29 +34,30 @@ public bool Matches(BinaryExpression expression)
 
         public IWhereFragment Parse(IQueryableDocument mapping, ISerializer serializer, BinaryExpression expression)
         {
-            var isValueExpressionOnRight = expression.Right.IsValueExpression();
+            var areBothMemberExpresions = !expression.Left.IsValueExpression() && !expression.Right.IsValueExpression();
```
mysticmind: Please rename the variable from `areBothMemberExpresions` to `areBothMemberExpressions`

src/Marten/Linq/Parsing/SimpleBinaryComparisonExpressionParser.cs
```diff
@@ -34,29 +34,30 @@ public bool Matches(BinaryExpression expression)
...
-		        value = fieldOther.SqlLocator;
-	        }
-	        else
-	        {
-		        memberAccess = null;
-				value = field.GetValue(valueExpression);
-	        }
-
-	        var jsonLocator = field.SqlLocator;
+            var field = mapping.FieldFor(members);
```
mysticmind: There are no changes in this code path, not sure why it is showing a diff

oskardudycz: @mysticmind I have autoformatting turned on in my visual studio. It seems that file had previously mixed tabs and spaces. Now it fixed it to have spaces.

src/Marten/Linq/Parsing/SimpleBinaryComparisonExpressionParser.cs
oskardudycz: Fixed

oskardudycz:
@mysticmind comments fixed

Issue #870, opened 2017-09-07
linq support for child collection and child collection of child collection


KopIlya:
Hello. I know that linq support is very difficult. From documentation I understood some queries are supported for searching in child collections. 
But what about if child collection is not simple type array? Like this:

```
    public class BaseClass : PocoStorableObject
    {
        public List<SubChild> SubChildren { get; set; }
        public SubChild SingleSubChild { get; set; }

        public string Name { get; set; }
    }

    public class SubChild
    {
        public List<SubSubChild> SubSubChildren { get; set; }
        public SubSubChild SingleSubSubChild { get; set; }

        public int SubAge { get; set; }
        public string SubName { get; set; }
    }

    public class SubSubChild
    {
        public string SubSubName { get; set; }
    }
```

And queries like this doesn't work:
```
var query1 = martenQueryable.Where(o => o.SubChildren.Any()).ToList();
var query2 = martenQueryable.Where(o => o.SubChildren.Any(c => c.SubSubChildren != null)).ToList();
var query3 = martenQueryable.Where(o => o.SubChildren.Any(c => c.SubSubChildren.Any())).ToList();
var query4 = martenQueryable.Where(o => o.SubChildren.Any(c => Marten.Util.StringExtensionMethods.Contains(c.SubName, "IVAN", StringComparison.OrdinalIgnoreCase))).ToList();
var query5 = martenQueryable.Where(o => o.SubChildren.Any(c => Marten.Util.StringExtensionMethods.Contains(c.SingleSubSubChild.SubSubName, "IVAN", StringComparison.OrdinalIgnoreCase))).ToList();

string[] names = new string[] {"name1, name2"};
var query6 = martenQueryable.Where(o => o.SubChildren.Any(c => c.SingleSubSubChild.SubSubName.IsOneOf(names))).ToList();
```

How difficult is it to implement query support for such child collections?
Will there be support for such requests in the near future?

robert-skarzycki:
I've just faced same issue. :/

oskardudycz:
See currently supported scenarios. https://github.com/JasperFx/marten/pull/1107 We added some improvements for the inner collections, but it's not yet finished. 
I added tests for @KopIlya scenarios in https://github.com/JasperFx/marten/pull/1265.

@jeremydmiller do you plan to solve them in your Linq overhaul?

jeremydmiller:
Hey folks, this is **done** in a v4 branch.

jeremydmiller:
[...]

PR #1265, opened 2019-05-11
Added failing tests for #870 - linq support for child collection and child collection of child collection

oskardudycz:
Failing tests for scenarios from #870

### DIFF
--- a/src/Marten.Testing/Linq/query_with_inner_query_with_global_CollectionStorage_WithArray.cs
+++ b/src/Marten.Testing/Linq/query_with_inner_query_with_global_CollectionStorage_WithArray.cs
@@ -28,6 +28,8 @@ namespace Marten.Testing.Linq
 
         public IReadOnlyCollection<TypeWithInnerCollections> IReadonlyCollectionOfInnerClasses { get; set; }
 
+        public TypeWithInnerCollections Child { get; set; }
+
         public static TypeWithInnerCollections Create(params string[] array)
         {
             return new TypeWithInnerCollections()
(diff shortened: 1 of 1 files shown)


mysticmind:
@oskardudycz Looks like the recent fixes on #1334 have resolved this issue. Can you please fix the file conflict and run a test on this?

oskardudycz:
@mysticmind I rebased the master, unfortunately, it seems that those queries are still not supported.

mysticmind:
@jeremydmiller @oskardudycz @jokokko I was looking at ToroDB to learn more since it also works on top of Postgres JSONB, really like the model they adopted to not store document with nested structures but break it down into sub documents and store it separately. See the details from the ToroDB internals as below:
> JSON documents are split by hierarchy levels into “subdocuments”, which contain no nested structures. Each subdocument level is stored separately
Subdocuments are classified by “type”. Each “type” maps to a different table.

It will be interesting to read this [doc](https://info.citusdata.com/rs/235-CNE-301/images/ToroDB_internals_How_to_Create_NoSQL_on_top_of_SQL_-_Alvaro_Hernandez.pdf)

Possibly if we build upon this idea, I think we can easily deal with querying nested child collection across multiple Postgres versions without using JSONPath(SQL/JSON) which is available only from PG12+.

Please let me know your thoughts.

jeremydmiller:
I'm grabbing this into a local branch today. I'm planning to finally beat this with some CTE magic

jeremydmiller:
Got this in locally now.

PR #1267, opened 2019-05-12
Updated packages to most recent version. Added Benchmark project to s…

oskardudycz:
Updated packages to most recent version. 

Added Benchmark project to solution, fixed compilation errors

### DIFF
--- a/azure-pipelines-benchmarks.yml
+++ b/azure-pipelines-benchmarks.yml
@@ -57,8 +57,9 @@ jobs:
         displayName: Create db and add plv8 extension
       - script: ./build.sh ci      
       - script: ./build.sh benchmarks
+        displayName: BuildAndBenchmark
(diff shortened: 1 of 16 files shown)


oskardudycz:
@mysticmind it seems that JSON.NET `12.x` has some performance issues :| 
Here are the details (see `BuildAndBenchmark` log)
- version with `12.x` - https://dev.azure.com/jasperfx-marten/marten/_build/results?buildId=3
- version with `11.x` - https://dev.azure.com/jasperfx-marten/marten/_build/results?buildId=41
- version from master - https://dev.azure.com/jasperfx-marten/marten/_build/results?buildId=33
I run those benchmarks few times, although we should give them some love, as the results are fluctuating a bit from one build to another.

I downgraded the JSON.NET back to `11.0.2` (as it is on master). I'll try tomorrow to update build configuration to copy benchmarks artifacts.

PR #1268, opened 2019-05-13
Added Auto LF and .editorconfig settings for aligning formatting rules

oskardudycz:
@jeremydmiller @mysticmind @jokokko if you're fine with this `.editorsettings` then I'll do code clean up and change warnings settings into errors.

See rules explanation: https://docs.microsoft.com/en-US/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2017

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,6 @@
 * text=auto *.sh eol=lf
+*.png binary
+*.ttf binary
+*.jpg binary
+*.jpeg binary
+*.pdf binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
-* text=auto *.sh eol=lf
+* text=auto 
+*.sh eol=lf
 *.png binary
 *.ttf binary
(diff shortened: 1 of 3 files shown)


.gitattributes
```diff
@@ -0,0 +1,6 @@
+* text=auto *.sh eol=lf
```
mysticmind: Please move `*.sh eol=lf` to a new line

jeremydmiller:
I don't see anything in there that would bug me. Might also be worthwhile to add R# settings too though

jeremydmiller:
I don't see anything in there that would bug me. Might also be worthwhile to add R# settings too though

oskardudycz:
@jeremydmiller I'm not using currently Resharper, but reading that it appears that Resharper should respect `.editorconfig`

oskardudycz:
@jokokko imho it's better to keep it in new line, for me it's not readable and confusing when we have it in single line.

PR #1281, opened 2019-06-05
Add functionality to get pg version using document store

mysticmind:
- Implement `GetPostgresVersion` method in `DocumentStore`
- Add unit test


oskardudycz:
@jokokko @mysticmind I like the idea of `IDiagnostics`, probably we could have that and extend, as this is also area that we need to improve.

PR #1294, opened 2019-06-20
Test attribute for targeting min and/or max versions of PG.

jokokko:
Allows for less setup than SkippableFact & PG version looked up only once / test run.


jokokko:
@oskardudycz 👍 done (seems to be non-related test failure... Can requeue checks).

PR #1298, opened 2019-06-25
Fix #1296. Return updates AND upserts in `IUnitOfWork.Updates` and `I…

### LINKED ISSUE #1296: custom listener not working for .Update but works for .Store
we have a custom listener setup like in this article: https://visualstudiomagazine.com/articles/2017/03/01/marten-monitor-respond-aspnet.aspx

the listener looks like this: 
```
namespace EvictionAPISite.Services.Repository
{
    public class CustomDocumentSessionListener : DocumentSessionListenerBase
    {
        public override void BeforeSaveChanges(IDocumentSession session)
        {
            IUnitOfWork pendingChanges = session.PendingChanges;

            var firmUpdates = pendingChanges.UpdatesFor<Firm>();
            foreach (var update in firmUpdates)
            {
				// code goes here
            }            
        }
    }
}
```
elsewhere, when we want to save changes to an updated document we have code like this:

```
var firm = dbSession.Query<Firm>().SingleOrDefault(foo => foo.Id == args.Id);
firm.foo = bar;
dbSession.Update(firm);
dbSession.SaveChanges();
```

and the save works fine, but in the listener, the firmUpdates collection is empty.

if we simply change:

- FROM: dbSession.Update(firm);
- TO: dbSession.Store(firm);

then the save still works, but the firmUpdates collection knows about the change and does its job accordingly.

this behavior is the same whether we use .LightweightSession or .OpenSession or even .OpenSession with dirty tracking.

jokokko:
…UnitOfWork.UpdatesFor`.

### DIFF
--- a/src/Marten.Testing/Bugs/Bug_1296_sessionlistener_updates.cs
+++ b/src/Marten.Testing/Bugs/Bug_1296_sessionlistener_updates.cs
@@ -1,0 +1,56 @@
+using System;
+using System.Linq;
+using Marten.Services;
+using Marten.Testing.Documents;
+using Xunit;
+
+namespace Marten.Testing.Bugs
+{
+    public class Bug_1296_sessionlistener_updates : IntegratedFixture
+    {
+        [Fact]
+        public void bug()
+        {
+            var user = new User();
+
+            using (var session = theStore.OpenSession())
+            {
+                session.Insert(user);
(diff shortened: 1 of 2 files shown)


jokokko:
Should be non-related 2 test failures... `can_select_DateTimeOffset_and_will_return_localtime` being flaky :(

PR #1299, opened 2019-06-25
Updated Scenario `Aggregates, Events, Repositories` to be simplified to not provide additional infrastructure wrapping around Marten

oskardudycz:
Based on the Discussion on Gitter Updated Scenario `Aggregates, Events, Repositories` to be simplified to not provide additional infrastructure wrapping around Marten. That will make it more straightforward for new users (eg. to not think why did we used custom aggregation instead of the default one. 

Fixed also version handling for the optimistic concurrency.

Question raised on Gitter:
```
ddivita
@ddivita
Jun 24 21:39
Int the repository example here: http://jasperfx.github.io/marten/documentation/scenarios/aggregates_events_repositories/ The method : protected void Register<T>(Action<T> handle) is manually setting the handlers as opposed to Marten automatically registering the Applymethods on the Aggregate Root?
```

@ddivita fyi

### DIFF
--- a/src/Marten.Testing/Events/ScenarioAggregateAndRepository.cs
+++ b/src/Marten.Testing/Events/ScenarioAggregateAndRepository.cs
@@ -123,7 +123,7 @@ namespace Marten.Testing.Events
 
             var @event = new InvoiceCreated(invoiceNumber);
             // Instantiation creates our initial event, capturing the invoice number
-            RaiseEvent(@event);
+            Enqueue(@event);
             Apply(@event);
         }
 
@@ -141,7 +141,7 @@ namespace Marten.Testing.Events
             }
 
             var @event = new LineItemAdded(price, vat, description);
-            RaiseEvent(@event);
+            Enqueue(@event);
             Apply(@event);
         }
 
@@ -231,7 +231,7 @@ namespace Marten.Testing.Events
             uncommittedEvents.Clear();
         }
 
-        protected void RaiseEvent(object @event)
+        protected void Enqueue(object @event)
         {
             Version++;
             if (@event is IVersionedEvent versionedEvent)
--- a/src/Marten.Testing/Events/ScenarioAggregateAndRepository.cs
+++ b/src/Marten.Testing/Events/ScenarioAggregateAndRepository.cs
@@ -121,7 +121,7 @@ namespace Marten.Testing.Events
                 throw new ArgumentException("Invoice number needs to be positive", nameof(invoiceNumber));
             }
 
-            var @event = new InvoiceCreated(invoiceNumber);
+            var @event = new InvoiceCreated(invoiceNumber, ++Version);
             // Instantiation creates our initial event, capturing the invoice number
             Enqueue(@event);
             Apply(@event);
@@ -140,7 +140,7 @@ namespace Marten.Testing.Events
                 throw new ArgumentException("Description cannot be empty", nameof(description));
             }
 
-            var @event = new LineItemAdded(price, vat, description);
+            var @event = new LineItemAdded(price, vat, description, ++Version);
             Enqueue(@event);
             Apply(@event);
         }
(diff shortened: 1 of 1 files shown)


src/Marten.Testing/Events/ScenarioAggregateAndRepository.cs
```diff
@@ -70,6 +73,28 @@ public void CanGuardVersion()
...
+            var invoice = CreateInvoice();
+            invoice.Version.ShouldBe(3);
+            repository.Store(invoice);
+
+            // assert version was incremented properly
+            var invoiceFromRepository = repository.Load<Invoice>(invoice.Id);
+            invoiceFromRepository.Version.ShouldBe(3);
+
+            // update aggregate
+            invoiceFromRepository.AddLine(100, 23, "Some nice with with 23% VAT");
```
mysticmind: Please fix the description, there are repeating words

jokokko:
Some remarks regarding the changes:
* The aggregate is about protecting its invariants -> I find the public setter on `Id` confusing, as I'd expect it to not change.
* Each public method on the aggregate now has to remember to call two methods (Enqueue, Apply).
* Each `Apply` method now has to remember to restore the aggregates version.
* The aggregate version is restored from the version stored in the event. Now, while not a typical scenario, what if I decided to rewrite (or copy) the stream and e.g. delete events. Should the version of the aggregate match that of the number of events it is indexing in the stream as that would no longer be the case.

oskardudycz:
@jokokko see my answers below

> The aggregate is about protecting its invariants -> I find the public setter on Id confusing, as I'd expect it to not change.
I can remove it in this example, although this is current Marten limitation for projections to have public setter with Id, so inline projection might not work correctly.

> Each public method on the aggregate now has to remember to call two methods (Enqueue, Apply).
That's correct, but it's the same number of calls as the registrations in constructor. Also it's the common pattern used in Event Sourcing. Having them in the methods we don't need to do custom aggregation. Imho previous solution with custom aggregation might be good or even better solution at the end, but imho we should show in scenarios the most common scenarios. So showing that Marten user needs to write custom aggregator might be confusing and might bring the thoughts "why Marten doesn't have it out of the box? Why do I need to write my own wrappers?", while it's doable without that. Also I personally prefer explit calls where you can debug without reflection and see precisely what's going on instead of reflection.

> The aggregate version is restored from the version stored in the event. Now, while not a typical scenario, what if I decided to rewrite (or copy) the stream and e.g. delete events. Should the version of the aggregate match that of the number of events it is indexing in the stream as that would no longer be the case.
Yes, but without that optimistic concurrency won't work properly, as you need to read from somewhere the version number. Imho we should give the possibility in Marten to apply stream and event metadata to the stream/events field during aggregation (as optional of course). So eg. when aggregate has field Version or event has field version then it will be mapped from the metadata. I know that it's possible to use Event<SomeEventType> but it's cumbersome, as it won't be easily used during the event creation in public method. The other option is to split IEvent into smaller intefaces. 

@jokokko thoughts?

oskardudycz:
@jokokko any thoughts for my response? ;)

oskardudycz:
@jokokko any updates?

jokokko:
@oskardudycz Sorry for the delay. I'm fine with merging this. Although, I am not a fan of events having mutable surface, in this case the `Version` property. Given that events, by definition, should be immutable. There's also the option to have both implementations in the samples if it's not thought too confusing, e.g. they are offered as alternative implementations of each other.

oskardudycz:
@jokokko I removed IVersionedEvent interface with public setter from the Event to not make it mutable (See more in commit: https://github.com/JasperFx/marten/pull/1299/commits/9200cb85639fb3738630eb4d8f62fe1818dd22a1).
Although I don't fully like it, as it's some kind of redundancy of data and need to do it manually (while it's going on under the hood). I created new issue to make possible automatic mapping of metadata information to event: https://github.com/JasperFx/marten/issues/1337

@jokokko what do you think?

jokokko:
@mysticmind I'm fine with merge. Ideally, I'd want the aggregate not to have the mutable `Id` and version propagating to each event & each `Apply` being responsible for "restoring" the aggregate version. But better get this PR off the table.

oskardudycz:
@jokokko public mutable `Id` is current Marten limitation for the Inline Projection. It has to be property named `Id` with public getter and setter. I'd like to change that in `v4.0` but currently it is as it is.

mysticmind:
#1355 has been merged in lieu of this. Closing this one.

PR #1304, opened 2019-06-29
Updated Custom Projection Documentation. 

### LINKED ISSUE #1249: NpgsqlOperationInProgressException during Rebuild when running async query in a ViewProjection
If you are using ViewProjection, with a binding like this: 
```csharp
this.ProjectEvent<ProjectionEvent<UserNew>>(async (s, v, e) => await this.Apply(v, e, s));
```

and a projection method similar to this:
```csharp
private async Task Apply(User v, ProjectionEvent<UserNew> e, IQuerySession s)
{
    var account = await s.LoadAsync<Account>(e.Data.AccountId);
    v.DatabaseName = account.DatabaseName;
}
```

When you run a Rebuild using this projection, the LoadAsync line throws a ```NpgsqlOperationInProgressException```. 

If you use Load instead (non-async) everything is fine.

Hopefully there is enough information here to track this bug(?) down, or to help me understand what I'm doing wrong. :)

oskardudycz:
Updated Custom Projection Documentation.

Added sample for ProjectEventAsync with loading related document with documentSession and using event meta data.

Fixes #1249

@tetious fyi

### DIFF
--- a/documentation/documentation/events/projections/custom.md
+++ b/documentation/documentation/events/projections/custom.md
@@ -12,11 +12,47 @@ or through a class like:
  
 <[sample:viewprojection-from-class]> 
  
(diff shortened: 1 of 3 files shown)


oskardudycz:
@tetious thank you for the feedback :) I'm glad that you like it 👍

PR #1321, opened 2019-07-20
Fix to ensure functions and storage features are added only once - fixes #1306

### LINKED ISSUE #1306: An item with the same key has already been added. Key: patch_doc
Ciao!

So, using Marten 3.3.0, I have this problem that happens pretty rarely, but which softlock my web app and I cannot use it until I restart the cluster.

you can see there:

image

I've registered everything as SingleInstance.

I have handlers that are registered InstancePerRequest which are injecting the Repository and calling his methods, each methods of Repository creates a session, use it and dispose it. So it should be fine.

However, when the softlock happens is because Autofac CANNOT create the DocumentStore instance because of the error:

 An item with the same key has already been added. Key: patch_doc

which is happening in the constructor of DocumentStore with StoreOptions as parameter.

My quick thinking: It's possible that the DocumentStore failed for some reason the very first time (maybe DB was still booting up) and then when it tries to recreate the instance it reuses the SAME StoreOptions which already has some shit in memory and it explodes.

What I should do to prevent this?

mysticmind:
- Fix to ensure transforms functions, system functions, storage features
are added only once to the dictionary
- Add unit test

Fixes #1306

### DIFF
--- a/src/Marten.Testing/StoreOptionsTests.cs
+++ b/src/Marten.Testing/StoreOptionsTests.cs
@@ -111,7 +111,10 @@ namespace Marten.Testing
             var store1 = new DocumentStore(options);
             // pass with the same options and check it does not throw ArgumentException
             // "An item with the same key has already been added. Key: <transform function name/feature schema name>"
-            var store2 = new DocumentStore(options);
+            Should.NotThrow(() =>
+            {
+                var store2 = new DocumentStore(options);
+            });
         }
 
         public class FakeUserStorage : IDocumentStorage, IdAssignment<User>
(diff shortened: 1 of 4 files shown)


src/Marten.Testing/StoreOptionsTests.cs
```diff
@@ -102,6 +102,18 @@ public void default_ddl_rules()
...
+        [Fact]
+        public void ensure_patch_system_transform_functions_and_feature_schemas_are_added_only_once()
+        {
+            var options = new StoreOptions();
+            options.Connection(ConnectionSource.ConnectionString);
+
+            var store1 = new DocumentStore(options);
+            // pass with the same options and check it does not throw ArgumentException
+            // "An item with the same key has already been added. Key: <transform function name/feature schema name>"
+            var store2 = new DocumentStore(options);
```
oskardudycz: @mysticmind could you add explicit check with `Should.NotThrow` method? Then the comment wouldn't be needed.

mysticmind: @oskardudycz god suggestion, I have added `Should.NotThrow`. Albeit, it does not have the means to pass a specific exception.

Issue #1190, opened 2019-02-14
Add ability to create gin indexes on specific jsonb fields


barryhagan:
For an object in my database, I need to use a containment query like this:
```
var matches = new[] {"abc", "xyz"};
context.Query<T>().Where(x => x.MemberIds.Any(m => matches.Contains(m)));
```
Marten has specific support for this query, and produces a Postgres search like this:
```
data -> 'MemberIds' ?| array['abc', 'xyz']`
```

The problem is that this type of query will not be able to use either a ComputedIndex on this field via `For<T>.Index(x => x.MemberIds)`, nor the complete gin index on data that can be added using `For<T>.GinIndexJsonData()`.

The computed index does not work because it selects using `data ->> MemberIds` and the gin index does not work because the query needs `data -> MemberIds` instead of just `data`.

It seems like the best solution here would be the ability to create an `IndexDefinition` instead of a `ComputedIndex` for this field, which could be setup as gin with jsonb_ops to support the query.

Right now the only way I've found to hack this is to directly modify the Expression on the gin index of data, so the specific field is also available with jsonb_ops instead of jsonb_path_ops to support the query:

```
// ugly, ugly hack
For<T>().GinIndexJsonData(def => 
 def.Expression = "data jsonb_path_ops, ((data -> 'MemberIds'::text))"
);
```
This hack is very fragile because the formatting of the Expression is used to when doing a SchemaUpdate check.  Those extra parens around the field and spacing is critical or it will try to rebuild the index on each schema check.

Ideally I'd like to have something like this which would create an index that supports `?|` for the specific Any().Contains() query on MemberIds
```
public class DocumentMappingExpression<T>
{
    public DocumentMappingExpression<T> GinIndex(Expression<Func<T, object>> expression, Action<IndexDefinition> configure = null);
}
```

oskardudycz:
@barryhagan  I think that what you're trying to achieve is already supported. See `Customizing a Calculated Index` section of documentation: http://jasperfx.github.io/marten/documentation/documents/customizing/computed_index/

So you could do:

```
var store = DocumentStore.For(_ =>
{
    _.Connection(ConnectionSource.ConnectionString);

    // The second, optional argument to Index()
    // allows you to customize the calculated index
    _.Schema.For<YourTypes>().Index(x => x.MemberIds, x =>
    {
        // Change the index method to "gin"
        x.Method = IndexMethod.gin;
    });
});

```

barryhagan:
[...]

PR #1334, opened 2019-08-12
Issue #1190 -  ensure the locator for a CollectionAnyContainment query matches the field's index

barryhagan:
This required changes the JsonLocatorField constructors so the generated locators were consistent regardless of field depth.

Before this change, an array would change casting behavior depending on depth: 

```
public class T
{
    public String[] MyArray { get; set; }
    public T Inner { get; set; }
}
```
 `CAST(d.data ->> 'MyArray' as jsonb)`
vs
 `CAST(d.data -> 'Inner' ->> 'MyArray' as varchar[])`
 
I'm still unclear why arrays as top level properties were always being cast to jsonb, but I left it that way.

### DIFF
--- a/src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
+++ b/src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
@@ -22,6 +22,7 @@ namespace Marten.Linq
         private readonly SubQueryExpression _expression;
         private readonly IQueryableDocument _mapping;
 
+        [Obsolete("Use the constructor that takes IQueryableDocument instead. This might be removed in v4.0.")]
         public CollectionAnyContainmentWhereFragment(MemberInfo[] members, ISerializer serializer, SubQueryExpression expression) : this(members, serializer, expression, null)
         {
         }
@@ -203,7 +204,7 @@ namespace Marten.Linq
             var members = visitor.Members;
             if (!members.Any())
                 throwNotSupportedContains();
-            var path = _mapping?.FieldFor(members).SqlLocator ?? CommandBuilder.BuildJsonStringLocator("d.data", members.ToArray(), _serializer.Casing);
+            var path = _mapping?.FieldFor(members).SqlLocator ?? $"CAST ({CommandBuilder.BuildJsonStringLocator("d.data", members.ToArray(), _serializer.Casing)} as jsonb)";
             return $"{path} ?| :{fromParam.ParameterName}";
         }
 
--- a/src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
+++ b/src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using Baseline;
+using Marten.Schema;
 using Marten.Util;
 using NpgsqlTypes;
 using Remotion.Linq.Clauses;
@@ -19,12 +20,14 @@ namespace Marten.Linq
         private readonly MemberInfo[] _members;
         private readonly ISerializer _serializer;
         private readonly SubQueryExpression _expression;
+        private readonly IQueryableDocument _mapping;
 
-        public CollectionAnyContainmentWhereFragment(MemberInfo[] members, ISerializer serializer, SubQueryExpression expression)
+        public CollectionAnyContainmentWhereFragment(MemberInfo[] members, ISerializer serializer, SubQueryExpression expression, IQueryableDocument mapping)
         {
             _members = members;
             _serializer = serializer;
             _expression = expression;
+            _mapping = mapping;
         }
 
         public void Apply(CommandBuilder builder)
@@ -196,8 +199,7 @@ namespace Marten.Linq
             var members = visitor.Members;
             if (!members.Any())
                 throwNotSupportedContains();
-            var path = members.Select(m => m.Name.FormatCase(_serializer.Casing)).Join("'->'");
-            return $"d.data->'{path}' ?| :{fromParam.ParameterName}";
+            return $"{_mapping.FieldFor(members).SqlLocator} ?| :{fromParam.ParameterName}";
         }
 
         private void throwNotSupportedContains()
--- a/src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
+++ b/src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
@@ -22,6 +22,10 @@ namespace Marten.Linq
         private readonly SubQueryExpression _expression;
         private readonly IQueryableDocument _mapping;
 
+        public CollectionAnyContainmentWhereFragment(MemberInfo[] members, ISerializer serializer, SubQueryExpression expression) : this(members, serializer, expression, null)
+        {
+        }
+
         public CollectionAnyContainmentWhereFragment(MemberInfo[] members, ISerializer serializer, SubQueryExpression expression, IQueryableDocument mapping)
         {
             _members = members;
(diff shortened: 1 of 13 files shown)


src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
```diff
@@ -19,12 +20,14 @@ public class CollectionAnyContainmentWhereFragment: IWhereFragment
         private readonly MemberInfo[] _members;
         private readonly ISerializer _serializer;
         private readonly SubQueryExpression _expression;
+        private readonly IQueryableDocument _mapping;
 
-        public CollectionAnyContainmentWhereFragment(MemberInfo[] members, ISerializer serializer, SubQueryExpression expression)
+        public CollectionAnyContainmentWhereFragment(MemberInfo[] members, ISerializer serializer, SubQueryExpression expression, IQueryableDocument mapping)
```
oskardudycz: **Note:** This will be breaking change. I'm not sure if that's fine, as this class should be more internal rather than public. @jeremydmiller @mysticmind @jokokko thoughts?

src/Marten/Schema/JsonLocatorField.cs
```diff
@@ -14,38 +15,55 @@ public static JsonLocatorField For<T>(EnumStorage enumStyle, Casing casing, Expr
...
             var property = ReflectionHelper.GetProperty(expression);
 
-            return new JsonLocatorField("d.data", new StoreOptions(), enumStyle, casing, property);
+            return new JsonLocatorField("d.data", null, enumStyle, casing, property);
         }
 
         private readonly Func<Expression, object> _parseObject = expression => expression.Value();
 
-        public JsonLocatorField(string dataLocator, StoreOptions options, EnumStorage enumStyle, Casing casing, MemberInfo member) : base(enumStyle, member)
+        public JsonLocatorField(string dataLocator, string databaseSchemaName, EnumStorage enumStyle, Casing casing, MemberInfo member, string pgType = null) :
```
oskardudycz: **FIX:** Same comment as in https://github.com/JasperFx/marten/pull/1334/files#r313276676. But here I think that we should keep backward compatibility, so at least keep the constructor with Store options.

src/Marten/Schema/JsonLocatorField.cs
```diff
@@ -14,38 +15,55 @@ public static JsonLocatorField For<T>(EnumStorage enumStyle, Casing casing, Expr
...
+
+            if (MemberType == typeof(string) || isStringEnum)
+            {
+                SqlLocator = $"{locator}";
+            }
+            else if (TypeMappings.TimespanTypes.Contains(MemberType))
             {
-                SqlLocator = $"{options.DatabaseSchemaName}.mt_immutable_timestamp({dataLocator} ->> '{memberName}')";
-                SelectionLocator = $"CAST({dataLocator} ->> '{memberName}' as {PgType})";
+                SqlLocator = $"{databaseSchemaName ?? StoreOptions.DefaultDatabaseSchemaName}.mt_immutable_timestamp({locator})";
```
oskardudycz: **FIX:** If the `options` parameter from the constructor was removed then `StoreOptions` won't be initialized and available.

src/Marten/Schema/JsonLocatorField.cs
```diff
@@ -14,38 +15,55 @@ public static JsonLocatorField For<T>(EnumStorage enumStyle, Casing casing, Expr
...
+        public JsonLocatorField(string dataLocator, string databaseSchemaName, EnumStorage enumStyle, Casing casing, MemberInfo member, string pgType = null) :
+            this(dataLocator, databaseSchemaName, enumStyle, casing, new[] { member }, pgType)
         {
-            var memberType = member.GetMemberType();
-            var memberName = member.Name.FormatCase(casing);
+        }
+
+        public JsonLocatorField(string dataLocator, string databaseSchemaName, EnumStorage enumStyle, Casing casing, MemberInfo[] members, string pgType = null) : base(enumStyle, members)
+        {
+            var locator = new StringBuilder(dataLocator);
```
oskardudycz: **SUGGESTION:** This location building logic would be nice to move to some separate method/class.

src/Marten/Schema/JsonLocatorField.cs
```diff
@@ -63,30 +81,6 @@ public JsonLocatorField(string dataLocator, StoreOptions options, EnumStorage en
             }
         }
 
-        public JsonLocatorField(string dataLocator, EnumStorage enumStyle, Casing casing, MemberInfo[] members) : base(enumStyle, members)
```
oskardudycz: **NOTE:** this is also breaking change.

src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
mysticmind: It would be better to ensure the changes are binary compatible.

barryhagan: Should I mark this obsolete then?  I'll avoid the breaking change, but IDocumentQuery is needed to build a consistent locator for the query to fix the bug.

src/Marten/Schema/JsonLocatorField.cs
barryhagan: restored

src/Marten/Schema/JsonLocatorField.cs
barryhagan: `options` restored, but this was just using a static const for the default value.

src/Marten/Schema/JsonLocatorField.cs
barryhagan: Moved out into refactored `CommandBuilder`

src/Marten/Schema/JsonLocatorField.cs
barryhagan: restored

src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
oskardudycz: @barryhagan yes, I think that you can mark it as obsolete with the note that it might be removed in `v4.0`

src/Marten/Schema/JsonLocatorField.cs
```diff
@@ -63,28 +74,9 @@ public JsonLocatorField(string dataLocator, StoreOptions options, EnumStorage en
...
+        public JsonLocatorField(string dataLocator, EnumStorage enumStyle, Casing casing, MemberInfo[] members) :
+            this(dataLocator, new StoreOptions(), enumStyle, casing, members, null)
```
oskardudycz: @barryhagan I think that for that case it'd be better to pass just `null`. Sending default StoreOptions might be misleading, as they don't relate with the StoreOptions that user have set up in `DocumentStore` initialization. Imho having null would be more clear indication that for this case we're not sending any options.

src/Marten/Linq/CollectionAnyContainmentWhereFragment.cs
barryhagan: added `[Obsolete()]` to this and other dead constructors

src/Marten/Schema/JsonLocatorField.cs
barryhagan: done

src/Marten/Util/TypeMappings.cs
```diff
@@ -310,17 +310,6 @@ public static bool HasTypeMapping(Type memberType)
             return ResolvePgType(memberType) != null || memberType.IsEnum;
         }
 
-        public static string ApplyCastToLocator(this string locator, EnumStorage enumStyle, Type memberType)
```
mysticmind: Can we just mark this method as obsolete rather than removing it to make it a non-breaking change in terms of binary compatibility? Also, include a comment that it will be removed in v4.0.

barryhagan: done

PR #1341, opened 2019-08-20
Fixes an issue where the `ViewProjection` class can process events out of order

ericgreenmix:
When Delete and Store events on the same stream were on the same page of the ViewProjection, Delete would always happen after calling Store, causing the events to potentially be processed out of order


mysticmind:
I have re-run PG11 build, noticed a failing test related to events (possibly a flaky one).

PR #1349, opened 2019-09-05
ViewProjection - Only Delete view if it exists

ericgreenmix:
There is a bug in my last ViewProjection changes that would attempt to delete a view if it was `null`. This fixes that issue by checking if the view exists before calling delete.


ericgreenmix:
Done

oskardudycz:
@ericgreenmix  thank you 👍

PR #1350, opened 2019-09-05
Make the search for upsert write method more precise

baronfel:
While investigating the update to npgsql 4.1.0, I found that the ordering of the members on the NpgsqlBinaryImporter had changed so that the `Write<T>(T, NpgsqlDbType)` overload was no longer the first two-parameter overload.  After some investigation in the repl I found the intended overloads in a way that works across versions.

<details>
<summary>Repl code to verify the 'is the same method' assertion</summary>

```fsharp

// uncomment one of these
#r "/Users/chethusk/.nuget/packages/npgsql/4.1.0-preview2/lib/netcoreapp3.0/Npgsql.dll"
//or
//#r "/Users/chethusk/.nuget/packages/npgsql/4.0.9/lib/netstandard2.0/Npgsql.dll"

open Npgsql
open System.Reflection

let methods = typeof<NpgsqlBinaryImporter>.GetMethods()

// existing code in UpsertArgument.cs
let method = methods |> Array.tryFind (fun x -> x.GetParameters().Length = 2);

// new proposed algorithm
let meth =
    methods
    |> Array.tryFind (fun x -> 
        x.Name = "Write" 
        && x.IsGenericMethod 
        && x.GetParameters().Length = 2
        && x.GetParameters().[0].ParameterType.IsGenericParameter
        && x.GetParameters().[1].ParameterType = typeof<NpgsqlTypes.NpgsqlDbType>)

method = meth

```
</details>

The output of this when referencing the 4.0.9 dll is `true`, meaning the overloads are the same for both functions, whereas on 4.1.0 the output is `false`. The overload that's found for the current detection is `System.Threading.Tasks.Task WriteAsync[T](T, System.Threading.CancellationToken)` on the 4.1.0 previews.

This isn't a huge change, but it should hopefully smooth the way towards 4.1.0 and make the intent of the `writeMethod` member more clear.

### DIFF
--- a/src/Marten/Schema/Arguments/UpsertArgument.cs
+++ b/src/Marten/Schema/Arguments/UpsertArgument.cs
@@ -12,7 +12,7 @@ namespace Marten.Schema.Arguments
     public class UpsertArgument
     {
         protected static readonly MethodInfo writeMethod =
-            typeof(NpgsqlBinaryImporter).GetMethods().FirstOrDefault(x => x.GetParameters().Length == 2);
+            typeof(NpgsqlBinaryImporter).GetMethods().FirstOrDefault(x => x.Name == "Write" && x.GetParameters().Length == 2 && x.GetParameters()[0].ParameterType.IsGenericParameter && x.GetParameters()[1].ParameterType == typeof(NpgsqlTypes.NpgsqlDbType));
 
         protected static readonly MethodInfo _paramMethod = typeof(SprocCall)
             .GetMethod("Param", new[] { typeof(string), typeof(object), typeof(NpgsqlDbType) });


baronfel:
When coupled with the changes to npgsql in https://github.com/npgsql/npgsql/pull/2608, this change should be all that's required to ensure that marten works with the 4.x and 4.1 versions of npgsql.

baronfel:
I'd be happy to when npgsql 4.1.0 goes stable. Right now it's on 4.1.0-preview2, and I was testing it in my work application specifically for connection-pooling contention improvements when I discovered this incompatibility.  If you're interested, the mentioned PR has grown quite a bit larger because of several (un)intended API surface area changes, some of which touch npgsql type handling that may impact marten as well.

oskardudycz:
@baronfel ah ok, I thought that I missed 4.1.0 release and it’s there.
Thank you for the heads up. After we merge your PR I’ll  do Some branch with this upgrade and see what might be broken.

baronfel:
In particular I think the `Transaction.IsCompleted()` checks that are done are going to be a problem: https://github.com/npgsql/npgsql/issues/2416

This PR suggests that Marten would need to change to no longer rely on the `NpgsqlTransaction` itself knowing its state, which is a change that I think should be made _before_ 4.1.0 releases because it sounds like that particular API decision isn't going to be reverted: https://github.com/npgsql/npgsql/pull/2608#issuecomment-529193494.

PR #1351, opened 2019-09-06
Add functionality to define duplicate field with not null constraint

### LINKED ISSUE #1198: Allow Duplicated fields to be marked as NOT NULL

mysticmind:
- Add functionality to define duplicate field with not null constraint
using attribute and fluent interface
- Add unit tests
- Add documentation
- fixes #1198

### DIFF
--- a/src/Marten/Storage/DocumentTable.cs
+++ b/src/Marten/Storage/DocumentTable.cs
@@ -178,7 +178,7 @@ namespace Marten.Storage
     {
         private readonly DuplicatedField _field;
 
-        public DuplicatedFieldColumn(DuplicatedField field) : base(field.ColumnName, field.PgType)
+        public DuplicatedFieldColumn(DuplicatedField field) : base(field.ColumnName, field.PgType, field.NotNull ? "NOT NULL": "NULL")
         {
             CanAdd = true;
             _field = field;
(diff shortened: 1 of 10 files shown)


src/Marten/Storage/DocumentTable.cs
```diff
@@ -178,7 +178,7 @@ public class DuplicatedFieldColumn: TableColumn
     {
         private readonly DuplicatedField _field;
 
-        public DuplicatedFieldColumn(DuplicatedField field) : base(field.ColumnName, field.PgType)
+        public DuplicatedFieldColumn(DuplicatedField field) : base(field.ColumnName, field.PgType, field.NotNull ? "NOT NULL": "NULL")
```
oskardudycz: **SUGGESTION:** Maybe it would be worth to make the `NOT NULL` and  `NULL` constants? I personally prefer to have them for such case, it's usually less error prone.

mysticmind:
> 
> 
> Looks good @mysticmind!
> Could you add some acceptance tests veryfing that documents with such `DuplicatedField` configuration are added/updated correctly?

I did add tests around the SQL generation in my PR. I will add acceptance tests as well sometime later today.

PR #1361, opened 2019-09-30
Updated minimum version of Npgsql to 4.0.9 based on the users feedback

oskardudycz:
Because the version of the Npgsql was lowered to `4.0.4` then users without doing a cleanup might have failed first build with cached nugets.
Raised the version to limit such issues.


oskardudycz:
Closing because it'll be obsolete with https://github.com/JasperFx/marten/pull/1362 and version `3.9.0`

PR #1363, opened 2019-10-01
Allow case invariance for netcore apps

eouw0o83hf:
We just ran into an issue where `StringComparison.InvariantCultureIgnoreCase` was not being applied to queries in our netcore 2.2 app. This should make query case invariance work as expected for netcore apps.

I was unable to divine the intend behind the `#if NET46` directive - I really wanted to just delete it entirely, but I'm sure it's allowing for some other edge case somewhere.

### DIFF
--- a/src/Marten.Testing/Linq/InvariantCultureIgnoreCase_filtering.cs
+++ b/src/Marten.Testing/Linq/InvariantCultureIgnoreCase_filtering.cs
@@ -1,0 +1,28 @@
+using System;
+using System.Linq;
+using Marten.Testing.Documents;
+using Shouldly;
+using Xunit;
+
+namespace Marten.Testing.Linq
+{
+    public class InvariantCultureIgnoreCase_filtering: IntegratedFixture
+    {
+        [Fact]
+        public void can_search_case_insensitive()
+        {
+            var user = new User {UserName = "TEST_USER"};
+
+            using (var session = theStore.OpenSession())
+            {
+                session.Store(user);
+                session.SaveChanges();
+            }
+
+            using (var query = theStore.QuerySession())
+            {
+                query.Query<User>().Single(x => x.UserName.Equals("test_user", StringComparison.InvariantCultureIgnoreCase)).Id.ShouldBe(user.Id);
+            }
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


mysticmind:
@eouw0o83hf Can you also please add a unit test for this?

oskardudycz:
@eouw0o83hf @mysticmind I also don't see the reason why to differentiate that, I'm fine with removing #if condition.

I also second request for adding the unit test(s).

@jeremydmiller do you have more insights?

mysticmind:
> the InvariantCulture comparisons were missing from DotNet Core until 2.0, so it was necessary before netstandard1.3 was dropped as a target.

Above is the comment from @barryhagan in Gitter. 

We can go ahead and remove the `#if` condition and include it by default.

mysticmind:
I shall look at removing the `#if` conditions and also add the unit test to move this forward.

oskardudycz:
@mysticmind we could try to remove that check and see if tests are not broken. If they're fine we could generate prerelease version and send to our users so they could play with that and do verification (if they'd be willing).

mysticmind:
@oskardudycz I removed the condition `#if`'s and tests are all green. Let me look at adding a unit test for this case.

mysticmind:
@eouw0o83hf Can you please share an example use case of how you used StringComparison.InvariantCultureIgnoreCase?

eouw0o83hf:
@mysticmind I'm using the string comparison in queries of projection data, particularly for case-invariant search term matching. For instance,

```
var matches = await session
    .Query<SearchProjection>()
    .Where(a => a.Name.StartsWith("searchTerm", StringComparison.InvariantCultureIgnoreCase))
    .ToListAsync();
```

mysticmind:
@eouw0o83hf thanks for the reply. I am more looking at an example search team and value in docs which you are searching as an example using `StringComparison.InvariantCultureIgnoreCase` comparison.

eouw0o83hf:
@mysticmind any string with case mismatches. I'd recommend storing `"Test"` in the datastore and searching against it with `"test"`, `"TEST"`, and `"TeSt"`

mysticmind:
> 
> 
> [CLA assistant check](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1363)
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1363) before we can accept your contribution.
> 
> ❌ eouw0o83hf [sign now](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1363)
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/JasperFx/marten?pullRequest=1363) it.

@eouw0o83hf Please sign the CLA.

mysticmind:
@jeremydmiller @oskardudycz @jokokko please review the PR.

PR #1368, opened 2019-10-16
Use async versions of dispose and close methods

YohDeadfall:
Npgsql starting version 4.1 supports async versions of `Close` and `Dispose` methods.

### DIFF
--- a/src/Marten/Marten.csproj
+++ b/src/Marten/Marten.csproj
@@ -37,8 +37,8 @@
     </ItemGroup>
     <ItemGroup>
         <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
-        <PackageReference Include="Npgsql" Version="[4.1.1,4.2.0)" />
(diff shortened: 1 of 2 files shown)


YohDeadfall:
Marten is targeting .NET Framework and .NET Standard 2.0, so it isn't able to consume this API and Npgsql unfortunately doesn't provide this methods while for others it does.

@roji Looks like we have missed something.

YohDeadfall:
@jeremydmiller @oskardudycz would you like to have access to async methods in the next patch release if @roji agrees with it?

roji:
@YohDeadfall not exposing async versions for .NET Standard 2.0/.NET Framework was semi-intentional. If people actually request this I guess we could expose them in a patch - nobody's ever asked :)

oskardudycz:
@YohDeadfall thank you for giving the heads up. Currently, we're not supporting (yet) `IAsyncDisposable` anywhere in our codebase. Technically from SemVer perspective adding new feature should be minor, but if it's not breaking anything then I'm fine with exposing that via a patch release.

Issue #1054, opened 2018-07-23
Single HiLo id generator per document store


oakio:

mdissel:
Resulting in unique ID's for each record in every table? If yes, nice enhancement :+1:

jeremydmiller:
@oakio I'd be worried about contention if you do this. Why is this valuable?

oakio:
The main reason is if I want to merge two tables into one, I will not have conflicts by id

jeremydmiller:
@oakio Okay, but this one is "I take pull requests";) Some kind of way to declare that the Hilo sequence is shared?

mysticmind:
> The main reason is if I want to merge two tables into one, I will not have conflicts by id
I would argue that this is *not* a typical use case Marten users would encounter on a regular basis. 

So going ahead and closing this. Feel free to re-open if someone wants to send us a PR :-)

mysticmind:
Follow on note, please have a look at https://dba.stackexchange.com/questions/193788/what-can-go-wrong-using-the-same-sequence-across-multiple-tables-in-postgres. I would veto this feature in Marten.

mdissel:
@mysticmind Would you accept a pull request that changes the current HiLoSequence behaviour to make it possible to override the entitiyname? by default the enitityname equals the doctype, but it can be overridden so you can have a sequence that is used by multiple tables.

mdissel:
@oakio have you already started working on this request?

oakio:
@mdissel no

mdissel:
Still relevant for you?

mdissel:
At first sight it seems that this is necessary:
- extend HiLoSettings with an optional SequenceName property
- use SequenceName if set in this method https://github.com/JasperFx/marten/blob/c8016cd903fc1e349163ed44d19df21cf88fe726/src/Marten/Schema/Identity/Sequences/SequenceFactory.cs#L61
and https://github.com/JasperFx/marten/blob/c8016cd903fc1e349163ed44d19df21cf88fe726/src/Marten/Schema/Identity/Sequences/SequenceFactory.cs#L68

PR #1370, opened 2019-10-16
Optionally global unique ID for multiple document types via HiLoSequence, see #1054

mdissel:


src/Marten/Schema/Identity/Sequences/SequenceFactory.cs
```diff
@@ -55,17 +55,20 @@ public void WritePermissions(DdlRules rules, StringWriter writer)
...
         {
-            return _sequences.GetOrAdd(documentType,
-                type => new HiloSequence(_tenant, _options, documentType.Name, settings));
+            return _sequences.GetOrAdd(GetSequenceName(documentType, settings),
+                sequence => new HiloSequence(_tenant, _options, sequence, settings));
+        }
+
+        private string GetSequenceName(Type documentType, HiloSettings settings)
+        {
+            if (!string.IsNullOrEmpty(settings.SequenceName))
```
oskardudycz: **QUESTION:** Is it doublechecking if the `SequenceName` might be `string.Empty` or is it something that may happen? If it's doublechecking then maybe it would be good to verify also if it's not whitespace?

mysticmind: @oskardudycz Default behavior is that the `SequenceName` is not set and will be `null` as per the initialization. So checking for `!string.IsNullOrEmpty(settings.SequenceName)` is good enough. I agree that we have to validate the sequence name for the below
> Sequence name is a string of 31 or fewer characters[1]. A name must start with a letter or an underscore; the rest of the string can contain letters, digits, and underscores.

mdissel: There's no validation for a default sequencename (equals to the <documentype>.name)

ps. we don't talk about postgresql sequences, but the value of the sequece_name column in the table hilo

mysticmind: Understood, please ignore my comment in the validation part

PR #1371, opened 2019-10-17
Fixes #1369 - Projections inherited from a base class that hides the Id property, projects to an empty Id.

### LINKED ISSUE #1369: Projections inherited from a base class that hides the Id property, projects to an empty Id.
We ran into an issue where our base class contains a property named `Id`. Our projections are inheriting from that base class. When querying the projection the `Id` property is empty.

```
public class Root {
     public Guid Id {get;set;}
}

public class Employee: Root {
     public string Name {get;set;}
     public string EmployeeId {get;set;}
}
```

`var emp = _session.Query<Employee>().FirstOrDefault(e => e.Id = Guid.Parse("df0916ac-c00d-4843-8b58-fbb8325e0aeb"));`

The employee object does come back, but the `Id` is an empty Guid.

oskardudycz:
Updated GetProperties to also return private and protected to allow eg. having a private or protected setter for ID.

Added `NonPublicMembersStorage` to allow `private` and `protected` setters usage during properties (de)serialization and non-default constructor usage (although full support - so removing the limitation of the view having public constructor will be changed in the separate PR).

Added documentation for missing Newtonsoft.NET serializer options.

Fixes #1369

@jediwarpraptor @ddivita fyi


src/Marten/StoreOptions.cs
```diff
@@ -219,9 +219,22 @@ public void Serializer(ISerializer serializer)
         /// <param name="enumStyle"></param>
         /// <param name="casing">Casing style to be used in serialization</param>
         /// <param name="collectionStorage">Allow to set collection storage as raw arrays (without explicit types)</param>
-        public void UseDefaultSerialization(EnumStorage enumStyle = EnumStorage.AsInteger, Casing casing = Casing.Default, CollectionStorage collectionStorage = CollectionStorage.Default)
+        /// <param name="nonPublicMembersStorage">Allow to use non public members should be used during deserialization</param>
```
mysticmind: Change to "Allow non public members to be used during deserialization"

src/Marten/ISerializer.cs
```diff
@@ -75,4 +86,12 @@ public enum CollectionStorage
         Default,
         AsArray
     }
-}
\ No newline at end of file
+
+    [Flags]
+    public enum NonPublicMembersStorage
```
mysticmind: Add XML documentation for each option.

src/Marten/StoreOptions.cs
oskardudycz: Fixed 👍

oskardudycz:
@mysticmind thank you for the review. I'll update the docs and ping you when they're ready.

oskardudycz:
@mysticmind documentation updated in https://github.com/JasperFx/marten/pull/1371/commits/060a09945248b07033cd847f7e33bc9a37f80a9a

oskardudycz:
Thank you @mysticmind - updated 👍

PR #1375, opened 2019-10-29
QuerySession implements finalizer, but Dispose does not call into `GC…

jokokko:
….SuppressFinalize` -> our connections needlessly always end up in the finalizer queue.

Fixed by calling into `GC.SuppressFinalize`. Now, do we want the finalizer in the first place? By quick look, Npgsql has no finalizers (though
they call into `GC.SuppressFinalize` in a few ctors where they derive from `Component` or other types with finalizers). Noticed by incident when going over a mem dump...


oskardudycz:
Seems legit 👍

Issue #1378, opened 2019-11-04
Calls to mt_immutable_timestampt/mt_immutable_timestamptz fail with duplicated fields


cocowalla:
Similar to #837, after upgrading Marten I'm getting the following error message when updating a type with a duplicated column:

`schema "public" does not exist`

I can see a statement being executed like:

`update myschema.mt_doc_mytype as d set message_timestamp = public.mt_immutable_timestamptz(data -> 'Message' ->> 'Timestamp') where (d.tenant_id = :tenantid and d.id = :arg9)`

^ Note the `public.mt_immutable_timestamptz` - also note that when creating my document store I am setting `DatabaseSchemaName = myschema`. 

The problem seems to be here in [DuplicatedField](https://github.com/JasperFx/marten/blob/0839ec6d364cdfd07ead7ccba15785ae29e72cb8/src/Marten/Schema/DuplicatedField.cs#L101) - a _new_, default `StoreOptions` is being created, instead of using the configured one.

cocowalla:
Will send a PR for this shortly

stvoidmain:
Nice to see this worked out!
Letme know if you need help to try something.
(sorry for not camoing up before, was on a very tight spot with work)

PR #1379, opened 2019-11-04
fix: #1378, use configured StoreOptions

cocowalla:


src/Marten.Testing/ConnectionSource.cs
```diff
@@ -4,8 +4,9 @@ namespace Marten.Testing
 {
     public class ConnectionSource : ConnectionFactory
     {
-        public static readonly string ConnectionString = Environment.GetEnvironmentVariable("marten_testing_database")
-            ?? "Host=localhost;Port=5432;Database=postgres;Username=postgres;password=postgres";
+        //public static readonly string ConnectionString = Environment.GetEnvironmentVariable("marten_testing_database")
```
oskardudycz: @cocowalla please revert this leftover

src/Marten/Schema/DuplicatedField.cs
```diff
@@ -14,18 +14,20 @@ namespace Marten.Schema
     public class DuplicatedField: Field, IField
     {
         private readonly Func<Expression, object> _parseObject = expression => expression.Value();
+        private readonly StoreOptions _storeOptions;
         private readonly bool useTimestampWithoutTimeZoneForDateTime;
         private string _columnName;
 
-        public DuplicatedField(EnumStorage enumStorage, MemberInfo[] memberPath, bool useTimestampWithoutTimeZoneForDateTime = true, bool notNull = false)
-            : base(enumStorage, memberPath, notNull)
+        public DuplicatedField(StoreOptions storeOptions, MemberInfo[] memberPath, bool useTimestampWithoutTimeZoneForDateTime = true, bool notNull = false)
```
oskardudycz: This is breaking change. `DuplicatedField` in theory should be internal detail, but as this is a popular feature then IMHO it might break our users' code.
I'd suggest adding a new constructor with store options parameter and keep this one calling newly added with the current usage.

src/Marten.Testing/ConnectionSource.cs
cocowalla: Hah, I did the same thing last time I sent a Marten PR (too lazy to restart my IDE after settling an environment variable), and this time I told myself I'd definitely change it back before pushing the code!

src/Marten/Schema/DuplicatedField.cs
cocowalla: I thought about this at the time, but figured I'd just push the code and see what you thought; I'll go with my gut next time ;)

oskardudycz: Normally it would be a good idea to not have redundant code, but we need to be careful for the breaking changes. In v4 we should definitely make internal as much as we can, then such refactoring would be much easier.

cocowalla:
@oskardudycz changes made as per your review

src/Marten.Testing/ConnectionSource.cs
```diff
@@ -5,7 +5,7 @@ namespace Marten.Testing
     public class ConnectionSource : ConnectionFactory
     {
         public static readonly string ConnectionString = Environment.GetEnvironmentVariable("marten_testing_database")
-            ?? "Host=localhost;Port=5432;Database=marten;Username=postgres;password=postgres";
+            ?? "Host=localhost;Port=5432;Database=postgres;Username=postgres;password=postgres";
```
cocowalla: Gah, sorry!

oskardudycz: That was also happening to me quite often until I set up the env variable ;)

oskardudycz:
@cocowalla I've let myself to add one more commit https://github.com/JasperFx/marten/pull/1379/commits/b4150886a69b6295e614ceeef432f08fe0dc3824 - it was faster than explaining via PR comment - I hope that you don't mind 😄

cocowalla:
@oskardudycz of course not - thanks for reviewing so quickly after I submitted this PR!

src/Marten.Testing/Schema/DuplicatedFieldTests.cs
```diff
@@ -135,6 +142,32 @@ public void pg_type_is_used_for_string_list(string pgType, string expectedUpdate
             field.DbType.ShouldBe(NpgsqlDbType.Array | NpgsqlDbType.Text);
         }
 
+        [Theory]
+        [InlineData(null, "date = public.mt_immutable_timestamp(data ->> 'Date')")]
+        [InlineData("myergen", "date = myergen.mt_immutable_timestamp(data ->> 'Date')")]
```
mysticmind: `myergen` looks like a very cryptic name for schema. Suggest to change it.

cocowalla: Haha, it means "meh", and it's just something I always use without thinking - I just checked and it seems I've used it a [couple](https://github.com/JasperFx/marten/blob/0ccc54748849236c49ccb38fa3f1984b4a2f6efe/src/Marten.Testing/Bugs/Bug_980_bulk_insert_with_subclass.cs#L24) of [times ](https://github.com/JasperFx/marten/blob/0ccc54748849236c49ccb38fa3f1984b4a2f6efe/src/Marten.Testing/Bugs/Bug_899_operations_out_of_order_when_types_use_inheritance.cs#L24) before in previous PRs!

I don't mind changing it if you want though?

cocowalla:
> what if a user wants to specifically override the StoreOptions EnumStorage for a specific duplicate field or to set a varying EnumStorage for a specific duplicate field? Thoughts?

It would be a bit odd to use different enum storage styles in the same database, but I suppose it's conceivable. I guess there are 3 approaches you could use:

1. Use the [obsoleted overload](https://github.com/JasperFx/marten/pull/1379/files#diff-e8e0cf443036c7fa3e51325b1310fc96R21)
2. Use the new ctor, and pass in a customised `StoreOptions` for each field
3. Add yet another overload that takes both a `StoreOptions` and an `EnumStorage` that overrides that of the `StoreOptions`

src/Marten.Testing/Schema/DuplicatedFieldTests.cs
mysticmind: Alright, we can leave it "as is".

mysticmind:
we could possibly use the obsoleted overload methods, we will review this again during the 4.0 work. Going ahead and merging this.

oskardudycz:
Personally I'd prefer to remove all of the properties and methods that are marked as obsolete - eg. `DuplicatedFieldEnumStorage` and `DuplicatedFieldUseTimestampWithoutTimeZoneForDateTime`. My intention for introducing them was to fix the compatibility bugs without introducing breaking changes.

I don't see much value of having a separate `EnumStorage` config for duplicated fields in general and especially for the single one. Personally I'd prefer to keep the single settings if the user won't demand to have it separately. We can also ask users about their thoughts.

PR #1390, opened 2019-12-01
Fixes #1386 - not being able to serialize IReadonlyCollection<int> while using CollectionStorage.AsArray

### LINKED ISSUE #1386: Question regarding use of CollectionStorage.AsArray
Hi!

I'm trying to get the Serializer option "CollectionStorage" to work, but when I set it to "AsArray", I only get null in the database. The current implementation is a IReadOnlyCollection&lt;int> wrapping a List&lt;int>, but I have tried with several others and none of them work:

- IReadOnlyCollection&lt;int> wrapping List&lt;int>
- IEnumerable&lt;int> wrapping List&lt;int>
- IEnumerable&lt;int> wrapping int[]

```
public class Foo
{
	public IReadOnlyCollection<int> Categories { get; private set; }

	public Foo(IEnumerable<int> categories)
	{
		Categories = categories.ToList();
	}
}

```

```
var documentStore = DocumentStore.For(_ =>
{
	_.Serializer(new JsonNetSerializer
	{
		CollectionStorage = CollectionStorage.AsArray
	});
	_.Schema.For<Foo>();
        ...
}
```

It only works if the property type is an array, but I don't want the property to be changeable from outside the object (DDD). I.e. this works:
```
public class Foo
{
	public int[] Categories { get; private set; }

	public Foo(IEnumerable<int> categories)
	{
		Categories = categories.ToArray();
	}
}

```

The reason for me wanting to set CollectionStorage.AsArray is that collections get rather unwieldy in the db otherwise:
```
"Categories": {
     "$type": "System.Collections.Generic.List`1[[System.Int32, System.Private.CoreLib]], System.Private.CoreLib",
     "$values": [
         1,
         3
     ]
}
```

I'm using Marten 3.9.0, PostgreSQL for Windows v12, .Net Core 3.0.

oskardudycz:
As in the description - see more details in #1386 

Besides providing the fix and adding test for the described case, I updated inner collection queries tests to be sure that this case won't be missed again.

### DIFF
--- a/src/Marten.Testing/Documents/User.cs
+++ b/src/Marten.Testing/Documents/User.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using Baseline;
 
 namespace Marten.Testing.Documents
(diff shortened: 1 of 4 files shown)


jeremydmiller:
So first off, big kudos for the test coverage. Other than that, as long as the tests all pass, I can't see any issue with the change.

Issue #1385, opened 2019-11-12
Patch API doesn't respect property names casing


nikodemrafalski:
Marten 3.10.0

Given  DocumentStore configured for `camelCase` property naming,

The change
```session.Patch<Doc>(where).Set(x => x.Dog.Name, "Milo");```
renders invalid sql command  paremeter.
```arg4: {"type":"set","value":null,"path":"Dog.Name"}```
instead of 
```arg4: {"type":"set","value":null,"path":"dog.name"}```

preventing the patching from happening

PR #1405, opened 2019-12-05
Fix patch expression to use right property name casing

mysticmind:
- Fix patch expression to use right property name casing based on serializer settings
- Add unit test

fixes GH-1385

### DIFF
--- a/src/Marten.Testing/Patching/PatchExpressionTests.cs
+++ b/src/Marten.Testing/Patching/PatchExpressionTests.cs
@@ -3,6 +3,7 @@ using Marten.Patching;
 using Marten.Schema;
 using Marten.Services;
 using Marten.Storage;
+using Marten.Testing.Documents;
 using NSubstitute;
 using Shouldly;
 using Xunit;
@@ -13,7 +14,7 @@ namespace Marten.Testing.Patching
     {
         private readonly PatchExpression<Target> _expression;
         private readonly ITenant _schema = Substitute.For<ITenant>();
-        
+
 
         public PatchExpressionTests()
         {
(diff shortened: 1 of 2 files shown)


src/Marten.Testing/Patching/PatchExpressionTests.cs
```diff
@@ -400,5 +401,22 @@ public void duplicate_property_no_target()
...
+
+        [Fact]
+        public void check_camel_case_serialized_property()
+        {
+            var store = TestingDocumentStore.For(_ =>
+            {
+                _.UseDefaultSerialization(casing: Casing.CamelCase);
+            });
+
+            var expr1 = new PatchExpression<Target>(null, _schema, new UnitOfWork(store, store.Tenancy.Default), store.Serializer);
```
oskardudycz: **SUGGESTION:** Could we make variables naming more descriptive? eg.
```suggestion
            var enumExpression = new PatchExpression<Target>(null, _schema, new UnitOfWork(store, store.Tenancy.Default), store.Serializer);
```

src/Marten.Testing/Patching/PatchExpressionTests.cs
```diff
@@ -400,5 +401,22 @@ public void duplicate_property_no_target()
...
+            expr1.Set(x => x.Color, Colors.Blue);
+            expr1.Patch["path"].ShouldBe("color");
+
+            var expr2 = new PatchExpression<Target>(null, _schema, new UnitOfWork(store, store.Tenancy.Default), store.Serializer);
```
oskardudycz:
```suggestion
            var stringExpression = new PatchExpression<Target>(null, _schema, new UnitOfWork(store, store.Tenancy.Default), store.Serializer);
```

PR #1408, opened 2019-12-16
Another `IQuerySession.QueryAsync` method

i00lii:
Hi. 

I'm trying to extend my awesome `DocumentStore` with custom db function based on `UpsertFunction` logic. I've found how you use `SprocCall` to fill `NpgsqlCommand`'s argument list during Insert/Update/... documents. Looks pretty reusable... 

But for now there is no way to invoke custom `IQueryHandler<T>` from the user code (https://github.com/JasperFx/marten/blob/master/src/Marten/QuerySession.cs#L78).

So it might be extreamly usefult to extend `IQuerySession.QueryAsync` api which acceps 'IQueryHandler<IReadOnlyList<T>>' as paramters. No sooner said than done ^____^

### DIFF
--- a/src/Marten/IQuerySession.cs
+++ b/src/Marten/IQuerySession.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Marten.Linq;
+using Marten.Linq.QueryHandlers;
 using Marten.Schema;
 using Marten.Services.BatchQuerying;
 using Marten.Storage;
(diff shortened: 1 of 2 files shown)


i00lii:
Also, is there is a way to extract this (https://github.com/JasperFx/marten/blob/master/src/Marten/Schema/Resolver.cs#L81)
code somewhere into public place?

mysticmind:
The whole code base has evolved a lot with v4 with big bang changes. Please raise a separate issue on your requirement for us to eyeball and look at what we can do. Closing this PR.

PR #1410, opened 2019-12-18
Make DeleteAllEventData safe for empty database.

tetious:
`truncate` fails when the tables don't exist. This tweak prevents the problem, similar to how it is prevented elsewhere in DocumentCleaner.

### DIFF
--- a/src/Marten/Schema/DocumentCleaner.cs
+++ b/src/Marten/Schema/DocumentCleaner.cs
@@ -110,8 +110,14 @@ WHERE  s.sequence_name like 'mt_%' and s.sequence_schema = ANY(?);";
         {
             using (var connection = _tenant.OpenConnection(CommandRunnerMode.Transactional))
             {
-                connection.Execute($"drop table if exists {_options.Events.DatabaseSchemaName}.mt_events cascade;" +
-                                   $"drop table if exists {_options.Events.DatabaseSchemaName}.mt_streams cascade");
+                connection.Execute("DO $$ BEGIN " +
+                                        "IF EXISTS(SELECT * FROM information_schema.tables " +
+                                        $"WHERE table_name = 'mt_events' AND table_schema = '{_options.Events.DatabaseSchemaName}') " +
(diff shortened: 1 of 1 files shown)


src/Marten/Schema/DocumentCleaner.cs
```diff
@@ -110,8 +110,8 @@ public void DeleteAllEventData()
         {
             using (var connection = _tenant.OpenConnection(CommandRunnerMode.Transactional))
             {
-                connection.Execute($"truncate table {_options.Events.DatabaseSchemaName}.mt_events cascade;" +
-                                   $"truncate table {_options.Events.DatabaseSchemaName}.mt_streams cascade");
+                connection.Execute($"drop table if exists {_options.Events.DatabaseSchemaName}.mt_events cascade;" +
```
mysticmind: If you update the code as below, it will yield the same result and the unit tests will pass as well. Can you please take a look at the below:
```
connection.Execute("DO $$ BEGIN " +
                           "IF EXISTS(SELECT * FROM information_schema.tables " +
                           $"WHERE table_name = 'mt_events' AND table_schema = '{_options.Events.DatabaseSchemaName}') " +
                           $"THEN TRUNCATE TABLE {_options.Events.DatabaseSchemaName}.mt_events CASCADE; END IF;" +
                           "IF EXISTS(SELECT * FROM information_schema.tables " +
                           $"WHERE table_name = 'mt_streams' AND table_schema = '{_options.Events.DatabaseSchemaName}') " +
                           $"THEN TRUNCATE TABLE {_options.Events.DatabaseSchemaName}.mt_streams CASCADE; END IF; " +
                           "END; $$;");
```

mysticmind:
@tetious Thanks for the PR. Since the unit tests are failing, I have proposed a slightly varying method in the review comments, please check the same.

src/Marten/Schema/DocumentCleaner.cs
tetious: This is perfect. Thanks! Updated!

PR #1412, opened 2020-01-15
Added trim to TableColumn Name and Type

malscent:
Trim column name and type to get rid of any extraneous white space.  It can cause marten to fail to derive a schema patch.


mysticmind:
@oskardudycz All tests look to be passed so I am not sure what you are indicating as failing tests.

oskardudycz:
@mysticmind I meant that we were this bug not covered in our tests (are there were passing) so it would be nice to add test for failing scenario that @malscent is fixing so we won't have regression in the future.

malscent:
I'll see if i can get you some unit tests in the next day or so

malscent:
Added unit tests and also found a case where a space IN the table name caused an exception as well.  Replaced spaces with underscores.

src/Marten.Testing/Bugs/Bug_PR_1412_spaces_in_table_name_cause_schema_patch_to_fail.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            StoreOptions(_ =>
+            {
+                _.AutoCreateSchemaObjects = AutoCreate.CreateOrUpdate;
+                _.Storage.Add(new spaceAfterTableNameSchema());
+            });
+            try
+            {
+                theStore.Schema.ApplyAllConfiguredChangesToDatabase();
+                theStore.Schema.ApplyAllConfiguredChangesToDatabase();
```
oskardudycz: **CONCERN:** @malscent why is it needed to apply changes to database twice? Same concern applies for the tests below.

src/Marten.Testing/Bugs/Bug_PR_1412_spaces_in_table_name_cause_schema_patch_to_fail.cs
```diff
@@ -0,0 +1,143 @@
...
+            }
+            catch (Exception e)
```
oskardudycz: **FIX:** @malscent coud you use `Should.NotThrow` (see [sample](https://github.com/JasperFx/marten/blob/1128d6f6a63188087dd34645f735b015f76bf4c7/src/Marten.Testing/StoreOptionsTests.cs#L114)). Instead of the catch? Same apply for other places.

src/Marten.Testing/Bugs/Bug_PR_1412_spaces_in_table_name_cause_schema_patch_to_fail.cs
malscent: Because the error here is that the first application of the changes is s uccessful and Postgres truncates the the whitespace.  The second apply is the one that throws, due to Marten not being able to generate the schema diff.

src/Marten.Testing/Bugs/Bug_PR_1412_spaces_in_table_name_cause_schema_patch_to_fail.cs
malscent: @oskardudycz I could not find an example in your link.  Did you send the correct one?

src/Marten.Testing/Bugs/Bug_PR_1412_spaces_in_table_name_cause_schema_patch_to_fail.cs
oskardudycz: Thank you for an explanation. Maybe it would be worth to add a comment with information or update the test name to mention this second call?

src/Marten.Testing/Bugs/Bug_PR_1412_spaces_in_table_name_cause_schema_patch_to_fail.cs
oskardudycz: It should lead you to:
obraz

src/Marten.Testing/Bugs/Bug_PR_1412_spaces_in_table_name_cause_schema_patch_to_fail.cs
mysticmind: I was also wondering about the a call twice, thanks for the clarification. A comment will add clarity.

src/Marten.Testing/Bugs/Bug_PR_1412_spaces_in_table_name_cause_schema_patch_to_fail.cs
malscent: Haha it did.. i'm just blind.  I'll get that corrected ASAP

src/Marten.Testing/Bugs/Bug_PR_1412_spaces_in_table_name_cause_schema_patch_to_fail.cs
mysticmind: @malscent I think the second call can be changed to `theStore.Schema.AssertDatabaseMatchesConfiguration()`. This results in the same check which is equivalent and generates the schema diff. This way an explicit comment/explanation is not required. Thoughts?

malscent: > I added a comment to use `AssertDatabaseMatchesConfiguration()` for the second call, rest all looks fine.

Yup.. Adding now

mysticmind:
@malscent I have merged the PR, will take care to update the unit tests to use `AssertDatabaseMatchesConfiguration()`

PR #1419, opened 2020-01-30
Bug#1404 fixed Hilo concurrency error

### LINKED ISSUE #1404: Concurrent update problem when updating mt_hilo
Hi, I started to have a lot of these:
```
[106350] ERROR:  could not serialize access due to concurrent update
[106350] CONTEXT:  SQL statement "update documentos.mt_hilo set hi_value = next_value where entity_name = entity"
[106350] STATEMENT:  SELECT * FROM documentos.mt_get_next_hi("entity" := $1)
```
Which I didn't have before, is there anything that changed on that handling that can make this error arise?
As i see it, this shouldn't happen at all, and its making application to throw exception and abort current operation, any ideas?

Miggleness:
Fixes concurrency error with HiloSequence generator in high concurrency case.

1. Added test that consistently reproduces the problem

2. Updated `get_next_hi` sproc update condition to include check that the current hilo value hasn't changed since the last select. If it did, then return -1

3. Updated transaction isolation to read committed.  We do not need to run in serialization isolation now because of #2. 

4. Removed `try..catch` block surrounding sproc call. It's put in place before to address exceptions regarding concurrency issues in serialization isolation mode, hence no longer necessary.

Things for reviewer to look at
1. I'm not sure about is whether those who already have Marten running live would get the updated sproc (e.g. AutoCreate option is set to none).  How do we force sproc update?

2. It might be a better idea to retry getting next hi within the sproc.  Depends on maintainers' stance on this and sproc updates to deployments, we can implement it in this way instead

2. Throwing type `Exception`.  Should I create a new exception type for HiloSequence?

### DIFF
--- a/src/Marten/Schema/Identity/Sequences/HiLoSequence.cs
+++ b/src/Marten/Schema/Identity/Sequences/HiLoSequence.cs
@@ -94,12 +94,12 @@ namespace Marten.Schema.Identity.Sequences
 
                         CurrentHi = Convert.ToInt64(raw);
 
-                    } while (CurrentHi < 0 && attempts < 30);
+                    } while (CurrentHi < 0 && attempts < _options.HiloSequenceDefaults.MaxAdvanceToNextHiAttempts);
 
                     // if CurrentHi is still less than 0 at this point, then throw exception
                     if (CurrentHi < 0)
                     {
-                        throw new Exception("Unable to advance hi sequence");
+                        throw new HiloSequenceAdvanceToNextHiAttemptsExceededException();
                     }
                 }
                 finally
--- a/src/Marten/Schema/Identity/Sequences/HiLoSequence.cs
+++ b/src/Marten/Schema/Identity/Sequences/HiLoSequence.cs
@@ -65,17 +65,21 @@ namespace Marten.Schema.Identity.Sequences
         {
             lock (_lock)
             {
-                if (ShouldAdvanceHi())
+                var retryCount = 0;
+                while(ShouldAdvanceHi())
                 {
                     try
                     {
                         AdvanceToNextHi();
                     }
-                    catch (Exception)
+                    catch (Exception ex)
                     {
-                        // Retry once.
                         Thread.Sleep(50);
-                        AdvanceToNextHi();
+                        retryCount++;
+                        if(retryCount> 3)
+                        {
+                            throw ex;
+                        }
                     }
                 }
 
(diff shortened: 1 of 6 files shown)


Miggleness:
I realize the retry increment shouldnt be in the exception block. Will fix later

src/Marten/Schema/Identity/Sequences/HiLoSequence.cs
```diff
@@ -91,14 +81,26 @@ public void AdvanceToNextHi()
...
+                        // Sproc is expected to return -1 if it's unable to
+                        // atomically secure the next hi
+                        var raw = conn.CreateCommand().CallsSproc(GetNextFunction)
+                            .With("entity", _entityName)
+                            .Returns("next", NpgsqlDbType.Bigint).ExecuteScalar();
+
+                        CurrentHi = Convert.ToInt64(raw);
 
-                    tx.Commit();
+                    } while (CurrentHi < 0 && attempts < 30);
```
mysticmind: I think it would help to pass the `attempts` value as in `StoreOptions` with a default value of `30` and use it so that end users can configure it as they see it fit.

Also I think adding some sleep time between attempts would help.

src/Marten/Schema/Identity/Sequences/HiLoSequence.cs
```diff
@@ -91,14 +81,26 @@ public void AdvanceToNextHi()
...
 
-                    CurrentHi = Convert.ToInt64(raw);
+                    // if CurrentHi is still less than 0 at this point, then throw exception
+                    if (CurrentHi < 0)
+                    {
+                        throw new Exception("Unable to advance hi sequence");
```
mysticmind: As you indicated, a specific exception for this case will be better rather than generic exception.

Miggleness: Added exception `HiloSequenceAdvanceToNextHiAttemptsExceededException`

src/Marten/Schema/Identity/Sequences/HiLoSequence.cs
Miggleness: Added new property under `StoreOptions.HiloSequenceDefaults` named `MaxAdvanceToNextHiAttempts`.

No point in adding sleep time. You would want to retry as soon as possible.

Miggleness:
@mysticmind thanks. added commit

mysticmind:
@jeremydmiller @oskardudycz @jokokko could you please review this PR?

jeremydmiller:
@mysticmind @Miggleness I think this looks good to me. I like the optimistic concurrency check. Good call.

oskardudycz:
Looks legitimate to me. 

@Miggleness: Regarding the question:

> I'm not sure about is whether those who already have Marten running live would get the updated sproc (e.g. AutoCreate option is set to none). How do we force sproc update?

If the AutoCreate none is set then user should manage the updates manually and prepare the sql script. Currently, it's possible to do it with `ToPatch` method (see https://github.com/JasperFx/marten/blob/fcdecf7aadd889ed2e55cb421b166da8cda79295/src/Marten/Schema/IDocumentSchema.cs#L44)

PR #1420, opened 2020-01-31
Fix ComputedIndex DDL generation when non-default casing is present

lahma:
While trying to enforce a business rule of uniqueness for certain set of columns regardless the content casing I ran into a DDL generation bug. The code was generating something like 

```
CREATE UNIQUE INDEX mt_doc_measurand_uidx_departmentname ON public.mt_doc_measurand (lower(  (data ->> 'Department'), (data ->> 'Name')));
```
The problem was that the column part was determined in the ComputedIndex constructor where there is no way to know any custom casing. The generated DDL had thus lower call for all fields where it should have had lower/upper call for each individual column.

Fixed this by moving column generation as part of DDL generation where casing could be used with desired value.

### DIFF
--- a/src/Marten.Testing/Acceptance/computed_indexes.cs
+++ b/src/Marten.Testing/Acceptance/computed_indexes.cs
@@ -181,6 +181,32 @@ namespace Marten.Testing.Acceptance
         }
 
         [Fact]
+        public void create_multi_property_string_index_with_casing()
+        {
+            StoreOptions(_ =>
+            {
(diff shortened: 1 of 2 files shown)


lahma:
I'm unsure if the CI errors are valid or not, is there something to fix here?

mysticmind:
@lahma CI build failure looks like not related to your fixes, restarted the failed matrix build pertaining to  PG11.

lahma:
Thank you and you're welcome! 🙂

PR #1427, opened 2020-02-08
Add support for In & NotIn Linq operators for queries.

apereiratl:
Added to provide compatibility for users coming from RavenDb to ease the transition.


oskardudycz:
As @jeremydmiller agreed on Gitter and it's a small change then I'm merging.

PR #1428, opened 2020-02-09
Add overrides for LoadMany to support List<T> as opposed to T[].

### LINKED ISSUE #1484: Allow passing `IEnumerable<T>` to `IDocumentSession.Store()`
This applies to the both the untennanted and tennanted methods.

Currently they only accept `params T[]`, there should be versions which take `IEnumerable<T>`

apereiratl:
Fixes #1484

### DIFF
--- a/src/Marten/QuerySession.cs
+++ b/src/Marten/QuerySession.cs
@@ -147,7 +147,17 @@ namespace Marten
             return LoadMany<T>().ById(ids);
         }
 
+        public IReadOnlyList<T> LoadMany<T>(IEnumerable<string> ids)
+        {
+            return LoadMany<T>().ById(ids);
+        }
+
         public IReadOnlyList<T> LoadMany<T>(params Guid[] ids)
+        {
+            return LoadMany<T>().ById(ids);
+        }
+
+        public IReadOnlyList<T> LoadMany<T>(IEnumerable<Guid> ids)
         {
             return LoadMany<T>().ById(ids);
         }
@@ -157,7 +167,17 @@ namespace Marten
             return LoadMany<T>().ById(ids);
         }
 
+        public IReadOnlyList<T> LoadMany<T>(IEnumerable<int> ids)
+        {
+            return LoadMany<T>().ById(ids);
+        }
+
         public IReadOnlyList<T> LoadMany<T>(params long[] ids)
+        {
+            return LoadMany<T>().ById(ids);
+        }
+
+        public IReadOnlyList<T> LoadMany<T>(IEnumerable<long> ids)
         {
             return LoadMany<T>().ById(ids);
         }
@@ -167,7 +187,17 @@ namespace Marten
             return LoadMany<T>().ByIdAsync(ids);
         }
 
+        public Task<IReadOnlyList<T>> LoadManyAsync<T>(IEnumerable<string> ids)
+        {
+            return LoadMany<T>().ByIdAsync(ids);
+        }
+
         public Task<IReadOnlyList<T>> LoadManyAsync<T>(params Guid[] ids)
+        {
+            return LoadMany<T>().ByIdAsync(ids);
+        }
+
+        public Task<IReadOnlyList<T>> LoadManyAsync<T>(IEnumerable<Guid> ids)
         {
             return LoadMany<T>().ByIdAsync(ids);
         }
@@ -177,7 +207,17 @@ namespace Marten
             return LoadMany<T>().ByIdAsync(ids);
         }
 
+        public Task<IReadOnlyList<T>> LoadManyAsync<T>(IEnumerable<int> ids)
+        {
+            return LoadMany<T>().ByIdAsync(ids);
+        }
+
         public Task<IReadOnlyList<T>> LoadManyAsync<T>(params long[] ids)
+        {
+            return LoadMany<T>().ByIdAsync(ids);
+        }
+
+        public Task<IReadOnlyList<T>> LoadManyAsync<T>(IEnumerable<long> ids)
         {
             return LoadMany<T>().ByIdAsync(ids);
         }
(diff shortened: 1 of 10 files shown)


jeremydmiller:
I vote extension methods as well. I'm going to say again that I'm really not enthusiastic about these going into our core.

oskardudycz:
@jeremydmiller I agree, that this is something that probably our user could write on their own (eg. by adding the extensions methods on their codebase), but if that helps the migration for the RavenDB users (that I assume were our initial userbase) then I think that it's fine. I think that if it's in the different namespace than the original interface then it shouldn't be a huge maintenance burden.

mysticmind:
I echo the same thought as @jeremydmiller  and @oskardudycz . We could possibly look at adding a docs section to outline these so that users porting code base from RavenDB to Marten.

@jeremydmiller @oskardudycz should we close this PR or wait to hear from @apereiratl?

oskardudycz:
I'd propose to wait a few days and see if @apereiratl still would like to come up with the extension methods solution.

apereiratl:
Sorry I've been a bit busy but will convert them to extension methods and resend the patch.

oskardudycz:
@apereiratl there is no rush on that. We just wanted to know if we may expect the updates in the nearest future 👍

mysticmind:
@apereiratl we are planning to release v3.12.0 on 8th, is it something which you could look to squeeze in?

oskardudycz:
@jeremydmiller @mysticmind I checked and unfortunately, it won't work with extension methods. Params methods will still take precedence and IEnumerable or IList will be used.

I think that it's fine to add such methods, as it's a common error of wrongly passing IEnumerable - especially in Append events function.

Same case for https://github.com/JasperFx/marten/pull/1485 provided by @trullock.

After consideration, I'm fine with that (however our interface will be even bigger). Probably suggestion made by @trullock to make the IEnumerable default method and params extension method would also work, but such a switch can be made in v4.

I can add necessary changes.

Thoughts?

jeremydmiller:
I don't think this adds a huge amount of value to worry about what's an extension method and what's not. I say just add the overloads to `LoadMany()` and call it a day. 

I do want to reevaluate `IDocumentSession` in v4, but there I'm more worried about moving the plain text searching out of the root interface.

oskardudycz:
I was also referring to other places where we have params [] eg. Store, Append events etc. 
I think that having a smaller interface and implementation and overloads as Extension methods might make it easier to maintain and less prone for breaking changes.

So if there are no votes against I’ll da overloads and we’ll evaluate removing some of them in V4

jeremydmiller:
@oskardudycz Gotcha, so just that. It's a breaking binary change, so you have to leave the changes to thin it down for v4

oskardudycz:
@jeremydmiller sounds good to nr 👍

oskardudycz:
@jeremydmiller @mysticmind @jokokko I cherrypicked @trullock changes, aligned both implementations and added other overloads eg. to Insert, Update and added for EventStore.

Could you verify if you're fine with that?

src/Marten/QuerySession.cs
```diff
@@ -119,13 +119,10 @@ private T load<T>(object id)
...
             var mapping = Tenant.MappingFor(typeof(T));
-            if (id.GetType() != mapping.IdType)
+            if (id.GetType() != mapping.IdType && (id.GetType() != typeof(int) || mapping.IdType != typeof(long)))
             {
-                if (id.GetType() == typeof(int) && mapping.IdType == typeof(long))
-                    return;
-
                 throw new InvalidOperationException(
-                    $"The id type for {typeof(T).FullName} is {mapping.IdType.Name}, but got {id.GetType().Name}");
+                    "The id type for {typeof(T).FullName} is {mapping.IdType.Name}, but got {id.GetType().Name}");
```
mysticmind, line 125: Why was the `$` removed, isn't it not required for string interpolation?

src/Marten/QuerySession.cs
```diff
@@ -119,13 +119,10 @@ private T load<T>(object id)
         private void assertCorrectIdType<T>(object id)
         {
             var mapping = Tenant.MappingFor(typeof(T));
-            if (id.GetType() != mapping.IdType)
+            if (id.GetType() != mapping.IdType && (id.GetType() != typeof(int) || mapping.IdType != typeof(long)))
```
mysticmind, line 122: Adding this ` && (id.GetType() != typeof(int) || mapping.IdType != typeof(long)` in the parent `if` block looks to be changing the behaviour of earlier implementation. As per my understanding of earlier logic, if `id.GetType()` and `mapping.IdType` did not match, the subsequent check inside `if (id.GetType() == typeof(int) && mapping.IdType == typeof(long))` means that `int` and `long` are still compatible hence exception is not required to be thrown.

Can you please let me know your thoughts on the above?

src/Marten/QuerySession.cs
oskardudycz: You have good eyes! Nice catch. Fixed.

src/Marten/QuerySession.cs
oskardudycz: @mysticmind I just simplified the previous condition.

So I at first inverted the condition from line https://github.com/JasperFx/marten/pull/1428/files#diff-2f5f5939e30dd61deec7ee3492372bc9L124.

Then it was:
```
if (id.GetType() != mapping.IdType)
{
    if (id.GetType() != typeof(int) || mapping.IdType != typeof(long))
        throw new InvalidOperationException(
            $"The id type for {typeof(T).FullName} is {mapping.IdType.Name}, but got {id.GetType().Name}");
}
```
having that I joined two conditions to the final form

mysticmind: The code in master is as below:
https://github.com/JasperFx/marten/blob/master/src/Marten/QuerySession.cs#L119|
```
        private void assertCorrectIdType<T>(object id)
        {
            var mapping = Tenant.MappingFor(typeof(T));
            if (id.GetType() != mapping.IdType)
            {
                if (id.GetType() == typeof(int) && mapping.IdType == typeof(long))
                    return;

                throw new InvalidOperationException(
                    $"The id type for {typeof(T).FullName} is {mapping.IdType.Name}, but got {id.GetType().Name}");
            }
        }
```

oskardudycz: Reverted the change - I agree my version was too confusing,

mysticmind: @oskardudycz  with ref to our DM. I was bit sleep deprived. The changes result in equivalent logic but the earlier logic was bit explicit on why we are running those checks. Thanks for reverting it.

PR #1456, opened 2020-04-19
Add possibility to start typed event streams with non-generic StartStream

scrodde:
I have a use case where I'm starting new event streams without knowing the type at compile time. It's essential for me that the stream type info gets persisted. This change declares two new StartStream methods on the IEventStore interface which allows you to pass the the aggregate type as a parameter.


scrodde:
@oskardudycz will do!

oskardudycz:
I added an additional overload to keep consistency and updated documentation.

oskardudycz:
@scrodde congratulations. You just become 103 Marten contributor 😃

scrodde:
haha, thx @oskardudycz ;)

PR #1463, opened 2020-05-05
Test Performance Optimization

jeremydmiller:
* Introduced some new testing base classes to centralize integration testing
* Trying to reuse a `DocumentStore` over most of the integration tests with `IntegrationContext`
* Separated the *Schema* tests that tend to be slow and destructive of the database structure to *Marten.Schema.Testing*
* Running the main test library with xUnit parallelization turned on
* Selectively turns on Storyteller retry for some of the not perfectly reliable async daemon tests

I'm a little disappointed by the progress here, but I see a little better than a 25% reduction in execution time for the unit tests.


martenbuild.cs
```diff
@@ -10,6 +12,9 @@ internal class MartenBuild
     {
         private const string BUILD_VERSION = "3.11.0";
 
+        private const string DockerConnectionString =
```
oskardudycz, line 15: **CONCERN:** That's minor, but wouldn't it be better to use the same config as in the pipeline tests? https://github.com/JasperFx/marten/pull/1463/files#diff-fec826feae04e51c0d94076385408bdcR36 It might be also good to use different database than default `postgres` as it might interfere with other user configuration.

src/Marten.Schema.Testing/IntegrationContext.cs
```diff
@@ -0,0 +1,152 @@
...
+#endif
+
+        protected IntegrationContext(ITestOutputHelper output = null)
+        {
+            _output = output;
+
+            UseDefaultSchema();
+            _store.Advanced.Clean.CompletelyRemoveAll();
+
+#if NET461
```
oskardudycz, line 29: **CONCERN:** Why is it needed?

src/Marten.Testing/Harness/IntegrationContext.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        protected override DocumentStore theStore
+        {
+            get
+            {
+                if (_store != null) return _store;
+
+                if (!_hasBuiltStore)
+                {
+                    base.theStore.Advanced.Clean.DeleteAllDocuments();
```
oskardudycz, line 56: **CONCERN:** Shouldn't we use here also `CompletelyRemoveAll` like in the code below: https://github.com/JasperFx/marten/pull/1463/files#diff-8f22a8190d9167240cb5140d183b18fbR41 ? If we'd like to remove only data here then we should probably call also `DeleteAllEventData` as integration tests may be call also for Event Store.

src/Marten.Testing/Harness/IntegrationContext.cs
```diff
@@ -0,0 +1,84 @@
...
+                    _hasBuiltStore = true;
+                }
+
+                return base.theStore;
+            }
+        }
+
+        protected virtual IDocumentSession theSession
```
oskardudycz, line 64: **FIX:** I think that this should be removed as it hides the same property from the base class. We should also consider removing [_session](https://github.com/JasperFx/marten/pull/1463/files#diff-8f22a8190d9167240cb5140d183b18fbR10) from this class and using it from the base.

src/Marten.Testing/Harness/OneOffConfigurationsContext.cs
```diff
@@ -0,0 +1,120 @@
...
+using Marten.Testing.Events;
+using Xunit;
+
+namespace Marten.Testing.Harness
+{
+    /// <summary>
+    /// Use this if the tests in a fixture are going to use
+    /// all custom StoreOptions configuration
+    /// </summary>
+    public abstract class OneOffConfigurationsContext
```
oskardudycz, line 14: **SUGGESTION:** Maybe it would be more explanatory to name it "CustomDocumentStoreConfigurationsContext"?

src/Marten.Testing/Harness/OneOffConfigurationsContext.cs
```diff
@@ -0,0 +1,120 @@
...
+            }
+        }
+
+        public void Dispose()
+        {
+            foreach (var disposable in _stores)
+            {
+                disposable.Dispose();
+            }
+            _session?.Dispose();
```
oskardudycz, line 114: **CONCERN:** Shouldn't we dispose session before disposing _stores?

src/Marten.Testing/Harness/StoreFixture.cs
```diff
@@ -0,0 +1,46 @@
...
+                    _store.Advanced.Clean.CompletelyRemoveAll();
+                }
+
+                return _store;
+            }
+        }
+
+        public void Dispose()
+        {
+
```
oskardudycz, line 42: **FIX:** Redundant empty line.
```suggestion
```

src/Marten.Testing/Harness/IntegrationContext.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        private bool _hasBuiltStore = false;
+
+        protected override DocumentStore theStore
+        {
+            get
+            {
+                if (_store != null) return _store;
+
+                if (!_hasBuiltStore)
```
oskardudycz, line 54: **CONCERN:** Is this code reachable? We're setting the _store in the constructor https://github.com/JasperFx/marten/pull/1463/files#diff-8f22a8190d9167240cb5140d183b18fbR39. `_store` is private and not exposed anywhere then it makes it readonly in fact.

src/Marten.Testing/Harness/DestructiveIntegrationContext.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public DestructiveIntegrationContext(DefaultStoreFixture fixture) : base(fixture)
+        {
+            theStore.Advanced.Clean.CompletelyRemoveAll();
+        }
+
+        public override void Dispose()
+        {
+            base.Dispose();
+            theStore.Advanced.Clean.CompletelyRemoveAll();
```
oskardudycz, line 20: **CONCERN:** Do we need to call this also on dispose if we're running it also in constructor? That seems to be potentially redundant.

martenbuild.cs
```diff
@@ -78,9 +89,42 @@ private static void Main(string[] args)
             Target("pack", DependsOn("compile"), ForEach("./src/Marten", "./src/Marten.CommandLine", "./src/Marten.NodaTime"), project =>
                 Run("dotnet", $"pack {project} -o ./../../artifacts --configuration Release"));
 
+            Target("init-db", () =>
```
oskardudycz, line 92: Nice, I like that idea 👍

src/Marten.Testing/Harness/IntegrationContext.cs
```diff
@@ -0,0 +1,100 @@
...
+
+                return _session;
+            }
+        }
+
+        public override void Dispose()
+        {
+            if (_overrodeStore)
+            {
+                Fixture.Store.Advanced.Clean.CompletelyRemoveAll();
```
oskardudycz, line 92: **QUESTION:** Do we need this clean up if we're doing that both in the store options and default store initialization? This seems redundant (I understand that there might be some leftovers but they should be wiped by other integration tests if there is collision with the schema name).

src/Marten.Testing/Harness/IntegrationContext.cs
```diff
@@ -0,0 +1,100 @@
...
+namespace Marten.Testing.Harness
+{
+    [CollectionDefinition("integration")]
+    public class IntegrationCollection : ICollectionFixture<DefaultStoreFixture>
+    {
+
+    }
+
+    [Collection("integration")]
+    public class IntegrationContext : StoreContext<DefaultStoreFixture>
```
oskardudycz, line 14: **CONCERN:** Why do we need few IntegrationContexts? Can't we reuse them? If they have different purposes then it would be good if they have different naming.

src/Marten.NodaTime.Testing/IntegrationContext.cs
```diff
@@ -0,0 +1,149 @@
+using System;
+using Marten.Services;
+using Marten.Testing.Harness;
+using Xunit.Abstractions;
+
+namespace Marten.NodaTime.Testing
+{
+    public abstract class IntegrationContext : IDisposable
```
oskardudycz, line 8: **CONCERN:** Cannot we reuse the `IntegrationContext` class from the main project? Maybe we could create a separate base project with base classes to reuse them between different test projects?

martenbuild.cs
jeremydmiller: I'll make it use "marten_testing" as the database name.

src/Marten.Testing/Harness/IntegrationContext.cs
jeremydmiller: Done.

src/Marten.Testing/Harness/IntegrationContext.cs
jeremydmiller: Done.

src/Marten.Testing/Harness/OneOffConfigurationsContext.cs
jeremydmiller: `_session` in this case isn't from the additional stores, so it doesn't really matter.

src/Marten.Testing/Harness/IntegrationContext.cs
jeremydmiller: It is reachable, and it's getting around whether or not you're using the default store that's shared or using something overridden by StoreOptions

src/Marten.Testing/Harness/DestructiveIntegrationContext.cs
jeremydmiller: We don't do the remove all in the constructor function, and this only applies if someone called `StoreOptions()` that could quite possibly make destructive changes to the public schema. This was a workaround for IntegrationContext sometimes being used for altered configuration in the same class.

martenbuild.cs
jeremydmiller: I use that for something similar at work, but w/ Sql Server

src/Marten.Testing/Harness/IntegrationContext.cs
jeremydmiller: You're not doing exactly the same thing here and in the constructors. We clean off the data no matter what, but the big clean off everything is only when someone potentially changes the public schema.

src/Marten.Testing/Harness/IntegrationContext.cs
jeremydmiller: There's a different integration context in the Schema project and yet another in the Noda, but they behave differently than the one, single context here

src/Marten.NodaTime.Testing/IntegrationContext.cs
jeremydmiller: It wasn't playing nicely here, and I was lazy. I don't see it doing any harm

src/Marten.Testing/Harness/OneOffConfigurationsContext.cs
jeremydmiller: I'd rather not tbh.

src/Marten.Testing/Harness/IntegrationContext.cs
oskardudycz: Can we rename them to eg. HarnessIntegrationContext etc.?

oskardudycz:
@jeremydmiller it would be also good to made the base classes abstract and their constructor protected ;)

jeremydmiller:
I think the base classes are all protected, and I'm saying no to making the constructor protected. That's super annoying 'cause R# tries to make the subclass constructors protected. Can we just get this thing in if the tests pass?

Issue #1432, opened 2020-02-28
CombGuid setting not respected for event streams


lahma:
Noticed that the current implementation of starting a stream with events just uses Guid.NewGuid(), should event store also respect the global setting?

https://github.com/JasperFx/marten/blob/d9294b1348491c53457612ee4e42975cd0204f9a/src/Marten/Events/EventStore.cs#L165-L173

mysticmind:
Currently, it only applies it for document store and not event store. Events stored in `mt_events` uses CombGuid for `id`  since the events are contiguous within a stream. But I am not sure how useful CombGuid will be for stream id in `mt_streams` since the nature of it does not need to be contiguous. 

Given the above, I feel that the current implementation is fine. That said, `options.DefaultIdStrategy = (mapping, storeOptions) => new CombGuidIdGeneration();` gives an impression that everything will use CombGuid but in reality, only the document store uses this.

Thoughts?

mysticmind:
@jeremydmiller @oskardudycz @jokokko Could you please share your your thoughts/comments?

oskardudycz:
@mysticmind @lahma I think that event store should respect the global settings.

Btw. it's weird and really bad that we're using the `Guid.NewGuid` for the streams and events :( I was sure that we're using sequential guids for that... We need to change that asap.

I think that we can include that even before the v4 with the feature flag as we're doing for the `DuplicatedFieldEnumStorage` and change that in v4 to respect it by default.

jeremydmiller:
@mysticmind I think this was just an oversight on my part originally. I'll take the itty, bitty one here.

lahma:
Thank you 👍🏻

PR #1480, opened 2020-05-21
Event Store uses sequential guids for the stream ids. Closes GH-1432

jeremydmiller:
I think this was mostly an oversight on my part to begin with


jeremydmiller:
I vote to deal with making the identity more variable in v4, so I'm pulling this in.

Issue #1466, opened 2020-05-11
Import Marten.AspNetCore into the official repository


jeremydmiller:
With @jokokko 's sign off, let's import the [Marten.AspNetCore](https://github.com/jokokko/marten.aspnetcore) library into the official mothership repository and add documentation for that here.

## Tasks

- [ ] Import the code
- [ ] Add documentation
- [ ] Target `IHostBuilder` in .Net Core 3.* apps
- [ ] Target `IWebHostBuilder` in all targets (PITA to switch between the two things, but what can you do?)
- [ ] Optionally start the Async Daemon as an `IHostedService` (might break this out to a separate issue)
- [ ] Register container resolved `IDocumentSessionListener` objects that would be added to newly created `IDocumentSession` objects at resolution time (might break this out later)
- [ ] Might need to allow users to register an all custom factory for `IDocumentSession` building

jeremydmiller:
I think now we'll just stick the extension methods on `ServiceCollection` into the main Marten lib

jeremydmiller:
This was done through other pull requests to v3.12

PR #1490, opened 2020-05-26
Pool in existing code for #1466. Separate DI & Asp.Net Core integration?

jokokko:
Skeletons

### DIFF
--- a/src/Marten.AspNetCore.Tests/Marten.AspNetCore.Tests.csproj
+++ b/src/Marten.AspNetCore.Tests/Marten.AspNetCore.Tests.csproj
@@ -1,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>        
+        <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
+        <AssemblyName>Marten.AspNetCore.Tests</AssemblyName>
+		<GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
+		<GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
+        <LangVersion>7.2</LangVersion>
+    </PropertyGroup>	
+    <PropertyGroup>
+        <NoWarn>1570;1571;1572;1573;1574;1587;1591;1701;1702;1711;1735</NoWarn>
+    </PropertyGroup>    
+    <ItemGroup>
+        <ProjectReference Include="..\Marten.AspNetCore\Marten.AspNetCore.csproj" />
+		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.3.0" />
+		<PackageReference Include="xunit.runner.visualstudio" Version="2.4.1">
+		  <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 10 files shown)


jeremydmiller:
Some thoughts:

* Let's add any extensions to `IServiceCollection` to the main Marten library itself
* Any extension method on `IServiceCollection` should be `AddMarten()` just to conform to .Net Core idioms. 
* I'm on the fence for having a separate library for extensions to `IHostBuilder` or `IWebHostBuilder`. I thought it was a pain in other OSS projects to support both across different .Net Core versions. I can be convinced either way, especially if someone else does the work;) There is prior art in Alba or Oakton to look at in this regard.
* I think I'd register the `IDocumentSession` and `IQuerySession` by default. I'd vote for `LightweightSession()` as the default for the `IDocumentSession`
* I like what @jokokko did for chaining the session creation options
* For the `AddMarten()` overloads, I'd recommend at least `AddMarten(string connectionString)` as a super quick start and a more usable `AddMarten(Action<StoreOptions> configure)`. With maybe a 3rd `AddMarten(StoreOptions options)`. And all three chain a fluent interface with additional options to customize the session creation
* For the session creation customization (the registration of IDocumentSession or IQuerySession), I think we should support some kind of optional builder like so:

```
public interface ISessionBuilder
{
    IDocumentSession BuildSession();
    IQuerySession BuildQuerySession();
}
```

so it could be registered like:

```
public void ConfigureServices(IServiceCollection services)
{
    services.AddMarten(...).BuildSessionsWith<MySessionBuilder>();
}

public class MySessionBuilder : ISessionBuilder
{
    public MySessionBuilder(IDocumentStore store, IDocumentSessionListener[] listeners)
    {
          // do special stuff. Might be using some other services injected in here
          // for multi-tenancy. Or doing some metadata trackign when we get that.
          // or adding listeners that need to come from a DI container
     }
}
```

I'm suggesting using a builder class here instead of using a Lambda service registration to optimize the DI mechanics instead of doing repetitive service location calls.

* I'd keep the option to open sessions with a custom `SessionOptions` too

* For the hosted service for the Marten async daemon, we've got to also do something for ownership in a cluster. Maybe w/ postgresql advisory locks?


I'm sure we'll think of other things later

oskardudycz:
@jeremydmiller @jokokko do we plan to have it in v3.12? I think that at least an initial method for configuring document store and document session would be nice to have plus documentation as this is a huge pain and most often asked question by our users.

I can work on that if you don't have time.

oskardudycz:
As we discussed on Gitter. I agree with @jeremydmiller suggestions. I'd add that in my opinion for the current version it'd be better to:
- keep the DocumentSession as default. People are probably using that most of the time. We could consider using Lightweight in v4 with doing some explanation to the people
- not add host builders extension currently to do not need to do cross-version support (`IWebHostBuilder` in netcoreapp2.x and `IHostBuilder` in netcoreapp3.x`). As 3.1 is currently LTS then we can drop 2.x support in v4 and add just support for  `IHostBuilder` and not have the maintenance pain.

For sure we need to add also good documentation around that.

mysticmind:
#1501 addresses the requirements around DI registration. Closing this one.

PR #1501, opened 2020-06-08
DI registrations

jeremydmiller:
The new `AddMarten()` service registrations we've been talking about for v3.12. I have not finished the documentation yet, but the code is ready to look at.

I'm punting on the async daemon to v4 if that's okay. 

I'll be adding more docs in my tomorrow. Sorry, but that was just a bit too much tonight.

### DIFF
--- a/src/Marten.Testing/MartenServiceCollectionExtensionsTests.cs
+++ b/src/Marten.Testing/MartenServiceCollectionExtensionsTests.cs
@@ -1,0 +1,142 @@
+using Lamar;
+using Marten.Testing.Harness;
+using Microsoft.Extensions.DependencyInjection;
+using Shouldly;
+using Xunit;
+
+namespace Marten.Testing
+{
+
+
+
+    public class MartenServiceCollectionExtensionsTests
+    {
+        // Using Lamar for testing this because of its diagnostics
+
+        [Fact]
+        public void add_marten_with_just_connection_string()
+        {
+            using (var container = Container.For(x =>
+            {
+                x.AddMarten(ConnectionSource.ConnectionString);
+            }))
+            {
+                ShouldHaveAllTheExpectedRegistrations(container);
+            }
+        }
+
+        [Fact]
+        public void add_marten_by_store_options()
+        {
+            using (var container = Container.For(x =>
+            {
+                var options = new StoreOptions();
+                options.Connection(ConnectionSource.ConnectionString);
+                x.AddMarten(options);
+            }))
+            {
+                ShouldHaveAllTheExpectedRegistrations(container);
+            }
+        }
+
+        [Fact]
+        public void add_marten_by_configure_lambda()
+        {
+            using (var container = Container.For(x =>
+            {
+                x.AddMarten(opts => opts.Connection(ConnectionSource.ConnectionString));
+            }))
+            {
+                ShouldHaveAllTheExpectedRegistrations(container);
+            }
+        }
+
+        [Fact]
+        public void eager_initialization_of_the_store()
+        {
+            IDocumentStore store = null;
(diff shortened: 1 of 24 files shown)


src/Marten.Testing/MartenServiceCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,142 @@
+using Lamar;
+using Marten.Testing.Harness;
+using Microsoft.Extensions.DependencyInjection;
+using Shouldly;
+using Xunit;
+
+namespace Marten.Testing
+{
+
```
oskardudycz, line 9: **SUGGESTION:** It would be nice to remove those empty lines.

src/Marten/MartenServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// <param name="services"></param>
+        /// <param name="options">The Marten configuration for this application</param>
+        /// <returns></returns>
+        public static MartenConfigurationExpression AddMarten(this IServiceCollection services, StoreOptions options)
+        {
+            services.AddSingleton(options);
+            services.AddSingleton<IDocumentStore, DocumentStore>();
+
+            // This can be overridden by the expression following
+            services.AddSingleton<ISessionFactory, DefaultSessionFactory>();
```
oskardudycz, line 36: **SUGGESTION:** We should use `TryAddSingleton` method to allow user to override this implemntation. See more in [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.extensions.servicecollectiondescriptorextensions.tryaddsingleton?view=dotnet-plat-ext-3.1).
```suggestion
            services.TryAddSingleton<ISessionFactory, DefaultSessionFactory>();
```

src/Marten/MartenServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            /// <summary>
+            /// Use an alternative strategy / configuration for opening IDocumentSession or IQuerySession
+            /// objects in the application with a custom ISessionFactory type registered as a singleton
+            /// </summary>
+            /// <typeparam name="T"></typeparam>
+            /// <returns></returns>
+            public MartenConfigurationExpression BuildSessionsWith<T>() where T : class, ISessionFactory
+            {
+                _services.AddSingleton<ISessionFactory, T>();
```
oskardudycz, line 79: Same comment as above https://github.com/JasperFx/marten/pull/1501/files#r436521253.

Also `ISessionFactory` not always should be Singleton. For a non-default case I'd rather expect it to be Scoped - eg. if the user would like to start the session with Tenant. The tenant probably will be taken from the request.

src/Marten/MartenServiceCollectionExtensions.cs
jeremydmiller: I don't think that's necessary here Oskar.

src/Marten.Testing/MartenServiceCollectionExtensionsTests.cs
jeremydmiller: Let's please not invest time in this kind of code formatting in the test code

jeremydmiller:
Hey all, I think I've addressed all the comments in the PR and I want to declare victory. I've:

* flushed out the guide for the new extensions
* added a super, quick sample to the Getting Started page and linked to the bigger guide on AddMarten()
* replaced the StructureMap example with Lamar as the alternative IoC registration

mysticmind:
This is super good docs, thanks @jeremydmiller ❤️ 

Also this addresses the most frequent question asked by users!

Issue #1472, opened 2020-05-11
Use Oakton.AspNetCore for Marten.CommandLine


jeremydmiller:
The `Marten.CommandLine` package already uses Oakton for command line parsing. For easier integration in .Net Core applications, we could shift that to using the [Oakton.AspNetCore](https://jasperfx.github.io/oakton/documentation/aspnetcore/) package so the command line support can be added to any ASP.net Core 2.* or .Net Core 3.* project by installing the Nuget. This might simplify the usage because you'd no longer need a separate project for the command line support.

There are some long standing stories about extending the command line support for the event store projection rebuilding. I think that would be more effective if it switches over to Oakton.AspNetCore.

PR #1505, opened 2020-06-10
Updated command line lib to use Oakton.AspNetCore, also updated Basel…

jeremydmiller:
This wil close GH-1472.

* Change Marten.CommandLine to use Oakton.AspNetCore for easier usage, if limited to .Net Core 3.0
* Updates Marten itself to Baseline 2.*, and removed some duplication of utilities that moved from Marten to Baseline 2


jeremydmiller:
This code is already in master, not sure how it got orphaned.

PR #1508, opened 2020-06-16
Call the correct TransactionState constructor

### LINKED ISSUE #1507: Connection is disposed on SafeChanges(Async)
We would like to manage the Npgsql connection by ourselves and we therefore used the possibility to inject a connection to the SessionOptions.

This approach is working fine as long as we are only reading. But after calling SaveChanges or SaveChangesAsync the session is not usable anymore.

It looks like the session disposes the connection and sets it to Null on commits and rollbacks.

I would expect the session to not close the connection when getting a connection via the sessionOptions and keep it alive at least until the session is disposed.

Instead it seems like the session tries to create a new Connection for the next database interaction and throws a NullReferenceException in the process.

Do I have any other options to bypass this behaviour and keep the connection and the session alive?
Is this behaviour intended or may this be fixed?

I created a test that shows my problem:
```C#
  class TestObject
{
    public Guid Id { get; set; }
    public string Name { get; set; }
}

[Test]
public void Inject_Connection()
{
    var connectionString = $"host=host;database=db;password=pw;username=un;";
    var store = DocumentStore.For(connectionString);

    var connection = new NpgsqlConnection(connectionString);
    connection.Open();
    var session = store.OpenSession(new SessionOptions { Connection = connection });

    var testObject = new TestObject{Id = new Guid(), Name = "testObject"};

    session.Store(testObject);
    session.SaveChanges();

    // The following call will throw a NullReferenceException
    session.Query<TestObject>().Any(t => t.Id == testObject.Id);
}
```

Thanks! Looking forward to your reply!

TobiasSchaffner:
The session raised a NullReferenceException on usage after a SaveAsync call when constructed with Session options.

The problem can be found in the ManagedConnection when constructed with SessionOptions.
After a dispose of the TransactionState the ManagedConnection will try to rebuild it on further usage in the buildConnection method, The TransactionState constructor used relies on, that the factory has been set before.

This behaviour has been fixed by using the correct constructor and providing the fields needed.

The following test replicates the behaviour:
```C#
  class TestObject
{
    public Guid Id { get; set; }
    public string Name { get; set; }
}

[Test]
public void Inject_Connection()
{
    var connectionString = $"host=host;database=db;password=pw;username=un;";
    var store = DocumentStore.For(connectionString);

    var connection = new NpgsqlConnection(connectionString);
    connection.Open();
    var session = store.OpenSession(new SessionOptions { Connection = connection });

    var testObject = new TestObject{Id = new Guid(), Name = "testObject"};

    session.Store(testObject);
    session.SaveChanges();

    // The following call will throw a NullReferenceException without this patch
    session.Query<TestObject>().Any(t => t.Id == testObject.Id);
}
```

fixes #1507

### DIFF
--- a/src/Marten.Testing/CoreFunctionality/SessionOptionsTests.cs
+++ b/src/Marten.Testing/CoreFunctionality/SessionOptionsTests.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using Baseline;
 using Marten.Services;
 using Marten.Testing.Harness;
@@ -130,18 +131,18 @@ public void ConfigureCommandTimeout(IDocumentStore store)
 
             var options = new SessionOptions() { Connection = connection };
 
-            var testobject = new FryGuy();
+            var testObject = new FryGuy();
 
             using (var session = documentStore.OpenSession(options))
             {
-                session.Store(testobject);
+                session.Store(testObject);
                 session.SaveChanges();
-                session.Load<FryGuy>(testobject.Id).ShouldNotBeNull();
+                session.Load<FryGuy>(testObject.Id).ShouldNotBeNull();
             }
         }
 
         [Fact]
-        public void session_with_custom_connection_reusable_after_saveChangesAsync()
+        public async Task session_with_custom_connection_reusable_after_saveChangesAsync()
         {
             var connectionStringBuilder = new NpgsqlConnectionStringBuilder(ConnectionSource.ConnectionString);
 
@@ -155,13 +156,14 @@ public void ConfigureCommandTimeout(IDocumentStore store)
 
             var options = new SessionOptions() { Connection = connection };
 
-            var testobject = new FryGuy();
+            var testObject = new FryGuy();
 
             using (var query = documentStore.OpenSession(options))
             {
-                query.Store(testobject);
-                query.SaveChangesAsync().Wait();
-                query.LoadAsync<FryGuy>(testobject.Id).GetAwaiter().GetResult().ShouldNotBeNull();
+                query.Store(testObject);
+                await query.SaveChangesAsync();
+                var loadedObject = await query.LoadAsync<FryGuy>(testObject.Id);
+                loadedObject.ShouldNotBeNull();
             }
         }
 
--- a/src/Marten.Testing/CoreFunctionality/SessionOptionsTests.cs
+++ b/src/Marten.Testing/CoreFunctionality/SessionOptionsTests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using Baseline;
 using Marten.Services;
 using Marten.Testing.Harness;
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1508) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1508) before we can accept your contribution.<br/><br/>:x: TobiasSchaffner [sign now](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1508)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/JasperFx/marten?pullRequest=1508) it.</sub>

src/Marten.Testing/CoreFunctionality/SessionOptionsTests.cs
```diff
@@ -115,6 +115,56 @@ public void can_override_pgcstring_timeout_in_sessionoptions()
...
+
+            var options = new SessionOptions() { Connection = connection };
+
+            var testobject = new FryGuy();
+
+            using (var query = documentStore.OpenSession(options))
+            {
+                query.Store(testobject);
+                query.SaveChangesAsync().Wait();
+                query.LoadAsync<FryGuy>(testobject.Id).GetAwaiter().GetResult().ShouldNotBeNull();
```
oskardudycz, line 164: **SUGGESTION**
```suggestion
                const loadedObject = await query.LoadAsync<FryGuy>(testobject.Id);
                loadedObject.ShouldNotBeNull();
```

src/Marten.Testing/CoreFunctionality/SessionOptionsTests.cs
```diff
@@ -115,6 +115,56 @@ public void can_override_pgcstring_timeout_in_sessionoptions()
...
+            using (var session = documentStore.OpenSession(options))
+            {
+                session.Store(testobject);
+                session.SaveChanges();
+                session.Load<FryGuy>(testobject.Id).ShouldNotBeNull();
+            }
+        }
+
+        [Fact]
+        public void session_with_custom_connection_reusable_after_saveChangesAsync()
```
oskardudycz, line 144:
```suggestion
        public async Task session_with_custom_connection_reusable_after_saveChangesAsync()
```

src/Marten.Testing/CoreFunctionality/SessionOptionsTests.cs
```diff
@@ -115,6 +115,56 @@ public void can_override_pgcstring_timeout_in_sessionoptions()
...
+            connection.Open();
+
+            var options = new SessionOptions() { Connection = connection };
+
+            var testobject = new FryGuy();
+
+            using (var query = documentStore.OpenSession(options))
+            {
+                query.Store(testobject);
+                query.SaveChangesAsync().Wait();
```
oskardudycz, line 163:
```suggestion
                await query.SaveChangesAsync();
```

src/Marten/Services/ManagedConnection.cs
```diff
@@ -19,6 +19,7 @@ public class ManagedConnection: IManagedConnection, IDatabase
         private TransactionState _connection;
         private readonly bool _ownsConnection;
         private readonly IRetryPolicy _retryPolicy;
+        private NpgsqlConnection _npgsqlConn;
```
oskardudycz, line 22: **SUGGESTION:** I think that `_externalConnection` should be more descriptive.
```suggestion
        private NpgsqlConnection _externalConnection;
```

Issue #1519, opened 2020-07-14
Duplicated field related generated sql, doesn't respect casing.


nikodemrafalski:
Hi, probably a bug in Marten (3.12)

```
public class Something
{
   public Guid Id{get;set;}
   public Guid OtherId{get;set;
}
```

```
var store = DocumentStore.For(options =>
            {
                options.PLV8Enabled = true;
                options.Connection(config.ConnectionString);
                options.UseDefaultSerialization(EnumStorage.AsString, Casing.CamelCase);
                options.DatabaseSchemaName = config.SchemaName;
                options.AutoCreateSchemaObjects = AutoCreate.CreateOrUpdate;
                options.Schema.For<Something>()
                   .Duplicate(x => x.OtherId)
                   .VersionedWith(x => x.Version);
            });
```

When marten generates migration patch to populate duplicated  `other_id` it references to `OtherId` as `data ->> 'OtherId'` instead of `data -->> 'otherId'`.
At first id did not have duplicated column it is a migration to existing data.
 same patch like 
```
   session.Patch<Something>(x => x.OtherId== @event.StreamId && x.AnyTenant()).Set("anything", (string)null);
```
also generates SQL like that ```;update mt_doc_something as d set other_id= CAST(data ->> 'OtherId' as uuid) where (d.other_id= :arg19 and 1=1)```

I suspect there's a bug here (DefaultCasing hardcoded in both if branches)
https://github.com/JasperFx/marten/blob/61200af89aaafa1a71e76bcd95d20f2eb24ae9de/src/Marten/Schema/DuplicatedField.cs#L106

mysticmind:
Fixed via #1520

PR #1520, opened 2020-07-14
Fix duplicate field schema generation to respect casing

mysticmind:
- Fix the duplicate field schema generation to respect casing
- Add unit tests
- fixes GH-1519


jeremydmiller:
Thumbs up from me, but I'll need to watch this when it hits the 4.0 codebase. Really wish folks wouldn't do this kind of thing.

PR #1542, opened 2020-08-10
Support case-insensitive order by for strings

lahma:
I asked about this on Gitter with little feedback, it seemed to be missing so I decided to try how one could implement order by semantics case-insensitively. Some remarks:

* To order normal column one would use `citext` type or in Postgres v12 there's new collation support, either of if these are not available on Marten context
* This solution uses `lower(field)` which should work for most use cases, with slight performance penalty
* Because Remotion.Linq doesn't support the comparer parameter, a new code path needs to be taken
    * This is why you cannot combine "regular" OrderBy and OrderBy that has comparer

This was my first try against Remotion.Linq so not sure if the approach is ideal.

I did this against 3.12 as I would need this feature, if you find this reasonable addition. Otherwise it would be nice to find a way to plug this in on consumer side. Basically the `writeOrderClause` and friends should somehow be virtual.

### DIFF
--- a/src/Marten/Linq/Model/SelectManyQuery.cs
+++ b/src/Marten/Linq/Model/SelectManyQuery.cs
@@ -171,20 +171,7 @@ namespace Marten.Linq.Model
 
         private string determineOrderClause(ChildDocument document)
         {
-            var orders = bodyClauses()
-                .SelectMany<IBodyClause, (Ordering Clause, bool CaseSensitive)>(x =>
-                {
-                    switch (x)
-                    {
-                        case OrderByClause orderByClause:
-                            return orderByClause.Orderings.Select(o => (o, true));
-                        case OrderByComparerClause orderByComparerClause:
-                            return orderByComparerClause.Orderings.Select(o => (o, orderByComparerClause.CaseSensitive));
-                        default:
-                            return Enumerable.Empty<(Ordering, bool)>();
-                    }
-                })
-                .ToArray();
+            var orders = bodyClauses().GetStringOrderingClauses();
 
             if (!orders.Any())
                 return string.Empty;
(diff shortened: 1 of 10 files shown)


src/Marten/Linq/Model/SelectManyQuery.cs
```diff
@@ -171,16 +171,35 @@ public void ConfigureCommand(IIncludeJoin[] joins, ISelector selector, CommandBu
 
         private string determineOrderClause(ChildDocument document)
         {
-            var orders = bodyClauses().OfType<OrderByClause>().SelectMany(x => x.Orderings).ToArray();
+            var orders = bodyClauses()
+                .SelectMany<IBodyClause, (Ordering Clause, bool CaseSensitive)>(x =>
```
mysticmind, line 175: This block of code looks like a repetitive one as in `LinqQuery.cs` hence suggest to add a class to under `Util` and use it in both the places. Thoughts?

lahma: I can certainly do that 👍 

One other thing came to my mind, should we also support:

```
session.Query<Target>()
                .Where(x => x.Id.In(ids))
                .OrderBy(x => x.String.ToLower())
                .ThenBy(x => x.AnotherString.ToLower())
```

This has the upside of allowing the chaning without the `IComparer` requirement that I explained in the description, but is less idiomatic in LINQ sense, both would produce the same SQL. (Of course Invariant versions and ToUpper also would be supported.

lahma:
One other thing came to my mind, should we also support:

```
session.Query<Target>()
                .Where(x => x.Id.In(ids))
                .OrderBy(x => x.String.ToLower())
                .ThenBy(x => x.AnotherString.ToLower())
```

This has the upside of allowing the chaining without the `IComparer` requirement that I explained in the description, but is less idiomatic in LINQ sense, both would produce the same SQL. (Of course `Invariant()` versions and `ToUpper()` also would be supported.

src/Marten/Linq/Model/SelectManyQuery.cs
lahma: there's now OrderingExtensions helper

PR #1543, opened 2020-08-13
Introduce DefaultRetryPolicy

lahma:
Per Gitter discussion, this is the [documentation sample retry policy](https://martendb.io/documentation/documents/advanced/retrypolicy/) with some tweaks.

* filter defaults to checking for `x is NpgsqlException npgsqlException && npgsqlException.IsTransient`
* renamed `NTimes` to just `Times`
* new parameter `sleep` that allows to return `TimeSpan` to sleep given current retry number, defaults to retry number as seconds
* made `IRetryPolicy` interface implementation explicit to hide members not relevant for public API

### DIFF
--- a/src/Marten/IRetryPolicy.cs
+++ b/src/Marten/IRetryPolicy.cs
@@ -76,51 +76,52 @@ namespace Marten
     /// </remarks>
     public sealed class DefaultRetryPolicy: IRetryPolicy
     {
-        private readonly int _maxTries;
+        private readonly int _maxRetryCount;
         private readonly Func<Exception, bool> _filter;
         private readonly Func<int, TimeSpan> _sleep;
 
         private static readonly Func<Exception, bool> DefaultFilter = x => x is NpgsqlException npgsqlException && npgsqlException.IsTransient;
         private static readonly Func<int, TimeSpan> DefaultSleep = x => TimeSpan.FromSeconds(x);
 
-        private DefaultRetryPolicy(int maxTries, Func<Exception, bool> filter, Func<int, TimeSpan> sleep)
+        private DefaultRetryPolicy(int maxRetryCount, Func<Exception, bool> filter, Func<int, TimeSpan> sleep)
         {
-            _maxTries = maxTries;
+            _maxRetryCount = maxRetryCount;
             _filter = filter ?? DefaultFilter;
             _sleep = sleep ?? DefaultSleep;
         }
 
         /// <summary>
-        /// Initializes a retry policy that will retry once after failure.
+        /// Initializes a retry policy that will retry once after failure that matches the filter.
         /// </summary>
         /// <param name="filter">Optional filter when to apply, default to checking for <see cref="NpgsqlException.IsTransient"/></param>
         /// <param name="sleep">Optional sleep after exception, gets retry number 1-N as parameter, defaults to sleeping retry number seconds</param>
         /// <returns>The configured retry policy.</returns>
         public static IRetryPolicy Once(Func<Exception, bool> filter = null, Func<int, TimeSpan> sleep = null)
         {
-            return new DefaultRetryPolicy(2, filter, sleep);
+            return new DefaultRetryPolicy(1, filter, sleep);
         }
 
         /// <summary>
-        /// Initializes a retry policy that will retry twice after failure.
+        /// Initializes a retry policy that will retry twice after failure that matches the filter.
         /// </summary>
         /// <param name="filter">Optional filter when to apply, default to checking for <see cref="NpgsqlException.IsTransient"/></param>
         /// <param name="sleep">Optional sleep after exception, gets retry number 1-N as parameter, defaults to sleeping retry number seconds</param>
         /// <returns>The configured retry policy.</returns>
         public static IRetryPolicy Twice(Func<Exception, bool> filter = null, Func<int, TimeSpan> sleep = null)
         {
-            return new DefaultRetryPolicy(3, filter, sleep);
+            return new DefaultRetryPolicy(2, filter, sleep);
         }
 
         /// <summary>
         /// Initializes a retry policy that will retry given amount of times after failure.
         /// </summary>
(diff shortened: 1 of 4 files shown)


lahma:
I still need add incremental sleep between retries.

src/Marten/IRetryPolicy.cs
```diff
@@ -66,4 +67,115 @@ public Task<TResult> ExecuteAsync<TResult>(Func<Task<TResult>> operation, Cancel
...
+        /// <param name="sleep">Optional sleep after exception, gets retry number 1-N as parameter, defaults to sleeping retry number seconds</param>
+        /// <returns>The configured retry policy.</returns>
+        public static IRetryPolicy Times(int times, Func<Exception, bool> filter = null, Func<int, TimeSpan> sleep = null)
+        {
+            return new DefaultRetryPolicy(times + 1, filter, sleep);
+        }
+
+        void IRetryPolicy.Execute(Action operation)
+        {
+            Try(() =>
```
oskardudycz, line 128: @lahma could you use [NoSynchronizationContextScope](https://github.com/JasperFx/marten/blob/master/src/Marten/Util/NoSynchronizationContextScope.cs) in non-async methods to prevent deadlocks? Here's sample usage https://github.com/JasperFx/marten/blob/master/src/Marten/Events/Projections/ViewProjection.cs#L773

src/Marten/IRetryPolicy.cs
```diff
@@ -66,4 +67,115 @@ public Task<TResult> ExecuteAsync<TResult>(Func<Task<TResult>> operation, Cancel
...
+            _sleep = sleep ?? DefaultSleep;
+        }
+
+        /// <summary>
+        /// Initializes a retry policy that will retry once after failure.
+        /// </summary>
+        /// <param name="filter">Optional filter when to apply, default to checking for <see cref="NpgsqlException.IsTransient"/></param>
+        /// <param name="sleep">Optional sleep after exception, gets retry number 1-N as parameter, defaults to sleeping retry number seconds</param>
+        /// <returns>The configured retry policy.</returns>
+        public static IRetryPolicy Once(Func<Exception, bool> filter = null, Func<int, TimeSpan> sleep = null)
```
oskardudycz, line 99: **KUDOS** I really like what you did with Once, Twice and Times. Nice API design 👍

src/Marten/IRetryPolicy.cs
lahma: Added that, cool little helper!

src/Marten/IRetryPolicy.cs
lahma: Can't take the credit, much based on @jokokko 's work 🙂

oskardudycz:
@lahma will you be also so kind and update the documentation with information about the built-in `DefaultRetryPolicy`? 

I'm also fine with updating sample to be retrying transient errors.

src/Marten/IRetryPolicy.cs
oskardudycz: Thanks!

src/Marten/IRetryPolicy.cs
```diff
@@ -66,4 +67,118 @@ public Task<TResult> ExecuteAsync<TResult>(Func<Task<TResult>> operation, Cancel
...
+                {
+                    operation();
+                    return Task.CompletedTask;
+                }, CancellationToken.None).GetAwaiter().GetResult();
+            }
+        }
+
+        TResult IRetryPolicy.Execute<TResult>(Func<TResult> operation)
+        {
+            return Try(() => Task.FromResult(operation()), CancellationToken.None).GetAwaiter().GetResult();
```
oskardudycz, line 140: @lahma this one would be also good to wrap with `NoSynchronizationContextScope`.

lahma: Ah, missed that one, will amend to earlier commit.

lahma:
> @lahma will you be also so kind and update the documentation with information about the built-in `DefaultRetryPolicy`?

I tried to do that in `update samples` commit, don't know how that ends looking

> I'm also fine with updating sample to be retrying transient errors.

Updated the sample to check that it has to be transient, test case will use old logic after the sample.

src/Marten/IRetryPolicy.cs
```diff
@@ -66,4 +67,121 @@ public Task<TResult> ExecuteAsync<TResult>(Func<Task<TResult>> operation, Cancel
...
+    public sealed class DefaultRetryPolicy: IRetryPolicy
+    {
+        private readonly int _maxTries;
+        private readonly Func<Exception, bool> _filter;
+        private readonly Func<int, TimeSpan> _sleep;
+
+        private static readonly Func<Exception, bool> DefaultFilter = x => x is NpgsqlException npgsqlException && npgsqlException.IsTransient;
+        private static readonly Func<int, TimeSpan> DefaultSleep = x => TimeSpan.FromSeconds(x);
+
+        private DefaultRetryPolicy(int maxTries, Func<Exception, bool> filter, Func<int, TimeSpan> sleep)
```
mysticmind, line 86: I understand the code has been adapted based on what @jokokko put in the docs. For the value of `maxTries`, user will typically put the exact number of retries i.e. for once, it 1, for twice, it is 2. 

With the current implementation, we are expecting the user to pass the value incremented by 1 which could lead to a bit of confusion. I would suggest that we change the code logic internally as `catch (Exception e) when (tries++ <= _maxTries && _filter(e)) {}` rather than `catch (Exception e) when (++tries < _maxTries && _filter(e)) {}` which requires the value to be incremented by 1 for the condition to work. 
In summary,  the change in code is `++tries` to `tries++`.  Now, user can pass the exact number of retries.

lahma: Sounds very reasonable, tweaked code with https://github.com/JasperFx/marten/pull/1543/commits/a3f8a7e2bdffb7404795307c33487e65495e0b0f

Issue #1276, opened 2019-06-03
NodaTime Instant can't be used in a select projection


tetious:
On current master, it looks like there is a problem when using an Instant in a select. If you specify an instant in your type like this (where CreatedOn is the Instant):

```csharp
s.Query<Client>().Select(c => new { c.Id, c.Name, c.CreatedOn })
```

It builds the query like this:
```sql
select jsonb_build_object('Id', d.id, 'Name', d.data ->> 'Name', 
'CreatedOn', CAST(d.data ->> 'CreatedOn' as timestamp)) as json from public.mt_doc_client as d
```

The problem with casting to timestamp is that this omits the Z at the end, breaking NodaTime's Instant parser.

```
NodaTime.Text.UnparsableValueException: The value string does not match a quoted string in the pattern. Value being parsed: '2019-05-09T02:45
:25.299385^'. (^ indicates error position.)
```

Possibly the correct thing to do would be to leave it as a string, which is how it behaves when you leave out the select. Another option would be to tweak the parser itself when configuring NodaTime's serializer. I took a peek at the code, but I'm not sure of the best way to handle this such that it won't break things when not using NodaTime. 

Hopefully someone more familiar with the codebase might know exactly how to address this. :) If tweaking the serializer config makes most sense, I can likely figure this out and do a PR.

Thanks for all your awesome hard work on Marten! It's a pleasure to use.

oskardudycz:
@tetious I'll doublecheck that. But imho after my fixes it should work fine in version `3.6.0` that will be finally publish in next few days.

oskardudycz:
Closed by: https://github.com/JasperFx/marten/pull/1290
Added InstantJsonConverter to properly deserialize Instants when db won't return data in TZ format.
The need come up with #1276.
This is workaround for NodaTime issues nodatime/nodatime#154 and nodatime/nodatime#1237.

Revisit this when NodaTime 3.0.0 will be officialy released - then this class might be not needed anymore.

PR #1545, opened 2020-08-20
Fix projecting nullable Instant

lahma:
This adds upon earlier fix https://github.com/JasperFx/marten/pull/1290 for https://github.com/JasperFx/marten/issues/1276 , unfortunately the converter didn't check for Nullable<Instant> and projecting such non-null value from database fails with error: 

```
Newtonsoft.Json.JsonSerializationException: Cannot convert value to System.Nullable`1[NodaTime.Instant]
 ---> NodaTime.Text.UnparsableValueException: The value string does not match a quoted string in the pattern. Value being parsed: '2020-08-20T11:11:32.467173^'. (^ indicates error position.)
```

I extended the earlier test case a bit and added missing type check. Hoping to get this to next servicing release 🙏 

/cc @oskardudycz


mysticmind (approved):
Thanks for the fix @lahma . Looks good to me. 🚀

PR #1560, opened 2020-09-29
Add NGram search functionality. Generates unigrams, bigrams & trigram…

apereiratl:
…s since that allows us the best search experience. Previously, search was not able to accurately match strings for multi search terms if the term was within a multi word string and within the words. Example, we now accurately match 'rich com text' within 'Communicating Across Contexts (Enriched)'. Currently, defaults to english. Which needs to be fixed later.


jokokko:
Neat! Maybe some overlap with pg_trgm (https://www.postgresql.org/docs/current/pgtrgm.html)? Just wondering what the perf implications are over it...

apereiratl:
pg_trgm generates trigrams and as such if you're going to do a search as you type query, your first two (single and double character) searches will result in table lookups, not index lookups since trgm will not generate grams of length < 3. I can provide performance comparisons but from my review of the code with a table with a million or so documents, it was just as fast as pg_trgm.

mdissel:
Can you add some tests?

apereiratl:
@mdissel Will do. There's also one item I need to resolve. The indexes generated by the feature are not properly supported when performing migrations. I'm currently working on this. I will push a fix with updates and with documentation as well. This might take a few days.

apereiratl:
I've create a 4.x PR for this https://github.com/JasperFx/marten/pull/1990. Should I close this one?

mysticmind:
yes, closing the same.

PR #1566, opened 2020-10-09
Remove incorrect comment

jchannon:


oskardudycz (approved):
Good catch @jchannon 👍

PR #1579, opened 2020-10-18
Update minimum npgsql to 4.1.5

Hawxy:
Changes: https://github.com/npgsql/npgsql/milestone/70?closed=1

Most importantly, this fixes the long-running problem of combining NodaTime with BCL DateTime types.


mysticmind (approved):
@Hawxy Looks good to me. Also thanks for updating the unit tests 👏

Hawxy:
Note: The custom type mappings in the nodatime package should be able to be binned once 4.1.6 is released, as 4.1.5 missed a mapping that the event stream requires. I'll create a PR whenever it's available.

mysticmind:
This change is already added to 4.x so cherry picking to merge from 3.x is not required.

PR #1585, opened 2020-10-21
Allow null enum values to be parsed in duplicated field resolution

eouw0o83hf:
This caused a nasty production bug for me when constructing a query against duplicated columns using optional args. For instance, given the document setup:

```
class ExampleDocument
{
    Colors Color { get; set; }
    string Name { get; set; }
}
```

I had a method which was something to the tune of:

```
IEnumerable<ExampleDocument> ListDocuments(Colors? color, string? name)
    =>  _querySession
        .Query<ExampleDocument>()
        .Where(a => (color != null && a.Color == color)
                 && (name == null || a.Name == name))
        .ToList();
```

That caused a hard error because a `null` value for `color` was causing the duplicated field handler to error out, despite the fact that it would never be evaluated against the column itself due to the OR evaluation.

### DIFF
--- a/src/Marten.Testing/Linq/Fields/DuplicatedFieldTests.cs
+++ b/src/Marten.Testing/Linq/Fields/DuplicatedFieldTests.cs
@@ -62,6 +62,25 @@ namespace Marten.Testing.Linq.Fields
             field.GetValueForCompiledQueryParameter(constant).ShouldBe(Colors.Blue.ToString());
         }
 
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


mysticmind (approved):
@eouw0o83hf  Looks good to me. Thank you for the fix. This PR is against master, we will also cherry pick this for 3.x.

mysticmind:
@jeremydmiller is okay with this, merged the same.

Issue #1582, opened 2020-10-19
Strongly Typed Ids


markjerz:
It would be nice if Marten was able to use strongly typed ids for each document (and the references to them). For example, a POCO class Blog would have an id of type BlogId:

```
public class Blog {
    public BlogId Id { get; set; }

    public string Title { get; set; }
}

public struct BlogId {
    public Guid Id { get; }

    public BlogId(Guid id) {
        this.Id = id;
    }
}
```

For background, strongly typed Ids are widely used in software following DDD practices. There is also a library [here](https://github.com/andrewlock/StronglyTypedId) to implement them, and the author of that library has also written a [series of blog posts](https://andrewlock.net/using-strongly-typed-entity-ids-to-avoid-primitive-obsession-part-1/) on them as well.

I would be very happy to implement the code for supporting these. I have started doing some work but wanted to open up the issue before I got too far in to it ....

# Solution so far

Supporting the strongly typed id for the id column in the database has been fairly simple so far. I have made use of the npgsql type mapping in order to, for example, map the above BlogId type to the underlying uuid postgres type. The code below (which is WIP) hopefully shows how that could work.

At the moment the set up of this handler is done manually but it would be fairly simple to dynamically add these at run time when a DocumentMapping is generated.

The only other thing that I can see as being a problem is that we should also override the Json (de)serialization so that these custom ids are serialized using the underlying primitive (note that the repo above already provides these). That way, LINQ queries `.Where(x => x.Id == new BlogId())` would work with minimal LINQ provider re-work?

As I've not used Marten in anger (barely at all actually!) can anyone see any other potential issues for implementing this functionality?

Is a solution along these lines appropriate?

Thanks,
Mark

```
// TODO make this non guid specific
    public class WrappedGuidHandler : NpgsqlTypeHandler<Guid>
    {
        private UuidHandler uuidHandler;

        public WrappedGuidHandler(PostgresType postgresType) : base(postgresType)
        {
            this.uuidHandler = new UuidHandler(postgresType);
        }

        public override async ValueTask<Guid> Read(NpgsqlReadBuffer buf, int len, bool async,
            FieldDescription? fieldDescription = null)
        {
            await buf.Ensure(len, async);
            return this.uuidHandler.Read(buf, len, fieldDescription);
        }

        protected override async ValueTask<TAny> Read<TAny>(NpgsqlReadBuffer buf, int len, bool async, FieldDescription? fieldDescription = null)
        {
[...]

PR #1587, opened 2020-10-22
Strongly typed custom ids

markjerz:
Enables the use of strongly type ids (see https://andrewlock.net/using-strongly-typed-entity-ids-to-avoid-primitive-obsession-part-1/ for details).

Provides npgsql type handlers for conversion to underlying primitive type as well as json converters for document serialization to primitive type.

Strongly typed ids are registered through various StoreOptions extension methods named UseStronglyTypedId(s)

For understanding what the calling code looks like see src/Marten.Schema.Testing/Identity/StronglyTyped/load_and_store_strongly_typed_ids.cs Hopefully everything flows from there.

Accompanying issue is here: https://github.com/JasperFx/marten/issues/1582

As a first time contributor, and someone who's not actually used marten in anger (although I want to!), I imagine I've missed a bunch of use-cases. So please let me know of anything that you might be aware of as a possible issue and I'll be sure to add extra tests/code. I was pleasantly surprised at how little change I actually had to make to the core library - well done!

At the moment, there's no documentation. Feels like something that can wait until the PR is in a decent state?

### DIFF
--- a/src/Marten.Schema.Testing/Identity/StronglyTyped/DefaultPrimitiveTypeFinderTests.cs
+++ b/src/Marten.Schema.Testing/Identity/StronglyTyped/DefaultPrimitiveTypeFinderTests.cs
@@ -1,0 +1,55 @@
+using Marten.Schema.Identity.StronglyTyped;
+using System;
+using Xunit;
+
+namespace Marten.Schema.Testing.Identity.StronglyTyped
+{
+    public class DefaultPrimitiveTypeFinderTests
+    {
+        [Fact]
+        public void SingleParamConstructor()
+        {
+            var type = DefaultPrimitiveTypeFinder.FindPrimitiveType(typeof(FieldWrapper));
+            Assert.Equal(typeof(Guid), type);
+        }
+
+        [Fact]
+        public void DefaultConstructor()
+        {
+            var type = DefaultPrimitiveTypeFinder.FindPrimitiveType(typeof(DefaultWrapper));
+            Assert.Equal(typeof(Guid), type);
+        }
+
+        [Fact]
+        public void SingleParamPropConstructor()
+        {
+            var type = DefaultPrimitiveTypeFinder.FindPrimitiveType(typeof(PropertyWrapper));
+            Assert.Equal(typeof(Guid), type);
+        }
+
+        public class FieldWrapper
+        {
+            private Guid id;
+
+            public FieldWrapper(Guid id)
+            {
+                this.id = id;
+            }
+        }
+
+        public class PropertyWrapper
+        {
+            public Guid Id { get; }
+
+            public PropertyWrapper(Guid id)
+            {
+                this.Id = id;
+            }
+        }
+
+        public class DefaultWrapper
+        {
+            public Guid Id { get; set; }
+        }
+    }
+}
--- a/src/Marten.Schema.Testing/Identity/StronglyTyped/Model/Blog.cs
+++ b/src/Marten.Schema.Testing/Identity/StronglyTyped/Model/Blog.cs
@@ -1,0 +1,15 @@
+namespace Marten.Schema.Testing.Identity.StronglyTyped.Model
+{
+    public class Blog
+    {
+        public Blog(BlogId id, string name)
+        {
+            this.Id = id;
+            this.Name = name;
+        }
+
+        public BlogId Id { get; }
+
+        public string Name { get; }
+    }
+}
--- a/src/Marten.Schema.Testing/Identity/StronglyTyped/Model/BlogId.cs
+++ b/src/Marten.Schema.Testing/Identity/StronglyTyped/Model/BlogId.cs
@@ -1,0 +1,38 @@
+using System;
+
+namespace Marten.Schema.Testing.Identity.StronglyTyped.Model
+{
+    public struct BlogId
+    {
+        public Guid Id { get; }
+
+        public BlogId(Guid id)
+        {
+            Id = id;
+        }
+        public bool Equals(BlogId other)
+        {
+            return Id.Equals(other.Id);
+        }
+
+        public override bool Equals(object obj)
+        {
+            return obj is BlogId other && Equals(other);
+        }
+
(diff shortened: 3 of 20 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1587) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1587) before we can accept your contribution.<br/><br/>:x: markjerz [sign now](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1587)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/JasperFx/marten?pullRequest=1587) it.</sub>

mysticmind (commented):
Sorry about the delay in reviewing this. First of all, thank you very much for taking the time to send us this PR. Good work, overall it looks fine to me from a conceptual standpoint.

Few comments/thoughts:
- The implementation assumes that the `Id` will be the property/field name of the id within the Strongly typed id class (which is okay, by and large) and the document within which the strongly typed id is used. When you use the Strongly typed id as part of a document `class Blog { public BlogId Id { get; } }`, Marten supports to use a different identity name by using `[Identity]` attribute or the fluent interface `storeOptions.Schema.For<Blog>().Identity(x => x.CustomId);`. I think if the implementation can cater for also marking an `Identity` or the fluent interface for both the strongly type id class and the document, it will make the functionality more feature complete. Please check https://martendb.io/documentation/documents/identity/ for more details.

- In terms of store options setup, rather than `storeOptions.UseStronglyTypedId<PostId>();`, it would be intuitive to use `storeOptions.Schema.For<Blog>.StronglyTypedId<PostId>(x => x.Id);`. In addition we can add `StronglyTypeIdentityAttribute`. So both fluent interface and attribute based setup is possible.

- Please check util class `ReflectionExtensions.cs` for the code pertaining to getting property, fields etc and use it wherever possible. If something is not available, possibly it is a candidate which you can add to this util class.

You have added a good amount of unit tests coverage which is a big thumps up from me! 🏅 

Please consider the above suggestions/review comments. I will await for other reviewers to complete their reviews and then this will shape up better for next steps. 👍

markjerz:
Hi, thanks.

- I hope that the assumption around "id" is only in the DefaultPrimitiveTypeFinder. I don't think I've assumed that for the document and was assuming you could override the class behaviour by providing your own PrimitiveTypeFinder. However, at the moment, that probably involves too much effort than supporting it better out of the box. I'll take another look at making that easier to configure.

- I think the reason I wanted to do it this way was so that you could indicate all of your strongly typed id classes through reflection or something so that you wouldn't have to explicitly set up all of the types. That way, your basically just adding to the list of types that are allowed to be Ids e.g.
```
storeOptions.UseStronglyTypeIds(typeof(BlogId).Assembly.GetTypes().Where(t => <something>));
```
Maybe it should be renamed to e.g. `storeOptions.AddStronglyTypedIdTypes()` (or something with the word Type a few less times! - any suggestion?)

- Sure. It's always tricky finding the right stuff in a large codebase!

jeremydmiller:
@markjerz Did you consider depending on implicit casts on the strong typed ids instead? That might be a simpler, less intrusive approach. If that worked out, it would streamline the API signatures.

markjerz:
@jeremydmiller I had but some reason assumed it wouldn't work with the LoadMany variants. However, it appears to so, yes, seems like a legitimate thing to investigate. Mandating implicit operator overrides would make it less generic but given what we're talking about I think that's actually a sensible requirement. The primitiveIdAccessor already attempts to do a cast. It would certainly slim down the api.

I'll take a look....

mysticmind:
Adding this below solution for clarity around implicit casts for someone stumbling upon a similar problem. Read further details [here](https://andrewlock.net/using-strongly-typed-entity-ids-to-avoid-primitive-obsession-part-3/)
```
public readonly struct BlogId
{
    public static implicit operator Guid(BlogId blogId) => blogId.Value;
    ....
    ....
}

```

mysticmind:
@markjerz Given where we stand on this with an alternative non-intrusive solution, can we please close this PR and also the issue?

markjerz:
@mysticmind I don’t think what you are proposing as a workaround actually works though? You still need to make changes to the library or am I missing something? Just for starters you only allow IDs of type guid/int/long/string so the document mapping fails. 

Shall I keep going in that direction until I’ve got something functional for review?

jeremydmiller:
@markjerz I think it would take *some* changes to the way that Marten validates Id types to accommodate the strong typed ids, yes. But just to not waste any more of your time, I'm probably going to be opposed to bringing in the bigger set of changes. There's still a *lot* of other functionality that might be impacted by the strong typed ids.

markjerz:
Ok, that's a shame. 

I had been looking for a long time to find something capable of actually supporting a proper DDD model with full persistence ignorance and thought I had found it (minus the strongly typed IDs).

oskardudycz:
From my perspective - I'm open for doing the strongly typed IDs. It's a common pattern in DDD community. If @jeremydmiller and @babuannamalai agree I think that we could continue the discussion to find the suitable way. I could try to "fight" with our identity map.

cocowalla:
This is something that is clearly useful to strict DDD'ers, not so much for everyone else. I wonder if there might be a suitable extensibility point such that this functionality could be provided as a separate/community package? (I realise identity is central, and Marten likely makes a lot of assumptions about it, so it might be unrealistic; just thinking out loud 😄)

jeremydmiller:
@cocowalla I think it would take at least some mild changes to the library no matter what. 

@oskardudycz Just try the implicit cast first and see how that goes. But I'm still opposed to making bigger structural changes to accommodate a feature that I don't believe will be used to any significant degree. I think the bigger changes are somewhat likely to cause quite a bit of support overhead for us over time.

PR #1592, opened 2020-10-31
Remove NET461 leftovers and multi-target net5

Hawxy:
There's a bunch of stuff in here:

- Drop remaining net461 support items and cleanup csprojs
- Fix tests that fail if you live east of UTC
- Multi-target net5 to enable new roslyn analyzers and let contributors poke a stick at C# 9 features
- Multi-target test projects to cover range of supported .NET Core versions
- Port the .NET 5 reflection fix from 3.x
- Add a note to docs regarding compatibility
- Update dependencies across the board (fixes some weirdness with the VS test runner)


lahma:
Just wanted to say thank you for your work and especially,

> Fix tests that fail if you live east of UTC

Cheers from Finland 🚀

src/Marten.Testing/Acceptance/document_inserts.cs
```diff
@@ -51,6 +52,26 @@ public void can_insert_a_mixed_bag_of_documents()
                 query.Query<User>().Count().ShouldBe(4);
             }
         }
+#if NET
```
mysticmind, line 55: addictive aspiration 🥇

README.md
```diff
@@ -18,15 +18,11 @@ Marten also provides .NET developers with an ACID-compliant event store with use
 
 Before getting started you will need the following in your environment:
 
-**1. .NET Core SDK 3.1 (or higher)**
+**1. .NET Core SDK 5.0 (or higher)**
```
mysticmind, line 21: Should this include .NET Core SDK 3.1 as well?

mysticmind (approved):
Looks good to me 👍 . Few minor comments, please have a look.

README.md
Hawxy: Not required, .NET Core SDKs are backwards compatible with lower targets.

Hawxy: I think the only exception to that would be for pointing the tests at the specific runtime version.

mysticmind: we can leave it as `.NET Core SDK 3.1 (or higher)`, this also means that it includes net5. thoughts?

Hawxy: I adjusted the wording to be more accurate

oskardudycz (approved):
Great stuff @Hawxy! Once again thank you for keeping an eye on the .NET 5 compatibility - really appreciated! 🙇

jeremydmiller (approved):
Looks good to me. Good catch w/ the Datetime to DateTimeOffset conversions in the tests

Issue #1594, opened 2020-11-12
BulkInsert performance hit


toanxyz:
https://github.com/JasperFx/marten/blob/4e36f49f8470a9cfc414a804b044f44f04bc9ef8/src/Marten/Internal/CodeGeneration/BulkLoaderBuilder.cs#L100

`create temporary table {_tempTable} as select * from {_mapping.TableName.QualifiedName};truncate table {_tempTable}`
-> From what I'm understanding, the 1st statement will create a temp table with existing data from the source table and truncate it at the 2nd step. The purpose is to create a temp table with source table schema right?

I suggest we can use `create temporary table {_tempTable} as select * from {_mapping.TableName.QualifiedName} limit 0` to create a temp table with source table schema? "limit 0" will avoid the temp table with existing data from source table.

We're facing the performance hit when multiple requests are using BulkInsert, the source table is huge and the process of create temporary table above is killing the database.

mysticmind:
@toanxyz good catch on the issue and looks to be a definitely a performance hit since all data gets copied to the temp table. We will look at applying your proposed suggestion.

@jeremydmiller thoughts?

jeremydmiller:
@mysticmind @toanxyz Yeah, let's take this in ASAP. Good catch. 🤦

toanxyz:
@mysticmind @jeremydmiller  I have created a PR for this https://github.com/JasperFx/marten/pull/1596 to contribute to the project.

mysticmind:
Fixed via https://github.com/JasperFx/marten/commit/05bf0ae469ef186195c4578e4de86f047ce56b35

mysticmind:
Quick note, this fix also aapplied to 2.x (2.10.3 released) and 3.x branches (we will release the fix in this next bug fix release 3.13.3).

PR #1597, opened 2020-11-13
Add limit 0 to create a temp table with schema only

toanxyz:
Fix for issue #1594 in 2.0 branch


toanxyz:
@jeremydmiller  could we apply this to 2.0 branch and build a 2.x version, our team is using 2.7.1 and upgrading to 4.x is quite challenging right now. We're in prod mode so if would be great if we can have this fix in 2.x branch -> from 2.10.2 to 2.10.3

mysticmind:
@toanxyz I will take a look at your request and revert. We have not touched the 2.x codebase in a while.

mysticmind:
@toanxyz v2.10.3 is published now and available in NuGet.

PR #1600, opened 2020-11-16
Update to use Npgsql 5.0.0

mysticmind:
- Update `Npgsql`, `Npgsql.Json.NET`, `Npgsql.NodaTime` to v5.0.0
- Fix and handle a Npgsql breaking change of `NpgsqlTransaction.IsCompleted` not
available in public API. Refer to https://github.com/npgsql/npgsql/blob/d5aee643792fad71b9220d06c16a3612ccc10e0d/src/Npgsql/NpgsqlTransaction.cs#L412 based on which the fix is added
- Update `Newtonsoft.Json` to  12.0.3 (12.0.2 is the minimum requirement for
`Npgsql.Json.NET 5.0.0`)

### DIFF
--- a/src/Marten.NodaTime/Marten.NodaTime.csproj
+++ b/src/Marten.NodaTime/Marten.NodaTime.csproj
@@ -37,7 +37,7 @@
     </PropertyGroup>
     <ItemGroup>
         <PackageReference Include="NodaTime.Serialization.JsonNet" Version="2.3.0" />
-        <PackageReference Include="Npgsql.NodaTime" Version="[4.1.5,4.2.0)" />
+        <PackageReference Include="Npgsql.NodaTime" Version="5.0.0" />
         <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
     </ItemGroup>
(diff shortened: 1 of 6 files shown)


jeremydmiller (approved):
I'm surprised now that the empty command text thing worked at all:)

Looks good to me.

kdlslyv:
Is it possible to release with this?
Would be awesome because of the peer dependency on npgsql 5 regarding ef core 5 😄

oskardudycz:
@kdlslyv we can release prerelease v4 version. If you're using the v3 the I'm not sure if we'll be able to provide the support there.

kdlslyv:
@oskardudycz thanks for the quick reply - yes I'd happily start working with the v4 preview :)

oskardudycz:
@kdlslyv thanks to @mysticmind you should have it in https://www.nuget.org/packages/Marten/4.0.0-alpha.3 👍

kdlslyv:
yup, it has been working splendid today 😎 👍
thank you @mysticmind

oskardudycz:
@kdlslyv thank you for the feedback. If you find something - feel free to send us the info 👍

PR #1604, opened 2020-11-17
Remove type mappers from nodatime implementation

Hawxy:
As this is now supported in Npgsql v5 + v4.1.6 out of the box, this should no longer be required.


mysticmind (approved):
@Hawxy Thank you, looks good to me 👍

oskardudycz (approved):
@Hawxy thank you! I love the perspective of removing this code - I wasn't proud of it. I'm glad that Npgsql team implemented this out of the box.

PR #1643, opened 2020-12-05
Remove some redundant code

SimonCropp:
i am not sure about this one, so it needs a proper review.


SimonCropp:
give v4 is beta, and there is unlikely to be another release of 3, should i change `removed in version 4` to be `removed in version 5`?

PR #1649, opened 2020-12-06
pass cancellation tokens

SimonCropp:
I am assuming cancellation tokens should be passes in these scenarios. 

If not, it might be work adding a comment as to why (which i am happy to do)


mysticmind (approved):
I do think that the cancellation tokens need to be passed for the code blocks in question. Looks good to me.

oskardudycz (approved):

oskardudycz:
@SimonCropp I rebased the changes - it seems that 2 files were removed in the meantime by @jeremydmiller ;) I'll merge it when pipelines finish.

Thank you for again catching those small issues. Highly appreciated.

PR #1652, opened 2020-12-06
check for ViewProjection.projectEvent.handler null when UpdateOnly

SimonCropp:
will prevent a null ref when UpdateOnly and handler=null

given handler is assumed to not be null when ProjectionEventType.UpdateOnly. then null check should be inside the 

    type == ProjectionEventType.CreateAndUpdate || type == ProjectionEventType.UpdateOnly

also better perf by removing a redundant check if a different ProjectionEventType

### DIFF
--- a/src/Marten/Events/Projections/ViewProjection.cs
+++ b/src/Marten/Events/Projections/ViewProjection.cs
@@ -680,12 +680,12 @@ namespace Marten.Events.Projections
         {
             if (viewIdSelector == null && viewIdsSelector == null)
                 throw new ArgumentException($"{nameof(viewIdSelector)} or {nameof(viewIdsSelector)} must be provided.");
-            if (handler == null && type == ProjectionEventType.CreateAndUpdate)
-                throw new ArgumentNullException(nameof(handler));
 
             EventHandler eventHandler;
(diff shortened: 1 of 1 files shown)


src/Marten/Events/Projections/ViewProjection.cs
```diff
@@ -680,12 +680,12 @@ public enum ProjectionEventType
         {
             if (viewIdSelector == null && viewIdsSelector == null)
                 throw new ArgumentException($"{nameof(viewIdSelector)} or {nameof(viewIdsSelector)} must be provided.");
-            if (handler == null && type == ProjectionEventType.CreateAndUpdate)
```
mysticmind, line 683: @SimonCropp existing code is actually throwing an `ArgumentNullException` for the case of `handler` being `null` and type being `ProjectionEventType.CreateAndUpdate`. But the code changes looks to throw an exception for `handler` being `null` and for `type` being `ProjectionEventType.CreateAndUpdate` or `type == ProjectionEventType.UpdateOnly`. This changes the existing behaviour event though unit tests are green. Thoughts?

mysticmind (approved):
You are right, the handler is assumed to be not null looking at the code block. Looks good to me 👍

oskardudycz (approved):

PR #1654, opened 2020-12-06
use First when result is assumed to exist

SimonCropp:
in these scenarios First is more correct than FirstOrDefault since the result is assumed not to be null


jeremydmiller:
One test was off, but otherwise good. I pulled this in via the CLI with a rebase

PR #1655, opened 2020-12-06
Updated code generation to set `Marten.Generated` as the assembly name

oskardudycz:


oskardudycz:
Ugh - https://github.com/dotnet/runtime/issues/12454. It seems that `InternalsVisibleTo` won't work up until .NET 6...

oskardudycz:
Closing that as it seems that it won't be fixable until .NET 6...

PR #1657, opened 2020-12-08
move to DateTimeOffset

### LINKED ISSUE #1656: Replace all public DateTime properties with DateTimeOffset in the public API
For example, `StreamState` has these:

```
        /// <summary>
        /// The last time this stream was appended to
        /// </summary>
        DateTime LastTimestamp { get; }

        /// <summary>
        /// When this stream was originally created
        /// </summary>
        DateTime Created { get; }
```

Probably be better to make those DateTimeOffset. Does make us need to watch the Postgresql data types on columns though.

SimonCropp:
fixes #1656

### DIFF
--- a/src/Marten/Events/Projections/ViewProjection.cs
+++ b/src/Marten/Events/Projections/ViewProjection.cs
@@ -923,14 +923,14 @@ namespace Marten.Events.Projections
     {
         public Guid Id { get; protected set; }
         public int Version { get; protected set; }
-        public DateTime Timestamp { get; protected set; }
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@SimonCropp,

This one's gonna need a little more love. We'll need to switch up the underlying database schema to use a different date time type, and get the database migration stuff to be able to support that. I'm putting this one off because that's unpleasant work:)

Hey, when you're into all this code, be aware that `ViewProjection` is getting completely gutted and rebuilt for V4. I just haven't started that yet, pending some other conversations w/ @oskardudycz et al.

PR #1667, opened 2020-12-16
use c# discards

SimonCropp:
> Because there is only a single discard variable, and that variable may not even be allocated storage, discards can reduce memory allocations

https://docs.microsoft.com/en-us/dotnet/csharp/discards


oskardudycz (approved):
Looks legit 👍

PR #1685, opened 2020-12-27
System.Text.Json support

oskardudycz:
Added 
- System.Text.Json support,
- test matrix configuration for different serializer,
- updated ISerializer public API to allow passing Stream and DBDataReader

### DIFF
--- a/src/Marten.Storyteller/Fixtures/EventStore/InlineAggregationFixture.cs
+++ b/src/Marten.Storyteller/Fixtures/EventStore/InlineAggregationFixture.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using Baseline;
 using Marten.Testing;
@@ -74,7 +75,7 @@ namespace Marten.Storyteller.Fixtures.EventStore
         }
 
         [FormatAs("Members should be {Members}")]
-        public string[] Members()
+        public List<string> Members()
         {
             return CurrentObject.As<QuestParty>().Members;
         }
--- a/src/Marten.Testing/Documents/Target.cs
+++ b/src/Marten.Testing/Documents/Target.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.Json.Serialization;
 using Baseline;
 
 namespace Marten.Testing.Documents
@@ -119,6 +120,7 @@ namespace Marten.Testing.Documents
 
         public bool Flag { get; set; }
 
+        [JsonInclude]
         public string StringField;
 
         public double Double { get; set; }
@@ -126,7 +128,7 @@ namespace Marten.Testing.Documents
         public DateTime Date { get; set; }
         public DateTimeOffset DateOffset { get; set; }
 
-        public float Float;
+        public float Float { get; set; }
 
         public int[] NumberArray { get; set; }
 
--- a/src/Marten.Testing/Documents/Target.cs
+++ b/src/Marten.Testing/Documents/Target.cs
@@ -120,7 +120,7 @@ namespace Marten.Testing.Documents
 
         public bool Flag { get; set; }
 
-        [JsonInclude]
+        [JsonInclude] // this is needed to make System.Text.Json happy
         public string StringField;
 
         public double Double { get; set; }
@@ -128,7 +128,8 @@ namespace Marten.Testing.Documents
         public DateTime Date { get; set; }
         public DateTimeOffset DateOffset { get; set; }
 
-        public float Float { get; set; }
+        [JsonInclude] // this is needed to make System.Text.Json happy
+        public float Float;
 
         public int[] NumberArray { get; set; }
 
--- a/src/Marten.Testing/Events/Projections/QuestMonsters.cs
+++ b/src/Marten.Testing/Events/Projections/QuestMonsters.cs
@@ -13,28 +13,18 @@ namespace Marten.Testing.Events.Projections
     public interface IMonstersView
     {
         Guid Id { get; }
-        string[] Monsters { get; }
+        List<string> Monsters { get; }
     }
 
     public class QuestMonsters
     {
         public Guid Id { get; set; }
 
-        private readonly IList<string> _monsters = new List<string>();
+        public List<string> Monsters { get; set; } = new();
 
         public void Apply(MonsterSlayed slayed)
         {
-            _monsters.Fill(slayed.Name);
-        }
-
-        public string[] Monsters
-        {
-            get { return _monsters.ToArray(); }
-            set
-            {
-                _monsters.Clear();
-                _monsters.AddRange(value);
-            }
+            Monsters.Fill(slayed.Name);
         }
     }
 
@@ -42,21 +32,11 @@ namespace Marten.Testing.Events.Projections
     {
         public Guid Id { get; private set; }
 
-        private readonly IList<string> _monsters = new List<string>();
+        public List<string> Monsters { get; set; } = new();
 
         public void Apply(MonsterSlayed slayed)
         {
-            _monsters.Fill(slayed.Name);
-        }
-
-        public string[] Monsters
-        {
-            get { return _monsters.ToArray(); }
-            set
-            {
-                _monsters.Clear();
-                _monsters.AddRange(value);
-            }
+            Monsters.Fill(slayed.Name);
         }
     }
 
@@ -64,41 +44,21 @@ namespace Marten.Testing.Events.Projections
     {
         public Guid Id { get; protected set; }
 
-        private readonly IList<string> _monsters = new List<string>();
+        public List<string> Monsters { get; set; } = new();
 
         public void Apply(MonsterSlayed slayed)
         {
-            _monsters.Fill(slayed.Name);
-        }
-
-        public string[] Monsters
-        {
-            get { return _monsters.ToArray(); }
-            set
-            {
-                _monsters.Clear();
-                _monsters.AddRange(value);
-            }
+            Monsters.Fill(slayed.Name);
         }
     }
 
     public class QuestMonstersWithBaseClass: Root, IMonstersView
     {
-        private readonly IList<string> _monsters = new List<string>();
+        public List<string> Monsters { get; set; } = new();
 
         public void Apply(MonsterSlayed slayed)
         {
-            _monsters.Fill(slayed.Name);
-        }
-
-        public string[] Monsters
-        {
-            get { return _monsters.ToArray(); }
-            set
-            {
-                _monsters.Clear();
-                _monsters.AddRange(value);
-            }
+            Monsters.Fill(slayed.Name);
         }
     }
 
@@ -106,21 +66,11 @@ namespace Marten.Testing.Events.Projections
     {
         public new Guid Id { get; set; }
 
-        private readonly IList<string> _monsters = new List<string>();
+        public List<string> Monsters { get; set; } = new();
 
         public void Apply(MonsterSlayed slayed)
         {
-            _monsters.Fill(slayed.Name);
-        }
-
-        public string[] Monsters
-        {
-            get { return _monsters.ToArray(); }
-            set
-            {
-                _monsters.Clear();
-                _monsters.AddRange(value);
-            }
+            Monsters.Fill(slayed.Name);
         }
     }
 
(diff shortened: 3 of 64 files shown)


src/Marten.Testing/Documents/Target.cs
```diff
@@ -119,14 +120,15 @@ public Target()
...
+        [JsonInclude]
         public string StringField;
 
         public double Double { get; set; }
         public decimal Decimal { get; set; }
         public DateTime Date { get; set; }
         public DateTimeOffset DateOffset { get; set; }
 
-        public float Float;
+        public float Float { get; set; }
```
jeremydmiller, line 131: There's purposely a mix of properties and public fields to ensure that Marten works with both. If we do this, we just need to make sure there are tests specifically for Newtonsoft against public fields. So, ability to use a field as an identity and Linq querying.

Annoying, I know.

src/Marten.Storyteller/Fixtures/EventStore/InlineAggregationFixture.cs
```diff
@@ -74,7 +75,7 @@ public IGrammar QuestPartyShouldBe()
         }
 
         [FormatAs("Members should be {Members}")]
-        public string[] Members()
+        public List<string> Members()
```
jeremydmiller, line 78: Why did you have to do this? This wouldn't be serialized anywhere.

src/Marten.Testing/Events/Projections/QuestMonsters.cs
```diff
@@ -13,150 +13,88 @@ public class Root
     public interface IMonstersView
     {
         Guid Id { get; }
-        string[] Monsters { get; }
+        List<string> Monsters { get; }
```
jeremydmiller, line 16: STJ doesn't support arrays?

src/Marten.Testing/performance_tuning.cs
```diff
@@ -18,24 +13,36 @@ public class JilSerializer : ISerializer
...
 
-        public T FromJson<T>(TextReader reader)
+        public T FromJson<T>(Stream stream)
         {
-            return JSON.Deserialize<T>(reader, _options);
+            return JSON.Deserialize<T>(new StreamReader(stream), _options);
         }
 
-        public object FromJson(Type type, TextReader reader)
+        public Task<T> FromJsonAsync<T>(Stream stream, CancellationToken cancellationToken = default)
```
jeremydmiller, line 33: We should change the signature of this method to return a ValueTask

src/Marten.Testing/performance_tuning.cs
```diff
@@ -18,24 +13,36 @@ public class JilSerializer : ISerializer
...
+        }
+
+        public object FromJson(Type type, Stream stream)
+        {
+            return JSON.Deserialize(new StreamReader(stream), type, _options);
+        }
+
+        public Task<object> FromJsonAsync(Type type, Stream stream, CancellationToken cancellationToken = default)
+        {
+            return Task.FromResult(FromJson(type, stream));
```
jeremydmiller, line 45: We should probably be using a ValueTask

src/Marten/Linq/Selectors/SerializationSelector.cs
```diff
@@ -15,14 +15,14 @@ public SerializationSelector(ISerializer serializer)
 
         public T Resolve(DbDataReader reader)
         {
-            using var json = reader.GetTextReader(0);
+            using var json = reader.GetStream(0);
             return _serializer.FromJson<T>(json);
         }
 
-        public Task<T> ResolveAsync(DbDataReader reader, CancellationToken token)
+        public async Task<T> ResolveAsync(DbDataReader reader, CancellationToken token)
```
jeremydmiller, line 22: And let's look to see about using ValueTask. Might also want the actual reading from the Stream to be async. We should look at how the ASP.net JsonFormatter uses STJ and see if it buffers data from the Stream or anything like that. Not saying this is necessarily wrong, but we should do our homework.

src/Marten/Linq/SqlGeneration/JsonSelectClause.cs
```diff
@@ -103,15 +104,16 @@ public string Handle(DbDataReader reader, IMartenSession session)
...
+                builder.Write(text.GetStreamReader().ReadToEnd());
             }
 
             while (reader.Read())
             {
-                using var text = reader.GetTextReader(0);
+                using var text = reader.GetStream(0);
                 builder.Write(',');
-                builder.Write(text.ReadToEnd());
+                builder.Write(text.GetStreamReader().ReadToEnd());
```
jeremydmiller, line 116: Let's reevaluate this one to see if there's a way to not have to read and create the string. See if there's a way to pipe the Stream bytes directly to the CommandBuilder

src/Marten/Marten.csproj
```diff
@@ -31,6 +31,7 @@
         <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.11.1" />
         <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
         <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="[2.1.0, 6.0.0]" />
+        <PackageReference Include="System.Text.Json" Version="5.0.0" />
```
jeremydmiller, line 34: We'll make this an add on lib, right? Not be in the box?

src/Marten/Services/CommandRunnerExtensions.cs
```diff
@@ -29,7 +29,7 @@ public static QueryPlan ExplainQuery(this IManagedConnection runner, NpgsqlComma
             cmd.CommandText = string.Concat($"explain ({config} format json) ", cmd.CommandText);
             using (var reader = runner.ExecuteReader(cmd))
             {
-                var queryPlans = reader.Read() ? serializer.FromJson<QueryPlanContainer[]>(reader.GetTextReader(0)) : null;
+                var queryPlans = reader.Read() ? serializer.FromJson<QueryPlanContainer[]>(reader.GetStream(0)) : null;
```
jeremydmiller, line 32: One thing TODO here might be to make the ISerializer interface take in the DbDataReader and index, then let each serializer to whatever they need to do. So you don't have to make the serializers conform to something that isn't idea.

src/Marten/Services/JsonNetSerializer.cs
```diff
@@ -13,24 +16,22 @@ public class JsonNetSerializer: ISerializer
         private readonly ArrayPool<char> _charPool = ArrayPool<char>.Create();
         private readonly JsonArrayPool<char> _jsonArrayPool;
 
-
-
-        private readonly JsonSerializer _clean = new JsonSerializer
+        private readonly JsonSerializer _clean = new()
```
jeremydmiller, line 19: I'm not sure we use the containment operator anymore, so this might no longer be needed. Not 100% sure about that though.

src/Marten/Services/JsonNetSerializer.cs
```diff
@@ -61,54 +62,50 @@ public void Customize(Action<JsonSerializer> configure)
...
 
-            writer.Flush();
+            jsonWriter.Flush();
         }
 
         public string ToJson(object document)
         {
-            var writer = new StringWriter();
-            ToJson(document, writer);
+            using var stream = new MemoryStream();
```
jeremydmiller, line 81: Why are we having to write to a stream like this?

src/Marten/Services/JsonNetSerializer.cs
```diff
@@ -61,54 +62,50 @@ public void Customize(Action<JsonSerializer> configure)
...
                 ArrayPool = _jsonArrayPool,
-                CloseInput = false
+                CloseInput = false,
             };
 
             return _serializer.Deserialize<T>(jsonReader);
         }
 
-        public T FromJson<T>(TextReader reader)
+        public Task<T> FromJsonAsync<T>(Stream stream, CancellationToken cancellationToken = default)
```
jeremydmiller, line 101: We need to be using ValueTask.

jeremydmiller:
That looks like a ton of work.

With the `ISerializer` interface, we might be better off having it take in `DbDataReader` and an index, then letting each serializer do what works best for it, rather than trying to make an `ISerializer` conforming interface that isn't ideal. And we've gotta switch to `ValueTask` instead of returning `Task.FromResult()` in so many places.

src/Marten/Services/JsonNetSerializer.cs
oskardudycz: Updated to ValueTask in https://github.com/JasperFx/marten/pull/1685/commits/1ef5e80e9875fa88b5cd902e75eca143937d6488. We should probably review if there are no places in our API that should return ValueTask (however, that's out of scope of this PR).

src/Marten.Testing/performance_tuning.cs
oskardudycz: Used ValueTask in https://github.com/JasperFx/marten/pull/1685/commits/1ef5e80e9875fa88b5cd902e75eca143937d6488.

src/Marten.Testing/performance_tuning.cs
oskardudycz: Used ValueTask in https://github.com/JasperFx/marten/pull/1685/commits/1ef5e80e9875fa88b5cd902e75eca143937d6488.

src/Marten/Services/CommandRunnerExtensions.cs
oskardudycz: Refactored in https://github.com/JasperFx/marten/pull/1685/commits/be80fc6c0b93a4b901f3984ac1e8b9df31b94789.

src/Marten/Marten.csproj
oskardudycz: This we need to discuss - in .NET 5 it's already built-in. I added NuGet, as built-in version in .NET Core 3.1 is too old and missing features. We can distribute that as separate package, but then maintainance might be harder as we'll have growing start of packages (NodaTime, Console, etc.).

src/Marten/Services/JsonNetSerializer.cs
oskardudycz: I tried to use `ToJson` instead of `ToCleanJson` but there were set of failing tests around querying interface types collection.

src/Marten/Services/JsonNetSerializer.cs
oskardudycz: I reverted those changes and brought back old implementation after changing method signature with DBReader etc.

src/Marten/Linq/Selectors/SerializationSelector.cs
oskardudycz: If it's the right one - then it looks like is doing that the same way as we do: https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonOutputFormatter.cs#L77 (plus they're doing additional serialisation from object that we don't want to do).

oskardudycz: I think that we should add method `ToStream` and `ToStreamAsync` to `IQueryable` etc. Plus some custom output formatter that'll just return stream with proper encoding.

src/Marten.Testing/Documents/Target.cs
oskardudycz: System.Text.Json requires explicit attribute on fields. I reverted the change and added attribute with a comment informing that's only needed for it.

src/Marten.Storyteller/Fixtures/EventStore/InlineAggregationFixture.cs
oskardudycz: This was done to align other changes, but it's reverted.

src/Marten.Testing/Events/Projections/QuestMonsters.cs
oskardudycz: It wasn't supporting it properly when behind the Array was another type. It's fixed in the System.Text.Json 5, so I reverted the change here with the dependency added.

oskardudycz:
@jeremydmiller @mysticmind PR is ready for the final review.

src/Marten/Services/JsonNetSerializer.cs
```diff
@@ -95,20 +109,31 @@ public T FromJson<T>(Stream stream)
...
 
-            return _serializer.Deserialize<T>(jsonReader);
+            return _serializer.Deserialize(jsonReader, type);
         }
 
-        public object FromJson(Type type, TextReader reader)
+        public object FromJson(Type type, DbDataReader reader, int index)
         {
-            using var jsonReader = new JsonTextReader(reader)
+            using var textReader = reader.GetTextReader(index);
```
jeremydmiller, line 135: For later. If we took in the native Npgsql DbDataReader type instead, you get access to async versions of `GetTextReader()` and `GetStream()`

jeremydmiller (approved):
My vote is to pull this in now.

My only comment beyond "thank you for doing all of this and I'm shocked at how much work it really required" is that we need to think seriously about either passing around `NpgsqlReader` instead of `DbDataReader`, or downcast (ick) as appropriate to use the async `GetStreamAsync()` everywhere.

mysticmind (approved):
Looks good to me. Apart from all the changes required to make STJ work, I really liked that you have spruced up the code with C# 9 lang features 👍 

BTW, my changes pertaining to changing the code samples markers for migrating to docs to vitepress looks to have added conflicts in few files, will address it and the merge the same.

PR #1692, opened 2021-01-05
Projection Overhaul for V4

jeremydmiller:
## V4 Event Projection Model

This pull request completely replaces the supported projections in Marten. **Temporarily**, this pull request completely removes all support for the "async daemon", and a replacement should be the next work as a follow on.

The projection types are:

* Implement your own `IInlineProjection` class to do whatever you wanna do. This is simplified from the old `IProjection` model in V3
* "Self-Aggregates": Superset of the V3 ability to stick `Apply()` methods directly on the aggregate type. There are some other improvements like being able to use constructor functions or a static `Create(event, IQuerySession) : Self` method.
* `EventProjection` -- base class. Implement this to do whatever the heck you wanna do. Supports both a limited selection of inline lambda signatures, or a `Project(EventType @event, IDocumentOperations ops)` / `Transform(EventType) : DocumentType` conventional method set
* `AggregateProjection<T>` -- new base class for aggregating by stream. Inline lambdas + conventional method discovery for `Apply()`, `Create()`, and `ShouldDelete()`
* `ViewProjection<T, TId>` replacement. Inherits from `AggregateProjection<T>`. This differs a bit from the existing V3 `ViewProjection`, so we'll need some docs on migrating to V4. *I* think it has a much easier to use API.


mysticmind:
@jeremydmiller I was going through the PR changes earlier today, it was a bit overwhelming for me to run through it in the GH web view so was using Rider to run through it. Given that you have merged it now, I can run few tests etc and can learn the changes better.

PR #1695, opened 2021-01-07
Change event version from int->long

### LINKED ISSUE #1080: Use long data type for Event Version
For the next version of Marten, it would be good to upgrade the version for events to long instead of int. This will give streams with high event counts more room to grow

mysticmind:
Change event version from int->long
resolves #1080


mysticmind:
Quick note, for users migrating from 3.x to 4.x, we have to write a migration guide around the schema changes.

jeremydmiller:
@mysticmind That looks good from a code perspective, but I think we need something that can make the database migration as part of our normal diff detection. That's the only reason why I've put off doing this. I'd ask that we enhance the `Table` & `TableColumn` so that it can create the `ALTER TABLE` statement.

Same thing w/ removing the now obsolete `mt_append_events` function

oskardudycz (changes requested):
@mysticmind I agree with @jeremydmiller that we need to make sure that the migration will happen with alter table and no data will be lost.

Besides that, I think that you missed updates to:
- Append event function - e.g. https://github.com/JasperFx/marten/blob/master/src/Marten/Events/Schema/AppendEventFunction.cs#L36, https://github.com/JasperFx/marten/blob/master/src/Marten/Schema/SQL/mt_stream.sql#L35
- mt_events table - https://github.com/JasperFx/marten/blob/master/src/Marten/Schema/SQL/mt_stream.sql#L21
- mt_streams table - https://github.com/JasperFx/marten/blob/master/src/Marten/Schema/SQL/mt_stream.sql#L6

mysticmind:
@oskardudycz with regards to your review comments, those files under question anyway had to be removed. The recent commit will reflect the same. Only thing pending now is the migration which I am still looking at.

mysticmind:
Quick note though, there is one failing test only in PG9.6, I am yet to have a look.

PR #1696, opened 2021-01-07
Fix to close connection on rollback only if it owns the connection

### LINKED ISSUE #1620: Externally managed database connections are disposed by ManagedConnection
I am trying to manage some of my connections to the database externally in order to be able to manage the transaction lifecycle.

Basically I do something like this to setup the transaction:
```csharp
var npgsqlConnection = new NpgsqlConnection(ConnectionSource.ConnectionString);
await npgsqlConnection.OpenAsync();
var npgsqlTransaction = await npgsqlConnection.BeginTransactionAsync();
var sessionOptions = SessionOptions.ForTransaction(npgsqlTransaction);
```

Then I use a DocumentSession to update the database.

1. The DocumentSession is disposed, calling ManagedConnection.Dispose
2. This Disposes the underlying Connection, regardless if it was supplied externally or not.

I would suggest, that the ManagedConnection should only dispose the underlying connection, when OwnsConnection is true, changing the Dispose of ManagedConnection to

```csharp
public void Dispose()
{
    if (_ownsConnection)
         _connection?.Dispose();
}
```

I could prepare a PR if you are interested

mysticmind:
Fix to close connection on rollback only if it owns the connection
fixes #1620

### DIFF
--- a/src/Marten.Testing/Services/ManagedConnectionTests.cs
+++ b/src/Marten.Testing/Services/ManagedConnectionTests.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Data;
+using System.Threading;
 using System.Threading.Tasks;
 using Marten.Services;
 using Marten.Testing.Harness;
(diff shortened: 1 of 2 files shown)


mysticmind:
Quick note, we already fixed this in 4.x

oskardudycz (approved):
@babuannamalai Can we add test for that?

PR #1697, opened 2021-01-07
Fix to throw an exception for unique index without any properties on 3.x

### LINKED ISSUE #1581: Mapping with empty UniqueIndex() results in invalid DDL
Given a MartenRegistry:

```csharp
public class XRegistry : MartenRegistry
    {
      public XRegistry()
      {
        For<X>()
          .UniqueIndex();
      }
    }
```
UniqueIndex does not specify an expression. 

This creates the DDL

```sql
CREATE UNIQUE INDEX mt_doc_x_uidx_ ON public.mt_doc_x ();
```
The unique index would be created, but as no columns are specified it fails.

It would be nice to have an exception if you call `.UniqueIndex()` without providing any expressions.

mysticmind:
Fix to throw an exception for unique index without any properties
fixes #1581


oskardudycz (approved):
Looks good 👍

PR #1706, opened 2021-01-14
Fix issue 1703: symmetric equality

mkooiman:
This PR fixes:
https://github.com/JasperFx/marten/issues/1703

Also includes several tests to test for this issue

PR #1709, opened 2021-01-20
Bugfix/symmetric equality (for objects)

### LINKED ISSUE #1703: String equals in LINQ expression is not symmetric
In Marten linq Queries, it seems that queries are not always symmetric, 
if I query with:
`.Where( x => x.String.Equals("someString")`
I get the expected result, rewriting the query to:
`.Where( x => "someString".Equals(x.String)`
Results in a Marten.Exceptions.BadLinqExpressionException

When expanding the query to ignore casing:
`.Where( x => "someString".Equals(x.String, StringComparison.InvariantCultureIgnoreCase)`
This results in the query being executed but the result always being empty.

The cause seems to be the assumption that the expression is always in the format `<target.Member>.Equals(<constant>)`, in the case of the string comparison with casing ignored, the argument `InvariantCultureIgnoreCase` is assumed to be the constant to compare the member value to. 

using the equals operator ( == ) results in a query that is symmetric:
`.Where( x => x == "someString")`
executes the same as
`.Where( x => "someString" == x.String)`

The reason this came up is that my member variable is nullable, using the `<target.Member>.Equals` construction causes the nullable warning, I assumed checking for null to remove the warning would affect the resulting sql query with an explicit null check which would be redundant, using the failing construction `"".Equals(x.s)` causes no warnings, doesn't require me to disable the warning and doesn't require the explicit null check. 

I've created a test case & a solution on my own fork, it can be found here:
https://github.com/JasperFx/marten/compare/master...mkooiman:bugfix/symmetric-equality

This issue affects both 3.13.3 and the current master branch.

mkooiman:
Fixes #1703, but for objects

As noted by @jeremydmiller, the previous fix (#1706) only  applied to strings. This change adds the same fix to the generic Equals method of System.Object. 
I've also moved the test class to better fit the naming scheme.

If there are other places where this issue occurs, I'd be happy to take a look.

### DIFF
--- a/src/Marten.Testing/Bugs/Bug_1703_Equality_Not_Symmetric.cs
+++ b/src/Marten.Testing/Bugs/Bug_1703_Equality_Not_Symmetric.cs
@@ -8,14 +8,14 @@ using Xunit;
 
 namespace Marten.Testing.Bugs
 {
-    public sealed class equality_not_symmetric: IntegrationContext
+    public sealed class Bug_1703_Equality_Not_Symmetric: IntegrationContext
     {
-        public equality_not_symmetric(DefaultStoreFixture fixture) : base(fixture)
+        public Bug_1703_Equality_Not_Symmetric(DefaultStoreFixture fixture) : base(fixture)
         {
         }
         
         [Fact]
-        public void equality_equals_operator_should_be_symmetric()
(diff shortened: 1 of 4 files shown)


mysticmind (approved):
Looks good to me 👍

oskardudycz (approved):

PR #1712, opened 2021-01-22
First round of the V4 Async Daemon

jeremydmiller:
It's not all the way back to 100% of what the current async daemon is, but the foundation is in place and I'd like this pulled in soon so we can knock down some issues.

What is working:

* The new `EventProjection` can run in the new async daemon
* The new `ViewProjection` and new `AggregateProjection` both work in the async daemon. Little different set of optimizations compared to more, "inline" projection types
* `ProjectionAgent` is the new equivalent to the old `ProjectionTrack`. What was `Fetcher` is really combined into `ProjectionAgent`, and it's more "pull-based" to keep enough events queued up to keep the real projection humming along
* Isn't tested yet, but the new async daemon can handle multi-tenanted events and documents
* The new daemon uses the now standard .Net `ILogger` abstraction
* Closes 3-4 issues along the way


src/Marten/Events/Daemon/ProjectionAgent.cs
```diff
@@ -0,0 +1,191 @@
...
+            _hopper = _projectionShard.Start(this, _logger, _cancellationSource.Token);
+            _loader.LinkTo(_hopper);
+
+            var lastCommitted = await _store.Events.ProjectionProgressFor(_projectionShard.ProjectionOrShardName);
+
+            _commandBlock.Post(Command.Started(tracker.HighWaterMark, lastCommitted));
+
+            _subscription = _tracker.Subscribe(this);
+
+            _logger.LogInformation($"Projection agent for '{_projectionShard.ProjectionOrShardName}' has started from sequence {lastCommitted} and a high water mark of {tracker.HighWaterMark}");
```
Hawxy, line 122: I'd be much appreciated if you used the structured logging overloads for the `ILogger` integration. It's useful for those of us using structured sinks and log aggregation platforms like Sumologic.

```suggestion
            _logger.LogInformation("Projection agent for '{ProjectionOrShardName}' has started from sequence {lastCommitted} and a high water mark of {HighWaterMark}", _projectionShard.ProjectionOrShardName, lastCommitted, tracker.HighWaterMark);
```

jeremydmiller: @Hawxy I think that's a great idea, and had thought vaguely about that. Any chance you'd be interested in taking that on?

Hawxy: Sure 👍

src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
```diff
@@ -0,0 +1,169 @@
...
+        private readonly NpgsqlCommand _updateStatus;
+        private readonly NpgsqlParameter _newSeq;
+        private readonly NpgsqlCommand _findSafeSequence;
+        private readonly NpgsqlParameter _safeTimestamp;
+
+        public HighWaterDetector(ITenant tenant, EventGraph graph)
+        {
+            _tenant = tenant;
+
+            _findSafeSequence = new NpgsqlCommand($@"select min(seq_id) from {graph.DatabaseSchemaName}.mt_events where mt_events.timestamp >= :timestamp");
```
oskardudycz, line 26: We should consider inedxing strategy for those columns (at least as opt-in). For huge number of events those queries may take some time. On the other hand indexes will slow down appending...

oskardudycz: Also I'm not sure (yet) about all the assumptions (as I didn't go through all of he changes), but we might think about introducing some numeric offset commit instead of using the timestamp. Timestamp for multinode environments might be hit with time skew issues between nodes.

src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
```diff
@@ -0,0 +1,169 @@
...
+            IManagedConnection conn)
+        {
+            // If the last high water mark is the same as the highest number
+            // assigned from the sequence, then the high water mark cannot
+            // have changed
+            if (statistics.LastMark == statistics.HighestSequence)
+            {
+                statistics.CurrentMark = statistics.LastMark;
+            }
+            else if (statistics.HighestSequence == 0)
```
oskardudycz, line 99: @jeremydmiller is this case when projection was resetted and is being rebuilt from scratch?

src/Marten/Events/Daemon/HighWater/HighWaterAgent.cs
```diff
@@ -0,0 +1,144 @@
...
+                    case HighWaterStatus.Changed:
+                        await markProgress(statistics, _settings.FastPollingTime);
+                        break;
+
+                    case HighWaterStatus.CaughtUp:
+                        await markProgress(statistics, _settings.SlowPollingTime);
+                        break;
+
+                    case HighWaterStatus.Stale:
+                        var safeHarborTime = _current.Timestamp.Add(_settings.LeadingEdgeBuffer);
```
oskardudycz, line 90: Same consideration as above about the timestamp vs numeric offset and time skew as above https://github.com/JasperFx/marten/pull/1712#discussion_r563291852.

src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            var statistics = await loadCurrentStatistics(conn, token);
+
+
+            await calculateHighWaterMark(token, statistics, conn);
+
+            return statistics;
+        }
+
+        private async Task calculateHighWaterMark(CancellationToken token, HighWaterStatistics statistics,
```
oskardudycz, line 89: Suggestion. I think that it would be good to keep those calculations in the HighWaterStatistics class nad make setters for those properties private. It would be easier to track what's changed where in the future.

src/Marten/Events/Daemon/HighWater/HighWaterStatistics.cs
```diff
@@ -0,0 +1,34 @@
...
+        public long HighestSequence { get; set; }
+        public long CurrentMark { get; set; }
+        public bool HasChanged => CurrentMark > LastMark;
+        public DateTimeOffset? LastUpdated { get; set; }
+        public long SafeStartMark { get; set; }
+        public DateTimeOffset Timestamp { get; set; }
+
+        public HighWaterStatus InterpretStatus(HighWaterStatistics previous)
+        {
+            if (HighestSequence == 1 && CurrentMark == 0) return HighWaterStatus.CaughtUp;
```
oskardudycz, line 17: **QUESTION**: @jeremydmiller is it because HighestSequence numbered from 1 in database?

jeremydmiller: Yep. Confused me a little bit to be honest, but Postgresql sequences start w/ 1 by default. So the initial state is "HighestSequence = 1", which is really the current value of the postgres sequence.

src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
jeremydmiller: The timestamp is assigned by the database, and the only possible usage of it is in this one case to detect stale, missing values in the sequence. We don't key off that otherwise.

And yes to the indexing. Another option is to shard the table by sequence range. I don't see how a materialized view would help with the frequency that this would be getting hit.

src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
jeremydmiller: It's there because it came up in the tests. The only thing that can possibly be is if the system is in its initial state

src/Marten/Events/Daemon/HighWater/HighWaterAgent.cs
jeremydmiller: Same comment. This is to slide around "stale missing sequence" values. We're not using this in any way where we need precision. The "safe harbor" time would be something like "just assume anything older than 5 seconds is good". Easier to talk about this in Zoom.

When the HighWaterDetector "finds" the statistics, it starts looking from the last previously known sequence. Keep that in mind here.

So the case here is:

1. fetch the statistics once, the last good number is 1000, but the sequence is at 1300.
2. fetch the statistics again on the regular polling interval, the last good number is 1000, but the sequence has advanced to 1500. We can tell that there's stale data where transactions are either being *really slow finishing up, but have already reserved the sequence numbers. The current async daemon *sucks* because you always have to be very careful about whether or not missing sequence numbers are real and just in flight, or the result of a failure. The new tombstone event thing *should* make that problem mostly go away, but only mostly. So at some point we have to assume that it's safe to pick at any events older than the "safe harbor time". 
3. wait a calculated moment, and look for the highest sequence number that w/o gaps higher than the "safe harbor" timestamp

At no time are we using that timestamp as the determination for what evens have been published or not, except for the stale sequence gap issue. Which shouldn't happen unless you get funky database connectivity issues. And even there, you're not needing a lot of precision in the timestamp values, so I'm not concerned with the node drift issue.

src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
jeremydmiller: I completely understand why you're saying that, but I think that's just some extra complexity that would never pay off.

PR #1743, opened 2021-02-19
Hook up projection daemon in marten extensions

Hawxy:
This was missing and causing any daemon setting to throw on application startup when the hosted service attempted to resolve.

Due to the constructor on the projection daemon, building the instance via the store appears to be the simplest option.


jeremydmiller:
@Hawxy Why was this necessary? We had tests on registering the async daemon hosted service before. What were you seeing as far as an exception?

Hawxy:
The `AsyncProjectionHostedService` takes `IProjectionDaemon` in the constructor, so it just instantly throws on startup as no `IProjectionDaemon` is registered.

The tests checked registration, not that the hosted service was able to be resolved.

PR #1747, opened 2021-03-04
Update loading document documentation with possible null return value…

jedidja:
… and async sample.


mysticmind (approved):
@jedidja Thanks for the PR, looks good to me 👍

Issue #1659, opened 2020-12-10
Allow all projection types to accept abstract classes or interfaces for the event type


jeremydmiller:
This was a frequent request, but I'm kicking it out to its own issue so it doesn't get lost and I don't have to worry about it right now.

Some thoughts:

* Probably need to mark `Event<T>` as co/contra variant
* Messes with the async daemon's ability to filter event fetching by the expected type
* Complicates the code generation in the new async & view projections
* Worry about do you apply both the handler for a specific event type and a superclass event type if both exist

PR #1749, opened 2021-03-06
Added `new T Data` to IEvent<T> and removed GetData to simplify the end usage and make it backward compatible

oskardudycz:


src/Marten.AsyncDaemon.Testing/event_projections_end_to_end.cs
```diff
@@ -171,7 +171,7 @@ public override void Apply(IDocumentOperations operations, IReadOnlyList<StreamA
                 switch (@event.Data)
                 {
                     case IEvent<Travel> e:
-                        var travel = e.GetData();
```
jeremydmiller, line 174: You have to use the method for co/contravariance on the event wrappers. You can't make this change, or it'll break other functionality.

jeremydmiller:
Yeah, you can't do this without breaking the co/contra variance on `IEvent<T>`. It's not as simple as just reverting back to the property unless we want to redo the functionality from https://github.com/JasperFx/marten/issues/1659

src/Marten/Events/Event.cs
```diff
@@ -72,7 +72,7 @@ public interface IEvent
 
     public interface IEvent<out T> : IEvent
     {
-        T GetData();
+        new T Data { get; }
```
jeremydmiller, line 75: Does this actually work? If the test class `using_non_concrete_types_in_projections` is actually passing, I'm perfectly good with this PR.

jeremydmiller:
@oskardudycz I feel dumb now. Sorry for the friction this caused.

PR #1769, opened 2021-03-13
Nullability - Public surface

Hawxy:
This first PR mostly targets DocumentSession + Queryable extensions, as well as few spots that were simple to annotate. If the Marten maintainers are fine with this direction I'll continue into further areas. It might be valuable to consider any additional generic constraints that could be added as part of this work.

Some example projects have had NRTs enabled in order to serve as a test bed when required.

The build will fail for the 3.1 tests until the pipeline is updated to build with the .NET 5 SDK, as unconstrained nullable generics from C# 9 are required.

### DIFF
--- a/src/Marten/IQuerySession.cs
+++ b/src/Marten/IQuerySession.cs
@@ -8,7 +8,7 @@ using Marten.Services.BatchQuerying;
 using Marten.Storage;
 using Marten.Storage.Metadata;
 using Npgsql;
-
+#nullable enable
 namespace Marten
 {
     public interface IQuerySession: IDisposable, IAsyncDisposable
@@ -19,7 +19,7 @@ namespace Marten
         /// <typeparam name="T"></typeparam>
         /// <param name="id"></param>
         /// <returns></returns>
-        T Load<T>(string id);
+        T? Load<T>(string id);
 
         /// <summary>
         /// Asynchronously find or load a single document of type T by a string id
@@ -28,7 +28,7 @@ namespace Marten
         /// <param name="id"></param>
         /// <param name="token"></param>
         /// <returns></returns>
-        Task<T> LoadAsync<T>(string id, CancellationToken token = default(CancellationToken));
+        Task<T?> LoadAsync<T>(string id, CancellationToken token = default);
 
         /// <summary>
         /// Load or find a single document of type T with either a numeric or Guid id
@@ -36,7 +36,7 @@ namespace Marten
         /// <typeparam name="T"></typeparam>
         /// <param name="id"></param>
         /// <returns></returns>
-        T Load<T>(int id);
+        T? Load<T>(int id);
 
         /// <summary>
         /// Load or find a single document of type T with either a numeric or Guid id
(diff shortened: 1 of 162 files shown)


src/Marten/IQuerySession.cs
```diff
@@ -8,7 +8,7 @@
 using Marten.Storage;
 using Marten.Storage.Metadata;
 using Npgsql;
-
+#nullable enable
 namespace Marten
 {
     public interface IQuerySession: IDisposable, IAsyncDisposable
```
Hawxy, line 14: Can all generics on this interface be constrained to `class`?

oskardudycz (commented):
@Hawxy I'm a fan of the Nullable Reference Types. My only concern is that I'm afraid that eventually, we'll need to support all those weird attributes like NullableIfNullable, etc, which might be a maintenance nightmare.

However, the best would be to introduce such changes in the major version, so probably that's the right time. From my side, I'm fine with continuing working on that, but I'd like to see @jeremydmiller and @mysticmind opinion.

Regarding the PR, I'll need to go through that and check case by case. I'm wondering about checking that with tool like NDepend.
However,

Hawxy:
> My only concern is that I'm afraid that eventually, we'll need to support all those weird attributes like NullableIfNullable, etc, which might be a maintenance nightmare.

I'm yet to come across a class within the Marten codebase that has a complex enough null relationship to warrant usage of the attributes (ie lazy initialization or complexity within getters/setters), so I don't really expect much use of them. Even if they are used, they're set and forget until you need to refactor that particular bit of code. 

> I'm wondering about checking that with tool like NDepend.

I'm unsure if there's much value there. Most of this nullability work will be in areas of the codebase that have root nullability (where the code branches out from a nullable context and generates a warning), or is already documented as returning null (ie `Load<T>`, `SingleOrDefault`) so the risk of an annotation being incorrect is relatively low.

jeremydmiller:
@Hawxy I say "full speed ahead", but I'd ask you to focus on public APIs first.

Hawxy:
This _should_ be ready to review, I believe I've covered the vast majority of the public API. 

I may require a hand from @mysticmind to update the 3.1 build pipeline to build with the 5.0 SDK whilst still testing with the 3.1 runtime.

mysticmind:
@Hawxy I will have a look at the build pipeline related changes today and keep you posted.

mysticmind:
@Hawxy I have fixed the 3.1 target builds. You can have a look at the current builds in progress.

Hawxy:
Perfect, thanks!

jeremydmiller (approved):
My vote is to get this in as soon as everyone else is comfortable with it just because of the high potential for merge conflicts.

mysticmind (approved):
LGTM. Great work @Hawxy.

Quick observation, I see few warnings as below:
`src/Marten/Util/TypeMappings.cs` - Possible null reference argument for parameter 'value' in 'ImHashMap<Type, string> ImHashMap<Type, string>.AddOrUpdate(Type key, string value)

`src/Marten/Events/CodeGeneration/MethodCollection.cs` - The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.

I think overall new contributors may have issues handling nullable annotations unless they have a good grasp around all these new nullable related things.

I am going ahead and merging this PR.

PR #1777, opened 2021-03-23
Async daemon error handling improvements

jeremydmiller:


jeremydmiller:
This is the poison pill event detection. And a lot of preparatory work just to get there.

PR #1785, opened 2021-03-29
Added sample showing multiple streams projections.

oskardudycz:
@DariuszFirek, could you confirm that it covers your cases?


DariuszFirek:
Yes, that's something very similar to things I'm working on. 
But as far I saw, in marten v4 we don't have ProjectEvent(IList<Guid>, Apply), there are only ProjectEvent selecting single projection item's id.

PR #1815, opened 2021-05-11
Fix local docs serve

Hawxy:
The latest version of Vite released a few hours ago has a breaking change in its file serving due to a security issue, thereby breaking Vitepress.


mysticmind:
Thanks @Hawxy, I did see this thread regarding Vite.

Issue #1821, opened 2021-05-17
Schema validation fails always if document has indexes or foreign keys


VilleHakli:
In Marten v4 alpha 11 there seems to be a bug which causes `documentStore.Schema.AssertDatabaseMatchesConfiguration()` to fail if document has either index or foreign key defined.

Test case to reproduce the issue:
```csharp
using System;
using System.Threading.Tasks;

using Bug1821;

using Marten.Testing.Harness;

using Weasel.Postgresql;

using Xunit;

namespace Marten.Testing.Bugs
{
    public class Bug_1821_schema_validation_fails_if_document_has_indexes_or_foreign_keys: BugIntegrationContext
    {
        [Fact]
        public async Task schema_validation_should_succeed_when_document_has_index()
        {
            using var documentStore = SeparateStore(x =>
            {
                x.AutoCreateSchemaObjects = AutoCreate.All;
                x.Schema.For<TestDocument>()
                    .Index(y => y.Name);
            });

            await documentStore.Advanced.Clean.CompletelyRemoveAllAsync();
            await documentStore.Schema.ApplyAllConfiguredChangesToDatabase(AutoCreate.All);

            await documentStore.Schema.AssertDatabaseMatchesConfiguration();
        }

        [Fact]
        public async Task schema_validation_should_succeed_when_document_has_foreign_key()
        {
            using var documentStore = SeparateStore(x =>
            {
                x.AutoCreateSchemaObjects = AutoCreate.All;
                x.Schema.For<TestDocument>()
                    .ForeignKey<TestDocument>(y => y.ParentId);
            });

            await documentStore.Advanced.Clean.CompletelyRemoveAllAsync();
            await documentStore.Schema.ApplyAllConfiguredChangesToDatabase(AutoCreate.All);

            await documentStore.Schema.AssertDatabaseMatchesConfiguration();
        }
    }
}

namespace Bug1821
{
    public class TestDocument
    {
        public Guid Id { get; set; }
        public string Name { get; set; }

        public Guid? ParentId { get; set; }
    }
}
```

jeremydmiller:
Sigh. There's a **lot** of tests on the new schema management code already, and it's not that different. I'll check this out today.

jeremydmiller:
@VilleHakli What version of Postgresql are you using btw? It might be vulnerable on that axis.

VilleHakli:
[...]

PR #1831, opened 2021-05-23
Fix for GH-1821

VilleHakli:
This is second part of fixes for GH-1821. Most of the fixes required was in Weasel and the added tests will not pass until Marten uses Weasel with fixes done in JasperFx/weasel#26.


VilleHakli:
These changes were committed in a6baecbf4744fe1dadfc628396a4ec1a42029b61

PR #1840, opened 2021-06-04
Add failing multi-tenancy events scenario

Hawxy:


jeremydmiller:
I've got this.

jeremydmiller:
Got it. Regression bug from when I introduced the event archiving, and we didnt' have a test for it. Fixed some other related issues while I was in there. Fix is already into master.

Issue #1845, opened 2021-06-24
Async Daemon skips event on StopAll exception handling strategy.


imorphr:
Async Daemon skips event on StopAll exception handling strategy.

To showcase the problem, I have created a sample solution with two projects. One is a producer, and another one is a consumer of the events using async Daemon.

To reproduce the issue using the sample:
1. Create a blank PostgreSQL database called "testdb".
2. Run producer project and generate a few events following the on-screen instruction
3. Run consumer project. It should consume the events. When it encounters the event that causes failure, the application should retry three times and finally end with "StopAll".

At this stage, the entry in the mt_event_progression table reflects the latest event sequence number. My understanding was that the stop all should stop at the point it failed, and the progression sequence should be at or before the event that caused the failure.

[Sample.zip](https://github.com/JasperFx/marten/files/6708604/Sample.zip)

imorphr:
On further investigation, it is found that calling complete on the ActionBlock does not stop processing pending messages in the queue. Therefore, the ActionBlock needs to be faulted to drop the pending messages, and it can be done by rethrowing exception.

oskardudycz:
@imorphr, thank you for the fix! :) It's available in v3.13.5. See more:
- https://github.com/JasperFx/marten/releases/tag/3.13.5
- https://www.nuget.org/packages/Marten/3.13.5

oskardudycz:
Fixed by #1857

PR #1857, opened 2021-07-12
1845 async projection skips events on stop

imorphr:
As demonstrated in the issue https://github.com/JasperFx/marten/issues/1845 Async Daemon was skipping events on StopAll exception handling strategy. Fixed it by rethrowing the exception, that sets the ActionBlock faulted, resulting in the ActionBlock dropping all pending events.


oskardudycz (approved):

oskardudycz:
@imorphr, thank you for the fix! I won't forward port it to Marten v4, as Async Daemon internals were completely rewritten by @jeremydmiller.

PR #1861, opened 2021-07-19
Added Async suffix to Marten.Schema methods

oskardudycz:
Added Async suffix to Marten.Schema methods (e.g., ApplyAllConfiguredChangesToDatabase`) as they do not follow the convention method (e.g. SaveChangesAsync, RunSqlAsync etc.) which may be dangerous for the database setup if someone skips await. I personally stumbled on that.

I would be all for removing the suffix in all places if we were all async.

### DIFF
--- a/docs/guide/documents/advanced/extensions.md
+++ b/docs/guide/documents/advanced/extensions.md
@@ -72,7 +72,7 @@ public class FakeStorage : FeatureSchemaBase
     }
 }
 ```
-<sup><a href='https://github.com/JasperFx/marten/blob/master/src/Marten.Testing/CoreFunctionality/ability_to_add_custom_storage_features.cs#L49-L68' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_creating-a-fake-schema-feature' title='Start of snippet'>anchor</a></sup>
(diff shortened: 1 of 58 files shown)


mysticmind (approved):

jeremydmiller:
I'm not enthusiastic about this, but don't really care that much. You don't use these methods that often, and I'd push folks to use the command line tooling for schema management where they'd never see it. I vote to just take this in now, but make it clear in the docs that the command line tooling is probably the way to go in real usage.

oskardudycz:
@jeremydmiller, I agree. I'd also like to get rid of those `Async` suffixes, but I think that for now, it's important to be consistent. Maybe we could also mark all sync API as obsolete with a description that it might be removed in the next major version? I agree about describing that in docs.

jeremydmiller:
@oskardudycz "Maybe we could also mark all sync API as obsolete with a description that it might be removed in the next major version?" -- yeah, I agree with that. **Some** of this was caused by the multiplexing connection thing. We had to go async all the way down to make that work.

oskardudycz:
OK, I'll send a follow-up PR with Obsolete markers.

PR #1865, opened 2021-07-22
Added layout of Event Store docs

oskardudycz:


docs/.vitepress/config.js
```diff
@@ -350,42 +350,125 @@ function getDocumentDbSidebar() {
...
     {
-      text: 'Schema Objects',
-      link: '/guide/events/schema'
+      text: 'Introduction',
+      link: '/guide/events/index'
     },
     {
-      text: 'Stream identity',
-      link: '/guide/events/identity'
+      text: 'Appending events',
```
jeremydmiller, line 357: I think the outline looks good. I think I saw a note that you'd talk about features for dealing with concurrency in this slot, and I think that's probably fine.

jeremydmiller (approved):
Looks good to me as is. I couldn't think of anything that's missing. We do have a lot more command line support for the daemon now, but I think that should fit into rebuilding projections or the daemon conversation anyway. Make everything flow from "what is the user trying to do" rather than the technical implementation of Marten.

oskardudycz:
@jeremydmiller, thank you for checking that. For sure the Async Daemon will need to have more considerations. I was event thinking about the separate dedicated node with subsection. However, I shaped it like that as I'd like to take the user through the typical journey learning those concepts in the context of Marten.

I also added a separate section for Multitenancy, as not many tools have it out of the box, so it's good to highlight that feature.

docs/.vitepress/config.js
oskardudycz: Yup, I'd like to add it there. If it grows too big, then we could nest it.

Issue #1879, opened 2021-09-03
Events loss in Marten.Events.Projections.Async.Fetcher under load


simbadltd:
## Issue
There is events loss under load in 3.x, e.g. several transactions appending event (minimum 2 concurrent transactions). The same issue has been described here: [ Async daemon can lose events in a highly concurrent scenario #1629 ](https://github.com/JasperFx/marten/issues/1629).

## Root Cause
When an event is added to an aggregate in our code, `mt_append_event` function has been called. This function opens new transaction inside the database. Each transaction allocate it's own unique `seq_id` from `mt_events_sequence`. Finish transactions order is not defined and not guaranteed by PostgreSQL. This means that transaction which started later may commit sooner. And that's why we can observe records with greater `seq_id` sooner than records with smaller `seq_id`. 

`Marten.Events.Projections.Async.Fetcher` in the process of fetching next events page makes a request like:
```sql
select seq_id from public.mt_events where seq_id > [LAST] 
and seq_id <= [LIMIT] 
order by seq_id
```
This means that if  `Fetcher` will pick larger `seq_id` (because its' transaction committed sooner), then events with smaller `seq_id` (because transaction has committed later) will be skipped because of `[LAST]` parameter for the next fetch will be updated to `seq_id` number from second transation.

## Solution
The main idea is to use PostgreSQL `tx_id` (transaction id). It is monotonically increasing number handling by PostgreSQL itself. This means that we can have unique increasing number for each event in correct order: the later event the smaller number. We also have `txid_snapshot_xmin(txid_current_snapshot())` construction which returns minimum `tx_id` (transaction id) of any opened (not finished, not committed) transaction. So, we are going to use it in way that we can chuck out all committed later than minimum not finished transaction events: `tx_id < txid_snapshot_xmin(txid_current_snapshot())`. Once, not finished transaction committed, all this events become visible for `Fetcher` and will be handled:
```sql
select seq_id from public.mt_events where seq_id > [LAST]
and seq_id <= [LIMIT]
and tx_id < txid_snapshot_xmin(txid_current_snapshot())
order by seq_id
```

`events` table and `mt_append_event` function also require changes:

```sql
DROP TABLE IF EXISTS {databaseSchema}.mt_events;
CREATE TABLE {databaseSchema}.mt_events (
	seq_id		bigint CONSTRAINT pk_mt_events PRIMARY KEY,
	id		uuid NOT NULL,
[...]

PR #1880, opened 2021-09-03
Events loss in Fetcher under load

simbadltd:
Fix for [Events loss in Marten.Events.Projections.Async.Fetcher under load #1879 ](https://github.com/JasperFx/marten/issues/1879)

### DIFF
--- a/src/Marten/Events/AppendEventFunction.cs
+++ b/src/Marten/Events/AppendEventFunction.cs
@@ -66,9 +66,9 @@ BEGIN
 		body = bodies[index];
 
 		insert into {databaseSchema}.mt_events
-			(seq_id, id, stream_id, version, data, type, tenant_id, {DocumentMapping.DotNetTypeColumn})
+			(seq_id, id, stream_id, version, data, type, tenant_id, {DocumentMapping.DotNetTypeColumn}, tx_id)
 		values
-			(seq, event_id, stream, event_version, body, event_type, tenantid, dotnet_types[index]);
+			(seq, event_id, stream, event_version, body, event_type, tenantid, dotnet_types[index], txid_current());
 
 		index := index + 1;
 	end loop;
--- a/src/Marten/Events/EventsTable.cs
+++ b/src/Marten/Events/EventsTable.cs
@@ -19,6 +19,7 @@ namespace Marten.Events
             AddColumn("timestamp", "timestamptz", "default (now()) NOT NULL");
             AddColumn<TenantIdColumn>();
             AddColumn(new DotNetTypeColumn { Directive = "NULL" });
+            AddColumn("tx_id", "bigint", "default 0 NOT NULL");
 
             if (events.TenancyStyle == TenancyStyle.Conjoined)
             {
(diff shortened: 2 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1880) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1880) before we can accept your contribution.<br/><br/>:x: simbadltd [sign now](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=1880)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/JasperFx/marten?pullRequest=1880) it.</sub>

jeremydmiller:
@simbadltd Hey, we can take this into the 3.* branch, but be aware that the async daemon was 100% rewritten in the v4 release that's coming this month.

Leh2:
Given how critical an error this is, it would be very nice to get this is a 3.* release 🙏 

Can the table migration be done automatically or should it be opt in?

simbadltd:
> Given how critical an error this is, it would be very nice to get this is a 3.* release 🙏
> 
> Can the table migration be done automatically or should it be opt in?

There is no table migration for existing databases since that it might be not safe especially for production.

oskardudycz (approved):

Leh2:
> > Given how critical an error this is, it would be very nice to get this is a 3.* release 🙏
> > Can the table migration be done automatically or should it be opt in?
> 
> There is no table migration for existing databases since that it might be not safe especially for production.

@simbadltd @oskardudycz how do we utilize this change? Do we need to add the new column `tx_id` manually with default value `0`  and then everything is good to go?

oskardudycz:
@Leh2 if I'm not wrong, looking at the code, then that should be enough. If you'd like to, I can also make this feature opt-in.

simbadltd:
@Leh2 Yes, add the 'tx_id' column typed as `bigint` with default value `0` and not-null constraint. Also, you need to patch `mt_append_event` function as follows:

Append newly added `tx_id` column into the `INSERT` statement and also add the `txid_current()` function to the `INSERT` statement values part:
```
		insert into {databaseSchema}.mt_events
			(seq_id, id, stream_id, version, data, type, tx_id)
		values
			(seq, event_id, stream, event_version, body, event_type, txid_current());
```

Leh2:
Thanks for the answers 🙏 

@oskardudycz giving this is a breaking change, shouldn't this be a major change or added as opt-in?

oskardudycz:
@Leh2, fair, I'll work this week to make it opt-in in the `3.15.6`.

Leh2:
> @Leh2, fair, I'll work this week to make it opt-in in the `3.15.6`.

Hi @oskardudycz. Did you get the time to look at this :)

culaja:
@oskardudycz Is this fix available in Marten 4.0.4 ?
I can't see tx_id column in mt_events table.

oskardudycz:
@culaja nope, it's not. As I mentioned, this fix won't be forward ported. The whole Async Daemon was rewritten from scratch. It's much more reliable than it was before. @jeremydmiller would need to wage if that's something that we need to have in v4.

culaja:
Since I have a case where I manually read events from the mt_events I'm not sure how to be sure that I'll not skip some event in high-load scenarios.

oskardudycz:
You may try serialisable or repeatable read serialisation level for that case. Then it won't be possible to skip any events or have phantoms (https://www.postgresql.org/docs/9.5/transaction-iso.html).

jeremydmiller:
@culaja That's the whole point of the async daemon. What are you doing where you can't use that?

culaja:
I can't find API in Marten like there is in EventStoreDb: to subscribe to all streams by starting from a specific position (sequence).
`Subscibe(Type[] typeOfEventsToSusbcribeTo, long startSequence, Func<EventHandler, Task> eventHandler)`

So this is the reason why I'm reading events directly from mt_events.

Issue #1886, opened 2021-09-11
When a compiled query uses the same input parameter multiple times inside a query and the same compiled query is called twice with different parameters, the 2nd invocation retains values from previous ones.


rwasef1830:
Take a query like this. User class has NormalizedUserName and NormalizedEmail as duplicated indexed fields.

```cs
public class User
{
    public Guid Id { get; set; }
    public string NormalizedUserName { get; set; }
    public string NormalizedEmail { get; set; }
}

public class FindUserByNameQuery : ICompiledQuery<User, User?>
{
	public string NormalizedName { get; init; } = string.Empty;

	public Expression<Func<IMartenQueryable<User>, User?>> QueryIs()
	{
		return q => q.FirstOrDefault(
			x => x.NormalizedUserName == this.NormalizedName
				 || x.NormalizedEmail == this.NormalizedName);
	}
}
```

Then use it twice with different sessions in a row like this on the same document store:

```cs
var session1 = this._store.QuerySession();
await using (session1.ConfigureAwait(false))
{
    await session1
	.QueryAsync(
		new FindUserByNameQuery
		{
			NormalizedName = "1",
		});
}

var session2 = this._store.QuerySession();
await using (session2.ConfigureAwait(false))
{
    await session2
	.QueryAsync(
		new FindUserByNameQuery
		{
			NormalizedName = "2",
		});
}
```

Generates 2 queries that look this:
```
LOG:  execute <unnamed>: select d.data from public.mt_doc_user as d where (d.normalized_user_name = $1 or d.normalized_email = $2) LIMIT $3
DETAIL:  parameters: $1 = '1', $2 = '1', $3 = '1'

LOG:  execute <unnamed>: select d.data from public.mt_doc_user as d where (d.normalized_user_name = $1 or d.normalized_email = $2) LIMIT $3
DETAIL:  parameters: $1 = '2', $2 = '1', $3 = '1'
```

Notice how the 2nd usage of the same input got its value retained from the previous invocation.
Unit test PR coming up.

rwasef1830:
Workaround is avoid using the same input parameter multiple times in compiled queries.

PR #1887, opened 2021-09-11
Unit test demonstrating bug #1886.

rwasef1830:
In this test, the bug happens and it fails to find the inserted user due to incorrect retention of values in the compiled query from the previous invocation.


jeremydmiller:
I pulled this one in w/ a rebase. @rwasef1830 Thanks for the repro steps!!!

jeremydmiller:
And this is fixed in master now too.

PR #1889, opened 2021-09-14
Fix run_specification for Inline_Aggregation_by_Stream.md

0xced:
Before this commit:
❌ run_specification(path: "Event Store/Projections/Inline_Aggregation_by_Stream.md") would fail with the following error:
> System.Exception: Rights: 2, Wrongs: 0, Exceptions: 2, SyntaxErrors: 0

And this is what was displayed in the html result file:
> Cell Error in 'Members'
```
No converter found for type System.Collections.Generic.List`1[[System.String, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]
```

After this commit:
✅ All StorytellerRunner.Runner.run_specification theories pass

Note: this looks like a limitation of Storyteller.


mysticmind (approved):
Looks good to me.

PR #1899, opened 2021-09-23
Added sample Event Grouper that transforms events (Grouping + FanOut)

oskardudycz:
@jeremydmiller, do you think we could add some syntactic sugar to enable such a scenario when you need both grouper and fan out?


oskardudycz:
@jeremydmiller, I sprinkled the example with a bit of syntactic sugar. I added:
- [TransformsEvent](https://github.com/JasperFx/marten/pull/1899/files#diff-517b901d6afbec74a7286767e5e477ec7163452ee692b4bbbc2aa0489058b1e0R50) that marks transformed event to be handled without the hack with dummy Apply method,
- [WithData](https://github.com/JasperFx/marten/pull/1899/files#diff-8c311b10436b320f6b49f677bcb00c601d621a03a286c3150322a4103d9e1984R219) to make cloning events with different event data easier.

I also refactored a bit the transformation.

If you're not against that, then I'll go ahead and merge it. We could add a dedicated grouper with transformation, but I'll suggest something in separate PR.

Thoughts?

@kluhman FYI

PR #1904, opened 2021-09-24
Added readonly event store methods to IQuerySession

oskardudycz:
I achieved that by adding the `IQueryEventStore` interface and `QueryEventStore` implementation.

I'm not entirely proud with the [CreateEventStore](https://github.com/JasperFx/marten/pull/1904/files#diff-629665ba21293b01ff2d61f05c14faeb4da6cc434d3baaa436520fd9ca8ab9beR41) virtual method, but I haven't found a better (less invasive) option, as creating `EventStore` requires passing to-be-created session with `this`. Another alternative would be to have a public setter for the session, but that's even more ugly...

@jeremydmiller, if you have better suggestions, I'm open to updating PR.


jeremydmiller (approved):
This all looks fine with me.

PR #1923, opened 2021-10-10
Fix some exception logging

gkinsman:
Some log messages are currently using this overload:

`void LogError(this ILogger logger, string message, params object[] args)`

But in order to correctly capture and enrich the log event with the exception, the following overload should be used:

`void LogError(this ILogger logger, Exception exception, string message, params object[] args)`.

Currently the exception isn't rendered as MSEL ignores props that have no equivalent holes.


mysticmind:
@gkinsman  Thanks for the PR, looks good to me.

PR #1958, opened 2021-10-29
Removed duplicate code snippet in Events Quickstart

mhmd-azeez:
Docs page: https://martendb.io/events/quickstart.html


mysticmind:
@mhmd-azeez good catch, thanks for the PR.

mysticmind (approved):

mysticmind:
@mhmd-azeez on rechecking, the problem is a bit different, we have 2 code samples in code base using the same name i.e. `sample_sample-events` in https://github.com/JasperFx/marten/blob/b43ecd8a50d80e8d695469f3ff94485bded14e3a/src/Marten.Testing/Events/QuestTypes.cs#L12 and https://github.com/JasperFx/marten/blob/b43ecd8a50d80e8d695469f3ff94485bded14e3a/src/Marten.Schema.Testing/QuestTypes.cs#L12

The fix is to rename one of the sample name in code so that the docs don't produce repeating code. I will add a fix on this PR itself.

PR #1966, opened 2021-11-04
Brought back Event Type Mapping by adding a dedicated MapEventType method to IEventStoreOptions

oskardudycz:
Besides that enabled build script run if someone has .NET 6 installed (like me 😅 ).


mysticmind (approved):
Looks good to me 👍

PR #1971, opened 2021-11-06
Do not ignore .Take() when using .Single() in LINQ query

VilleHakli:
Following query will throw exception if there is more than one document with number 2:
```csharp
session.Query<Target>().Where(x => x.Number == 2).Take(1).Single();
```
When LINQ query with .Single() is parsed, the limit for SQL query is always set to two. I think that the Take(1) should be honored and not be overriden. This PR changes the query parsing to set the limit only if it is not set previously.


mysticmind (approved):
Looks good to me 👍

oskardudycz (approved):

PR #1972, opened 2021-11-07
Fix some VSTHRD002, VSTHRD103, VSTHRD105, VSTHRD110 threading issues

solyutor:
I'm back with a new bunch of fixes. Most of them are easy and simple, but a few should be reviewed carefully:
* ProjectionDaemon - `WaitForNonStaleData()` - completely rewritten, previous implementation never threw a `TimeoutException` (at least I wasn't able to fine how it could happen). Less important fix is for `StartAllShards()` there was a bug which forced the method to block on `Task.Unwrap()` call instead of running tasks in parallel. 
* Simplifed interaction between `TenantSliceRange` and `TenantSliceGroup`
* `AsyncProjectionBase.Apply()`  throws `NotSupportedException` as there's no good and safe way to call an async method from a sync one. 

I tested it twice with `Newtonsoft.Json` and `SystemTextJson` and all tests passed.


oskardudycz (approved):
Again, excellent work, @solyutor 👍 Thank you.

mysticmind (approved):
The changes look fine. But when I ran the async daemon unit tests, around 5 tests are consistently failing (I re-ran these few times as well). See details below:
Screenshot 2021-11-09 at 9 16 58 PM

These outlined issues look to be related to changes in `WaitForNonStaleData`.

solyutor:
I didn't realize async daemon tests don't run during build script execution. The problem with the last commit, I'll investigate it.

solyutor:
@mysticmind thanks for noticing async daemon test issues. It appeared to be `NullReferenceException` with nullified `_bulder` field of `TenantSliceGroup`. Added [thread-safe check for null](https://github.com/solyutor/marten/blob/04378a80d5033eafbfcbbdbd76872ad8d1b6bee9/src/Marten/Events/Aggregation/TenantSliceGroup.cs#L146). 
However, if `ShardAgent.Stop(Exception ex = null)` logged the `ex` it would be faster and easier to find the root cause. May be it worse adding some warn/error level message?

mysticmind (approved):
All async unit tests are green now. Good work @solyutor 👍 
Screenshot 2021-11-10 at 2 32 00 PM

mysticmind:
> However, if ShardAgent.Stop(Exception ex = null) logged the ex it would be faster and easier to find the root cause. May be it worse adding some warn/error level message?

I think we could add a log error if exception is not null.

PR #1976, opened 2021-11-11
Move IInitialData call to new IHostedService

solyutor:
I analyzed Marten code base and found  the only usage of `IInitialData` is inside the `DocumentStore` constructor. There two options possible to fix it:
* Move calls to `IInitialData` out of constructor to an async builder method. Will require relatively large changes to Marten code and is going to be a breaking change. Another issues is that most of initialization code related to ASP.NET CORE (and other frameworks/libraries, I guess)  is synchronous and isn't  going to change in the foreseeable future. 
* Make  `IInitialData` synchronous. It's a breaking change also but a lesser one. Initialization code won't be changed. 

As it's a breaking change so I guess the PR should wait until Marten 5.0 release.
Thoughts?


Hawxy:
It was made async as npgsql's multiplexing support requires all database calls to be async, however production-ready multiplexing has been pushed off to npgsql 7 so support for it _could_ be reintroduced in a future release.

> Another issues is that most of initialization code related to ASP.NET CORE (and other frameworks/libraries, I guess) is synchronous and isn't going to change in the foreseeable future.

If Marten v5 goes down the purist .NET 6 path, then the assumption could be made that everyone is going to be using the new ASP.NET Core templates where async initialization is implicitly available via top-level statements.

jeremydmiller:
Since we're about to do a V5 anyway to accommodate other breaking changes from Npgsql 6, I say that the initial data gets somewhat excised. Make it be async all the way, but push it's execution to an `IHostedService`. This'll make the usage be reliant on using AddMarten(), but I think I want to encourage that anyway. If we do that, IInitialData can be async all the way down.

jeremydmiller:
@solyutor Thank you so much for everything you've done for threading and async code on Marten, but I'd like to take this a bit different way and move the IInitialData into a new IHostedService for Marten v5. And keep it async all the way down.

I think the general direction is moving in the direction of everything being async and changing the Marten internals to reflect that.

PR #1978, opened 2021-11-13
Any query optimization (avoid counting all rows)

solyutor:
Current `Any()` implementation compiles to a query `select (count(*) > 0) as result from public.mt_doc_target as d`. It forces postgres to count all related rows which is obviously redundant work. Fastest query I could imaging would be `select exists as result from (select 1 from public.mt_doc_target as d) as temp`. I haven't figured out an easy way to implement this directly within Marten codebase. Luckily I found a workaround that works: `select 1=1 as result from public.mt_doc_target as d LIMIT 1` and gives almost same execution plan as `exists` based query (see below).

I also deduplicated  `previewing_the_command_from_a_queryable_Tests` and `previewing_the_command_from_a_queryable_in_a_different_schema_Tests` with some string interpolation magic.

These are the execution plans to shows the improvement on a 10k rows tables:
 
image

image

image

### DIFF
--- a/src/Marten/Linq/SqlGeneration/AnySelectClause.cs
+++ b/src/Marten/Linq/SqlGeneration/AnySelectClause.cs
@@ -25,7 +25,7 @@ namespace Marten.Linq.SqlGeneration
 
         public void WriteSelectClause(CommandBuilder sql)
         {
-            sql.Append("select (count(*) > 0) as result");
+            sql.Append("select TRUE as result");
             sql.Append(" from ");
             sql.Append(FromObject);
             sql.Append(" as d");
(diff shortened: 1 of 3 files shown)


oskardudycz (approved):
@solyutor, nice one! 👌

src/Marten/Linq/SqlGeneration/AnySelectClause.cs
```diff
@@ -25,7 +25,7 @@ public AnySelectClause(string from)
 
         public void WriteSelectClause(CommandBuilder sql)
         {
-            sql.Append("select (count(*) > 0) as result");
+            sql.Append("select 1 = 1 as result");
```
mysticmind, line 28: Nitpick: `1  = 1` could be changed to `TRUE` to make it clear and explicit.

mysticmind (approved):
Looks good to me. Please do check the comment from my end.

src/Marten/Linq/SqlGeneration/AnySelectClause.cs
solyutor: Good Idea, fixed.

PR #1987, opened 2021-11-19
Npgsql 6 upgrade

oskardudycz:
- Updated Weasel to 2.0.0 and fixed DateTime related tests to follow new Npgsql rules https://www.npgsql.org/doc/index.html (DateTimeOffset for UTC, DateTime for without time zone).
- Added NodaTime mappings.

See related PR for Weasel: https://github.com/JasperFx/weasel/pull/35

### DIFF
--- a/src/Marten.NodaTime/NodaTimeExtensions.cs
+++ b/src/Marten.NodaTime/NodaTimeExtensions.cs
@@ -1,5 +1,5 @@
 using System;
-using System.Linq;
+using System.Data;
 using Marten.Services;
 using NodaTime;
 using NodaTime.Serialization.JsonNet;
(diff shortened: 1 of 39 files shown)


jeremydmiller:
You already know this, but that's a 5.0 release. I think it's smart to replace the DateTime parameters w/ DateTimeOffset. I'm good with everything I see here.

jeremydmiller (approved):
I think this is good as is

src/Marten.NodaTime/NodaTimeExtensions.cs
```diff
@@ -18,30 +21,63 @@ public static class NodaTimeExtensions
...
-                storeOptions.Serializer(serializer);
+                    break;
+                default:
+                    throw new NotSupportedException("Current serializer cannot be automatically configured for NodaTime. Set shouldConfigureJsonSerializer to false if you're using your own serializer.");
             }
+
+            storeOptions.Serializer(serializer);
+        }
+
+        public static void SetNodaTimeMappings()
```
mysticmind, line 52: Suggestion: Rename this from `SetNodaTimeMappings` to `SetNodaTimeTypeMappings` to add clarity.

mysticmind (approved):
Looks good to me. Added a suggestion, please have a look.

src/Marten.NodaTime/NodaTimeExtensions.cs
oskardudycz: Fixed in: https://github.com/JasperFx/marten/pull/1987/commits/380b2b39d57ad6454e1fa770d20a3cb022931268

PR #2004, opened 2021-12-04
Added JsonNetObjectContractProvider to enable private non default constructor resolution for Newtonsoft Json.NET

oskardudycz:
- Added `JsonNetObjectContractProvider` to enable private non default constructor resolution for Newtonsoft Json.NET
- Added `NonPublicMembersStorage.NonPublicConstructor` option and updated `JsonNetContractResolver` accordingly
- Added tests and updated docs

### DIFF
--- a/src/Marten/ISerializer.cs
+++ b/src/Marten/ISerializer.cs
@@ -165,6 +165,8 @@ namespace Marten
     {
         Default = 0,
         NonPublicSetters = 1,
-        NonPublicDefaultConstructor = 2
+        NonPublicDefaultConstructor = 2,
+        NonPublicConstructor = 4,
+        All = Default | NonPublicSetters | NonPublicDefaultConstructor | NonPublicConstructor
     }
 }
--- a/docs/configuration/json.md
+++ b/docs/configuration/json.md
@@ -152,7 +152,7 @@ private readonly JsonSerializer _serializer = new()
     ContractResolver = new JsonNetContractResolver()
 };
 ```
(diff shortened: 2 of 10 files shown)


src/Marten/ISerializer.cs
```diff
@@ -165,6 +165,8 @@ public enum NonPublicMembersStorage
     {
         Default = 0,
         NonPublicSetters = 1,
-        NonPublicDefaultConstructor = 2
+        NonPublicDefaultConstructor = 2,
+        NonPublicConstructor = 4,
```
mysticmind, line 169: Minor thing, shouldn't the value be `3` rather than `4`?

mysticmind (approved):
Looks good to me 👍

src/Marten/ISerializer.cs
oskardudycz: I'm adding that to allow bit flag usage (e.g. `NonPublicMembersStorage.NonPublicSetters | NonPublicMembersStorage.NonPublicDefaultConstructor`). It needs to be 2^n

oskardudycz: See also: https://github.com/JasperFx/marten/pull/2004/files#diff-f899e2a4569989d2b4fbf225960c0a07bba52c31d6b434b964ffb02dfa4eb148R33

mysticmind: Okay, makes sense 👍

Issue #2005, opened 2021-12-07
Marten v4.3.0 - EjectAllOfType does not clear pending changes


mronnblom-cint:
For some reason, `EjectAllOfType` does not clear pending changes. The ItemMap is cleared though. I can't see any apparent problem with the code below.. Maybe `CanBeCastTo` doesn't work in this case? There are no tests for `EjectAllOfType`...

DocumentSessionBase:
```
public virtual void EjectAllOfType(Type type)
{
    ItemMap.Remove(type);
    ChangeTrackers.RemoveAll(x => x.Document.GetType().CanBeCastTo(type));
}
```

oskardudycz:
@mronnblom-cint, thank you for the information. I don't recall any changes in this area recently. Could you send a PR with a test reproducing your issue? It'd help us to tackle this case.

jeremydmiller:
`PendingChanges()` is based off the pending operations, which is independent of the identity map or automatic dirty checking tracking. Honestly, the functionality is working exactly as it was envisioned. 

To be clear, what you're really wanting is to clear our and remove previously *explicitly* registered actions to a session? I.e., you're wanting to remove pending operations resulting from `Store()` / `Insert()` / `Update()` calls to the same session? That honestly isn't something we ever anticipated.

jeremydmiller:
And I'm removing the "Bug" tag here, because it's not.

mronnblom-cint:
Is it your intention that `Eject` and `EjectAllOfType` should work differently? That seems very weird to me, and also not in line with the summary descriptions. The summary for `EjectAllOfType` clearly states that it should clear "unit of work tracking". Eject does. EjectAllOfType does not. I made this PR which also demonstrates this with tests:
https://github.com/JasperFx/marten/pull/2010

jeremydmiller:
Can you go farther and send the PR in with whatever changes in functionality you want to purge operations as well? Dude, to be clear, I don't really approve of the way you're doing things, but I'm okay with the extra functionality here if you'll do the PR to also purge operations. What you're wanting is never what I intended the code to do, but I never thought to use a session in a reentrant way like you seem to be doing here.

mronnblom-cint:
"The way I'm doing things"? And what way is that? Have you looked at the PR? If you had, you would have seen that it DOES contain the changes to the functionality, including tests for it (that you completely have omitted for `EjectAllOfType`). Also, if this was never your intention of how that method should work, can you explain how the summary for it should be interpreted?

```
[...]

PR #2010, opened 2021-12-11
EjectAllOfType also clears WorkTracker (same as Eject does)

mronnblom-cint:
Fixes https://github.com/JasperFx/marten/issues/2005

### DIFF
--- a/src/Marten.Testing/CoreFunctionality/ejecting_a_document_type.cs
+++ b/src/Marten.Testing/CoreFunctionality/ejecting_a_document_type.cs
@@ -71,8 +71,10 @@ namespace Marten.Testing.CoreFunctionality
 
             using (var session = theStore.OpenSession())
             {
-                session.Store(target1, target2);
-                session.Store(user1, user2);
+                session.Insert(target1);
+                session.Store(target2);
+                session.Insert(user1);
+                session.Store(user2);
 
                 session.EjectAllOfType(typeof(Target));
 
@@ -84,6 +86,47 @@ namespace Marten.Testing.CoreFunctionality
                 session.Query<Target>().ShouldBeEmpty();
                 session.Load<User>(user1.Id).ShouldNotBeNull();
                 session.Load<User>(user2.Id).ShouldNotBeNull();
+            }
+        }
+
+        [Fact]
+        public void eject_a_document_type_clears_updates_from_the_unit_of_work_regular()
+        {
+            var target1 = new Target { Number = 1 };
+            var target2 = new Target { Number = 2 };
+            var user1 = new User { Age = 10 };
+            var user2 = new User { Age = 20 };
+
+            using (var session = theStore.OpenSession())
+            {
+                session.Store(target1, target2);
+                session.Store(user1, user2);
+
+                session.SaveChanges();
+            }
+
+            using (var session = theStore.OpenSession())
+            {
+                target1.Number = 3;
+                target2.Number = 4;
+                user1.Age = 30;
+                user2.Age = 40;
+                session.Update(target1);
+                session.Update(target2);
+                session.Update(user1);
+                session.Update(user2);
+
+                session.EjectAllOfType(typeof(Target));
+
+                session.SaveChanges();
+            }
+
+            using (var session = theStore.QuerySession())
+            {
+                session.Load<Target>(target1.Id).ShouldNotBeNull().Number.ShouldBe(1);
+                session.Load<Target>(target2.Id).ShouldNotBeNull().Number.ShouldBe(2);
+                session.Load<User>(user1.Id).ShouldNotBeNull().Age.ShouldBe(30);
+                session.Load<User>(user2.Id).ShouldNotBeNull().Age.ShouldBe(40);
             }
         }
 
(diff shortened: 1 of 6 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=2010) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=2010) before we can accept your contribution.<br/><br/>:x: mronnblom-cint [sign now](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=2010)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/JasperFx/marten?pullRequest=2010) it.</sub>

mysticmind (approved):
Looks good to me 👍

oskardudycz (approved):

src/Marten/Internal/UnitOfWork.cs
```diff
@@ -219,6 +219,17 @@ public void Eject<T>(T document)
             }
         }
 
+        public void EjectAllOfType(Type type)
+        {
+            var operations = operationsFor(type);
+            var matching = operations.OfType<IDocumentStorageOperation>().ToArray();
+
+            foreach (var operation in matching)
```
jeremydmiller, line 227: You have to check whether the `IDocumentStorageOperation` involves the `type` here.

src/Marten.Testing/CoreFunctionality/ejecting_a_document_type.cs
```diff
@@ -0,0 +1,117 @@
...
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace Marten.Testing.CoreFunctionality
+{
+    public class ejecting_a_document_type : IntegrationContext
+    {
+        [Fact]
+        public void eject_a_document_type_clears_it_from_the_identity_map_regular()
```
jeremydmiller, line 11: You probably want tests for Insert() and Update() as well as Store(). And you need to add operations from some sort of document besides `Target` to ensure that only the expected document type is impacted.

jeremydmiller (changes requested):
This just needs the document type filtering and a test proving that.

src/Marten/Internal/UnitOfWork.cs
mysticmind: @jeremydmiller  `var operations = operationsFor(type);`, this line does that check, anything else needed?

jeremydmiller: Ah, sorry, I missed that. Still need a test to prove that it's filtering correctly though.

mronnblom-cint: Good point! Hopefully I have time later this week to add more tests!

src/Marten.Testing/CoreFunctionality/ejecting_a_document_type.cs
mronnblom-cint: I'm in the process of adding tests for `Update()`, but I don't understand why this test fails (even with `EjectAllOfType` commented out). I have never used the `DirtyTrackedSession` myself, so maybe I've misunderstood it, but shouldn't it track pending changes without needing to call `Update`?

```
[Fact]
public void eject_a_document_type_clears_updates_from_the_unit_of_work_dirty()
{
    var target1 = new Target { Number = 1 };
    var target2 = new Target { Number = 2 };
    var user1 = new User { Age = 10 };
    var user2 = new User { Age = 20 };

    {
        session.Store(target1, target2);
        session.Store(user1, user2);

        session.SaveChanges();
    }

    using (var session = theStore.DirtyTrackedSession())
    {
        target1.Number = 3;
        target2.Number = 4;
        user1.Age = 30;
        user2.Age = 40;

        //session.EjectAllOfType(typeof(Target));

        session.SaveChanges();
    }

    using (var session = theStore.QuerySession())
    {
        session.Load<Target>(target1.Id).ShouldNotBeNull().Number.ShouldBe(1);
        session.Load<Target>(target2.Id).ShouldNotBeNull().Number.ShouldBe(2);
        session.Load<User>(user1.Id).ShouldNotBeNull().Age.ShouldBe(30);
        session.Load<User>(user2.Id).ShouldNotBeNull().Age.ShouldBe(40);
    }
}
```

jeremydmiller: Just what you said, the dirty tracking checks all documents that were loaded through the session at the time SaveChanges() was called. Very inefficient, only useful as a convenience when perf doesn't matter much.

mronnblom-cint: Ok. So why doesn't the test above work? When you do e.g. `target1.Number = 3` inside the `DirtyTrackedSession` and then `SaveChanges`, why is that change not committed?

mronnblom-cint: I just realized something myself.. Do you need to load the objects from the actual `DirtyTrackedSession` for it to know about them? That would probably make sense.

mronnblom-cint: Yep, that was it. See latest commit.

jeremydmiller:
I'm adding a little bit more to the tests locally, then pulling this in today.

Issue #2018, opened 2021-12-21
Fulltext Search on string list: Index generation error


lngr:
Full Text Search index creation on multiple properties with one string field and a string list is broken on current master.  This used to work on Marten 3.x.

        public sealed class BugFullTextSearchFields
        {
            public Guid Id { get; set; }
            public string Text { get; set; }
            public List<string> Data { get; set; }
        }

        public Bug_XXX_fts_string_list()
        {
            StoreOptions(_ => _.Schema
                .For<BugFullTextSearchFields>()
                .GinIndexJsonData()
                .FullTextIndex(index =>
                    {
                        index.Name = "mt_custom_my_index_name_fulltext_search";
                        index.RegConfig = "english";
                    },
                    x => x.Text, x => x.Data)
                .UseOptimisticConcurrency(true));
        
This creates the following SQL:

    CREATE INDEX mt_custom_my_index_name_fulltext_search ON
    bugs.mt_doc_bug_xxx_fts_string_list_bugfulltextsearchfields USING gin
    (to_tsvector('english',((data ->> 'Text') || ' ' || (CAST(data ->> 'Data' as varchar[])))));`

With the following error:

    Npgsql.PostgresException 42883: Operator does not exist: text || character varying[]

In Marten 3.14, the following index was created instead (copied from the Postgres index definition):

    CREATE INDEX mt_custom_my_index_name_fulltext_search
    ON bugs.mt_doc_bug_2018_fts_string_list_bugfulltextsearchfields USING gin
    (to_tsvector('english'::regconfig, ((data ->> 'Text'::text) || ' '::text) || ((data ->> 'Data'::text)::jsonb)))
    TABLESPACE pg_default;

lngr:
We run a query against that index as follows:

    SELECT * from bugs.mt_doc_bug_2018_fts_string_list_bugfulltextsearchfields WHERE
    (to_tsvector('english'::regconfig, ((data ->> 'Text'::text) || ' '::text) || ((data ->> 'Data'::text)::jsonb)))
	@@ to_tsquery('english'::regconfig, 'Dar:* & Very:*')

![image](https://user-images.githubusercontent.com/1508058/146912697-e91290a4-a052-4178-9dcb-e0bf8a3e5f44.png)

![image](https://user-images.githubusercontent.com/1508058/146912767-6f88c814-266d-4bd1-b06c-5321cc8c7f64.png)

jeremydmiller:
[...]

PR #2019, opened 2021-12-21
Test case for #2018: Full Text Search on string list

lngr:


mysticmind:
I have pulled in the failing test with the other PR https://github.com/JasperFx/marten/pull/2105 with a fix.

Issue #2011, opened 2021-12-16
Every created stream has the same "created" timestamp


kdlslyv:
Marten version: 4.3.0

![image](https://user-images.githubusercontent.com/13890151/146375136-ffb5dd7f-cf58-4954-89ed-81a020a8944c.png)

using `sess.Events.StartStream()` with or without type, it'll always use the same `created` timestamp from the first stream ever created. 

Very weird behavior - the `timestamp` column works as expected.

Could it be related to `StreamsTable.cs`?

for the `created` column, it's using `DefaultValueByString`:
```csharp
            .AddColumn(new StreamTableColumn("created", x => x.Created)
            {
                Writes = false, Type = "timestamptz"
            }).NotNull().DefaultValueByString("(now())");
```

for the `timestamp` column, it's using `DefaultExpression`:
```csharp
DefaultExpression = "(now())"
```

jeremydmiller:
@kdlslyv I "fixed" it in code, but I'm not confident about the database migration. You might have to patch this directly into the database to fix it permanently.

kdlslyv:
Thank you for properly confirming and fixing the issue 👍

I couldn't ascertain if it'd be fixed automatically by database migration, but I'd doubt it after haphazardly digging around while certainly not being knowledgeable regarding that mechanism 😄 

The patch is easily applied:
![image](https://user-images.githubusercontent.com/13890151/147355137-1042b641-4c91-4066-88e3-83e6b4296cb5.png)

PR #2022, opened 2021-12-21
Fixed default value for mt_streams.created. Closes GH-2011

jeremydmiller:


jeremydmiller:
Blech. It's gonna take a database migration to address this when a user takes it in.

PR #2036, opened 2022-01-06
Ensure ViewProjection uses CrossStreamAggregationRuntime

### LINKED ISSUE #2031: EventSlice incorrectly being marked as new
I think this should have been fixed by this: https://github.com/JasperFx/marten/pull/1952 (https://github.com/JasperFx/marten/issues/1947)

However, I'm not seeing my `ViewProjection`s use the correct runtime, they are still using `AggregationRuntime<,>` instead of `CrossStreamAggregationRuntime<,>` and so I'm still hitting the issue where the current document isn't being loaded when the first event in a slice is from a new stream. I'm using v4.3.0 

When I code-gen ahead of time, this projection:

```
public class  OutcomeProjection : ViewProjection<Outcome, Guid>
{
    // ...
}
```

Results in handlers like this:

```
    // START: OutcomeProjectionInlineHandler
    public class OutcomeProjectionInlineHandler : Marten.Events.Aggregation.AggregationRuntime<Foo.Outcome, System.Guid>
    {
```

The projection is registered with `ProjectionLifecycle.Async` and when debugging it seems to us the InlineHandler which is why I was looking at that.

I don't really understand how this works so I might be completely wrong but should these lines: https://github.com/JasperFx/marten/blob/9a51fc5f1c93673dc7b3846b46934b5b65eae2ec/src/Marten/Events/Aggregation/AggregateProjection.CodeGen.cs#L234-L235 be using the new `baseTypeForAggregationRuntime` method?

```
var inlineBaseType =  baseTypeForAggregationRuntime()
```

Thanks!

Rob89:
Fixes #2031 

The test from the original fix has been modified to ensure that the daemon applies the events from the other streams in separate slices rather than all together. The test was flawed as it loaded all 3 events into a slice. The `IsNew` logic looks at the first event to determine that it's a new projection and then applies all 3 events on the same document resulting in no loss of data. When they are applied in separate slices, the data loss does occur as they are no longer all applied on the same document.

### DIFF
--- a/src/Marten.AsyncDaemon.Testing/view_projection_end_to_end.cs
+++ b/src/Marten.AsyncDaemon.Testing/view_projection_end_to_end.cs
@@ -57,19 +57,23 @@ namespace Marten.AsyncDaemon.Testing
                 await session.SaveChangesAsync();
             }
 
+            // We need to ensure that the events are not processed in a single slice to hit the IsNew issue on multiple
+            // slices which is what causes the loss of information in the projection.
+            await daemon.Tracker.WaitForShardState("UserIssue:All", 4, 15.Seconds());
+
             using (var session = theStore.LightweightSession())
             {
                 session.Events.Append(issue2, new IssueCreated {UserId = user1, IssueId = issue2});
                 await session.SaveChangesAsync();
             }
 
+            await daemon.Tracker.WaitForShardState("UserIssue:All", 5, 15.Seconds());
+
             using (var session = theStore.LightweightSession())
             {
                 session.Events.Append(issue3, new IssueCreated {UserId = user1, IssueId = issue3});
                 await session.SaveChangesAsync();
             }
-
-
 
             await daemon.Tracker.WaitForShardState("UserIssue:All", 6, 15.Seconds());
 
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=2036) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=2036) before we can accept your contribution.<br/><br/>:x: Rob89 [sign now](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=2036)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/JasperFx/marten?pullRequest=2036) it.</sub>

jeremydmiller:
@Rob89 I'm very sorry this has hung out there so long. Would you mind if we try to have this rolled out into a v5 alpha or even v5 soon? Or do you need this in a 4.* branch release?

Rob89:
Hey, no problem. We're happy to work around it for now so feel free to put it into v5.

PR #2067, opened 2022-01-31
Optimizes Safe Sequence Finder using Current Safe Sequence Id

kluhman:
Rather than doing a full table scan on the `mt_events` table to filter down to rows past a given timestamp, updates query to take advantage of index by also filtering out rows before the latest safe sequence id


oskardudycz (approved):
Looks good to me 👍

PR #2098, opened 2022-02-28
Support adding metadata to individual events in the same session

elexisvenator:
A set of methods have been added to IEventStore to allow metadata (correlationids, causationids and user defined headers) to be added to one or more specific events within a session. This is in addition to standard session metadata, and overrided values take precedence.  There is also a method that allows for event metadata to be copied from an existing IEvent to other events.

Documentation has been updated where applicable and new tests have been written to cover the new functionality.
I have also updated the copy and transform stream scenario to include usage of this new metadata - as this scenario is the motivation behind supporting this functionality.

Other fixes:
- Saving null objects when using SystemTextJson caused a null reference exception (eg headers enabled and write event with no headers). New behaviour is to detect nulls and return a hardcoded "null" string.
- When calling IEvent.GetHeader with headers disabled (or no headers on event), null would be returned. However if headers were enabled and the dictionary exists, a KeyNotFoundException is thrown. Fixed this to also return null.

### DIFF
--- a/docs/events/metadata.md
+++ b/docs/events/metadata.md
@@ -29,7 +29,7 @@ The actual metadata is accessible from the `IEvent` interface or `Event<T>` even
 <!-- snippet: sample_IEvent -->
 <a id='snippet-sample_ievent'></a>
 ```cs
-public interface IEvent
+public interface IEvent : IEventMetadata
 {
     /// <summary>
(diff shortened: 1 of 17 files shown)


elexisvenator:
Closing PR after discussion on a better approach (exposing apis for StreamAction)

Issue #1875, opened 2021-08-23
Duplicated array field not supported


KasperDamgaard:
Using Marten 4.x 

If I do the following:
`registry.For<TreatmentCourse>().Duplicate(tc => tc.PaymentGuaranteeIds, "uuid[]");`
where paymentGuaranteesIds are an array of Guid's.

And when I then try to run the following query:
`session.Query<TreatmentCourse>().FirstOrDefaultAsync(treatmentCourse => treatmentCourse.PaymentGuaranteeIds.Any(id => id == paymentGuaranteeId))`

I get the exception:
`Unable to cast object of type 'Marten.Linq.Fields.DuplicatedField' to type 'Marten.Linq.Fields.ArrayField'.`

The exception stems from this place in the code: https://github.com/JasperFx/marten/blob/8721b7fe55ff5e30fa409abeae1593b3058468b3/src/Marten/Linq/Parsing/WhereClauseParser.cs#L261

PR #2122, opened 2022-03-05
Fix GH-1875

mysticmind:
- Fixes GH-1875
- Add unit test

### DIFF
--- a/src/DocumentDbTests/Bugs/Bug_1875_duplicated_array_field_test.cs
+++ b/src/DocumentDbTests/Bugs/Bug_1875_duplicated_array_field_test.cs
@@ -1,0 +1,40 @@
+using System.Collections.Generic;
+using System.Linq;
+using Marten.Services;
+using Marten.Testing.Documents;
+using Marten.Testing.Harness;
+using NpgsqlTypes;
+using Shouldly;
+using Weasel.Core;
+using Xunit;
+
+namespace DocumentDbTests.Bugs
+{
+    public class Bug_1875_duplicated_array_field_test : BugIntegrationContext
+    {
+        [Fact]
+        public void query_on_duplicated_array_field_test()
+        {
+            StoreOptions(_ =>
+            {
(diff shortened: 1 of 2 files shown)


src/DocumentDbTests/Bugs/Bug_1875_duplicated_array_field_test.cs
```diff
@@ -0,0 +1,40 @@
...
+using Shouldly;
+using Weasel.Core;
+using Xunit;
+
+namespace DocumentDbTests.Bugs
+{
+    public class Bug_1875_duplicated_array_field_test : BugIntegrationContext
+    {
+        [Fact]
+        public void query_on_duplicated_array_field_test()
```
oskardudycz, line 16: **SUGGESTION:** @mysticmind, it'd be worth also checking the `Guid` array as in the original array. We had cases in the past that types like string, guids, dates were failing, while for numbers, it was fine, and we were lacking tests for them.

oskardudycz (approved):
Babu, you're on 🔥 😄

src/DocumentDbTests/Bugs/Bug_1875_duplicated_array_field_test.cs
mysticmind: @oskardudycz, the issue was reproducible for the unit test which I have added so we are good.

mysticmind: There is no harm in adding another test using `Guid` array, will add it.

mysticmind: Added another unit test for Guid array duplicate field.

oskardudycz: Thank you, Babu! 🙂

PR #2124, opened 2022-03-08
ViewProjections support `Identity<TEvent>()` with no function - uses stream id

elexisvenator:
If events are designed to not include the stream id on their event object (instead relying on IEvent metadata to store the id), it can be difficult to build view projections for theses events, as only a custom grouper/slicer exposes the IEvent.

This change adds a method `Idientity<TEvent>()` to view projections. This overload doesnt take a function and assumes that matching events will use their stream id/key to map to the document aggregate id.

Examples of how this works (from new tests):

```cs
public class UserGroupsAssignmentProjectionUsingSelfAggregate: ViewProjection<UserGroupsAssignment, Guid>
{
    public UserGroupsAssignmentProjectionUsingSelfAggregate()
    {
        // This is just specifying the aggregate document id
        // per event type. This assumes that each event
        // applies to only one aggregated view document
        Identity<UserRegistered>();
        Identity<SingleUserAssignedToGroup>(x => x.UserId);
    }

    public void Apply(UserRegistered @event, UserGroupsAssignment view)
        => view.Id = @event.UserId;

    public void Apply(SingleUserAssignedToGroup @event, UserGroupsAssignment view)
        => view.Groups.Add(@event.GroupId);
}
```

And with a one-to-many fanout (was quite happy to find this just works):

```cs
public class UserSocialCircleProjection: ViewProjection<UserSocialCircle, Guid>
{
    public UserSocialCircleProjection()
    {
        Identity<UserRegistered>();

        // Multiple Identity methods can be called for the same event
        // Useful for describing different projection mappings
        Identity<UserFollowsUsers>();
        Identities<UserFollowsUsers>(x => x.FollowingUsers);
    }

    public void Apply(UserRegistered @event, UserSocialCircle view)
        => view.Id = @event.UserId;

    public void Apply(UserFollowsUsers @event, UserSocialCircle view)
    {
        if (@event.UserId == view.Id)
        {
            view.FollowingUsers.AddRange(@event.FollowingUsers.Where(user => !view.FollowingUsers.Contains(user)));
        }

        if (@event.FollowingUsers.Contains(view.Id) && !view.FollowedByUsers.Contains(@event.UserId))
        {
            view.FollowedByUsers.Add(@event.UserId);
        }
    }
}
```

Other changes:
- Fixed DateTime>DateTimeOffset implicit conversions in some unit tests to get all tests green.
- Added codedoc for `Identity` and `Identities` methods

### DIFF
--- a/src/EventSourcingTests/Projections/ViewProjections/simple_multi_stream_projection_with_one_to_many.cs
+++ b/src/EventSourcingTests/Projections/ViewProjections/simple_multi_stream_projection_with_one_to_many.cs
@@ -1,0 +1,133 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Marten.Events.Projections;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests.Projections.ViewProjections.SimpleWithOneToMany
+{
+    #region sample_view-projection-simple-with-one-to-many
+    public class UserGroupsAssignmentProjection: ViewProjection<UserGroupsAssignment, Guid>
+    {
+        public UserGroupsAssignmentProjection()
+        {
+            Identity<UserRegistered>(x => x.UserId);
+            Identities<MultipleUsersAssignedToGroup>(x => x.UserIds);
+        }
+
+        public void Apply(UserRegistered @event, UserGroupsAssignment view)
+        {
+            view.Id = @event.UserId;
+        }
+
+        public void Apply(MultipleUsersAssignedToGroup @event, UserGroupsAssignment view)
+        {
+            view.Groups.Add(@event.GroupId);
+        }
+    }
+
+    #endregion
+
+    public class simple_multi_stream_projection_with_one_to_many: OneOffConfigurationsContext
+    {
+        [Fact]
+        public async Task multi_stream_projections_should_work()
+        {
+
+            // --------------------------------
+            // Create Groups
+            // --------------------------------
+            // Regular Users
+            // Admin Users
+            // --------------------------------
+
+            var regularUsersGroupCreated = new UserGroupCreated(Guid.NewGuid(), "Regular Users");
+            theSession.Events.Append(regularUsersGroupCreated.GroupId, regularUsersGroupCreated);
+
+            var adminUsersGroupCreated = new UserGroupCreated(Guid.NewGuid(), "Admin Users");
+            theSession.Events.Append(adminUsersGroupCreated.GroupId, adminUsersGroupCreated);
+
+            await theSession.SaveChangesAsync();
+
+            // --------------------------------
+            // Create Users
+            // --------------------------------
+            // Anna
+            // John
+            // Maggie
+            // Alan
+            // --------------------------------
+
+            var annaRegistered = new UserRegistered(Guid.NewGuid(), "Anna");
+            theSession.Events.Append(annaRegistered.UserId, annaRegistered);
+
+            var johnRegistered = new UserRegistered(Guid.NewGuid(), "John");
+            theSession.Events.Append(johnRegistered.UserId, johnRegistered);
+
+            var maggieRegistered = new UserRegistered(Guid.NewGuid(), "Maggie");
+            theSession.Events.Append(maggieRegistered.UserId, maggieRegistered);
+
+            var alanRegistered = new UserRegistered(Guid.NewGuid(), "Alan");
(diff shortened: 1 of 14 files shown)


src/EventSourcingTests/Projections/ViewProjections/simple_multi_stream_projection_with_one_to_many.cs
```diff
@@ -125,6 +125,7 @@ public simple_multi_stream_projection_with_one_to_many()
         {
             StoreOptions(_ =>
             {
+                _.DatabaseSchemaName = "simple_multi_stream_projection_one_to_many";
```
jeremydmiller, line 128: The test harness automatically sets the schema name to the class name of the test fixture

src/Marten/Events/Aggregation/TenantSliceGroup.cs
```diff
@@ -35,6 +34,12 @@ public interface ITenantSliceGroup<TId>
         /// <param name="events"></param>
         void AddEvents(TId id, IEnumerable<IEvent> events);
 
+        /// <summary>
+        /// Add events of type TEvent to streams by their self aggregate id (stream id/key)
+        /// </summary>
+        /// <param name="events"></param>
+        void AddEvents<TEvent>(IEnumerable<IEvent> events);
```
jeremydmiller, line 41: Can you just rename the method to make it clear you're sorting events by their stream identity? I'd never guess what the method actually does by the method name here.

src/Marten/Events/Projections/SingleStreamGrouperByEventStreamId.cs
```diff
@@ -0,0 +1,18 @@
...
+using Marten.Events.Aggregation;
+
+namespace Marten.Events.Projections
+{
+    /// <summary>
+    /// Assigns an event to only one stream based on its stream id
+    /// </summary>
+    /// <typeparam name="TId"></typeparam>
+    /// <typeparam name="TEvent"></typeparam>
+    internal class SingleStreamGrouperByEventStreamId<TId, TEvent>: IGrouper<TId> where TEvent : notnull
```
jeremydmiller, line 11: I think I'd vote for a unit test directly against this thing

src/Marten/Events/Projections/ViewProjection.cs
```diff
@@ -109,6 +115,25 @@ public void Identity<TEvent>(Func<TEvent, TId> identityFunc)
             _groupers.Add(new SingleStreamGrouper<TId, TEvent>(identityFunc));
         }
 
+        /// <summary>
+        /// Use the TEvent event's stream key/id as the aggregate document id
+        /// </summary>
+        /// <typeparam name="TEvent"></typeparam>
+        /// <exception cref="InvalidOperationException"></exception>
+        public void Identity<TEvent>()
```
jeremydmiller, line 123: We need to rename this method to make it clear that the identity is the stream identity

jeremydmiller:
@elexisvenator I'm still iffy on this. Is there just no way you could redesign the event stream a little bit so you can do a simple aggregate by stream? When you capture `SingleUserAssignedToGroup`, can you just capture an additional event at the exact same time to whatever stream type the `UserRegistered` event is?

There's a couple comments regardless.

elexisvenator:
@jeremydmiller Thanks for the review, happy to make changes based on feedback but probably worth discussing the suitability of this change first.

My team doesn't actually need this right now (we are moving towards having a logical id that is different from the stream id). But I still feel there is a gap of some sort here. If a `SelfAggregate` is the intended default path, then a view projection is a deviation that allows flexibility. - `Identity` allowing some, CustomGroupers/Slicers allowing more. It seems strange that if you need a ViewProjection for some of your events that it actually becomes a lot harder to do the "default" behaviour that SelfAggregate does (at minimum you need to have a custom grouper to get the streamid/key from the IEvent.

It may also make sense in a future version to increase the power of  SelfAggregates by letting them define more complicated event mapping. Eg the codegen could pick up a method like the following on a selfAggregate:

```csharp
public static void MapEvents(IEventMapper mapper)
{
    mapper.Identity<UserRegistered>();
    mapper.Identity<SingleUserAssignedToGroup>(x => x.UserId);
}
```

jeremydmiller:
@elexisvenator In that case I'd prefer to just shut this down. I don't think this should be a common need and right now it would just be a confusing addition to the API IMO without more precise naming.

elexisvenator:
Agreed, closing

PR #2130, opened 2022-03-11
Update FastExpressionCompiler and move to internal package

Hawxy:
Updates the years old copy of FEC and replaces the vendored code with the internal-only package that FEC provides.


mysticmind (approved):
Looks good to me 👍

jeremydmiller:
Are we even using that internally anymore? I might be wrong, but I can't think of any place where that's still being used.

Hawxy:
It only ends up getting called here via `MethodCollection`'s construction:
https://github.com/JasperFx/marten/blob/master/src/Marten/Events/CodeGeneration/MethodCollection.cs#L193

PR #2131, opened 2022-03-15
Fix link to IIdGeneration.cs

juchom:


mysticmind (approved):
Thanks for the PR @juchom

Issue #2134, opened 2022-03-17
documentStore.QuerySession(SessionOptions.ForConnectionString(...) Not respecting connectionString


darraghjones:
Hi,

I'm trying to write a custom ISessionFactory to allow me to use a differnent connection string for queries. With the code below, the session is still opened with the connection string from the default `StoreOptions`. 

Also, if I 'get' the connection from the session object before I use it (as in the commented out code below), this throws an error:
`System.InvalidOperationException: Connection is not open`

```
internal class AuroraSessionFactory : ISessionFactory
{
    private readonly IDocumentStore store;

    public AuroraSessionFactory(IDocumentStore store)
    {
        this.store = store;
    }

    public IQuerySession QuerySession()
    {
        var session = store.QuerySession(SessionOptions.ForConnectionString("host=readonly;password=password;username=postgres;database=test6"));

        // var connection = session.Connection;
        // var connectionString = connection.ConnectionString;
        // Console.WriteLine(connectionString);
        return session;
    }

    public IDocumentSession OpenSession()
    {
        return store.LightweightSession();
    }
}
```

**I'm using version 5.0.0-rc1**

PR #2135, opened 2022-03-17
Use provided connection string when opening session

VilleHakli:
If session is opened by providing connection string using `documentStore.OpenSession(SessionOptions.ForConnectionString(connectionString))` the given connection string is not used. The issue seems to be that `MartenControlledConnectionTransaction` does not use the connection created by `SessionOptions.ForConnectionString(connectionString)`.


VilleHakli:
There is also bug #2134 about this.

jeremydmiller:
@VilleHakli I was literally getting things set up this morning to address this when I saw your PR. Thanks for doing this!

PR #2141, opened 2022-03-21
Add support for asserting database configuration on startup

VilleHakli:
This PR adds support for asserting database configuration on startup. This feature is implemented the same way as applying all database changes on startup.

If `AssertDatabaseMatchesConfigurationOnStartup` is called for `MartenConfigurationExpression`, during startup `AssertDatabaseMatchesConfigurationAsync` is called for all known databases. As possible configuration mismatch will throw exception this also prevents the application from starting if any database does not match configuration.


jeremydmiller:
@VilleHakli It's not documented yet (next up on my list, no, seriously!), but there's already an environment check for Oakton + Marten in V5 that would serve the same function.

Issue #2143, opened 2022-03-22
AppendExclusive example code from documentation does not work


vmartos:
Hi!

The code sample from the following page https://martendb.io/events/appending.html#serialized-access-to-the-stream (AppendExclusive) does not seem to work, when saving changes it throws the following exception:

Marten.Exceptions.EventStreamUnexpectedMaxEventIdException
  HResult=0x80131500
  Message=Unexpected starting version number for event stream 'f09e05a1-df8f-4b83-a343-1a3e760a4093', expected 3 but was 1
  Source=Marten
  StackTrace:
   at Marten.Events.StreamAction.PrepareEvents(Int64 currentVersion, EventGraph graph, Queue`1 sequences, IMartenSession session)
   at Marten.Events.EventGraph.<ProcessEventsAsync>d__87.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Marten.Internal.Sessions.DocumentSessionBase.<SaveChangesAsync>d__54.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at TestMartenAppendExclusive.Program.<Main>d__0.MoveNext() in C:\Users\v.martos\source\repos\TestMartenAppendExclusive\TestMartenAppendExclusive\Program.cs:line 46

the code sample is the following:

```
using Marten;
using Microsoft.Extensions.DependencyInjection;
using System;
using System.Collections.Generic;
using System.Threading.Tasks;
using Weasel.Postgresql;

namespace TestMartenAppendExclusive
{
    internal class Program
    {
        static async Task Main(string[] args)
        {
            var serviceCollection = new ServiceCollection();
            var databaseSchema = "test_marten";

            var martenConfiguration = serviceCollection.AddMarten(options =>
            {
                options.Connection("Username=admin;Password=changeit;Host=localhost;Port=5432");
[...]

PR #2144, opened 2022-03-23
Failing unit test for bug #2143

vmartos:
This PR contains a failing unit test demonstrating that calling AppendExclusive with no events and then calling Append fails with `Marten.Exceptions.EventStreamUnexpectedMaxEventIdException`


jeremydmiller:
Got it, and thank you!

PR #2146, opened 2022-03-29
Test to support custom field sources.

elexisvenator:
Still a work in progress.

Using string-backed struct as a custom object.
Currently works with `CustomId` and `CustomId?`. Does not yet work with `CustomId[]`.

Once working, can be added to documentation.


src/DocumentDbTests/Reading/Linq/invoking_query_with_custom_field_source.cs
```diff
@@ -0,0 +1,263 @@
...
+                {
+                    field = new NullableTypeField(new CustomIdField(dataLocator, serializer.Casing, members));
+                    return true;
+                }
+
+                field = null;
+                return false;
+            }
+
+            private static Type? GetRawMemberType(MemberInfo member)
```
jeremydmiller, line 166: There's an extension method in either Marten or Baseline to do this already. GetMemberType() I believe.

elexisvenator: I updated the extension method to make it public

elexisvenator:
@jeremydmiller I might need some help to make the `CustomId[]` scenario work, I'm a bit lost with how to make detection of arrays work.

elexisvenator:
I cleared a bunch of issues by registering a mapping for `CustomId` to dbtype in the PostgresqlProvider.
That leaves array contains, which i know what this issue is - 
In most places where a constant value for the custom id is passed in, it hits the CustomIdField.CreateComparison() method. This allows me to coerce the constant CustomId that is passed in into the raw value type the db needs.
```cs
            var def = new CommandParameter(((CustomId)value.Value).Value);
            return new ComparisonFilter(this, def, op);
```
But for the array.contains scenario specifically this doesn't happen. Instead the field maps to `ArrayField`, and the constant is converted to a CommandParameter in the `ContainsIdSelectorStatement` constructor.  The CommandParameter correctly asserts what the dbparametertype should be but doesnt do any conversion on the CustomId, so the npgsql command fails when it executes, not knowing how to process `CustomId` as varchar.

I think what needs to happen is when registering a mapping in the PostgresProvider for the custom id, some sort of function needs to be passed in eg `Func<TInputType, TRawType>`. This could then be used to preprocess the custom id whenever a lookup is done. This would be a pretty significant change in weasel though.

jeremydmiller:
I think there's an evil downcast in the Linq parsing code to ArrayField that's messing this up.

jeremydmiller:
@elexisvenator Is the only thing missing at this point the Contains() operation?

jeremydmiller:
Oh, and I could probably try to finish this off on Monday if you wanted.

elexisvenator:
I went through all documented functions and this is the list of functions that fail:

- `CustomId,IsOneOf()`
- `CustomId[].Any()` (with predicate)
- `!CustomId[].Any()` (with predicate)
- `CustomId[].Contains()`
- `!CustomId[].Contains()`
- `CustomId[].IsOneOf()`

They all seem to be failing on the same reason. (Can't map a customid value to a varchar postgres value)

jeremydmiller:
Everything has changed since this one, just cleaning up old things

PR #2149, opened 2022-03-30
Parameterize tenantid when checking stream version

Jordan466:


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=2149) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=2149) before we can accept your contribution.<br/><br/>:x: Jordan466 [sign now](https://cla.dotnetfoundation.org/JasperFx/marten?pullRequest=2149)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/JasperFx/marten?pullRequest=2149) it.</sub>

src/Marten/Events/EventStore.cs
```diff
@@ -245,8 +247,9 @@ public async Task AppendExclusive(string streamKey, CancellationToken token, par
             _store.Events.EnsureAsStringStorage(_session);
             await _session.Database.EnsureStorageExistsAsync(typeof(IEvent), token).ConfigureAwait(false);
 
-            var cmd = new NpgsqlCommand($"select version from {_store.Events.DatabaseSchemaName}.mt_streams where id = :id for update")
-                .With("id", streamKey);
+            var cmd = new NpgsqlCommand($"select version from {_store.Events.DatabaseSchemaName}.mt_streams where id = :id and tenant_id = :tenant_id for update")
```
jeremydmiller, line 250: This is a good catch, but we'll have to do some conditional logic here and only use the tenant_id version of the command when the event store is configured to be multi-tenanted. The tenant_id column only exists in that case.

jeremydmiller:
@Jordan466 Thanks for taking the time to troubleshoot this and do the PR! Do see the comment though, you have to account for whether or not multi-tenancy is present for the event store.

jeremydmiller:
Hey all, just to speed this up, I'm taking it in, then making the easy fix just to get 5.1 out. Thanks!

src/Marten/Events/EventStore.cs
jeremydmiller: It is, got it, and getting it applied with just a little structural improvement to the bigger EventSTore class that's gotten too big. Thank you so much for doing this, I pulled it down before I noticed you'd made the fix. This will be in 5.1.0 sometime in the next week or so.

PR #2152, opened 2022-03-30
Filter out Live projections in command line rebuild

### LINKED ISSUE #2151: Command line tools rebuild attempts to run Live projections
When live projections are registered, using the `rebuild` flag in the command line tooling causes projections with the `Live` lifecycle to be run through the async daemon which fails when it tries to truncate a table that doesn't exist.

The workaround I'm using for now is to use the interactive flag which filters out live projections.

Example error:

```
ERROR: Marten.Exceptions.MartenCommandException: Marten Command Failure:$
truncate table public.mt_doc_timelineaggregate CASCADE;delete from strategy_execution_event.mt_event_progression where name = :p0;$
$
42P01: relation "public.mt_doc_timelineaggregate" does not exist
```

Output of projections list:
image

Rob89:
Fixes #2151

### DIFF
--- a/src/Marten.CommandLine/Commands/Projection/ProjectionCommand.cs
+++ b/src/Marten.CommandLine/Commands/Projection/ProjectionCommand.cs
@@ -46,7 +46,7 @@ namespace Marten.CommandLine.Commands.Projection
 
         private async Task<bool> Rebuild(ProjectionInput input, DocumentStore store)
         {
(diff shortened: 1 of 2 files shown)


Rob89:
I couldn't find any tests for this and wasn't sure how to write one - I'm happy to add some if someone can give me a hint to get started on it

PR #2162, opened 2022-04-03
Add Event Sourcing hello world sample

jchannon:
This moves some existing projects into a new SolutionItem folder called Samples. Also added a Event Sourcing hello world sample app in a new samples folder in the repo, a placeholder for future samples that can be added.

This is the output from this sample app:
Screenshot 2022-04-03 at 17 39 42


mysticmind (approved):
Looks good to me 👍

PR #2168, opened 2022-04-07
Fix concurrency error when calling append several times for string identity

Leh2:
Before the fix the new test will fail with `Unexpected starting version number for event stream 'xxxx', expected 3 but was 2`.
This happened because it wasn't aware of previous calls to `Append`.

Same fix is previous applied for guid identify in https://github.com/JasperFx/marten/commit/b860265961b414b3840bae8bdea0445015febe42


jeremydmiller:
Technically not a concurrency issue per se, I'd describe it as a "re-entrant issue". Gonna grab this, and also apply the same fix to Guid identified streams. Will be in 5.1 in the next couple days.

jeremydmiller:
Nevermind, no idea why that was only like that on the string identified side.

PR #2175, opened 2022-04-13
Stop daemon and release db lock when exception occur on shard start

Leh2:
When `startAllProjections` fails for whatever reason (properly mostly transient db errors) the hot cold coordinator is in a limbo state where some shards could be started and others not. The lock got attained in the first place but the hot cold coordinator still tries to get a lock, but never will.

The proposed solution is to stop the daemon, release the lock and let the hot cold coordinator try to start it again. 

https://github.com/JasperFx/marten/blob/3e6099a6c595bd98573bb57c3e97ba39902330d1/src/Marten/Events/Daemon/HotColdCoordinator.cs#L84-L94


jeremydmiller:
I see it, and yep. I'm pulling this in. There is some retry action happening on each shard start, but since it's all or nothing right now, it makes sense to do it this way at least for now.

PR #2176, opened 2022-04-13
Failing Test for Tenancy Not Being Set on Query Session on Async Projections

kluhman:
Adds a failing unit test to demonstrate that tenancy is not being properly set on query sessions injected into an async projection when using a conjoined tenancy model.


mysticmind:
This test is pulled in as part of bug fix https://github.com/JasperFx/marten/pull/2178. Closing this one.

Issue #2177, opened 2022-04-13
Async Daemon does not set Tenant Correctly on Query Sessions using Conjoined Multi-Tenacy


kluhman:
It appears that when using the Conjoined Multi-Tenancy model, the async daemon does not set the tenancy on `IQuerySession` injected into projections correctly, resulting it be unable to load data that should be available. A failing, reproducible unit test has been submitted as in PR #2176 

Here is a summary of the configuration and projection used

```csharp
using var documentStore = SeparateStore(options =>
{
    options.Policies.AllDocumentsAreMultiTenanted();
    options.Events.TenancyStyle = TenancyStyle.Conjoined;
    options.Events.StreamIdentity = StreamIdentity.AsGuid;
    options.Events.MetadataConfig.EnableAll();

    options.Projections.Add<TicketProjection>();
});

public class User
{
    public Guid Id { get; set; }
    public string FirstName { get; set; }
    public string LastName { get; set; }
}

public class Ticket
{
    public Guid Id { get; set; }
    public string Name { get; set; }
    public User User { get; set; }
}

public class TicketProjection: AggregateProjection<Ticket>
{
    public TicketProjection()
    {
        Lifecycle = ProjectionLifecycle.Async;
    }

    public Ticket Create(TicketCreated created) =>
        new() { Id = created.TicketId, Name = created.Name };

    public void Apply(Ticket ticket, TicketAssigned assigned, IQuerySession session) =>
        ticket.User = session.Load<User>(assigned.UserId);
}
```

PR #2178, opened 2022-04-14
Fix GH-2177 set tenancy properly for query session for async projections

mysticmind:


oskardudycz (approved):
Nice @babuannamalai 👍

PR #2186, opened 2022-04-25
Starts Async Daemon when Store is Eagerly Initialized

### LINKED ISSUE #2185: Async Daemon Does Not Run When Eager Initializing Store
Marten Version: 5.2.0

When using the `InitializeStore()` functionality to eagerly build the document store, the async daemon will not run.

```csharp
services.AddMarten(options =>
{
    ...
}).AddAsyncDaemon(DaemonMode.Solo).InitializeStore();
``` 

It looks like using `InitializeStore` uses the in-memory `StoreOptions` object built during the `AddMarten` call, where as the async daemon uses an object built using `services.Configure`. So calling `InitializeStore` builds and caches the store using options with `AsyncMode = Disabled`, and when the daemon starts, it fails to run. Removing the `InitializeStore()` call fixes the issue.

I'll try to create a PR to fix this tonight

kluhman:
Fixes #2185

Sets the `AsyncMode` on the in-memory `StoreOptions` when adding the async daemon so it operates when users use the `InitializeStore` functionality.

### DIFF
--- a/src/Marten.AsyncDaemon.Testing/DaemonServiceRegistrationTests.cs
+++ b/src/Marten.AsyncDaemon.Testing/DaemonServiceRegistrationTests.cs
@@ -130,5 +130,30 @@ namespace Marten.AsyncDaemon.Testing
             logger.Received().LogDebug("Stopping the asynchronous projection agent");
         }
 
+        [Fact]
+        public async Task should_start_when_eagerly_initializing_the_store()
+        {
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@kluhman I'd actually rather solve this by eliminating the InitializeStore() method tbh.

jeremydmiller:
I made that method [Obsolete] instead with instructions for alternative approaches. Just like we discussed, so I'm not taking this one in.

PR #2191, opened 2022-04-27
Combine sql and parameter(s) to one log statement

Leh2:
When enabling debug logs the sql parameters was quite spammy


jeremydmiller:
Got it, and I like that. Thank you!

Issue #2195, opened 2022-04-29
Projections don't work when changing JSON Settings


phillip-haydon:
I have not verified this using Newtonsoft.Json but it's a bug that affects System.Text.Json.

Given the following:

```
            x.Serializer(new SystemTextJsonSerializer
            {
                EnumStorage = EnumStorage.AsString,
                Casing = Casing.Default
            });
```

If you have a document such as:

```
public class Player
{
    [JsonInclude]
    public int Id { get; private set; }

    public string Name { get; set; }

    public decimal Level { get; set; }

    public PlayerStatus Status { get; set; }
}
```

And you want to project this to say:

```
public class PlayerResponse
{
    public string Name { get; set; }

    public decimal Level { get; set; }
}
```

Querying like:

```
var skills = await conn.Query<Player>()
                       .Where(x => x.Status == PlayerStatus.Banned)
                       .Select(x => new PlayerResponse
                       {
                           Name = x.Name,
                           Skill = x.Skill
                       })
                       .ToListAsync();
```

This works... HOWEVER, if you (wipe the database, and start fresh) change the configuration:

```
            x.Serializer(new SystemTextJsonSerializer
            {
                EnumStorage = EnumStorage.AsString,
                Casing = Casing.CamelCase
            });
```

Everything works perfectly until you get to the project. 

It issues a query like so:

```
select jsonb_build_object('Name', d.data ->> 'name', 'Level', CAST(d.data ->> 'level' as decimal)) as data
from game.mt_doc_player as d
where d.data ->> 'Status' = :p0;
```

This results in a collection of empty objects. `Name` is `null` and `Level` is `0`.

I believe this is due to `PropertyNameCaseInsensitive` not being set to `true` in the serializer.

So there's 2 ways we can fix this I believe...

1) Change the serializer to be case insensitive. (This is my preferred approach)

2) Change the querying to use the same casing as the properties of the JSON. i.e 

`jsonb_build_object('name', d.data ->> 'name')`

`Name` -> `name` since the data property is `name`.

jeremydmiller:
@phillip-haydon 

"
1. Change the serializer to be case insensitive. (This is my preferred approach)

1. Change the querying to use the same casing as the properties of the JSON. i.e
"

**Number 1 isn't feasible.** Number 2 is already the behavior, so I'm not sure what you want to happen here. The default casing means that the Json is assumed to exactly match the property or field name. 

[...]

PR #2197, opened 2022-04-29
Issue #2195

phillip-haydon:
This is a failing test for issue #2195

Added a new property to the serializer to assign `PropertyNameCaseInsensitive` and ensure it's not a break change to everything else.

image


jeremydmiller:
@phillip-haydon I don't think this is doable man. The postgresql operators that we use throughout *most* of the Linq provider are case sensitive. It's not feasible to try to rewrite that IMO.

Issue #2201, opened 2022-04-29
Projection Out of Order Exceptions Thrown when Projection Batch Contains Only Hard Deletes


kluhman:
**Marten Version:** 4.x and 5.x

I've observed individual projections stop running and throw the `ProgressionProgressOutOfOrderException`, however, it is not related to having multiple instances of the daemon running, or the shard starting multiple times. It appears to be a problem only under specific situations that alter the number of records affected in the same operation.

In my situation, I have narrowed this down to when a batch of events only contains hard deletes. Reproducible unit test is incoming

jeremydmiller:
I'm going to swear the root cause is an Npgsql error in how it calculates the `DbDataReader.RowsEffected` property when you have a composite command w/ only 2 commands. I changed the logic in UpdateProjectionProgress to accommodate it though.

PR #2202, opened 2022-04-29
Reproducible Test for Projection Out of Order Exceptions when Batch Contains Only Hard Deletes

kluhman:
A unit test reproducing behavior explained in #2201 and discussed in Gitter


jeremydmiller:
Got it with a rebase.

PR #2203, opened 2022-05-02
Fixes for 5.3

jeremydmiller:
Mostly to trigger off the CI:)


toanxyz:
Hi @jeremydmiller, do we need to update the LamarCodeGeneration version here as well?https://github.com/JasperFx/marten/blob/master/src/Marten.CommandLine/Marten.CommandLine.csproj#L34

Issue #2174, opened 2022-04-12
Codegen with MultiTenantedDatabases option


krmapowel:
When configuring Marten with MultiTenantedDatabases option:
```Csharp
services.AddMarten(opts =>
{
    var tenants = _configuration.GetSection("Tenants").GetChildren();
    opts.MultiTenantedDatabases(x =>
    {
        foreach (var tenant in tenants)
        {
            x.AddSingleTenantDatabase(tenant.Value, tenant.Key);
        }
    });
}
```

Running `dotnet run -- codegen write` throws an exception.
```csharp
ERROR: Marten.Storage.UnknownTenantIdException: Unknown tenant id '*DEFAULT*'
   at Marten.Storage.StaticMultiTenancy.GetTenant(String tenantId)
   
 at Marten.Services.SessionOptions.Initialize(DocumentStore store, CommandRunnerMode mode)
   at Marten.DocumentStore.openSession(SessionOptions opt
ions)
   at Marten.DocumentStore.OpenSession(DocumentTracking tracking, IsolationLevel isolationLevel)
   at Marten.DocumentStore.LightweightSessio
n(IsolationLevel isolationLevel)
   at Marten.DocumentStore.LamarCodeGeneration.ICodeFileCollection.BuildFiles()
   at LamarCodeGeneration.DynamicC
odeBuilder.WriteGeneratedCode(Action`1 onFileWritten)
   at LamarCodeGeneration.Commands.GenerateCodeCommand.Execute(GenerateCodeInput input)
   at
 Oakton.OaktonCommand`1.Oakton.IOaktonCommand.Execute(Object input)
   at Oakton.CommandRun.Execute()
   at Oakton.CommandExecutor.execute(CommandR
un run)
```

Current workaround is to just set 
```csharp
opts.Connection("")
```
instead of multi tenancy, for the time of running `codegen write`.

PR #2208, opened 2022-05-05
Handle when TenantId has not been set

krmapowel:
Solves https://github.com/JasperFx/marten/issues/2174

The issue there was that TenantId was never null during code generation, so it always tried to get tenant from StaticMultiTenancy, instead of getting default.


jeremydmiller:
@babuannamalai I'm going to pull this in and make a quick test for it just to clean things off.

PR #2209, opened 2022-05-05
Big integer support in System.Text.Json

### LINKED ISSUE #2194: Support for .NET's `BigInteger`
https://docs.microsoft.com/en-us/dotnet/api/system.numerics.biginteger?view=net-6.0

Would you be open to accepting a PR for supporting .NET's type BigInteger?

I'm using this, but currently I'm working around it by treating it as a string and parsing it on a 2nd property's get/set.

This fits into PostgreSQLs data type `numeric`... sorta

> up to 131072 digits before the decimal point; up to 16383 digits after the decimal point

https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-INT

BigInteger has no bounds so according to the docs it can cause an OOM exception, which could be a number larger than 131072 digits long...

But knowing that and documenting that as a tradeoff allows it to be used within some weird scenarios people think of.

image

phillip-haydon:
- Test to verify it's working with Newtonsoft.Json
- Test to verify failure with System.Text.Json
- New converter for System.Text.Json
- Test to Serialize with Newtonsoft and Deserialize with STJ
- Test to Serialize with STJ and Deserialize with Newtonsoft

Labeled as Bug and Enhancement cos it enchances the usage of STJ and fixes the fact it was broken. (broken cos STJ is a bit turdy)

Fixes #2194

### DIFF
--- a/src/DocumentDbTests/Bugs/BigIntergerTests.cs
+++ b/src/DocumentDbTests/Bugs/BigIntergerTests.cs
@@ -8,6 +8,8 @@ namespace DocumentDbTests.Bugs
 {
     public class BigIntegerTests : BugIntegrationContext
     {
+        private static readonly string LargerThanLongValue = "123456789012345678901234567890123456789012345678901234567890";
+
         [Fact]
         public void When_Querying_Using_Newtonsoft_Json_Should_Persist_And_Fetch_BigInteger_Values()
         {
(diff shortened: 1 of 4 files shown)


oskardudycz (commented):
@phillip-haydon looks good to me, but could you add some tests on doing a filter on bigints (like Linq `Where`)?

jeremydmiller (approved):
Works for me

Issue #2211, opened 2022-05-07
Inner properties are not deserliazed when using Select projections


egorpavlikhin:
When trying to use a projection that expects an inner property to be passed as a whole object (as opposed to using just nested properties) projections fails with deserialization exceptions. See the code below:

Test to reproduce:
```csharp
/*class InnerTarget
{
    public InnerTarget(Target inner)
    {
        Inner = inner;
    }

    public Target Inner { get; set; }
}*/

var target = Target.Random(true);

theSession.Store(target);
theSession.SaveChanges();

var actual = theSession.Query<Target>()
    .Where(x => x.Id == target.Id)
    .Select(x => new InnerTarget(x.Inner))
    .First();

actual.Inner.Id.ShouldBe(target.Inner.Id);
actual.Inner.Number.ShouldBe(target.Inner.Number);
```

Exception:
```
Newtonsoft.Json.JsonSerializationException
Error converting value "{"Id": "2c571dcc-65b8-429a-bb42-9c45833c9604", "Date": "2005-12-29T00:00:00+11:00", "Flag": false, "Long": 1529706464, "Color": 1, "Float": 0.2573189, "Inner": null, "Double": 0.6195458437407138, "Number": 56753383, "String": "Gray", "UserId": "00000000-0000-0000-0000-000000000000", "Decimal": 0.0, "Children": null, "GuidArray": null, "OtherGuid": "00000000-0000-0000-0000-000000000000", "TagsArray": null, "DateOffset": "0001-01-01T00:00:00+00:00", "StringDict": {}, "StringList": [], "NumberArray": [7, 2, 3], "StringField": null, "TagsHashSet": null, "NullableEnum": null, "AnotherNumber": 1149552278, "AnotherString": "nine", "NullableColor": null, "NullableNumber": null, "NullableBoolean": null, "NullableDateTime": null}" to type 'Marten.Testing.Documents.Target'. Path 'inner', line 1, position 813.
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.EnsureType(JsonReader reader, Object value, CultureInfo culture, JsonContract contract, Type targetType)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
[...]

PR #2057, opened 2022-01-24
Select transform with dictionary fails

VilleHakli:
If you have query with select that selects dictionary from entity, the query fails because the JSON returned from database cannot be deserialized. Following exception is thrown
```
Newtonsoft.Json.JsonSerializationException
Error converting value "{"Key": "Value", "Key2": "Value2"}" to type 'System.Collections.Generic.Dictionary`2[System.String,System.String]'. Path 'Values', line 1, position 71.
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.EnsureType(JsonReader reader, Object value, CultureInfo culture, JsonContract contract, Type targetType)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.SetPropertyValue(JsonProperty property, JsonConverter propertyConverter, JsonContainerContract containerContract, JsonProperty containerProperty, JsonReader reader, Object target)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonSerializer.Deserialize(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonSerializer.Deserialize[T](JsonReader reader)
[...]

PR #2212, opened 2022-05-07
Fix inner classes not being deserialized when doing Select projections

egorpavlikhin:
Fix for https://github.com/JasperFx/marten/issues/2211

Not sure how appropriate the fix is, the issue does seem to be very similar to the one mentioned here https://github.com/JasperFx/marten/pull/2057 - nested class properties are not being deserialized as objects, but as strings and then fail to deserialize back to objects.

### DIFF
--- a/src/Marten/Linq/Parsing/SelectTransformBuilder.cs
+++ b/src/Marten/Linq/Parsing/SelectTransformBuilder.cs
@@ -104,7 +104,7 @@ namespace Marten.Linq.Parsing
                         locator = field.TypedLocator;
                     }
 
-                    if (field.FieldType.IsClass)
+                    if (field.FieldType.IsClass && field.FieldType != typeof(string) && field.FieldType != typeof(decimal))
                     {
                         // If the field is a class, we need to cast it to JSONB otherwise it will be serialized to plain string and fail to deserialize later on
                         locator = field.JSONBLocator;
--- a/src/Marten/Linq/Parsing/SelectTransformBuilder.cs
+++ b/src/Marten/Linq/Parsing/SelectTransformBuilder.cs
@@ -10,10 +10,10 @@ using Remotion.Linq.Parsing;
 
 namespace Marten.Linq.Parsing
 {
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@egorpavlikhin Can you look at the test failures?

egorpavlikhin:
@jeremydmiller Those ones are fixed, could you approve another run please?

mysticmind:
> @jeremydmiller Those ones are fixed, could you approve another run please?

Done, CI build has started now.

src/Marten/Linq/Parsing/SelectTransformBuilder.cs
```diff
@@ -104,6 +104,12 @@ public string ToJsonBuildObjectPair(IFieldMapping mapping, ISerializer serialize
                         locator = field.TypedLocator;
                     }
 
+                    if (field.FieldType.IsClass && field.FieldType != typeof(string) && field.FieldType != typeof(decimal))
```
mysticmind, line 107: @egorpavlikhin any specific reason why you are doing these additional checks i.e. `&& field.FieldType != typeof(string) && field.FieldType != typeof(decimal)`?

egorpavlikhin:
Yes, they are classes that are directly deserialized by Newtonsoft/Json.Net and don't need to be extracted as JSONB (one of the test runs failed because I was casting them to JSONB).

src/Marten/Linq/Parsing/SelectTransformBuilder.cs
jeremydmiller: And I'm gonna weigh in too. I'd much rather that this test be baked into the `IField` model rather than being a runtime test here. The Linq provider gets very unruly and hard to support when we've used copious amounts of "if/then" logic on exceptions like this.

jeremydmiller: This has hung out too long. We'll do Linq code cleanup later whenever we inevitably tackle more enhancements to the Select() functionality.

jeremydmiller (approved):

PR #2223, opened 2022-05-17
Query with "contains any" and include generates invalid SQL

VilleHakli:
If query has "contains any" where clause and include Marten generates invalid SQL.

Following query results in error from PostgreSQL:
```csharp
var otherIdsQuery = new[]
{
    Guid.NewGuid(),
    Guid.NewGuid()
};

var otherTestEntityLookup = new Dictionary<Guid, OtherTestEntity>();
var entities = await session.Query<TestEntity>()
    .Include(x => x.OtherIds, otherTestEntityLookup)
    .Where(x => x.OtherIds.Any(id => otherIdsQuery.Contains(id)))
    .ToListAsync();
```

PostgreSQL error details:
```
Severity: ERROR
SqlState: 42703
MessageText: column "ctid" does not exist
Hint: There is a column named "ctid" in table "d", but it cannot be referenced from this part of the query.
```

I fixed the issue by simply adding table alias into SQL generated by WhereCtIdInSubQuery. Not sure if this might cause problems elsewhere but at least all DocumentDbTests passed locally.

### DIFF
--- a/src/DocumentDbTests/Bugs/Bug_2223_list_contains_any_generates_invalid_sql.cs
+++ b/src/DocumentDbTests/Bugs/Bug_2223_list_contains_any_generates_invalid_sql.cs
@@ -1,0 +1,114 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+using Bug2223;
+
+using Marten;
+using Marten.Testing.Harness;
+
+using Shouldly;
+
+using Weasel.Core;
+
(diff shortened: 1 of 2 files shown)


mysticmind (approved):
Looks fine to me!

jeremydmiller:
And looks good to me too, I'm pulling this in. I don't see any adverse effects from this. I've tried from time to time to eliminate the "d." vestigal tail from teh Linq support, but it's way too tightly engrained.

Issue #2230, opened 2022-05-21
New recipe for "self-slicing cross-stream aggregates"


jeremydmiller:
Because what we need, what we really, really need, is *yet another* base class and approach for doing aggregations in Marten.

Also, I already have this working and it looks to give cleaner code in the long run than the existing `ViewProjection` approach.

PR #2231, opened 2022-05-21
New CrossStreamAggregateProjection recipe. 

jeremydmiller:
Closes GH-2230

### DIFF
--- a/src/EventSourcingTests/Examples/TeleHealth/Board.cs
+++ b/src/EventSourcingTests/Examples/TeleHealth/Board.cs
@@ -1,0 +1,112 @@
+using System;
+using System.Collections.Generic;
+
+namespace EventSourcingTests.Examples.TeleHealth;
+
+#if NET6_0_OR_GREATER
+public interface IBoardEvent
+{
+    Guid BoardId { get; }
+}
+
+public class Board
+{
+    private Board()
+    {
+    }
+
+    public Board(BoardOpened opened)
+    {
+        Name = opened.Name;
+        Activated = opened.Opened;
+        Date = opened.Date;
+    }
+
+    public void Apply(BoardFinished finished)
+    {
+        Finished = finished.Timestamp;
+    }
+
+    public void Apply(BoardClosed closed)
+    {
+        Closed = closed.Timestamp;
+        CloseReason = closed.Reason;
+    }
+
+    public Guid Id { get; set; }
+    public string Name { get; private set; }
+    public DateTimeOffset Activated { get; set; }
+    public DateTimeOffset? Finished { get; set; }
+    public DateOnly Date { get; set; }
+    public DateTimeOffset? Closed { get; set; }
+
+    public string CloseReason { get; private set; }
+}
+
+internal interface BoardStateEvent{}
+
+public record BoardOpened(string Name, DateOnly Date, DateTimeOffset Opened) : BoardStateEvent;
+
+public record BoardFinished(DateTimeOffset Timestamp) : BoardStateEvent;
+
+public record BoardClosed(DateTimeOffset Timestamp, string Reason) : BoardStateEvent;
+
+// Make this the target of a ViewProjection
+public class BoardView
+{
+    public DateOnly Date { get; set; }
+
+    public bool Active => Closed == null;
+
+    public DateTimeOffset Opened { get; set; }
+    public DateTimeOffset? Closed { get; set; }
+
+    public DateTimeOffset? Deactivated { get; set; }
+
+    public string CloseReason { get; set; }
+
+    public string Name { get; set; }
+
+    public Guid Id { get; set; }
+
+    public IList<BoardAppointment> Appointments { get; set; } = new List<BoardAppointment>();
+    public IList<BoardProvider> Providers { get; set; } = new List<BoardProvider>();
+
+    public int ReadyCount { get; set; }
+    public int RequestedCount { get; set; }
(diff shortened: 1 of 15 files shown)


src/Marten.AsyncDaemon.Testing/cross_stream_aggregation.cs
```diff
@@ -0,0 +1,152 @@
...
+        grouping.FanOutOnEach<Travel, Movement>(x => x.Movements);
+
+        return ValueTask.CompletedTask;
+    }
+
+    public void Apply(Day day, TripStarted e) => day.Started++;
+
+    public void Apply(Day day, TripEnded e) => day.Ended++;
+
+    public void Apply(Day day, Movement e)
```
oskardudycz, line 131: **COMMENT:** I think that this should also be named an event, e.g. `MovementRecorded`. Showing non-event in FanOut can confuse people why we're not using event but some dto.

src/Marten.AsyncDaemon.Testing/cross_stream_aggregation.cs
```diff
@@ -0,0 +1,152 @@
...
+    protected override ValueTask GroupEvents(IEventGrouping<int> grouping, IQuerySession session, List<IEvent> events)
+    {
+        // Tell the projection how to group the events
+        // by Day document
+        grouping.AddEventsWithMetadata<IDayEvent>(e => e.Data.Day, events);
+
+        // This just lets the projection work independently
+        // on each Movement child of the Travel event
+        // as if it were its own event
+        grouping.FanOutOnEach<Travel, Movement>(x => x.Movements);
```
oskardudycz, line 122: **COMMENT:** That might be only me, but for me, but looking at this GroupEvents function, it'd be hard for me to understand where all the `Movement` objects are coming from and how they'll be matched with Travel. I understand what's happening here, but even knowing Marten, I had to scroll up and down multiple times. Might be worth working on this API to make it more clear.

src/Marten.AsyncDaemon.Testing/cross_stream_aggregation.cs
```diff
@@ -0,0 +1,152 @@
...
+
+public class CrossStreamDayProjection: CrossStreamAggregateProjection<Day, int>
+{
+    public CrossStreamDayProjection()
+    {
+        ProjectionName = "Day";
+
+        // You have to specify this in order for the Travel events
+        // to flow through this projection.
+        IncludeType<Travel>();
```
oskardudycz, line 110: **SUGGESTION:** I think that it'd be better to rename `IncludeType` to `IncludeEventType` or just `Include` as I had a hard time remembering what it does. Also, we should rename `Travel` to something that's named more like an event, e.g. `TravelFinished`, as people will have a hard time understanding what it does. My first thought is that's some new feature to include other documents.

src/EventSourcingTests/Examples/TeleHealth/Board.cs
```diff
@@ -0,0 +1,112 @@
...
+    public string Name { get; private set; }
+    public DateTimeOffset Activated { get; set; }
+    public DateTimeOffset? Finished { get; set; }
+    public DateOnly Date { get; set; }
+    public DateTimeOffset? Closed { get; set; }
+
+    public string CloseReason { get; private set; }
+}
+
+internal interface BoardStateEvent{}
```
oskardudycz, line 46: I'm not a huge fan of interfaces for events. I see that it can be useful, but I did that in the past and saw that folks are trying to do when they're starting to use Event Sourcing/
Too often, that leads to wrong event modelling, as they're still thinking about making all things generic instead of treating each event as a different fact.
Especially putting `State` in the name is dangerous in samples, as already too many folks are trying just to reflect state changes instead of recording the business fact.

See also:
- https://event-driven.io/en/state-obsession/
- https://event-driven.io/en/property-sourcing/
That leads too often to CRUD sourcing and/or state obsession

src/EventSourcingTests/Examples/TeleHealth/BoardViewProjection.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace EventSourcingTests.Examples.TeleHealth;
+
+#if NET6_0_OR_GREATER
+public class BoardViewProjection: CrossStreamAggregateProjection<BoardView, Guid>
+{
+    protected override ValueTask GroupEvents(IEventGrouping<Guid> grouping, IQuerySession session, List<IEvent> events)
+    {
+        grouping.AddEventsWithMetadata<BoardStateEvent>(e => e.StreamId, events);
+        grouping.AddEvents<IBoardEvent>(x => x.BoardId, events);
```
oskardudycz, line 17: **QUESTION:** Is this intentional that we don't have any implementation of `IBoardEvent`?

src/EventSourcingTests/Examples/TeleHealth/BoardViewProjection.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Threading.Tasks;
+using Marten;
+using Marten.Events;
+using Marten.Events.Aggregation;
+using Marten.Events.Projections;
+
+namespace EventSourcingTests.Examples.TeleHealth;
+
+#if NET6_0_OR_GREATER
+public class BoardViewProjection: CrossStreamAggregateProjection<BoardView, Guid>
```
oskardudycz, line 12: **QUESTION:** Which events are coming from different streams?

PR #2242, opened 2022-05-25
Do not open transaction by default when using OpenSessionAsync

VilleHakli:
When Marten handles connection and session, OpenSession does open connection and begin transaction only when Serializable isolation level is used. Async version of OpenSession opens connection and begins transaction always. This PR changes the async version of OpenSession to work the same way as sync version.


mysticmind (approved):
Looks good to me.

jeremydmiller:
@VilleHakli This is fine, but there's absolutely no real reason to use OpenSessionAsync() outside of serializable transactions other than to get an IDE to shut up about async alternatives

VilleHakli:
I'm not actually using the async version anywhere but I was checking where Marten opens transactions and came across this one so thought to make a PR about this.

PR #2252, opened 2022-05-27
Overloads for StartStream that return IEventStream<T>

elexisvenator:
Works similar to the internal method `IEventIdentityStrategy<TId>.StartStream<TDoc>()`. Provides 3 ways to create a new stream, returning `IEventStream<T>` with the supplied aggregate set.

### DIFF
--- a/src/Marten/Events/EventStore.StartStream.cs
+++ b/src/Marten/Events/EventStore.StartStream.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Marten.Schema.Identity;
 
 #nullable enable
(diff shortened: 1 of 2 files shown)


elexisvenator:
The goal with this pr is to allow the same behaviour for adding events to a stream regardless of if that stream is new or existing. 

For example, I can write a service such as the following and call it from a create method and an update method.

```c#
public interface IDocumentAuthorService {
    Task AddAuthor(IEventStream<Document> stream, IQuerySession session, string userId);
    void RemoveAuthor(IEventStream<Document> stream, string userId);
}

elexisvenator:
@jeremydmiller is there anything I should add to this addition?

oskardudycz (approved):
I'm okay taking that in. However, I think that `IEventStore` interface is changing into a mammoth. I'd see those changes more as extension methods (that comment also applies to some other already existing). As it's following the current convention, I'm not super optimistic, but okay with that. I think that we should revisit this interface and slim it down in v6.

@elexisvenator, it'd be great if you could add more samples and docs for this change.

elexisvenator:
I am going to close this as it doesnt really matter for 2 reasons

1. Once a FetchForWriting is done, all subsequent calls to Events.Apply() are automatically optimistic. You dont actually have to call IEventStream.AppendOne() to get this benefit.
2. Calling FetchForWriting with a non-existent stream has the same effect as this pr's StartStream method.

PR #2259, opened 2022-06-01
Don't use parameters with to_tsvector to greatly improve performance

phinett:
Static values in the sql query perform much faster

### DIFF
--- a/src/Marten/Linq/Filters/FullTextWhereFragment.cs
+++ b/src/Marten/Linq/Filters/FullTextWhereFragment.cs
@@ -16,7 +16,7 @@ namespace Marten.Linq.Filters
         private readonly FullTextSearchFunction _searchFunction;
         private readonly string _searchTerm;
 
-        private string Sql => $"to_tsvector('{_regConfig}'::regconfig, {_dataConfig}) @@ {_searchFunction}('{_regConfig}'::regconfig, '{_searchTerm}')";
+        private string Sql => $"to_tsvector('{_regConfig}'::regconfig, {_dataConfig}) @@ {_searchFunction}('{_regConfig}'::regconfig, '{SanitizeSearchTerm(_searchTerm)}')";
 
         public FullTextWhereFragment(DocumentMapping mapping, FullTextSearchFunction searchFunction, string searchTerm, string regConfig = FullTextIndex.DefaultRegConfig)
         {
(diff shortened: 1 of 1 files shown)


src/Marten/Linq/Filters/FullTextWhereFragment.cs
```diff
@@ -16,7 +16,7 @@ internal class FullTextWhereFragment: ISqlFragment
         private readonly FullTextSearchFunction _searchFunction;
         private readonly string _searchTerm;
 
-        private string Sql => $"to_tsvector(:argRegConfig::regconfig, {_dataConfig}) @@ {_searchFunction}(:argRegConfig::regconfig, :argSearchTerm)";
+        private string Sql => $"to_tsvector('{_regConfig}'::regconfig, {_dataConfig}) @@ {_searchFunction}('{_regConfig}'::regconfig, '{_searchTerm}')";
```
elexisvenator, line 19: recommend doing a find/replace of `'` with `''` for safety

phinett: Ok will do, is there a extension / pre written function you have already for this type of thing?

elexisvenator: doesnt look like there is one in npgsql. Best thing I can find based on [this doco](https://www.postgresql.org/docs/8.3/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS) is to turn the string into an E string then find replace `'` with `''` and `\` with `\\`.

mysticmind:
https://github.com/npgsql/efcore.pg/pull/1888 - this is the EFCore issue for reference.

mysticmind:
https://github.com/npgsql/efcore.pg/pull/1888 - this is the EFCore issue for reference.

src/Marten/Linq/Filters/FullTextWhereFragment.cs
```diff
@@ -27,6 +27,12 @@ public FullTextWhereFragment(DocumentMapping mapping, FullTextSearchFunction sea
             _searchTerm = searchTerm;
         }
 
+        private string SanitizeSearchTerm(string searchTerm)
+        {
+            // edge case that will cause a sql exception with a single quote we need to handle also
+            return searchTerm == "'" ? "" : searchTerm?.Replace("'", "''");
```
mysticmind: If you find a single quote in text, it will need to replaced by double single quotes `''` but the above code looks to be incorrect i.e. if the search term contains single quote, it is returning empty string and if not then it is trying to do a replace. shouldn't it be just `return searchTerm?.Replace("'", "''");`? Please recheck this.

Could you please add a unit test specifically with search term containing semicolon?

mysticmind: > https://github.com/npgsql/efcore.pg/pull/1888 - this is the EFCore issue for reference.
@phinett Quick follow on note, I re-read the issue in the efcore.pg repo and observing that only the config doesn't need to be parametrized and search term can be parametrized. So I would suggest to keep the config as string and leave the search term as parameter i.e. `:argSearchTerm`. This way, we don't need to do any sanitization on the search term.

mysticmind: I have added the above suggested fix.

PR #2272, opened 2022-06-09
Added the ability to delete a mix of documents

### LINKED ISSUE #2271: Add the ability to delete a mix of documents
Having the `StoreObjects(IEnumerable<object> documents)` function has proven to be very useful in some situations. Especially in tests.

It would be nice to have a counterpart function to this which allows the deletion of a mix of documents. Such as `DeleteObjects(IEnumerable<object> documents)`.

For example:
```csharp
public interface IEntity
{
    Guid? Id { get; set; }
}

public class User : IEntity
{
    public Guid? Id { get; set; }
}

public class Order : IEntity
{
    public Guid? Id { get; set; }
}

var entities = new List<IEntity>
{
    new User(...),
    new Order(...),
};

// Store a mix of documents.
session.StoreObjects(entities);

// Delete a mix of documents.
session.DeleteObjects(entities);
```

I've used workarounds in the past using `storage.DeletionForEntity`. However this is unachievable with the changes in V4/V5.

jamievaughan:
Fixes #2271

The goal of this change is to implement a counterpart for the `StoreObjects(IEnumerable<object> documents)` function with a `DeleteObjects(IEnumerable<object> documents)` function.

### DIFF
--- a/src/Marten/IDocumentOperations.cs
+++ b/src/Marten/IDocumentOperations.cs
@@ -54,6 +54,12 @@ namespace Marten
         void DeleteWhere<T>(Expression<Func<T, bool>> expression) where T : notnull;
 
         /// <summary>
+        /// Delete an enumerable of potentially mixed documents
+        /// </summary>
+        /// <param name="documents"></param>
+        void DeleteObjects(IEnumerable<object> documents);
+
+        /// <summary>
         /// Explicitly marks multiple documents as needing to be inserted or updated upon the next call to SaveChanges()
         /// </summary>
         /// <typeparam name="T"></typeparam>
@@ -77,7 +83,7 @@ namespace Marten
         void Store<T>(T entity, Guid version) where T : notnull;
 
         /// <summary>
-        /// DocumentStore an enumerable of potentially mixed documents
+        /// Store an enumerable of potentially mixed documents
         /// </summary>
         /// <param name="documents"></param>
         void StoreObjects(IEnumerable<object> documents);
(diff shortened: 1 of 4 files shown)


src/Marten/IDocumentOperations.cs
```diff
@@ -53,6 +53,12 @@ public interface IDocumentOperations: IQuerySession
         /// <param name="expression"></param>
         void DeleteWhere<T>(Expression<Func<T, bool>> expression) where T : notnull;
 
+        /// <summary>
+        /// Delete an enumerable of potentially mixed documents
+        /// </summary>
+        /// <param name="documents"></param>
```
YarekTyshchenko, line 59: Missing param doc

src/Marten/Internal/Sessions/DocumentSessionBase.cs
```diff
@@ -319,26 +319,37 @@ public void EjectPatchedTypes(IUnitOfWork changes)
             foreach (var type in patchedTypes) EjectAllOfType(type);
         }
 
-        internal interface IHandler
+        internal interface IObjectHandler
```
mysticmind, line 322: The interface name change from `IHandler` to `IObjectHandler` is okay. But I felt that it could have been retained as `IHandler` itself (nitpicking here)

mysticmind (approved):
Looks good to me 👍

I did add a comment, please check.

src/Marten/Internal/Sessions/DocumentSessionBase.cs
jamievaughan: I do agree, and I did actually consider this while making the change. My intent was to make the interface more specific to the type handled. However, I'm not opposed to changing this back to `IHandler` if anyone feels strongly about this.

jeremydmiller (approved):
Looks good to me across the board. Wouldn't hurt to add a little bit to the docs for this too, but that can come later. I don't really care about the naming thing, but I'd ask that you don't rename things w/o more cause in the future. 

I vote to take this in.

PR #2278, opened 2022-06-22
Changed AggregateTo extensions receiver type

arielmoraes:
Changed the type from `IMartenQueryable` to `IQueryable` allowing calls to `AggregateTo` when chained Linq methods are used.

### DIFF
--- a/src/Marten/Events/AggregateToExtensions.cs
+++ b/src/Marten/Events/AggregateToExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System.Linq;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Baseline;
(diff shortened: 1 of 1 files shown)


src/Marten/Events/AggregateToExtensions.cs
```diff
@@ -15,7 +15,7 @@ public static class AggregateToExtensions
         /// <param name="state"></param>
         /// <typeparam name="T"></typeparam>
         /// <returns></returns>
-        public static T AggregateTo<T>(this IMartenQueryable<IEvent> queryable, T state = null) where T : class
```
mysticmind, line 18: I would suggest you to add new methods as below rather than changing the signature of the existing ones since it results in breaking changes in terms of binary compatibility
```cs
public static T AggregateTo<T>(this IQueryable<IEvent> queryable, T state = null) where T : class
{
    return AggregateTo(queryable.As<IMartenQueryable<IEvent>>(), state);
}

public static async Task<T> AggregateToAsync<T>(this IQueryable<IEvent> queryable, T state = null,
    CancellationToken token = new()) where T : class
{
    return await AggregateToAsync(queryable.As<IMartenQueryable<IEvent>>(), state, token).ConfigureAwait(false);
}
```

mysticmind (changes requested):

src/Marten/Events/AggregateToExtensions.cs
arielmoraes: Will do.

PR #2294, opened 2022-07-14
Flat table projections

jeremydmiller:
WIP, but I like to cheat and let CI run the slow tests


docs/configuration/prebuilding.md
```diff
@@ -133,12 +133,12 @@ As an example, here is the Marten configuration from the project we used to test
 <!-- snippet: sample_configuring_pre_build_types -->
 <a id='snippet-sample_configuring_pre_build_types'></a>
 ```cs
-public class Program
+public static class Program
```
oskardudycz, line 136: **FIX:** Something went wild here with formatting.

src/CommandLineRunner/Program.cs
```diff
@@ -30,12 +30,12 @@ public interface IOtherStore : IDocumentStore
 
     #region sample_configuring_pre_build_types
 
-    public class Program
+    public static class Program
```
oskardudycz, line 33: **FIX:** Something went wild here with formatting.

src/EventSourcingTests/Projections/Flattened/WriteTableWithGuidIdentifierProjection.cs
```diff
@@ -0,0 +1,152 @@
...
+{
+    public WriteTableWithGuidIdentifierProjection(): base("values", SchemaNameSource.DocumentSchema)
+    {
+        // Gotta have a primary key
+        Table.AddColumn<Guid>("id").AsPrimaryKey();
+
+        Project<ValuesSet>(cmd =>
+        {
+            // Derive column name by default, add column if does not exist
+            cmd.Map(x => x.A); // returns the column
```
oskardudycz, line 16: **QUESTION:** Wouldn't it be better to return here `StatementMap` to enable fluent chaining?  My understanding of the current `ColumnExpression` capabilities is to modify the schema, which I guess someone should not be doing here.

src/EventSourcingTests/Projections/Flattened/WriteTableWithGuidIdentifierProjection.cs
```diff
@@ -0,0 +1,152 @@
+using System;
+using Marten.Events.Projections.Flattened;
+
+namespace EventSourcingTests.Projections.Flattened;
+
+public class WriteTableWithGuidIdentifierProjection: FlatTableProjection
+{
+    public WriteTableWithGuidIdentifierProjection(): base("values", SchemaNameSource.DocumentSchema)
```
oskardudycz, line 8: **QUESTION:** I thought we would do something similar to what we have for regular projections, so allow someone to provide the view type and map by convention to the column names. We could even benefit from the code generation by doing that. We could enable people to configure mapping either through projection config and/or attributes and then just do upserts.

I think what's here is more flexible but also quite raw and different from what we have in other places. Yet, that doesn't mean that's bad. I think it'd be good to have something raw like that for someone who'd like to be more precise. So something like MongoDB API with their atomic operations. It might be even worth moving to regular JSON projections in the form of partial updates.

src/EventSourcingTests/Projections/Flattened/ValuesSubtracted.cs
```diff
@@ -0,0 +1,10 @@
+namespace EventSourcingTests.Projections.Flattened;
+
+public class ValuesSubtracted
+{
```
oskardudycz, line 4: **SUGGESTION:** I think that'd be good to add also at least fields with `DateTimeOffset` and `Guid`.

src/EventSourcingTests/Projections/Flattened/flat_table_projection_with_event_member_identifier_end_to_end.cs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    private async Task<Data> findData(string name)
+    {
+        using var conn = theStore.Storage.Database.CreateConnection();
+        await conn.OpenAsync();
+
+        var all = await conn.CreateCommand($"select * from {SchemaName}.member_values where name = :id")
+            .With("id", name)
+            .FetchList(readData);
```
oskardudycz, line 77: **SUGGESTION:** That's later on some potential for convention-based mapping with code generation.

src/Marten/Events/Projections/Flattened/StatementMap.cs
```diff
@@ -0,0 +1,170 @@
...
+using Baseline.Reflection;
+using LamarCodeGeneration;
+using Marten.Events.CodeGeneration;
+using Marten.Internal.CodeGeneration;
+using Weasel.Core;
+using Weasel.Postgresql.Tables;
+
+namespace Marten.Events.Projections.Flattened
+{
+    public class StatementMap<T> : IEventHandler
```
oskardudycz, line 15: **SUGGESTION:** I think that it'd be good to add people option to do custom SQL, then someone could do their own magic, and we wouldn't need to add a lot of new weird map permutations immediately.

src/Marten/Events/Projections/Flattened/FlatTableProjection.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// </summary>
+        public Table Table { get; }
+
+        /// <summary>
+        /// Write values for the designated event type to the table columns
+        /// </summary>
+        /// <param name="configure"></param>
+        /// <param name="tablePrimaryKeySource">If specified, tells the projection how to find the primary key value for the table from the event data. If missing, the projection uses the event's stream identity by default</param>
+        /// <typeparam name="T"></typeparam>
+        public void Project<T>(Action<StatementMap<T>> configure, Expression<Func<T, object>>? tablePrimaryKeySource = null)
```
oskardudycz, line 67: **QUESTION:** We'll support the right now single result value, right? (Not e.g. multiple identities)?

If we add custom SQL then it might be fine for people to not provide identity, but then that's probably a different projection type.

oskardudycz (commented):
I'm good with letting that in. Still, we'll need to watch for the feedback carefully. I think it'd be good to eventually make it a closer usability experience for the other types of projections. Rome wasn't built in a day 😉

src/CommandLineRunner/Program.cs
jeremydmiller: How so?

src/EventSourcingTests/Projections/Flattened/WriteTableWithGuidIdentifierProjection.cs
jeremydmiller: Not in this case IMO. The call to `Map()` actually returns the Weasel column expression to do whatever you want to fine tune that particular column. You *could* go with a fluent interface instead, and make any customizations to the column be done through an optional `Action<ColumnExpression>` lambda. I think I'd vote to leave things the way it is, but don't have a very strong opinion if you think the fluent interface approach is really better. My preference with FI's these days is to keep the chaining very short.

src/EventSourcingTests/Projections/Flattened/ValuesSubtracted.cs
jeremydmiller: As a test? Should be able to handle any type that Weasel + Npgsql handles as is.

src/EventSourcingTests/Projections/Flattened/flat_table_projection_with_event_member_identifier_end_to_end.cs
jeremydmiller: Something like what LLBLGen is today? Or do you mean something else? This was strictly for testing. Dapper would have helped, but I'm not a big Dapper guy.

src/Marten/Events/Projections/Flattened/StatementMap.cs
jeremydmiller: True, but they could also drop into an `EventProjection` and issue all the custom SQL they want. We could definitely do this later though.

Think about this though. The SQL generated by this functionality is doing upserts which are clumsy bits of code to write and would blow up the parameter count if you did it by hand. There is a reason why I went this way.

src/Marten/Events/Projections/Flattened/FlatTableProjection.cs
jeremydmiller: Like I said before, if you get out of the well lit path, you can go to an `EventProjection` and write custom SQL to your heart's content. Same with the aggregations, I don't want to even attempt to accommodate everything they could try to do in the OOTB recipe because that kills us with permutations. Let them go to custom, explicit code for anything weird or unusual.

src/EventSourcingTests/Projections/Flattened/WriteTableWithGuidIdentifierProjection.cs
jeremydmiller: Can you provide a concrete example? I'm not sure what about this solution isn't already pretty declarative. It's just "I want these members of these event types upserted into this row". By default, the column names are even derived from the member names. The point of this was just to write data to a straight SQL table. The examples I'm going to use it for are 1.) to help group a big cross-stream aggregation and 2.) to support metrics in a app where a flat table is advantageous to hit with SQL.

I'm definitely not going for a real ORM here.

And this does use code generation behind the scenes of course.

Issue #2184, opened 2022-04-24
Use IDocumentStore.BulkInsert inside a TransactionScope?


youqingz:
### Discussed in https://github.com/JasperFx/marten/discussions/2148

<div type='discussions-op-text'>

<sup>Originally posted by **youqingz** March 30, 2022</sup>
I can use IDocumentSession to enlist an ambient transaction inside a TransactionScope, but I also would like to use IDocumentStore.BulkInsert for some of the operations for performance reasons, but seems like it is not supported at the moment?

</div>

jeremydmiller:
Closed by #2303

PR #2303, opened 2022-07-24
#2184 Add bulk insert methods that can enlist an existing transaction

youqingz:


src/Marten/DocumentStore.cs
```diff
@@ -92,6 +91,14 @@ public virtual void Dispose()
             bulkInsertion.BulkInsert(documents, mode, batchSize);
         }
 
+        public void BulkInsertEnlistTransaction<T>(IReadOnlyCollection<T> documents,
```
mysticmind, line 94: Rather than adding new methods for enlisting transaction, I think it would be better to add an optional param to the existing methods to deal with enlisting. Thoughts?

youqingz: @jeremydmiller Suggested to add new method.
https://github.com/JasperFx/marten/discussions/2148#discussioncomment-2787127

mysticmind: Alight, didn't notice the discussion thread. Thanks for the heads up!

youqingz: Any other issues? Is it ready to be merged?

mysticmind: Ian good with it, will approve the same. After approval from another review either by @jeremydmiller or @oskardudycz, will merge this PR.

mysticmind (approved):
Looks fine to me.

src/Marten/DocumentStore.cs
jeremydmiller: Not gonna lie and say I exactly remember the conversation, but an optional parameter is a breaking change (requires a recompile even though the member usage is technically the same). That was probably why I asked for an overload instead. I'm catching up now, and probably pulling this in.

PR #2313, opened 2022-08-05
Discussion: Add test to show inline projection rebuilds stopping at gaps.

dionfoster:
This test is to reproduce an issue we noticed in one of our environments, where inline projection rebuilds would get stuck on gaps in the event stream.  Early gaps within the event stream were caused by some catastrophic failures, though I'd expect the projection rebuilds to be able to deal with gaps and move on.


jeremydmiller:
@dionfoster Got a fix coming in for this. I pulled your repro test in with a rebase, so closing this. Thank you for the reproduction steps, that's over half the battle for some things like this.

PR #2314, opened 2022-08-05
Upcasters

oskardudycz:
Pull request adds support for Upcasting events data. It provides built-in tooling for simple payload transformations. If your event schema has changed and you'd like to use new event types in your aggregates and projections, you could use that. Those transformations will be run during the serialisation phase, so upcasting on the fly old event data into the new one.

I added two types of upcasters:
- CLR types - when you're fine with keeping the old event payloadOKen it'll at first deserialise it, and you can map to the new one,
- Raw JSON transformations: more flexible, fewer allocations, but more tedious.

Technically, this is what I described in my blog article [Simple patterns for events schema versioning](https://event-driven.io/en/simple_events_versioning_patterns/).

**TODO:** The rest of the docs will be provided in the follow-up PR, not to create the _merge train_.


src/Marten/Events/IEventStoreOptions.cs
```diff
@@ -56,6 +57,52 @@ public interface IEventStoreOptions
...
+        /// <summary>
+        /// Maps CLR event type as particular event type name allowing to provide custom transformation.
+        /// This is useful for event type migration.
+        /// See more in docs: https://martendb.io/events/versioning.html#event-type-name-migration
+        /// </summary>
+        /// <param name="eventTypeName">Event type name</param>
+        /// <param name="jsonTransformation">Event payload transformation</param>
+        /// <typeparam name="TEvent">Mapped CLR event type</typeparam>
+        IEventStoreOptions Upcast<TEvent>(
+            string eventTypeName,
```
Hawxy, line 70: Given the type name is optional in the abstract class version, could it be made optional with the fluent syntax?

oskardudycz: Currently, it's not optional; it's only optional if you use upcasting by the specific CLR type. See: https://github.com/JasperFx/marten/pull/2314/files?diff=split&w=1#diff-aeac6bc19425c3715f9bf9d7e6436e1a6f7a8f51e8d8b14dff919691ae00fc79R39

I might try to provide overload with two event types that would not require passing that if that helps.

Hawxy: I see. Yes, that would be useful 👍

oskardudycz: Added new methods reusing the slimmed upcasting via CLR types: https://github.com/JasperFx/marten/pull/2314/commits/08873ff478c8ab469521476f699f75422ad7928c. Thoughts? :)

src/Marten/Events/EventGraph.cs
```diff
@@ -130,6 +129,69 @@ public void MapEventType(Type eventType, string eventTypeName)
             eventMapping.EventTypeName = eventTypeName;
         }
 
+        public IEventStoreOptions Upcast<TEvent>(
```
oskardudycz, line 132: @jeremydmiller, maybe it would be worth grouping those transformations into nested property instead of adding them straight into `EventGraph`? E.g. `options.Events.Transformations.Upcast(...)`. Thoughts?

src/Marten/Events/IEventStoreOptions.cs
Hawxy: Looks good, thanks!

oskardudycz: I thank you! 😅 Good call 🙂

src/Marten/Events/EventDocumentStorage.cs
```diff
@@ -267,6 +256,43 @@ public async Task<IEvent> ResolveAsync(DbDataReader reader, CancellationToken to
 
         public abstract Task ApplyReaderDataToEventAsync(DbDataReader reader, IEvent e, CancellationToken token);
 
+        private EventMapping SetEventMappingForDotNetTypeName(string dotnetTypeName, string eventTypeName)
```
jeremydmiller, line 259: It's nitpicking, but can you find a different name for this method that's more descriptive? This name implies a write, but it's not precisely that

src/Marten/Events/EventGraph.cs
jeremydmiller: Maybe, but that's a lot of dots and not super discoverable. I think I'd leave it on `EventGraph`, but maybe look at what could be hidden or simplified in v6.

src/Marten/Events/EventMapping.cs
```diff
@@ -301,4 +301,10 @@ public override IEvent Wrap(object data)
             };
         }
     }
+
+    public static class EventMappingExtensions
```
jeremydmiller, line 305: Is this new logic, or did you extract a method to get here?

src/Marten/Services/SystemTextJsonSerializer.cs
```diff
@@ -55,8 +59,18 @@ public void Customize(Action<JsonSerializerOptions> configure)
             configure(_options);
             configure(_optionsDeserialize);
             configure(_withTypes);
+
+            SyncDocumentDeserializeOptions();
         }
 
+        private void SyncDocumentDeserializeOptions() =>
```
jeremydmiller, line 66: camel casing on private/protected members to be a little more consistent

src/Marten/Events/EventDocumentStorage.cs
oskardudycz: I dropped `Set` prefix from the method. I was struggling tbt how to name it, as it's not a pure method and also setting the Event Mapping internally if it doesn't exist.

src/Marten/Events/EventMapping.cs
oskardudycz: Yeah, I just inlined what we had above: https://github.com/JasperFx/marten/pull/2314/files#diff-8d90b88a566421c9c25b2ebb7ebfff2a6c1996c2df677e25bd9bc69392891af2L55.
I updated it to internal access.

src/Marten/Services/SystemTextJsonSerializer.cs
oskardudycz: Renamed :+1:

PR #2342, opened 2022-08-30
Fixed mapping of the upcasting functions to allow multiple schema versions handling

oskardudycz:
PR fixed the mapping for upcasting function to allow multiple schema version handling.

Before that change, it was:
- old event type name => new CLR type,
- new CLR type => old event type name

Now it'll be:
- old event type name => new CLR type,
- new event type name => new CLR type,
- new CLR type => new event type name

That enables multiple schema version handling.

I also added a set of syntactic sugar for multiple schema version registration. It enables automated convention with the version suffix.

### DIFF
--- a/src/Marten/Events/CodeGeneration/DefaultAggregateConstruction.cs
+++ b/src/Marten/Events/CodeGeneration/DefaultAggregateConstruction.cs
@@ -34,6 +34,8 @@ namespace Marten.Events.CodeGeneration
 
         public IfStyle IfStyle { get; set; } = IfStyle.Else;
 
+        public string AdditionalNoConstructorExceptionDetails { get; set; } = string.Empty;
+
         public override IEnumerable<Variable> FindVariables(IMethodVariables chain)
         {
             _event = chain.FindVariable(typeof(IEvent));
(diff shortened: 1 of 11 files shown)


src/Marten/Events/EventGraph.cs
```diff
@@ -155,46 +156,36 @@ public void MapEventType(Type eventType, string eventTypeName)
             JsonTransformation jsonTransformation = null
         )
         {
-            var eventMapping = EventMappingFor(eventType);
+            var eventMapping = typeof(EventMapping<>).CloseAndBuildAs<EventMapping>(this, eventType);
```
oskardudycz, line 159: @jeremydmiller, this is the most important change. I'm setting the mapping here for the event type, to map the old event type name to the new CLR type. The mapping will looks like that now:
- old event type name => new CLR type,
- new event type name => new CLR type,
- new CLR type => new event type name

Before that change, it was:
- old event type name => new CLR type,
- new CLR type => old event type name

Because of that, it was hard (or almost impossible) to have different upcasting logic for different schema versions.

jeremydmiller: That makes sense.

src/Marten/Events/CodeGeneration/DefaultAggregateConstruction.cs
```diff
@@ -47,7 +47,10 @@ public override void GenerateCode(GeneratedMethod method, ISourceWriter writer)
             if (_constructor == null)
             {
                 writer.WriteLine(
-                    $"throw new {typeof(InvalidOperationException).FullNameInCode()}(\"There is no default constructor for {_returnType.FullNameInCode()}\");");
+                    $"throw new {typeof(InvalidOperationException).FullNameInCode()}(\"There is no default constructor or Create method for {_returnType.FullNameInCode()} event type." +
```
jeremydmiller, line 50: Good.

PR #2356, opened 2022-09-20
Add dynamic OrderBy extensions for batched queries

smbecker:


mysticmind (commented):
@smbecker could you please add a unit test as well?

smbecker:
> @smbecker could you please add a unit test as well?

I duplicated the existing dynamic order-by tests leveraging a batched query

mysticmind (approved):
Looks good to me 👍

src/Marten/QueryableExtensions.cs
```diff
@@ -492,43 +555,74 @@ public static IOrderedQueryable<T> ThenBy<T>(this IOrderedQueryable<T> queryable
...
+            return ApplyOrder(queryable, property, "ThenByDescending");
         }
 
         private static IOrderedQueryable<T> ApplyOrder<T>(
             IQueryable<T> queryable,
             string property,
             string methodName)
         {
+            var lambda = CompileOrderBy<T>(property, out var targetType);
+            var result = typeof(Queryable).GetMethods().Single(
```
oskardudycz, line 579: **COMMENT:** Not for this PR, but some memoisation will be good to add here.

oskardudycz (approved):

src/Marten/QueryableExtensions.cs
mysticmind: We will add a TODO comment for this, will merge the PR.

PR #2364, opened 2022-10-08
BUG: Incorrectly compiled query when child query in or clause

T0shik:
Failing test attached, to me it looks like it wipes the top level condition and attaches it to the nested one? I don't understand the SQL generation code in marten so wasn't confident making the fix. Here's additional information based on what I found during debugging:

Issue Location: DocumentStatement.CompileLocal
Issue Introduced: 48a9bfd378bed81ecba50c398c7921342ec5df58 (reversing this commit fixes the issue)

Generated failing SQL:
```
WITH mt_temp_id_list1CTE as (
    select ctid, unnest(CAST(ARRAY(SELECT jsonb_array_elements(CAST(d.data ->> 'Nested' as jsonb))) as jsonb[])) as data from my_table as d WHERE d.data ->> 'Name' = :p0
)
   , mt_temp_id_list2CTE as (
    select ctid, data from mt_temp_id_list1CTE as d where d.data ->> 'Name' = :p1
)
select d.id, d.data from my_table as d where d.ctid in (select ctid from mt_temp_id_list2CTE);
```

Working SQL:
```
WITH mt_temp_id_list1CTE as (
    select ctid, unnest(CAST(ARRAY(SELECT jsonb_array_elements(CAST(d.data ->> 'Nested' as jsonb))) as jsonb[])) as data from my_table as d
)
   , mt_temp_id_list2CTE as (
    select ctid, data from mt_temp_id_list1CTE as d where d.data ->> 'Name' = :p0
)
select d.id, d.data from my_table as d where d.data ->> 'Name' = :p1 or d.ctid in (select ctid from mt_temp_id_list2CTE);
```

### DIFF
--- a/docs/cSpell.json
+++ b/docs/cSpell.json
@@ -83,7 +83,8 @@
     "bigrams",
     "Blazor",
     "CQRS",
-    "stringly"
+    "stringly",
+    "hostbuilder"
   ],
   "ignoreRegExpList": [
     "\\((.*)\\)", // Markdown links


mysticmind:
This issue looks to be a similar/same one as https://github.com/JasperFx/marten/pull/2286

Issue #2370, opened 2022-10-28
Async Daemon - StaleSequenceThreshold not being honored properly


ericgreenmix:
The `StaleSequenceThreshold` gets added to the `Timestamp` of the last tracked (_current) `HighWaterStatistics`. If there is a time break in events larger than the last time the current statistics was updated, the `StaleSequenceThreshold` gets rendered useless. The `StaleSequenceThreshold` should be getting applied to the last time a non-stale status was polled.

Example of what happened to our production environment:

- `StaleSequenceThreshold` is set to 10 seconds.
- We had a time break in events of around 11 seconds.
- The next event that came in was a larger event, that does a bit more processing.
- Marten had pulled the nextval for the sequence of that event
- Highwater agent ran, detected stale because the HighwaterMark hadn't been updated to the nextval of the sequence yet
- `StaleSequenceThreshold` was compared against the Timestamp of the last change, around 11 seconds ago, making the safeHarborTime be less that the statistic Timestamp
- The statistics `CurrentMark` was the same as the tracker `HighWaterMark` (since the db update hadn't been applied yet, just a nextval sequence pull), meaning the agent looped on the 250 ms fast polling time until finally the agent could find a `SafeHarborSequence`, which ended up being one event after, causing an event to be skipped

I believe the fix for this is straight forward, and I am going to be submitting a PR shortly. I think all that needs to happen is the _current statistics need to be updated (really only the Timestamp) anytime progress is marked and the status is `CaughtUp`.

jeremydmiller:
Closed by #2371.

Thanks @ericgreenmix!

PR #2371, opened 2022-10-28
#2370 - Make sure the StaleSequenceThreshold gets applied to the timestamp of the last poll of detect changes

ericgreenmix:


jeremydmiller:
I'll take this one folks

Issue #2360, opened 2022-09-26
Provide an option to specify projection rebuild command timeout for Marten.CommandLine


NikiforovAll:
The default timeout is hardcoded and it is not possible to extend the migration period. 

Example: https://github.com/JasperFx/marten/blob/0315c77fdf883b1db16784de6aec35de06dad40d/src/Marten/Events/Daemon/ProjectionDaemon.cs#L256

Example application: 

```csharp
using Microsoft.Extensions.Hosting;
using Oakton;

await Host.CreateDefaultBuilder(args)
    .ConfigureServices((hostContext, services) => services.AddInfrastructure(hostContext.Configuration))
    .RunOaktonCommands(args);
```

mysticmind:
@NikiforovAll please run through the following pointers and you can attempt to send us a PR:
- https://github.com/JasperFx/marten/blob/c848c99067c43926cefb4263b4958f6765aac54c/src/Marten.CommandLine/Commands/Projection/ProjectionInput.cs#L11, Add a new property `ShardTimeout` as a TimeSpan
- https://github.com/JasperFx/marten/blob/c848c99067c43926cefb4263b4958f6765aac54c/src/Marten.CommandLine/Commands/Projection/IProjectionHost.cs#L17, Add an optional arg named `TimeSpan? shardTimeout=null`  to `IProjectionHost.TryRebuildShards`
- https://github.com/JasperFx/marten/blob/c848c99067c43926cefb4263b4958f6765aac54c/src/Marten.CommandLine/Commands/Projection/ProjectionHost.cs#L111, change this line to `await daemon.RebuildProjection(projectionName, shardTimeout ?? 5.Minutes(), token).ConfigureAwait(false))`
- Update the unit tests code under https://github.com/JasperFx/marten/blob/c848c99067c43926cefb4263b4958f6765aac54c/src/Marten.CommandLine.Tests/ProjectionControllerTests.cs#L42 and also add unit tests with and without passing ShartTimeout.

PR #2388, opened 2022-11-04
Provide an option to specify projection rebuild command timeout for Marten.CommandLine

mysticmind:
Fixes GH-2360

### DIFF
--- a/src/Marten.CommandLine/Commands/Projection/IProjectionHost.cs
+++ b/src/Marten.CommandLine/Commands/Projection/IProjectionHost.cs
@@ -1,5 +1,7 @@
+using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Baseline.Dates;
 using Marten.Events.Daemon;
 
 namespace Marten.CommandLine.Commands.Projection;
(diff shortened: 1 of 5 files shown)


src/Marten.CommandLine/Commands/Projection/IProjectionHost.cs
```diff
@@ -14,7 +16,7 @@ public interface IProjectionHost
 {
     IReadOnlyList<IProjectionStore> AllStores();
     void ListenForUserTriggeredExit();
-    Task<RebuildStatus> TryRebuildShards(IProjectionDatabase database, IReadOnlyList<AsyncProjectionShard> asyncProjectionShards);
+    Task<RebuildStatus> TryRebuildShards(IProjectionDatabase database, IReadOnlyList<AsyncProjectionShard> asyncProjectionShards, TimeSpan? shardTimeout=null);
```
jeremydmiller, line 19: Ugh, I'm sorry, but we've gotta do that as an overload, not as an optional argument. That's actually a breaking change that would require a code compilation, even though the code itself wouldn't have to change

jeremydmiller (approved):
Works for me, but we better put a sample of the CLI usage in the docs later

oskardudycz (approved):

NikiforovAll:
I would like to suggest reopening this story. Because the timespan "1h 30 mins" is not applied even when specified explicitly

@mysticmind, sorry if the screen doesn't tell all the required info, but I can assure you that it fails in 5 minutes or so.

image

cc: @jeremydmiller

mysticmind:
@NikiforovAll thanks for the details, I will check on this and keep you posted.

PR #2391, opened 2022-11-09
Gh issue 2327 fix identity extension

### LINKED ISSUE #2327: Using Identity extension on type with Id field with inappropriate type
There is an issue when using the `DocumentMappingExpression.Identity` method  to set up projection's Id field on a type which has another property with the name "Id" and an inappropriate type for Marten (other than int/long/string/Guid). When `DocumentMappingBuilder` attempts to create `DocumentMapping` instance, its constructor sets up `IdMember = FindIdMember(documentType);`, which finds that Id field and throws an `ArgumentOutOfRangeException` without invoking `DocumentMappingExpression`, which would override that `IdMember` property.

The problem doesn't occur when there isn't an Id field, as `FindIdMember` returns null and `DocumentMappingExpression` can set it up later or when Id field has a valid type for Marten (even if `StreamIdentity` is set to `string` and that field is `Guid`) as `FindIdMember` can return a valid field.

I ran into this error when trying to use the Identity extension instead of attribute in [Oskar's strongly-typed identifiers with Marten example](https://github.com/oskardudycz/EventSourcing.NetCore/blob/main/Marten.Integration.Tests/CompositeIds/CompositeIdsTests.cs), but it can be easily recreated with this test:
```c#
[Fact]
public async Task stream_id_is_set_using_identity_expression_when_there_is_id_field_with_inappropriate_type()
{
        Action act = () => StoreOptions(x =>
        {
            x.Schema.For<IdentityProjection>().Identity(x => x.Key);
        });
        act.ShouldNotThrow();
}

public class IdentityProjection
{
        public short Id { get; set; }

        public Guid Key { get; set; }
}
```

grzegorzorwat:
Closes #2327

### DIFF
--- a/docs/documents/identity.md
+++ b/docs/documents/identity.md
@@ -61,7 +61,7 @@ public class NonStandardDoc
     public string Name;
 }
 ```
-<sup><a href='https://github.com/JasperFx/marten/blob/master/src/DocumentDbTests/Writing/Identity/using_natural_identity_keys.cs#L72-L79' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_identityattribute' title='Start of snippet'>anchor</a></sup>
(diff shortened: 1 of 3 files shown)


mysticmind (approved):
Looks good to me 👍

oskardudycz (approved):

PR #2392, opened 2022-11-10
Create connection and enlist transaction for async operations

Leh2:
Adding `BeginTransaction()` for sync `Apply` method. `BeginTransactionAsync(token)` is already present in the async apply.

Before the change a sync query operation would result in the error `System.InvalidOperationException: 'Connection property has not been initialized.'` when using `SessionOptions.ForCurrentTransaction`.


mysticmind (approved):
Looks good to me 👍

oskardudycz (approved):

PR #2396, opened 2022-11-15
Fix for bug 2353

oskardudycz:


oskardudycz:
This seems to be a dead-end leading to distributed transactions. I'll send a dedicated PR disabling such usage for other types of tenancy than conjoined.

PR #2407, opened 2022-12-01
QueryableExtensions.WriteSingle use optional argument onFoundStatus

gfoidl:
The optional argument wasn't used before. This change sets the response statuscode to the given value.

Or should this argument be dropped? It's not available on the other methods, but that would be a breaking change.

### DIFF
--- a/src/IssueService/Controllers/IssueController.cs
+++ b/src/IssueService/Controllers/IssueController.cs
@@ -39,13 +39,17 @@ namespace IssueService.Controllers
         #region sample_write_single_document_by_id_to_httpresponse
 
         [HttpGet("/issue/{issueId}")]
-        public Task Get(Guid issueId, [FromServices] IQuerySession session)
+        public Task Get(Guid issueId, [FromServices] IQuerySession session, [FromQuery] string? sc = null)
         {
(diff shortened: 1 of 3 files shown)


oskardudycz (approved):
Thanks for the fix, good catch 👍 
I think that we should also include that in other endpoints, as sometimes you may need to return other success statuses like, e.g. 201.

gfoidl:
> we should also include that in other endpoints

👍🏻 

Shall I do this within this PR or in a separate one? (I'm asking as I see you approved this change)

oskardudycz:
@gfoidl, I'm fine with either way, depending on your preferences. Also, it'd be great if you could add test checking this change.

jeremydmiller:
Hey @oskardudycz, if we're going to apply this to v5, it needs to be a separate overload instead of an optional argument. Adding an optional argument is a breaking *binary* change even though it doesn't change the targeting code if y'all didn't already know that

oskardudycz:
I thought about doing that forward only unless there’s a significant need to retrofit it🙂

oskardudycz:
Thank you @gfoidl!

PR #1500, opened 2020-06-07
Added transformation of the unique constraint exception into EventStreamUnexpectedMaxEventIdException

oskardudycz:
Added transformation of the unique constraint exception into EventStreamUnexpectedMaxEventIdException when the streamid with version unique constraint were hit

Closes #1465

PR #2409, opened 2022-12-01
Fix transformation for unique contraint on events

Leh2:
A previous update to Npgsql (my best guess 😇) broke the exception transformer for optimistic concurrency errors on events.

As also noted [here](https://github.com/JasperFx/marten/pull/1500), test for this transform is a bit tricky but I haven't had any issues with these tests local but a couple of spins in the pipeline should show any improprieties 🤞


oskardudycz (approved):
Looks good to me.

oskardudycz:
More info in https://github.com/npgsql/npgsql/issues/2501.

mysticmind (approved):
Looks good to me 👍

Issue #1812, opened 2021-05-10
Extract all features that depend on PLV8 to an addon library


jeremydmiller:
PLv8 doesn't look like it's in the greatest health as a project (Google's fault, not the plv8 devs).

jeremydmiller:
Notes:

* Rename to Marten.PLV8 later
* Do something on IUnitOfWork so you can find patch operations. Maybe an extension method called Patches(), but you'll have to expose a helper method on IUnitOfWork

Tasks:

- [ ] Move over transforms
- [ ] Move over all patching code
- [ ] Async versions of all the transforms
- [ ] Publish the Nuget
- [ ] Some kind of extension on `StoreOptions` like the NodaTime model

PR #2412, opened 2022-12-01
Add migration note about PLV8 removal

joshuaflanagan:
The PLV8Enabled option no longer exists as of https://github.com/JasperFx/marten/issues/1812, so the mention in the FAQ is confusing.

Add a note to the migration guide to direct people to the separate nuget package.


jeremydmiller:
Hey @joshuaflanagan, long time man! We got it, and thank you.

PR #2425, opened 2022-12-11
Fix RebuildProjection<TView> shardTimeout param passing

horego:
Hello everyone,

when calling RebuildProjection<TView> shardTimeout is not passed in the subroutine in the case of
`typeof(TView).CanBeCastTo(typeof(ProjectionBase)) && typeof(TView).HasDefaultConstructor()`
and defaults always to 5.Minutes().

best regards


mysticmind (approved):
Looks good to me 👍

oskardudycz (approved):

oskardudycz:
Thanks @horego!

PR #2427, opened 2022-12-11
Fix side effect on postprocessUpdateAsync

horego:
Hello everyone,

I have not digged to deep into the issue but i observe an error when rebuilding the projection. I ran into the issue where i got an exception from postprocessUpdateAsync while trying to execute
`await reader.IsDBNullAsync(0, token).ConfigureAwait(false)`
after
`!await reader.ReadAsync(token).ConfigureAwait(false)`

since there is no data so IsDBNullAsync throw an exception. When compared to the non async method postprocessUpdate all is written in one line with conditional or operator ||. For me it looks like it should behave the same. What do you think?

best regards


jeremydmiller:
I see what you're saying, and I agree. Thanks, I'm pulling this in for the next 6.0 alpha

PR #2217, opened 2022-05-10
Fix a compatibility issue with .NET 6

Baune8D:

PR #2431, opened 2022-12-12
Fix a compatibility issue with .NET 7 and Marten 3.x

Baune8D:
Basically same thing as last time: https://github.com/JasperFx/marten/pull/2217

I understand you might be reluctant to publish this on such an old branch, but i thought i would give it a shot anyway.

As also stated last time, we have a lot of microservices and it is no easy task to migrate from v3 to v5. We have been migrating alot of our services since last time, but it is no small undertaking for us to convert every single one, we simply need more time.

### DIFF
--- a/src/Marten/Marten.csproj
+++ b/src/Marten/Marten.csproj
@@ -61,7 +61,7 @@
     </ItemGroup>
 
     <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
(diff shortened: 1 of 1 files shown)


mysticmind:
@Baune8D you are still in the same boat as in May, I do understand it from a practical standpoint, will do the needful. 😄

mysticmind (approved):

oskardudycz (approved):

mysticmind:
@Baune8D 3.14.4 is published and available in NuGet

Issue #2398, opened 2022-11-18
NgramSearch doesn't handle mt_grams_vector function outside of public schema


claudiolassala:
I had Marten configured to use separate read and write schemas (inspired by GoldenEye). 

NgramSearch fails to find the mt_grams_vector in my "read" schema. When building the WhereFragment, it doesn't handle looking for that function anywhere but in the "public" schema.

apereiratl:
Well, the functions are added using the built-in marten functionality. Are you saying you expect these functions to be added to all schemas?

mysticmind:
I have zeroed on the issue. This will require the following changes:
- Access to database schema in the https://github.com/JasperFx/marten/blob/81cf3c2b3c7992bbd199f2325d94dd81ab7d5c13/src/Marten/Linq/Parsing/Methods/NgramSearch.cs#L23 so that we can quality the ngram search related functions to use the one from the schema.
- Also the SQL files (`mt_grams_query.sql` and `mt_grams_vector.sql`) will need to be fixes to ensure they are qualified with the schema name for function usage.

PR #2434, opened 2022-12-19
Fix GH-2398

mysticmind:
Fix GH-2398

### DIFF
--- a/src/Marten/Linq/Parsing/EnumerableContains.cs
+++ b/src/Marten/Linq/Parsing/EnumerableContains.cs
@@ -19,10 +19,10 @@ public class EnumerableContains: IMethodCallParser
                expression.Arguments.Single().IsValueExpression();
     }
 
-    public ISqlFragment Parse(IFieldMapping mapping, ISerializer serializer, MethodCallExpression expression)
+    public ISqlFragment Parse(IFieldMapping mapping, IReadOnlyStoreOptions options, MethodCallExpression expression)
     {
         var value = expression.Arguments.Single().Value();
-        return ContainmentWhereFragment.SimpleArrayContains(FindMembers.Determine(expression.Object), serializer,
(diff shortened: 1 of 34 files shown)


oskardudycz (commented):
As we're near the next major release, then I think that it could be better to pass StoreOptions through `IMethodCallParser.Parse` method and `LinqParsing.BuildWhereFragment`. It should be able to get it in `WhereCallParser.VisitMethodCall` as it has StoreOptions. It's a bit more invasive but more explicit and logical than exposing it through Field Mapping. At least IMHO.

@jeremydmiller, thoughts?

jeremydmiller:
I agree with Oskar about passing StoreOptions into IMethodParser in v6

mysticmind:
okay, will make changes accordingly.

mysticmind:
@jeremydmiller @oskardudycz I have made the changes as suggested, this PR is ready for review.

jeremydmiller (approved):
I think this is good as is, and it's the right call in the long run to pass in the read only view of the StoreOptions

src/Marten/Linq/Parsing/EnumerableContains.cs
```diff
@@ -19,10 +19,10 @@ public bool Matches(MethodCallExpression expression)
                expression.Arguments.Single().IsValueExpression();
     }
 
-    public ISqlFragment Parse(IFieldMapping mapping, ISerializer serializer, MethodCallExpression expression)
+    public ISqlFragment Parse(IFieldMapping mapping, IReadOnlyStoreOptions options, MethodCallExpression expression)
```
oskardudycz, line 22: Nice, @mysticmind, good that you removed also the obsolete serialiser param :+1:

oskardudycz (approved):
Looks good to me :+1:

jeremydmiller:
I think this is ready to go in.

PR #2449, opened 2023-01-08
Hot cold coordinator: Don't stop polling on errors.

Leh2:
If an error occurs during polling for ownership, the daemon will just idle. This change will ensure the coordinator keeps polling for ownership on errors.

The errors we see is transient errors like `System.IO.EndOfStreamException: Attempted to read past the end of the stream.` , but I think it makes sense to keep polling on all kind of errors.


jeremydmiller:
@Leh2 Got it, and I see what you mean.

Issue #2450, opened 2023-01-09
Host is disposed too early when running projection rebuild command


davit-asryan-sveasolar:
Host is disposed too early when running projection rebuild command.
```cs
using System.Threading.Tasks;
using Oakton;

namespace Marten.CommandLine.Commands.Projection;

[Description("Marten's asynchronous projection and projection rebuilds")]
public class ProjectionsCommand: OaktonAsyncCommand<ProjectionInput>
{
    public override Task<bool> Execute(ProjectionInput input)
    {
        using var host = input.BuildHost();

        var controller = new ProjectionController(new ProjectionHost(host), new ConsoleView());
       // possibly should be awaited here?
        return controller.Execute(input);
    }
}
```

gfoidl:
You refer to https://github.com/JasperFx/marten/blob/99d25cdd4bd31e3506d5a4dbb76614fe3135797f/src/Marten.CommandLine/Commands/Projection/ProjectionCommand.cs#L11-L15

Then for
> `// possibly should be awaited here?`

Yes. Otherwise the `host` will get disposed on return from that function while the `Execute` is still executing. That's a 🐛.

Do you want to submit a PR for this? (Otherwise I'll send one).

PR #2453, opened 2023-01-11
Fixed ProjectionCommand dispose / using bug

gfoidl:
Fixes https://github.com/JasperFx/marten/issues/2450


gfoidl:
Is there any way to run / create a unit-test for this?

jeremydmiller:
Not a unit test. I think this would require something with a lot of data to try to reproduce the issue. I'm comfortable taking this in as is.

PR #2462, opened 2023-01-18
Fix multithreading issues in Patching API

### LINKED ISSUE #2460: Marten.PLv8.Patching API not thread safe?
I am working with a client who uses the `Marten.PLv8.Patching` API in a multi-threaded application and we have seen a lot of unhandled exceptions in the logs which read like this one:

```
Unhandled exception. System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Marten.PLv8.Transforms.TransformSchema.AddFunction(TransformFunction function)
   at Marten.PLv8.Transforms.TransformSchema.Load(TransformFunction function)
   at Marten.PLv8.Transforms.TransformSchema.loadPatchDoc()
   at Marten.PLv8.Transforms.TransformSchema.schemaObjects()+MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
   at Marten.PLv8.Transforms.TransformSchema.get_Objects()
   at Weasel.Core.Migrations.DatabaseBase`1.generateOrUpdateFeature(Type featureType, IFeatureSchema feature, CancellationToken token)
   at Weasel.Core.Migrations.DatabaseBase`1.ensureStorageExists(IList`1 types, Type featureType, CancellationToken token)
   at Weasel.Core.Migrations.DatabaseBase`1.EnsureStorageExists(Type featureType)
   at Marten.Storage.MartenDatabase.Marten.Storage.IMartenDatabase.EnsureStorageExists(Type documentType)
   at Marten.PLv8.Transforms.TransformExtensions.EnsureTransformsExist(IDocumentOperations operations)
   at Marten.PLv8.Patching.PatchingExtensions.patchById[T](IDocumentOperations operations, Object id)
   at Marten.PLv8.Patching.PatchingExtensions.Patch[T](IDocumentOperations operations, Guid id)
   at Program.<<Main>$>g__PatchItemAsync|0_2(IDocumentStore store, Guid itemId) in /Users/dustingorski/Documents/Temp/MartenApp/src/MartenApp/Program.cs:line 56
   at Program.<<Main>$>g__PatchItemAsync|0_2(IDocumentStore store, Guid itemId) in /Users/dustingorski/Documents/Temp/MartenApp/src/MartenApp/Program.cs:line 57
   at Program.<Main>$(String[] args) in /Users/dustingorski/Documents/Temp/MartenApp/src/MartenApp/Program.cs:line 47
   at Program.<Main>(String[] args)
```

The first thing that stood out to me was that it felt very odd that the `Add` method of an `IDictionary<TKey, TValue>` would throw a `System.IndexOutOfRangeException`:

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
```

After further investigating I learned that this is a very common error when a dictionary gets accessed from multiple threads at the same time. It is not a thread safe data structure which is why .NET has the [`ConcurrentDictionary`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-7.0) inside `System.Collections.Concurrent`.

I created a small demo application to reproduce this error which can be found here:

https://github.com/dustinmoris/marten-test-app

I believe that the culprit is the [`TransformSchema`](https://github.com/JasperFx/marten/blob/master/src/Marten.PLv8/Transforms/TransformSchema.cs#L16) class which has a local `IDictionary` where it stores all transformation functions.

This dictionary is accessed during a patch operation when `loadPatchDoc()` is being called. I am not 100% sure, but after browsing through the Marten code I believe that the reason why this dictionary is being shared and accessed via multiple threads is because an instance of type `TransformSchema` is stored inside the `StoreOptions` object and then loaded via this static extension method, essentially making the `TransformSchema` a singleton because if I'm not mistaken the `StoreOptions` is also a singleton as part of the singleton `IDocumentStore`:

https://github.com/JasperFx/marten/blob/master/src/Marten.PLv8/Transforms/TransformExtensions.cs#L23

However, I am a little bit out of my depth here if I'm honest. I don't have too much experience with Marten and I might be misunderstanding some core concepts here or we might have used the Patching API in a way which it was never intended to be used. Anyhow, I'm happy to help with the investigation or even help with a PR if you think that I've found an issue indeed.

Many thanks for this library and keep up the great work!

oskardudycz:
The Patching API database schema feature didn't use the concurrent dictionary but a regular one. That was causing issues if multiple requests were trying to access it at once. Added `ImHashMap` to have thread-safe access and additional locking not to have redundant loading of a patch function js file. Added necessary tests.

Fixes #2460 

_**Note:** Best to review with [whitespace changes hidden](https://github.com/JasperFx/marten/pull/2462/files?diff=split&w=1) as I did some small reformatting._

### DIFF
--- a/src/Marten.PLv8.Testing/Patching/Bug_1173_patch_typenamehandling_bug.cs
+++ b/src/Marten.PLv8.Testing/Patching/Bug_1173_patch_typenamehandling_bug.cs
@@ -23,53 +23,49 @@ public class Bug_1173_patch_typenamehandling_bug: BugIntegrationContext
     [Fact]
     public void can_support_typenamehandling()
     {
-        using (var store = SeparateStore(_ =>
-               {
-                   var serializer = new JsonNetSerializer();
-                   serializer.Customize(config =>
-                   {
-                       config.TypeNameHandling = Newtonsoft.Json.TypeNameHandling.Objects;
-                   });
-                   _.Serializer(serializer);
-                   _.AutoCreateSchemaObjects = AutoCreate.All;
-
-                   _.UseJavascriptTransformsAndPatching();
-               }))
+        using var store = SeparateStore(_ =>
         {
-            using (var session = store.OpenSession())
+            var serializer = new JsonNetSerializer();
+            serializer.Customize(config =>
             {
-                var obj = new PatchTypeA
-                {
-                    Id = "1",
-                    TypeB =
-                        new PatchTypeB
-                        {
-                            Name = "test"
-                        }
-                };
+                config.TypeNameHandling = Newtonsoft.Json.TypeNameHandling.Objects;
+            });
+            _.Serializer(serializer);
+            _.AutoCreateSchemaObjects = AutoCreate.All;
 
-                session.Store(obj);
-                session.SaveChanges();
-            }
-            using (var session = store.OpenSession())
+            _.UseJavascriptTransformsAndPatching();
+        });
+        using (var session = store.OpenSession())
+        {
+            var obj = new PatchTypeA
             {
-                var newObj = new PatchTypeB
-                {
-                    Name = "test2"
-                };
+                Id = "1",
+                TypeB =
+                    new PatchTypeB
+                    {
+                        Name = "test"
+                    }
+            };
 
-                session.Patch<PatchTypeA>("1").Set(set => set.TypeB, newObj);
-                session.SaveChanges();
-            }
-
-            using (var session = store.OpenSession())
+            session.Store(obj);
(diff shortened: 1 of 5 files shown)


jeremydmiller (approved):
LGTM

PR #2467, opened 2023-01-19
Bug fixes for V6

### LINKED ISSUE #2405: Ensure that code generated path is correct
There was a failing test on the main branch with a mismatch in the code-generated path. Check if the assertion needs to be updated or if there's a bug.
See more in: https://github.com/JasperFx/marten/blob/master/src/CoreTests/MartenServiceCollectionExtensionsTests.cs#L102.

### LINKED ISSUE #2459: PostgresException not transformed to ExistingStreamIdCollisionException for same stream Id
**Summary**
If trying to add a stream with the same ID as an existing stream, the expected behaviour is that an `ExistingStreamIdCollision` exception is thrown.  When an inline projection is being used, the underlying `PostgresException` is surfaced instead.

Using Marten 5.11.0

**Steps to reproduce**
An existing test shows that you should expect an `ExistingStreamIdCollision` exception if trying to create a new new stream with the same Id as an existing one:

https://github.com/JasperFx/marten/blob/master/src/EventSourcingTests/start_stream_should_enforce_that_it_is_a_new_stream.cs

However, the original PostgresException (`23505: duplicate key value violates unique constraint "pkey_mt_streams_id`) is returned when creating a new stream, when there is an inline projection configured for the stream's events.

To show this, you can adapt the existing test at: https://github.com/JasperFx/marten/blob/master/src/EventSourcingTests/Aggregation/fetching_inline_aggregates_for_writing.cs#L16 (see below).

When you run the modified test it will throw `PostgresException` rather than the expected `ExistingStreamIdCollision`.

The difference seems to be that the passing test throws the exception in `Marten.Internal.UpdateBatch.ApplychangesAsync -> session.ExecuteReaderAsync` which then gets transformed in the `catch`.  But the "failing" test throws the exception in the `Marten.Internal.UpdateBatch.ApplychangesAsync -> UpdateBatch.ApplyCallbacksAsync` code, where the PostgresException doesn't get transformed when there are three operations in the batch (Upsert the document, InsertStream and Insert Event to Stream).

```c#
public class fetching_inline_aggregates_for_writing : OneOffConfigurationsContext
{
    [Fact]
    public async Task fetch_new_stream_for_writing_Guid_identifier()
    {
        StoreOptions(opts => opts.Projections.SelfAggregate<SimpleAggregate>(ProjectionLifecycle.Inline));

        var streamId = Guid.NewGuid();

        var stream = await theSession.Events.FetchForWriting<SimpleAggregate>(streamId);
        stream.Aggregate.ShouldBeNull();
        stream.CurrentVersion.ShouldBe(0);

        stream.AppendOne(new AEvent());
        stream.AppendMany(new BEvent(), new BEvent(), new BEvent());
        stream.AppendMany(new CEvent(), new CEvent());

        await theSession.SaveChangesAsync();

        /********* ADD THESE FIVE LINES *********/
        var sameStream = theSession.Events.StartStream(streamId, new AEvent());
        Exception<ExistingStreamIdCollisionException>.ShouldBeThrownBy(async () =>
        {
        });

        var document = await theSession.Events.AggregateStreamAsync<SimpleAggregate>(streamId);
        document.ACount.ShouldBe(1);
        document.BCount.ShouldBe(3);
        document.CCount.ShouldBe(2);
    }
```

jeremydmiller:
* Generated source path test
* Multi-tenancy database tests that have test timing issues
* Bumped Weasel version to enable reconnection for migration code.

Fixes #2405
Fixes #2459


oskardudycz:
Merging as the only failing test was from https://github.com/JasperFx/marten/issues/2480.

PR #2470, opened 2023-01-22
Fix DocumentMapping.IdMember setter

grzegorzorwat:
While working on #2391, I found this bug that may cause some strange behaviors. When setting DocumentMapping.IdMember to an invalid type ArgumentOutOfRangeException was thrown, but IdMember was charged to this invalid type anyway.


mysticmind (approved):
Looks good to me 👍

oskardudycz (approved):
Looks good :+1: Thanks!

PR #2472, opened 2023-01-25
Add event projection rebuild failure scenario

Hawxy:
Running into some undefined behavior with `EventProjection`, I assumed that each call to `ProjectAsync` would be (somewhat) transactional, but it appears to end up in the operations queue as part of a batch when running rebuilds. This can result in an entity previously being created now being unavailable for a follow up handler.


jeremydmiller:
@Hawxy Couple comments:

* You'll need to use an identity session in order for that to work at all even in the inline mode. Otherwise there's no magic "find the previously stored document"
* Any reason why you're not using an aggregation? That's completely optimized for this kind of thing
* I "fixed" the rebuild, but it's going to require a change that will make other event projections slower. I'm iffy

PR #2478, opened 2023-02-03
Concurrency check for Update in DocumentSessionBase

luboshl:
I've added support of concurrency check when `Update` method of `DocumentSessionBase` is used. It is based on implementation used for `Store` method.

### DIFF
--- a/src/DocumentDbTests/Bugs/optimistic_concurrency_with_update_method.cs
+++ b/src/DocumentDbTests/Bugs/optimistic_concurrency_with_update_method.cs
@@ -1,0 +1,140 @@
+using System;
+using System.Threading.Tasks;
+using Marten.Exceptions;
+using Marten.Metadata;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace DocumentDbTests.Bugs;
+
+public class OptimisticConcurrencyWithUpdateMethodStoreFixture: StoreFixture
+{
+    public OptimisticConcurrencyWithUpdateMethodStoreFixture(): base("optimistic_concurrency_with_update_method")
+    {
+    }
+}
+
+public class optimistic_concurrency_with_update_method: StoreContext<OptimisticConcurrencyWithUpdateMethodStoreFixture>,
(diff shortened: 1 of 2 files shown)


oskardudycz (changes requested):
@luboshl, thank you for the changes. Could you add a test that’s checking if the versioning work for Update after this change?

luboshl:
> @luboshl, thank you for the changes. Could you add a test that’s checking if the versioning work for Update after this change?

@oskardudycz, I've added the tests. Two of them (`update_with_stale_version_throws_exception` and `update_with_stale_version_throws_exception_async`) are failing when you run them before the the commit with changes for Update method. After this commit all four of them succeeds.

What I don't understand is this condition that I copied from Store method:

```csharp
if (Concurrency == ConcurrencyChecks.Disabled && storage.UseOptimisticConcurrency)
````

I would expect something like
```csharp
if (Concurrency == ConcurrencyChecks.Disabled || storage.UseOptimisticConcurrency == false)
```

What am I missing?

oskardudycz (approved):
Looks good to me 👍

oskardudycz:
> What I don't understand is this condition that I copied from Store method:
> 
> ```cs
> if (Concurrency == ConcurrencyChecks.Disabled && storage.UseOptimisticConcurrency)
> ```
> 
> I would expect something like
> 
> ```cs
> if (Concurrency == ConcurrencyChecks.Disabled || storage.UseOptimisticConcurrency == false)
> ```
> 
> What am I missing?

Indeed it looks suspicious. It looks like it is either for the non-default scenarios when the optimistic concurrency setting is enabled on the document level or some mishap. @jeremydmiller, do you remember what the reason is? It looks lik

jeremydmiller:
@oskardudycz That was to be able to manually override the concurrency checks on a document type with concurrency protection. In that case, you had to switch to using a different document operation type altogether. That code is actually correct.

jeremydmiller:
I added an extra test for the "concurrency checks disabled", but it was all good, so this is in.

PR #2479, opened 2023-02-04
Bumped Weasel version to enable Advisory Locks retries when got Admin Shutdown (57P01) Postgres error was thrown

oskardudycz:
Admin Shutdown (`57P01`) error mean that database is not available (e.g. it's restarting). Npgsql from version 6 supports connection pool reset behind the scenes for this scenario (see https://github.com/npgsql/npgsql/issues/2896). I added the wrapper in Weasel (see: https://github.com/JasperFx/weasel/pull/79) to return `false` for this case and allow retry mechanisms (that we already have in Marten) for database creation.

That should (hopefully) fix failing tests in CI.
Bumped also dependencies to the latest versions.

@jeremydmiller fyi.

### DIFF
--- a/Analysis.Build.props
+++ b/Analysis.Build.props
@@ -6,7 +6,7 @@
         <AnalysisLevel>latest</AnalysisLevel>
     </PropertyGroup>
     <ItemGroup>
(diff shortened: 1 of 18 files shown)


oskardudycz:
Merging, as a test failing, is not related. Adding a dedicated issue for that https://github.com/JasperFx/marten/issues/2480.

PR #2514, opened 2023-03-01
Renamed asynchronous session creation to include explicit Serializable name

oskardudycz:
This will be more self-explanatory about intentions and won't get Rider and other IDEs to suggest async versions. Removed Obsolete attribute from methods with StoreOptions


jeremydmiller (approved):
Ship it!

PR #2517, opened 2023-03-04
Recommend immutable types when using a tracked EventProjection

Hawxy:


mysticmind (approved):
Looks good to me 👍

oskardudycz (approved):
Sounds good 👍

PR #2527, opened 2023-03-11
Moved EnableDocumentTrackingDuringRebuilds from IProjection to AsyncOptions

oskardudycz:
After reconsideration of changes in #2510, moved `EnableDocumentTrackingDuringRebuilds` from `IProjection` to `AsyncOptions` and renamed it to `EnableDocumentTrackingByIdentity` as it won't be only applicable to rebuilds but also the general asynchronous processing.

`AsyncOptions` is a better place for that, as it keeps all the settings related to asynchronous projection handling. `IProjection` is our main interface, so it should be kept as generic as possible. We may include more implementation detail into it when needed once we have the subscription API defined and ready, but for now, it's better to keep it generic.

Added also `AsyncOptions` to inline projection registration to add the possibility to define behaviour during the projection rebuild.

@Hawxy FYI, apologies for changing the API, but it's the last time to do it before we release v6. 

**Note:** It's a follow up to #2472, #2510 and #2517.

### DIFF
--- a/src/EventSourcingTests/EventStreamUnexpectedMaxEventIdExceptionTransformTest.cs
+++ b/src/EventSourcingTests/EventStreamUnexpectedMaxEventIdExceptionTransformTest.cs
@@ -54,7 +54,7 @@ public class EventStreamUnexpectedMaxEventIdExceptionTransformTest: IntegrationC
 
         async Task ForceEventStreamUnexpectedMaxEventIdException()
         {
-            await Parallel.ForEachAsync(Enumerable.Range(1, 50), async (_, token) =>
+            await Parallel.ForEachAsync(Enumerable.Range(1, 100), async (_, token) =>
             {
                 await using var session = theStore.LightweightSession();
                 session.Events.Append(streamId, departed);


src/Marten/Events/Daemon/TenantActionGroup.cs
```diff
@@ -22,8 +22,14 @@ public TenantActionGroup(Tenant tenant, IEnumerable<StreamAction> actions)
     public async Task ApplyEvents(ProjectionUpdateBatch batch, IProjection projection, DocumentStore store,
         CancellationToken cancellationToken)
     {
+        var tracking =
```
jeremydmiller, line 25: Not now, but I think the Daemon code gets some serious love soon. I'd prefer to be passing `AsyncOptions` around explicitly w/o the optional downcast. Do that when we tackle first class subscriptions. We'll use `AsyncOptions` with subscriptions I'd imagine.

oskardudycz: Yes, I wasn't fully happy about that, but I wanted to make it less invasive. I think that while working on Subscriptions, and especially on v7 we should revisit our abstractions on top of projections and maybe shuffle them a bit to make them more straightforward.

src/Marten/Events/Projections/ProjectionOptions.cs
```diff
@@ -177,13 +202,20 @@ public void Add<TProjection>(ProjectionLifecycle? lifecycle = null) where TProje
...
+        ProjectionLifecycle? lifecycle = null,
+        Action<AsyncOptions> asyncConfiguration = null
+    )
     {
         if (lifecycle.HasValue)
         {
             projection.Lifecycle = lifecycle.Value;
         }
 
+        asyncConfiguration?.Invoke(projection.Options);
```
jeremydmiller, line 217: Do you just want to go all the way for v6 and remove Lifecycle from the projection classes and make users explicitly specify that as part of registration?

jeremydmiller:
@oskardudycz I'm good with this getting pushed at any time

oskardudycz:
> @oskardudycz I'm good with this getting pushed at any time

Ok, then I'm taking that in, and we'll send PR covering in the following days https://github.com/JasperFx/marten/pull/2527#discussion_r1142148430.

PR #2549, opened 2023-04-10
Update introduction.md

mohsin-mehmood:
Updated text:
".NET Core 3.1 and .NET 5 was dropped as part of v6 release. If you require .NET Framework support, please use the latest Marten v5 release."


mysticmind (approved):
Looks good to me 👍

PR #2559, opened 2023-04-17
Ensure events with both `Create` and `Apply` are only ever handled once

### LINKED ISSUE #2528: Allow projections to be started and modified by the same event type
I have a use case where a projection is started by the same event type that will later modify it.

While event streams appear to support this design without issues, projections appear to invoke both `Create` and `Apply` for the initial event and all future events. For my use case, `Create` initializes a projection member with a value in that later is modified by new events. Since `Create` is also called for future events, it always resets this member to the initial value.

Known workaround: Provide a constructor on the projection type, setting the initial value. Remove `Create`, but keep the `Apply` method.

```cs
using System;
using System.Threading.Tasks;

using EventSourcingTests.Aggregation;

using Marten.Events.Aggregation;
using Marten.Testing.Harness;

using Shouldly;

using Xunit;
using Xunit.Abstractions;

namespace EventSourcingTests;

public class Allow_Create_and_Apply_with_same_event_type: AggregationContext
{
    private readonly ITestOutputHelper _output;

    public Allow_Create_and_Apply_with_same_event_type(DefaultStoreFixture fixture, ITestOutputHelper output):
        base(fixture: fixture)
    {
        _output = output;
    }

    [Fact]
    public async Task works()
    {
        var stream1 = Guid.NewGuid();

        UsingDefinition<CreateAndApplyProjection>();

        _output.WriteLine(_projection.SourceCode());

        // The initial command event "starts" the projection.
        // IMHO, CreateAndApplyProjection.Create() should be called (ACount == 1),
        // but CreateAndApplyProjection.Apply() should not be called (BCount == 0).
        await InlineProject(x =>
        {
            x.Streams[key: stream1].IsNew = true;
            x.Streams[key: stream1].Add(new StartsOrExtendsStream(Id: stream1));
        });

        await using var query1 = theStore.QuerySession();

        var readModel1 = await query1.LoadAsync<MyAggregate>(id: stream1);
        readModel1.ACount.ShouldBe(expected: 1);
        readModel1.BCount.ShouldBe(expected: 0);

        // Later a new emitted event "appends to" the current projection's snapshot.
        // IMHO, CreateAndApplyProjection.Create() should not be called (ACount == 1),
        // but CreateAndApplyProjection.Apply() should be called (BCount == 1).
        await InlineProject(x =>
        {
            x.Streams[key: stream1].Add(new StartsOrExtendsStream(Id: stream1));
        });

        await using var query2 = theStore.QuerySession();

        var readModel2 = await query2.LoadAsync<MyAggregate>(id: stream1);
        readModel2.ACount.ShouldBe(expected: 1);
        readModel2.BCount.ShouldBe(expected: 1);
    }
}

public record StartsOrExtendsStream(Guid Id);

public class CreateAndApplyProjection: SingleStreamAggregation<MyAggregate>
{
    private static int NumberOfCreateCalls = 0;

    public static MyAggregate Create(StartsOrExtendsStream ev)
    {
        return new MyAggregate { Id = ev.Id, ACount = ++NumberOfCreateCalls, BCount = 0};
    }

    public MyAggregate Apply(StartsOrExtendsStream ev, MyAggregate current)
    {
        return new MyAggregate() { Id = current.Id, ACount = current.ACount, BCount = current.BCount + 1 };
    }
}
```

elexisvenator:
Fixes #2528

Expected outcome: Each event passed into a projection is handled a maximum of once, either with a `Create` or an `Apply` method.

Old behaviour: If the first event in a stream has matching `Create` and `Apply` methods, both would be called.

New behaviour: If a `Create` method matches, that event is then skipped when iterating over apply event methods.

Example of issue:

```c#
    public record IncrementEvent{}

    public record CounterProjection(Guid Id, int Counter)
    {
        public static CounterProjection Create(IncrementEvent @event, IEvent metadata)
        {
            return new CounterProjection(metadata.StreamId, 1);
        }

        public CounterProjection Apply(IncrementEvent @event, CounterProjection current)
        {
            return current with {Counter = current.Counter + 1};
        }
    }

    [Fact]
    public async Task count_of_events_should_match_event_count()
    {
        var streamId = Guid.NewGuid();
        theSession.Events.Append(streamId, new IncrementEvent(), new IncrementEvent(), new IncrementEvent());
        await theSession.SaveChangesAsync();

        var aggregate = await theSession.Events.AggregateStreamAsync<CounterProjection>(streamId);

        Assert.NotNull(aggregate);
        Assert.Equal(3, aggregate.Counter); // Returned 4 before fix
    }
```

This required a bit of fiddling around with the codegeneration which I am definitly not super confident in. Pretty sure my changes are solid thanks to the tests covering them, but my approach may be off.

### DIFF
--- a/docs/configuration/hostbuilder.md
+++ b/docs/configuration/hostbuilder.md
@@ -246,7 +246,7 @@ public interface IConfigureMarten
     void Configure(IServiceProvider services, StoreOptions options);
 }
 ```
-<sup><a href='https://github.com/JasperFx/marten/blob/master/src/Marten/MartenServiceCollectionExtensions.cs#L732-L743' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_iconfiguremarten' title='Start of snippet'>anchor</a></sup>
+<sup><a href='https://github.com/JasperFx/marten/blob/master/src/Marten/MartenServiceCollectionExtensions.cs#L734-L745' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_iconfiguremarten' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
 You could alternatively implement a custom `IConfigureMarten` class like so:
--- a/docs/events/projections/aggregate-projections.md
+++ b/docs/events/projections/aggregate-projections.md
@@ -97,7 +97,7 @@ public class TripProjection: SingleStreamProjection<Trip>
     }
 }
 ```
-<sup><a href='https://github.com/JasperFx/marten/blob/master/src/Marten.AsyncDaemon.Testing/TestingSupport/TripProjectionWithCustomName.cs#L44-L74' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_tripprojection_aggregate' title='Start of snippet'>anchor</a></sup>
+<sup><a href='https://github.com/JasperFx/marten/blob/master/src/Marten.AsyncDaemon.Testing/TestingSupport/TripProjectionWithCustomName.cs#L43-L73' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_tripprojection_aggregate' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
 And register that projection like this:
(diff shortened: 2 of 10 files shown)


oskardudycz (commented):
This looks good to me in general, but I'm unsure if when we have https://github.com/JasperFx/marten/pull/2560 then if that'll be needed. Having static Apply method capabilities could be enough, as it will do upsert. The logic in this PR seems okay, but it'll add more magic to the usage pattern. @elexisvenator, @jeremydmiller thoughts?

jeremydmiller:
I'm asking #2560 to be changed, and I see this as orthogonal anyway. Good stuff, I'm pulling this in now.

agross:
Is this included in Marten 6.0?

mysticmind:
> Is this included in Marten 6.0?

yes @agross!

agross:
I was asking because after upgrading to 6.0 I tested this.

The second transaction affecting the projection still calls `Create` first and `Apply` second (like it did before 6.0 was released). 

Since the test above models a single transaction, could this be the reason why it does not work in my use case? 🤔

agross:
@mysticmind See https://github.com/JasperFx/marten/issues/2528#issuecomment-1534467386

PR #2560, opened 2023-04-17
Allow immutable aggregate types to use `Apply` methods

elexisvenator:
If an `Apply` method accepts the aggregate as a parameter, and returns the aggregate (or Task<Aggregate>), then it is allowed to be static.

The motivation behind this is to allow immutable self aggregates to not have any instance methods on them, which makes their usage a bit cleaner.

Examples:

```c#
public record TheAggregate(Guid Id, string Value)
{
    public static TheAggregate Create(CreateEvent @event, IEvent metadata)
    {
        return new TheAggregate(metadata.StreamId, @event.Value);
    }

    // Static apply method
    public static TheAggregate Apply(UpdateEvent @event, TheAggregate aggregate)
    {
        return aggregate with {Value = @event.Value};
    }
}

### DIFF
--- a/docs/configuration/hostbuilder.md
+++ b/docs/configuration/hostbuilder.md
@@ -246,7 +246,7 @@ public interface IConfigureMarten
     void Configure(IServiceProvider services, StoreOptions options);
 }
 ```
-<sup><a href='https://github.com/JasperFx/marten/blob/master/src/Marten/MartenServiceCollectionExtensions.cs#L732-L743' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_iconfiguremarten' title='Start of snippet'>anchor</a></sup>
+<sup><a href='https://github.com/JasperFx/marten/blob/master/src/Marten/MartenServiceCollectionExtensions.cs#L734-L745' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_iconfiguremarten' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
 You could alternatively implement a custom `IConfigureMarten` class like so:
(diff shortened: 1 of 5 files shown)


oskardudycz (approved):
Looks good to me, nice stuff @elexisvenator

mysticmind (approved):
Looks good to me. Adding docs pertaining to this would be helpful.

jeremydmiller (changes requested):
I'm a no on this one. It's a valid use case to want to use mutable aggregates with a static `Apply()` method. This creates a hidden rule that isn't necessary. 

*Now*, if you want to get fancier and detect somehow that the aggregate itself is immutable, **then** opt into that logic or declare on the projection that the aggregate is immutable, then I'm all good with this exact validation logic.

elexisvenator:
I think I was wrong to describe this in the context of immutable aggregates. While that is the context of where static apply methods should probably be used (in my opinion), the exact behavior here is:

> If an `Apply` method is pure, that is has no side effects and interacts only with state passed through parameters, then it should be ok for that method to be static.

elexisvenator:
I don't consider this to be a breaking change because:

* Existing `Apply` methods that are not static are not affected in any way
* If a projection currently has a static `Apply` method on it, this would already cause marten code generation to fail.

jeremydmiller:
Hey man, I honestly don't like to reject pull requests from folks that have obviously put in a lot of effort, but I don't think this one fits or is necessary. This could break currently existing code without being too obvious about why. I appreciate that you don't think the usage I worry about here is valid, but you have no earthly idea what *other* people might already be doing.

elexisvenator:
Apologies @jeremydmiller , I did some more testing and it turns out I was mistaken between create and apply methods. If an aggregate has a non-static method called `Create` then code generation will fail. A static method called `Apply` on an aggregate is ignored by code generation.

PR #2581, opened 2023-05-08
Ensure hot-cold coordinator exits if host cancellation occurs

Hawxy:
This was causing a deadlock in my test pipeline as the lock was being retried forever.


oskardudycz (approved):
Looks good to me; thank you for the fix!

PR #2603, opened 2023-06-12
Fix: Similarly named where clauses get incorrectly reused

elexisvenator:
When parsing expressions, marten will cache expression paths for future reuse. Unfortunately the caching key is not specific enough. For example, the following nested properties on document x:

* `x.Foo.Bar.Baz`
* `x.FooBar.Baz`
* `x.Foo.BarBaz`
* `x.FooBarBaz`

Will all be cached with a key of `FooBarBaz`. This can lead to incorrect query generation (see included tests).
The solution is to update the cache key to separate all properties with a `.`.

### DIFF
--- a/src/CoreTests/Bugs/Bug_2603_similar_named_props.cs
+++ b/src/CoreTests/Bugs/Bug_2603_similar_named_props.cs
@@ -33,14 +33,14 @@ public class Bug_2603_similar_named_props: BugIntegrationContext
 
         // actual query:
         //   select d.id, d.data
-        //   from bugs.mt_doc_bug_9003_similar_named_props_document as d
+        //   from bugs.mt_doc_bug_2603_similar_named_props_document as d
         //   where (d.tenant_id = $1  and  (d.data -> 'Foo' ->> 'BarBaz' = $2 or d.data -> 'Foo' ->> 'BarBaz' = $3))
         //                                                            incorrect--^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
         // parameters: $1 = '*DEFAULT*', $2 = 'aaa', $3 = 'ddd'
 
         // query should be:
         //   select d.id, d.data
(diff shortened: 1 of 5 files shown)


elexisvenator:
Ok looks like this breaks duplicate fields :(
I think the solution is to find where duplicate fields are mapped to and make the same change to the "key" there as well. Not sure where that is yet.

src/CoreTests/Bugs/Bug_2603_similar_named_props.cs
```diff
@@ -0,0 +1,58 @@
...
+        // actual query:
+        //   select d.id, d.data
+        //   from bugs.mt_doc_bug_9003_similar_named_props_document as d
+        //   where (d.tenant_id = $1  and  (d.data -> 'Foo' ->> 'BarBaz' = $2 or d.data -> 'Foo' ->> 'BarBaz' = $3))
+        //                                                            incorrect--^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+        // parameters: $1 = '*DEFAULT*', $2 = 'aaa', $3 = 'ddd'
+
+        // query should be:
+        //   select d.id, d.data
+        //   from bugs.mt_doc_bug_9003_similar_named_props_document as d
```
elexisvenator, line 43:
```suggestion
        //   from bugs.mt_doc_bug_2603_similar_named_props_document as d
        //   where (d.tenant_id = $1  and  (d.data -> 'Foo' ->> 'BarBaz' = $2 or d.data -> 'Foo' ->> 'BarBaz' = $3))
        //                                                            incorrect--^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        // parameters: $1 = '*DEFAULT*', $2 = 'aaa', $3 = 'ddd'

        // query should be:
        //   select d.id, d.data
        //   from bugs.mt_doc_bug_2603_similar_named_props_document as d
```

oskardudycz (approved):
Looks good to me; thank you!

PR #2625, opened 2023-07-02
Support upcasters within async projections

### LINKED ISSUE #2615: Async Daemon does not take upcasters into consideration.
Mentioned on Discord. Upcasters are currently limited to inline aggregation and do not get applied for async projections or rebuilds.

Hawxy:
Closes #2615

### DIFF
--- a/src/EventSourcingTests/SchemaChange/Upcasters.cs
+++ b/src/EventSourcingTests/SchemaChange/Upcasters.cs
@@ -6,6 +6,8 @@ using System.Threading;
 using System.Threading.Tasks;
 using Marten;
 using Marten.Events;
+using Marten.Events.Aggregation;
+using Marten.Events.Projections;
 using Marten.Services.Json;
 using Marten.Services.Json.Transformations;
 using Marten.Testing;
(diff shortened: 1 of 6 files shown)


src/EventSourcingTests/SchemaChange/Upcasters.cs
```diff
@@ -734,14 +736,14 @@ private void Apply(ShoppingCartOpened @event)
     // Events in new namespace `New`
     namespace New
     {
-        public class ShoppingCart: AggregateBase
+        public class ShoppingCartAggregate: AggregateBase
```
oskardudycz, line 739: **QUESTION:** @Hawxy, is there any reason for renaming this class? I'd prefer not to have suffixes in the type names, especially those that are used in the public documentation.

src/Marten/Events/Daemon/EventTypeFilter.cs
```diff
@@ -9,18 +12,19 @@ namespace Marten.Events.Daemon;
...
         EventTypes = eventTypes;
-        _typeNames = eventTypes.Select(x => graph.EventMappingFor(x).Alias).ToArray();
+        // We need to load events that are mapped to this event type, not just the event itself.
+        var additionalAliases = graph.AliasesForEvents(eventTypes);
+        _typeNames = eventTypes.Select(x => graph.EventMappingFor(x).Alias).Union(additionalAliases).ToArray();
     }
 
-    public Type[] EventTypes { get; }
-
+    public IReadOnlyCollection<Type> EventTypes { get; }
```
oskardudycz, line 27: **QUESTION:** Could you expand on the change from Array to ReadonlyCollection?

oskardudycz (approved):
Looks good to me, just minor comments around the naming and API changes.

src/EventSourcingTests/SchemaChange/Upcasters.cs
Hawxy: Good point, has been fixed.

src/Marten/Events/Daemon/EventTypeFilter.cs
Hawxy: Just avoids the `ToArray` when initialized from the projection base.

PR #2626, opened 2023-07-03
Add Dictionary ContainsKey support for value types

Hawxy:
Also permits any subclass of  `IDictionary`, as existing support was limited to the `IDictionary<string, x>` interface.

I loosened the type requirements for `.Contains` as well, but the current implementation only works if an interface is passed due to Remotion treating it as a different type of expression.


oskardudycz (approved):
Thnks @Hawxy!

PR #2634, opened 2023-07-06
Enable async projections to respect `EnableDocumentTrackingByIdentity` in regular mode, not only during the rebuild

oskardudycz:
`ProjectionDocumentSession` didn't handle `DocumentTracking` session options and always returned a lightweight session. Updated the code to return the correct tracking mode depending on the `EnableDocumentTrackingByIdentity` setting.

@jameswoodley, @Hawxy FYI

@ugugun, @alexander-paynova I think it might also be the fix issue we discussed around nulls in async projection handling.


oskardudycz:
@jeremydmiller agreed on Discord, so I'm merging that 👌😎

PR #2637, opened 2023-07-07
Automatically enable extensions for tenants

T0shik:
Expecting the extension to be automatically enabled in new tenants.

### DIFF
--- a/docs/configuration/cli.md
+++ b/docs/configuration/cli.md
@@ -2,7 +2,7 @@
 
 ::: warning
 The usage of Marten.CommandLine shown in this document is only valid for applications bootstrapped with the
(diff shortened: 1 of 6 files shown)


oskardudycz:
@T0shik, I added some tests and more docs around applying schema changes for multi-database setup.

If you didn't call the `ApplyAllDatabaseChangesOnStartup` method, Marten would still try to create a database upon the session creation. This action is invasive and can cause issues like timeouts, cold starts, or deadlocks. It also won't apply all defined changes upfront (so, e.g. indexes, custom schema extensions).

If you don't know the tenant upfront, you can create and apply changes dynamically by:

```csharp
var tenant = await theStore.Tenancy.GetTenantAsync(tenantId);
await tenant.Database.ApplyAllConfiguredChangesToDatabaseAsync();
```


You can place this code somewhere in the tenant initialization code. For instance:
- tenant setup procedure,
- dedicated API endpoint
- custom session factory, although that's not recommended for the reasons mentioned above.

I hope that clears things up. Thoughts?

PR #2642, opened 2023-07-11
Fixed querying for primitive types in custom SQL

Hawxy:
Permit querying any [npgsql mapped type](https://www.npgsql.org/doc/types/basic.html) within custom SQL.


oskardudycz (approved):
@Hawxy, thank you again for a nice improvement.

I added two more test cases: for integer and string.

Issue #2610, opened 2023-06-21
LiveStreamAggregation generates table not used


devployment:
Given the following config

```csharp
opts.AutoCreateSchemaObjects = AutoCreate.None;
opts.GeneratedCodeMode = TypeLoadMode.Static;
opts.SourceCodeWritingEnabled = false;
```

Adding a `LiveStreamAggregation` to satisfy code generation

`opts.Projections.LiveStreamAggregation<Issue>();`

Results in database changes to create a table that will not be used by executing

`dotnet run -- db-patch -d Marten ./marten-patch.sql`

I expected a `LiveStreamAggregation` to be handled in memory only. Not creating tables not being used. For my understanding this would be the case for a (persisted) snapshot. 

The interesting part is, that everything runs without applying the database changes as the table seems to be never used. So generating the table (migration) seems to be a side effect of the tooling.

oskardudycz:
The reason is that `LiveStreamAggregation` is reusing the code from `Snapshot` and defining `DocumentMapping` behind the scenes, which creates the table. This table won't be ever updated, but it's better not to have it, and it's a bug we should fix.

PR #2654, opened 2023-07-21
Added test for LiveStreamAggregation code file generation

### LINKED ISSUE #2645: `codegen write` does not generate `*Provider.cs` for `LiveStreamAggregation`
After upgrading from Marten 6.0.2 to 6.0.4 `dotnet run -- codegen write` no longer creates `*Provider.cs` files for `LiveStreamAggregation`s registered in the Marten config.

Probably related to the changes in 6.0.3 where tables for `LiveStreamAggregations` are no longer created in the DB schema. Generating providers might also have gotten lost in the process of making that change.

oskardudycz:
Changed LiveStreamAggregation implementation to set the `SkipSchemaGeneration` property for `DocumentMapping` instead of entirely removing it from storage.

Updated implementation for schema generation to skip such mappings.

It's more hotfix, than a real revamp, but without introducing breaking changes, it'll be harder to change it.

Fixes #2645

Relates to #2610

### DIFF
--- a/src/EventSourcingTests/Projections/CodeGeneration/ProjectionCodeGenerationTests.cs
+++ b/src/EventSourcingTests/Projections/CodeGeneration/ProjectionCodeGenerationTests.cs
@@ -5,6 +5,7 @@ using JasperFx.Core.Reflection;
 using Marten;
 using Marten.Events.Aggregation;
 using Marten.Events.Projections;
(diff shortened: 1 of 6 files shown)


oskardudycz:
@jeremydmiller said gave me no points for style but approved on Discord so I'm merging that 😉

Issue #2660, opened 2023-07-26
Deleting multiple document types in one session (Identity) doesn't work


migajek:
This might or might not be related to the issue I've reported recently #2653 

When calls `.Delete<T>(Id);` on multiple document types are issued within one session, they're not actually executed - the documents are not removed when SaveChangesAsync is called.

Please see the tests below:

```
public class FooModel
    {
        public Guid Id { get; set; }
    }

    public class BarModel
    {
        public Guid Id { get; set; }
        public Guid RelGuid { get; set; }
    }

    public class DeleteTests
    {
        // works
        [Fact]
        public async Task Deleting_Single_DocType_In_One_Session_Works()
        {
            using var store = DocumentStore.For(ConnectionStrings.Temp);

            // store & delete within same session
            using var sess = store.IdentitySession();
            var id = Guid.NewGuid();
            sess.Store(new FooModel() { Id = id });
            sess.Delete<FooModel>(id);
            await sess.SaveChangesAsync();

            using var sess2 = store.IdentitySession();
            var model = sess.Load<FooModel>(id);
            Assert.Null(model);
        }

        // fails
        [Fact]
        public async Task Deleting_Multiple_DocTypes_In_One_Session_Works()
        {
            using var store = DocumentStore.For(ConnectionStrings.Temp);

            // store & delete within same session
            using var sess = store.IdentitySession();
            var id = Guid.NewGuid();
            sess.Store(new FooModel() { Id = id });
            sess.Store(new BarModel() { Id = id });
            sess.Delete<FooModel>(id);
            sess.Delete<BarModel>(id);
            await sess.SaveChangesAsync();

            using var sess2 = store.IdentitySession();
            var model = sess.Load<FooModel>(id);
            Assert.Null(model);
        }

        // also fails
        [Fact]
        public async Task Delete_Where_Within_Same_Session_Doesnt_Affect_Actual_Delete()
        {
            using var store = DocumentStore.For(ConnectionStrings.Temp);

            // store & delete in the same session
            using var sess = store.IdentitySession();
            var id = Guid.NewGuid();
            sess.Store(new FooModel() { Id = id });
            sess.Store(new BarModel() { Id = Guid.NewGuid(), RelGuid = id });
            sess.Delete<FooModel>(id);
            sess.DeleteWhere<BarModel>(x => x.RelGuid == id);
            await sess.SaveChangesAsync();

            using var sess2 = store.IdentitySession();
            var model = sess.Load<FooModel>(id);
[...]

PR #2661, opened 2023-07-26
#2660 possible fix for deleting issues

migajek:
as communicated on discord, I'm not certain this is the right way to fix it.


jeremydmiller:
I don't think this is the right fix. I'd rather focus on making the "delete" also Eject() instead of having to worry about the ordering. The code you're touching has been very problematic in the past and I'm very loathe to revisit that.

jeremydmiller:
Yeah, sorry, we'll beat the delete issue another way. I don't want the very real regression bug risk here. But thank you for making the effort!

PR #2668, opened 2023-08-03
Rewrote introduction docs to make it more accessible and user friendly

oskardudycz:
I took inspiration from the Nest.js docs, which are praised by the community https://nextjs.org/docs:

Rewrote to show people first what we do, then how we do it and how they can benefit.

Outlined the most important features so people won't miss them.

Removed:
- link to obsolete videos,
- information about NRT, as it should be a standard now and is not adding much value for the reader of the introduction (too many tech details).


mysticmind:
@oskardudycz  Looks great to me, very clear, crisp and more structured to the point!

oskardudycz:
Thanks, @babuannamalai!

PR #2669, opened 2023-08-03
Added `readonly` modifier to fields that are never changed

Lehonti:


oskardudycz (approved):
Looks good; thanks @Lehonti, and welcome as the new Marten contributor!

PR #2672, opened 2023-08-04
Private methods that don't access any instance members were made `static`

Lehonti:


oskardudycz:
@Lehonti, could you rebase the branch instead of merging the master?

PR #2673, opened 2023-08-05
Added warning regarding connection bleed to the async-enumerable docs

FelixLorenz:


docs/documents/querying/linq/async-enumerable.md
```diff
@@ -40,5 +40,10 @@ public async Task query_to_async_enumerable()
     }
 }
 ```
+
+::: warning
+Be aware not to return the IAsyncEnumerable out of the scope in which the session that produces it is used. This would prevent the data base connection from being reused afterwards and thus leading to a connection bleed.
+:::
+
```
oskardudycz, line 47:
```suggestion

::: warning
Be aware not to return the IAsyncEnumerable out of the scope in which the session that produces it is used. This would prevent the database connection from being reused afterwards and thus lead to a connection bleed.
:::

```

oskardudycz (approved):
@FelixLorenz, thank you for the submission and welcome as the Marten contributor! 🙂

PR #2674, opened 2023-08-06
querying child collections impacts each other

T0shik:
Queries that work on their own, don't work if one runs before the other.

The docs are a little bit hazy on this example, should this work at all? 
or like nested in Any() both sides have to be arrays?
```
var guids = new List<Guid> { guid1, new Guid() }; 
var result = await query.Query<Two>()
    .Where(x => guids.Contains(x.Id))
    .ToListAsync();

result.Count.ShouldBe(1);
```

if Contains shouldn't be used like this I think it's important to throw an exception early, this was quite the head scratcher. 

There is a current work around using `IsOneOf` or `In` so no rush.

### DIFF
--- a/src/DocumentDbTests/Bugs/querying_child_collections_impact_each_other.cs
+++ b/src/DocumentDbTests/Bugs/querying_child_collections_impact_each_other.cs
@@ -1,0 +1,160 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Marten;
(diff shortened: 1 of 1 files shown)


jeremydmiller:
I do think I know what causes this, but I'm shoving to to the V7 release.

jeremydmiller:
This is actually fixed in the `linq` branch, but I'll still pull this into that branch as a test soon to verify

jeremydmiller:
And boom, this was actually fixed by the V7 changes

PR #2695, opened 2023-08-30
Documentation: Using Custom Grouper with Fan Out Feature for Event Projections

erdtsieck:


docs/events/projections/multi-stream-projections.md
```diff
@@ -453,3 +453,94 @@ public class DayProjection: MultiStreamProjection<Day, int>
...
+### The Scenario
+
+Imagine you have a system where `EmployeeAllocated` events contain a list of allocations for specific days. The goal is to project this information into a monthly summary. 
+
+### Custom Projection with Custom Grouper
+
+The `MonthlyAllocationProjection` class uses a custom grouper for this transformation. Here, `TransformsEvent<EmployeeAllocated>()` indicates that events of type `EmployeeAllocated` will be used even if there are no direct handlers for this event type in the projection.
+
+<!-- snippet: snippet-sample_showing_registering_custom_grouper_with_fan_out -->
+<a id='snippet-sample_showing_registering_custom_grouper_with_fan_out'></a>
```
oskardudycz, line 470: **FIX:** Marten gets snippets automatically thanks to [MarkdownSnippets](https://github.com/SimonCropp/MarkdownSnippets). To have such, you need to find the source code (usually located in tests) and wrap it with the exact region name and use it in docs with the `snippet-` prefix. The code sample is located here: https://github.com/JasperFx/marten/blob/4dafe48f4209a90b5c66ee8a9a2a60e15d7ca991/src/EventSourcingTests/Projections/MultiStreamProjections/CustomGroupers/custom_grouper_with_events_transformation.cs#L41.

docs/events/projections/multi-stream-projections.md
```diff
@@ -453,3 +453,94 @@ public class DayProjection: MultiStreamProjection<Day, int>
 ```
 <sup><a href='https://github.com/JasperFx/marten/blob/master/src/Marten.AsyncDaemon.Testing/ViewProjectionTests.cs#L124-L168' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_showing_fanout_rules' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
+
+## Using Custom Grouper with Fan Out Feature for Event Projections
+
+In Marten, the `MultiStreamProjection` feature allows for complex transformations and aggregations of events. However, there might be scenarios where a single event in your domain carries information that is more suitable to be distributed across multiple instances of your projected read model. This is where the combination of a Custom Grouper and the Fan Out feature comes into play.
```
oskardudycz, line 459: **COMMENT:** I think that in the below sample, we're not using Fan Out.

oskardudycz (changes requested):
Docs look good to me. Thank you for your contribution!
 
@erdtsieck, could you update the snippets with the guidance I placed in comment?

docs/events/projections/multi-stream-projections.md
erdtsieck: I'm not 100% sure if I did it correct now

erdtsieck: Should i add regions for the small blocks as code in the source as well, instead of copying them to the snippets in the docs and let tooling do the copying for me?
Like I tried in commit: https://github.com/JasperFx/marten/pull/2695/commits/062620a4b58c2f403d90bf0b333b286f235f862c

erdtsieck:
I'm not sure if I should make exact references to snippets and if the snippet refs should be empty

mysticmind:
When we deploy docs, the latest code snippet is always pulled. So once you add the code snippet in test project and the snippet placeholder without the code in docs page. When you run docs locally, the page with the snippet placeholder will get updated with the code snippet. 

In terms of commit, you can commit the changes to the test project which contains the source code snippet with the region marker. With regards to committing of the docs page, you can commit with just the snippet placeholder or you can commit with the snippet (which got added when you ran the docs locally). The latter is better, we can see the PR with it. 

In summary, the code snippet is not added manually in the docs page.

erdtsieck:
> In terms of commit, you can commit the changes to the test project which contains the source code snippet with the region marker. With regards to committing of the docs page, you can commit with just the snippet placeholder (keeping it empty)or you can commit with the snippet (which got added when you ran the docs locally). The latter is better, we can see the PR with it.

Thanks, I understand now. Will work on it!

oskardudycz (approved):
@erdtsieck, I hope that you don't mind that I fixed snippets and regions in https://github.com/JasperFx/marten/pull/2695/commits/4fbfe52adda58b45ddf3a55a5166e90b4f683c32.

In general, you need to add a named region around the code and use it in markdown. When you rebuild the solution the snippet code will be updated.

PR #2725, opened 2023-09-13
Aggregate projections with Create and Apply for the same event type

jeremydmiller:


jeremydmiller:
I think this is obsolete now

PR #2727, opened 2023-09-19
Made AsyncDaemonHealthCheck constructor public to enable proper usage in ASP.NET middleware conventions

daveHylde:
The internal `AsyncDaemonHealthCheck` dependencies (`AsyncDaemonHealthCheckSettings`) need to have public constructors for asp.net to add them to the middleware. The `AsyncDaemonHealthCheckSettings` and the `AsyncDaemonHealthCheck` has been moved to the namespace `Marten.Events.Daemon.Internal` to communicate the internal intent. The method users should use are still in the `Marten.Events.Daemon` namespace.


oskardudycz:
@daveHylde, I'm not against of making settings public, but shouldn't it take the instance that we're registering as singleton?

daveHylde:
@oskardudycz the settings are registered as a singleton service, so I'm not sure what you mean. This is how I register it: 
```cs
builder.Services.AddSingleton(new AsyncDaemonHealthCheckSettings(maxEventLag));
```

daveHylde:
I'm also not sure why the CI pipeline is failing, but I can see that's the case for a lot of other pr's as well.

oskardudycz:
> @oskardudycz the settings are registered as a singleton service, so I'm not sure what you mean. This is how I register it:
> 
> ```cs
> builder.Services.AddSingleton(new AsyncDaemonHealthCheckSettings(maxEventLag));
> ```

My question is that I do not fully understand why we need to expose them as public if we're registering them in the IoC. Could you elaborate more or send some reference that explains it?

daveHylde:
@oskardudycz your comment made me revisit the issue. You're absolutely right, and I realized that the issue was a dumb internal ctor on the HealthCheck. This internal ctor resulted in IoC not being able to initialize the class. 
My initial thought was that the internal classes was the issue, but it was just the constructor.

Sorry for the confusion. I've updated the PR with the fix.

daveHylde:
To sum it up: 
Internal constructor: 
bilde

public constructor you get 'Healthy'. 

Again - sorry for the confusion. And good catch! Contributing here is my first open source contribution - but it's given me a desire for more.

oskardudycz (approved):
Thank you, @daveHylde, looks good now. I also rebased your changes with the main branch. We prefer to use rebases instead of merged to make clearer git history and smoother conflict resolution.

Issue #2730, opened 2023-09-23
Add documentation about integration testing


erdtsieck:

oskardudycz:
Closed by #2732 :+1:

PR #2732, opened 2023-09-23
#2730 Add documentation about integration testing

erdtsieck:


erdtsieck:
I feel like, the failed test is not my fault? 🤔

docs/testing/integration.md
```diff
@@ -0,0 +1,330 @@
...
+<sup><a href='https://github.com/JasperFx/marten/blob/master/src/IssueService/Startup.cs#L32-L47' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_integration_use_scheme_name' title='Start of snippet'>anchor</a></sup>
+<!-- endSnippet -->
+
+::: warning
+Keep note that Marten can be configured to generate static code on startup that contains the scheme name, so it could be beneficial to turn that off for your integration tests:
+:::
+
+```cs
+var martenSettings = configuration.GetSection(MartenSettings.SECTION).Get<MartenSettings>() ?? new MartenSettings();
+if (martenSettings.FromTests)
```
oskardudycz, line 246: **QUESTION:** Why are you only applying database changes in test mode?

Also, my recommendation would be to always use `OptimizeArtifactWorkflow` as without that on CI you'll have a different configuration than in the regular deployment. Optimised Artefact Workflow will use dynamic generation for Development environment, so you should be fine to run tests locally without regenerating static types. That can be done on CI.

My suggestion would be to eliminate the `FromTests` setting and have the test settings the same as you'd have for production.

docs/testing/integration.md
```diff
@@ -0,0 +1,330 @@
...
+updated.Balance.ShouldBe(1300);
+```
+
+Furthermore it could be beneficial to disable all external wolverine transports to test in isolation, by adding this to the Alba host setup:
+
+```cs
+services.DisableAllExternalWolverineTransports();
+```
+
+## Testing Event Projections
```
oskardudycz, line 308: Personally, I'd recommend testing event projections also through the integration testing. The method below can also be a bit dangerous, as you cannot run that in parallel (unless you use a different database or lockid and distinct set of projections).

oskardudycz (commented):
@erdtsieck that's a nice piece of work! I left some comments for discussion.

docs/testing/integration.md
erdtsieck: Yeah, in my own code I use this. It is amazingly stable, but I did not want to put semaphores in the documentation 😆 .

If you point me to the right direction using a lock id, I will try it out and use that. Would you keep all events in one scheme and all projections in another? Or do you split the projections by lockid?

```cs
private readonly SemaphoreSlim _daemonInACage = new(1, 1);
/// <summary>
/// 1. Start generation of projections
/// 2. Wait for projections to be projected
/// </summary>
public async Task GenerateProjectionsAsync<TView>(CancellationToken cancellationToken)
{
    await _daemonInACage.WaitAsync(30.Seconds(), cancellationToken);
    try
    {
        using var daemon = await Store.BuildProjectionDaemonAsync();
        await daemon.StartAllShards();
        await daemon.RebuildProjection<TView>(10.Seconds(), cancellationToken);
    }
    finally
    {
        _daemonInACage.Release();
    }
}
```

docs/testing/integration.md
erdtsieck: The reasoning is to prevent code generation, since https://github.com/JasperFx/marten/discussions/2733 would be a problem if you run twice using two different schemes. That's why I do code generation in memory for integration tests.

If I understand correclty your alternative would be to use `ASPNETCORE_ENVIRONMENT` as `Development`. Which is default I think. I'm using `b.UseEnvironment("integration")` in my integration tests since I do not want appsettings.development.json in my tests.

Since I think most will use "development", I will make it a hint not to change that since code generation will not work with dynamic schemes.

docs/testing/integration.md
erdtsieck: For now I added a TODO, since documentation on most parts of integration testing is better then complete I suppose

erdtsieck:
> @erdtsieck that's a nice piece of work! I left some comments for discussion.

Thank you, I think, with your feedback, it is a good enough first version

oskardudycz (approved):
Thanks, @erdtsieck!

PR #2742, opened 2023-10-02
Add stream key variants of batched event query operations

elexisvenator:
Added missing `IBatchEvents.FetchStreamState(string streamKey)` 
and `IBatchEvents.FetchStream(string streamKey, long version, DateTime? timestamp, long fromVersion)`.

Nothing too complicated here, just filling in some gaps.


oskardudycz (commented):
@elexisvenator looks good to me. Could you add tests tho?

elexisvenator:
Can do :) I have duplicated the tests that exist for the stream ids and changed them to work on the string key versions.

oskardudycz (approved):
Thank you a lot, @elexisvenator; looks good!

PR #2748, opened 2023-10-10
Add IEvent<TEvent> fanout overload on multistream projection

### LINKED ISSUE #2747: Add IEvent<TEvent> fanout overload on multistream projection
The existing Fanout method looks like this:

```cs
public void FanOut<TEvent, TChild>(Func<TEvent, IEnumerable<TChild>> fanOutFunc,
        FanoutMode mode = FanoutMode.AfterGrouping)
```

And it would be beneficial to have an overload like this:

```cs
public void FanOutEvent<TEvent, TChild>(Func<IEvent<TEvent>, IEnumerable<TChild>> fanOutFunc,
    FanoutMode mode = FanoutMode.AfterGrouping) where TEvent : notnull
```

erdtsieck:
Fixes #2747


src/Marten/Events/Aggregation/EventListExtensions.cs
```diff
@@ -6,15 +6,20 @@ namespace Marten.Events.Aggregation;
 
 public static class EventListExtensions
 {
-    public static void FanOut<TSource, TChild>(this List<IEvent> events,
+    public static void FanOutEventData<TSource, TChild>(this List<IEvent> events,
```
jeremydmiller, line 9: Hey, you can't make breaking API changes. Can you move this back to what it was before?

src/Marten/Events/Aggregation/EventSlicer.cs
```diff
@@ -104,11 +104,27 @@ internal IEnumerable<Type> DetermineEventTypes()
     /// <param name="mode">Should the fan out operation happen after grouping, or before? Default is after</param>
     /// <typeparam name="TEvent"></typeparam>
     /// <typeparam name="TChild"></typeparam>
-    public EventSlicer<TDoc, TId> FanOut<TEvent, TChild>(Func<TEvent, IEnumerable<TChild>> fanOutFunc,
-        FanoutMode mode = FanoutMode.AfterGrouping)
+    public EventSlicer<TDoc, TId> FanOut<TEvent, TChild>(Func<TEvent, IEnumerable<TChild>> fanOutFunc, FanoutMode mode = FanoutMode.AfterGrouping)
```
jeremydmiller, line 107: Gotta add an overload rather than just using the optional argument 'cause semantic versioning. Irritating, but them's the rules

jeremydmiller (changes requested):
Just some semantic versioning stuff, the actual content is fine as is

src/Marten/Events/Aggregation/EventListExtensions.cs
oskardudycz: @jeremydmiller, it is not a breaking change but a new method, have a look below; GitHub is showing the diff weirdly.

src/Marten/Events/Aggregation/EventSlicer.cs
oskardudycz: Same as above, GitHub started misbehaving and showed diff wrongly.

oskardudycz:
@erdtsieck, could you rebase instead of merging the main branch? That'd be much appreciated.

erdtsieck:
> @erdtsieck, could you rebase instead of merging the main branch? That'd be much appreciated.

I'll try next time. Never rebased before. Can I still fix it at this point?

oskardudycz:
@erdtsieck, fine, I did that for you this time :wink:

src/Marten/Events/Aggregation/EventListExtensions.cs
oskardudycz: Ok, my bad. This is a breaking change, as the fanout function has changed in the function below. @erdtsieck you need to keep naming the same, the new FanOut need to have a different name.

erdtsieck: I think it's non-breaking now

src/Marten/Events/Aggregation/EventSlicer.cs
erdtsieck: I think it's non-breaking now

src/Marten/Events/Projections/MultiStreamProjection.cs
```diff
@@ -108,6 +108,21 @@ protected override void specialAssertValid()
...
+    /// <summary>
+    ///     Apply "fan out" operations to the given IEvent<TEvent> type that inserts an enumerable of TChild events right behind the
+    ///     parent
+    ///     event in the event stream
+    /// </summary>
+    /// <param name="fanOutFunc"></param>
+    /// <param name="mode">Should the fan out operation happen after grouping, or before? Default is after</param>
+    /// <typeparam name="TEvent"></typeparam>
+    /// <typeparam name="TChild"></typeparam>
+    public void FanOutEvent<TEvent, TChild>(Func<IEvent<TEvent>, IEnumerable<TChild>> fanOutFunc,
```
oskardudycz, line 120: **SUGGESTION:** My preference would be to name it as `FanOut` and just have overload instead of name with the suffix
```suggestion
    public void FanOut<TEvent, TChild>(Func<IEvent<TEvent>, IEnumerable<TChild>> fanOutFunc,
```

erdtsieck: hmm, I do not understand how it should know what overload to pick, but it seems to work now. I thought it did not before at some time during development.

oskardudycz: @erdtsieck .NET compiler is usually smart enough to select the best matching overload. So, as the new one has clear restriction on IEvent, then it should find it based on the criteria. That doesn't work if this would be an extension method, as regular methods have precedence over them.

erdtsieck: @oskardudycz I do understand that it picks the best overload, but since I call it from the example on a type that is not IEvent, I can not imagine how it should know.

```cs
FanOut<Travel, Stop>(x => x.Data.Stops);
```

Maybe because i try to access `.Data`, but that seems far fetched

src/Marten/Events/Aggregation/FanOutEventDataOperator.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using System.Collections.Generic;
+using Marten.Events.Projections;
+
+namespace Marten.Events.Aggregation;
+
+internal abstract class FanOutOperator<TSource>: IFanOutRule
```
oskardudycz, line 7: @erdtsieck, I apologise for such partial reviews, but one last ask: could you rename this file back to `FanOutOperator`? I think this name is more explicit, and it helps to keep track of the file changes history.

erdtsieck: That's no problem. I understand. Resharper can be helpfull, but you have to watch all changes

oskardudycz: Thanks!

jeremydmiller (approved):
Thanks for making the changes and doing this in the first place. Being strict, or at least trying to be, on SemVer is tedious:(

I'm pulling this in now

PR #2750, opened 2023-10-13
Allow customization of the snapshot projection

Xzelsius:
As discussed with @oskardudycz on discord, here is a little change that allows to customize the snapshot projection.
This helps for cases where one wants to use `.Snapshot<T>(..)` but likes to customize the default `SingleStreamProjection<T>`.

### DIFF
--- a/src/Marten/Events/Projections/ProjectionOptions.cs
+++ b/src/Marten/Events/Projections/ProjectionOptions.cs
@@ -175,7 +175,7 @@ public class ProjectionOptions: DaemonSettings
         Action<AsyncOptions> asyncConfiguration = null
     )
     {
-        var expression = singleStreamProjection<T>(ProjectionLifecycle.Live, asyncConfiguration);
+        var expression = singleStreamProjection<T>(ProjectionLifecycle.Live, null, asyncConfiguration);
 
         // Hack to address https://github.com/JasperFx/marten/issues/2610
         _options.Storage.MappingFor(typeof(T)).SkipSchemaGeneration = true;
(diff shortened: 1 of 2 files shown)


src/Marten/Events/Projections/ProjectionOptions.cs
```diff
@@ -190,17 +190,36 @@ internal IProjection[] BuildInlineProjections(DocumentStore store)
...
+    ///     projection within the async projection daemon
     /// </param>
     /// <returns>The extended storage configuration for document T</returns>
     public MartenRegistry.DocumentMappingExpression<T> Snapshot<T>(
         SnapshotLifecycle lifecycle,
+        Action<SingleStreamProjection<T>> configureProjection,
         Action<AsyncOptions> asyncConfiguration = null
     ) =>
-        singleStreamProjection<T>(lifecycle.Map(), asyncConfiguration);
+        singleStreamProjection<T>(lifecycle.Map(), configureProjection, asyncConfiguration);
```
Xzelsius, line 218: @oskardudycz I made an overload instead of adding an optional parameter to the existing method, like you said.

But this way, the overload does not receive the same amount of coverage like the existing one (unless I copy all  the test). 

If you are fine with this I leave it as is.
Otherwise, I would prefer to have it as an optional parameter on the existing method.

jeremydmiller: You have to make it an overload because of semantic version rules. The extra optional parameter would force users to recompile. Ticky tacky rule here, but the overload is necessary.

Xzelsius: So then it's just whether to copy some tests to get similar coverage or leave as is?

oskardudycz: Copying tests sounds fine to me. You could also add an integration test checking snapshot deletion for inline snapshot.

Xzelsius: @jeremydmiller I added a test that is equal to my actual use case. _(and I did not copy all other tests as discussed with you on discord)_

oskardudycz (approved):
Looks good to me :+1:

Xzelsius:
@oskardudycz Do you want me to create a follow-up PR that consolidates the overload for v7 release?

oskardudycz:
@Xzelsius, apologies for the late answer, but I somehow missed your question (GH notification sucks!). I'm fine with keeping this overload; it's useful for people who want to just configure snapshots. Most people configure it as inline, so having the overload with configure callback as the first parameter is useful IMHO.

Issue #2749, opened 2023-10-11
Projections generated code produces compilation errors due to non-deterministic sorting of switch statements


svrx:
After upgrade to version 6 some issue have been identified with compilation of generated code. 
When a projection depends on events from different levels of the same inheritance chain, the sorting of methods performed during code generation has been found to be non-deterministic.

Example:
```csharp
//Events
class BaseEvent {}
class DerivedEventA: BaseEvent {}
class DerivedEventB: BaseEvent {}

//Projection
class Aggregate{
   void Apply(DerivedEventA @event):
   void Apply(DerivedEventB @event):
   void Apply(BaseEvent @event):
}
```

Will sometimes throw a `Unreachable code detected` error during compilation of generated code, due to having `BaseEvent` appearing in switch order before any of the derived events.

svrx:
As discussed in Discord thread, further investigation has shown that root cause of the issue may reside in  `EventTypeComparer` implementation.  
At first glance it looks fine, although the resolution of comparison isn't granular enough, hence when being processed for sorting through .Net internals (depending on API flavor used, may be a QuickSort or Heap+Introspective sort), and optimizations are applied in regards divide and conquer nature of algorithms, having assumptions taken based on partitions and equality (as per result of comparer), which in turn results in some entries never being compared, since all the items in the partition being processed having been reported to be equal by the comparer (comparer returns 0).

In order to prove this theory I've manage to sidestep this issue on my tests by updating the implementation of the `EventTypeComparer` in the following way.
```csharp
internal class EventTypeComparer: IComparer<EventProcessingFrame>
{
    public int Compare(EventProcessingFrame x, EventProcessingFrame y)
    {
        if (x.EventType.CanBeCastTo(y.EventType))
        {
            return -1;
        }

        if (y.EventType.CanBeCastTo(x.EventType))
        {
            return 1;
        }

        return StringComparer.OrdinalIgnoreCase.Compare(x.EventType.Name,y.EventType.Name);
    }
}
 ```
The impact of this change is that when object are not from same family (assignable) it'll fall back to sorting by event type name.
Further testing is though required from my side. In general does this feel like the right direction?

svrx:
Meanwhile I've done additional testing and the root cause is confirmed, although previously proposed strategy still doesn't cover the requirement fully. 
In general `IComparer<T>` based implementation doesn't perform reliably when comparer logic runs based on disjoint comparison criteria. 
[...]

PR #2751, opened 2023-10-13
Projections generated code may fail compilation due to order of switch case statements

svrx:
After upgrade to version 6 some issue have been identified with compilation of generated code. 
When a projection depends on events from different levels of the same inheritance chain, the sorting of methods performed during code generation has been found to be non-deterministic.

Example:
```csharp
//Events
class BaseEvent {}
class DerivedEventA: BaseEvent {}
class DerivedEventB: BaseEvent {}

//Projection
class Aggregate{
   void Apply(DerivedEventA @event):
   void Apply(DerivedEventB @event):
   void Apply(BaseEvent @event):
}
```

Will sometimes throw a `Unreachable code detected` error during compilation of generated code, due to having `BaseEvent` appearing in switch case order before any of the derived events.

Further investigation has shown that root cause of the issue may reside in  `EventTypeComparer` implementation.  
At first glance it looks fine, although the resolution of comparison isn't granular enough, hence when being processed for sorting through .Net internals (depending on API flavour used, may be a QuickSort or Heap+Introspective sort), and optimizations are applied in regards divide and conquer nature of algorithms, having assumptions taken based on partitions and equality (as per result of comparer), which in turn results in some entries never being compared, since all the items in the partition being processed having been reported to be equal by the comparer (comparer returns 0).

Relates to #2749.

### DIFF
--- a/docs/events/projections/flat.md
+++ b/docs/events/projections/flat.md
@@ -163,13 +163,13 @@ A couple notes on this version of the code:
 
 * `FlatFileProjection` is adding columns to its table based on the designated column mappings. 
   You can happily customize the `FlatFileProjection.Table` object to add indexes, constraints, or defaults.
-* Marten is able to apply schema migrations and manage the table from the `FlatFileProjection` as long as it’s registered with Marten
-* When you call `Map(x => x.ActivityType)`, Marten is by default mapping that to a kebab-cased derivation of the member 
+* Marten is able to apply schema migrations and manage the table from the `FlatFileProjection` as long as it’s registered with Marten.
+* When you call `Map(x => x.ActivityType)`, Marten is by default mapping that to a snake_cased derivation of the member 
   name for the column, so “activity_type”. You can explicitly map the column name yourself.
 * The call to `Map(expression)` chains a fluent builder for the table column if you want to further customize the table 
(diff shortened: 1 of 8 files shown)


svrx:
Relates to #2749.

svrx:
The implementation was updated from prior version for simplicity and better encapsulation.
The sort method was encapsulated in `EventTypePatternMatchFrame` has that's where the logic is relevant and changed back to using simpler comparer originally used in combination with a different sort approach to avoids original issues.
All usages were updated accordingly.

jeremydmiller:
I'm working on this one now

jeremydmiller:
@svrx Just for your later reference, the usage of `SortedList` had some unexpected side effects where it would lose several event handlers. I switched the code to just use `List<T>.Sort()`, and that did the trick.

svrx:
@jeremydmiller you're right, apologies for overlooking that fact. In this case this PR is doing nothing. `EventTypePatternMatchFrameOrderTests` are failing the same way they would have before. Are these test run as part of your validation pipelines?
I can revert comparer to my interim implementation that didn't had that issue.
Question before I move with that, are interfaces that events may implement allowed to be used in projections? Or any plans to support something like that?

PR #2759, opened 2023-10-22
Bumped to 6.3.0

oskardudycz:
Updated also dependencies to latest versions.

### DIFF
--- a/src/EventSourcingTests/EventSourcingTests.csproj
+++ b/src/EventSourcingTests/EventSourcingTests.csproj
@@ -6,25 +6,25 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <ProjectReference Include="..\Marten\Marten.csproj" />
-        <ProjectReference Include="..\Marten.Testing.OtherAssembly\Marten.Testing.OtherAssembly.csproj" />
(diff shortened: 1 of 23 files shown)


src/EventSourcingTests/EventSourcingTests.csproj
```diff
@@ -156,9 +156,9 @@
     </ItemGroup>
 
     <ItemGroup>
-      <Reference Include="Microsoft.Extensions.Hosting">
-        <HintPath>..\..\..\..\..\..\usr\local\share\dotnet\shared\Microsoft.AspNetCore.App\6.0.21\Microsoft.Extensions.Hosting.dll</HintPath>
-      </Reference>
+        <Reference Include="Microsoft.Extensions.Hosting">
+            <HintPath>..\..\..\..\..\..\usr\local\share\dotnet\shared\Microsoft.AspNetCore.App\6.0.21\Microsoft.Extensions.Hosting.dll</HintPath>
```
Xzelsius, line 160: Why is this `<HintPath>` needed?

oskardudycz: Good question, but I don't know 🤷‍♂️ @jeremydmiller added it in https://github.com/JasperFx/marten/pull/2722. Of what I see it either looks redundant or like some DI voodoo. Maybe @jeremydmiller can expand on that.

PR #2760, opened 2023-10-22
Fixes of event type sorting for pattern matching, encapsulating logic in `EventTypePatternMatchFrame`

svrx:
After upgrade to version 6 some issue have been identified with compilation of generated code.
When a projection depends on events from different levels of the same inheritance chain, the sorting of methods performed during code generation has been found to be non-deterministic.

See more in #2749.

Current PR provides fixes and tests for such scenario.


svrx:
After discussion with @jeremydmiller, still pending providing an integration test for the scenario,  but submitting here for review of implementation and potential consideration for the upcoming 6.3.0 release.

svrx:
Integration test covering this scenario has been added, please let me know if it meets the expectations.

svrx:
Thanks for merging this one. When can I expect a release with it included?

oskardudycz:
@svrx, I think that it should be available in 6.4.1

PR #2497, opened 2023-02-23
Enabled the possibility of applying projections with different Conjoined Tenancy scopes for projections.

oskardudycz:
## Summary

1. Enabled the possibility of applying projections with different Conjoined Tenancy scopes for projections.
2. Enabled global projection for events with a conjoined tenancy style.
3. Added possibility to specify that document can be single tenanted (`SingleTenanted` option) when the global convention is to have multi-tenanted.

## Scenarios

| SESSION     | SLICE       | STORAGE   | RESULT          |
|-------------|-------------|-----------|-----------------|
| DEFAULT     | DEFAULT     | SINGLE    | THE SAME        |
| DEFAULT     | DEFAULT     | CONJOINED | THE SAME        |
| DEFAULT     | NON-DEFAULT | SINGLE    | THE SAME        |
| DEFAULT     | NON-DEFAULT | CONJOINED | NEW NON-DEFAULT |
| NON-DEFAULT | DEFAULT     | SINGLE    | NEW DEFAULT     |
| NON-DEFAULT | DEFAULT     | CONJOINED | THE SAME        |
| NON-DEFAULT | NON-DEFAULT | SINGLE    | NEW DEFAULT     |
| NON-DEFAULT | NON-DEFAULT | CONJOINED | THE SAME        |

Plus permutations related to:
- `session.SessionOptions.AllowAnyTenant`
- `session.Options.Advanced.DefaultTenantUsageEnabled`
- `session.DocumentStore.Options.Tenancy.IsTenantStoredInCurrentDatabase`

Fixes #2363

_**Note:** Suggested review with whitespace hidden https://github.com/JasperFx/marten/pull/2497/files?diff=split&w=1._

Issue #2363, opened 2022-10-04
DocumentSession.SaveChangesAsync() loads document from wrong tenant when applying events from another tenant to an inline projection


iamjustsam:
This issue is related to the functionality implemented for #2298 

When opening a session in `Tenant A` and then applying events for another tenant, Marten tries to load the projected document from the tenant of the session instead of the tenant the document actually belongs to.

Unit test to point out the issue:
```cs
using System;
using System.Threading.Tasks;
using Marten;
using Marten.Events.Projections;
using Marten.Storage;
using NUnit.Framework;

namespace TestProject1;

public class Tests
{
    private IDocumentStore _store;
    private Guid _aggregateId;

    [SetUp]
    public async Task Setup()
    {
        _aggregateId = Guid.NewGuid();
        _store = DocumentStore.For(options =>
        {
            options.Connection("Server=localhost;Database=multi-tenancy-issue;Port=5432;User Id=postgres;Password=postgres;");

            options.Events.TenancyStyle = TenancyStyle.Conjoined;
            options.Events.AddEventType<CreatedEvent>();
            options.Events.AddEventType<UpdatedEvent>();

            options.Projections.SelfAggregate<MyAggregate>(ProjectionLifecycle.Inline).MultiTenanted();
        });
        
        // Create a new aggregate in the default tenant
        await using (var session = _store.OpenSession())
        {
            var createdEvent = new CreatedEvent("Created");

            session.Events.StartStream<MyAggregate>(_aggregateId, createdEvent);
            await session.SaveChangesAsync();
        }
    }

    [Test]
    public async Task Appending_Events_To_Default_Tenant_From_Other_Tenant_Should_Succeed()
    {
        // Open a session for another tenant
        await using var session = _store.OpenSession("MyTenant");

        // Append new event in the default tenant
        session.ForTenant(Tenancy.DefaultTenantId).Events.Append(_aggregateId, new UpdatedEvent("Updated"));
        await session.SaveChangesAsync(); // SaveChanges loads the projected document from the wrong tenant when applying the new events
        
        // Load the projected document from the default tenant
        var aggregate = await session.ForTenant(Tenancy.DefaultTenantId).LoadAsync<MyAggregate>(_aggregateId);
        
        // Both the Created and Updated fields should be filled in
        Assert.NotNull(aggregate?.Updated, "aggregate.Updated is null"); // Updated is filled in because the UpdatedEvent is applied correctly
[...]

PR #2764, opened 2023-10-31
Fix wrong session-tenacity used in Custom Aggregation Projections

haefele:
This is a follow up to PR #2497 applying the method `UseTenancyBasedOnSliceAndStorage` also when using `Custom Aggregation Projections`, instead of only to `Single Stream Projections`.

This fixes the issue described in https://github.com/JasperFx/marten/issues/2363#issuecomment-1753734588.

There are some additional issues described in this other comment https://github.com/JasperFx/marten/issues/2363#issuecomment-1621043060. 

I think those should be fixed now too. The aggregate document is now always loaded using the correct tenant session (see [Single Stream Aggregations here](https://github.com/JasperFx/marten/blob/974a1cf1960bb13a176847ecd55c5cd1bc723545/src/Marten/Events/Aggregation/AggregationRuntime.cs#L70) and [Custom Aggregation Projections in this PR right now](https://github.com/JasperFx/marten/blob/409d8303cd2e70bf7a4c461dfbcab398dba86dba/src/Marten/Events/Aggregation/CustomProjection.cs#L56)).

Are there any other projections that would need this code-fix?

### DIFF
--- a/src/EventSourcingTests/Projections/MultiTenants/ConjoinedTenancyProjectionsTests.cs
+++ b/src/EventSourcingTests/Projections/MultiTenants/ConjoinedTenancyProjectionsTests.cs
@@ -1,8 +1,11 @@
 ﻿using System;
+using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Marten;
 using Marten.Events.Aggregation;
 using Marten.Events.Projections;
+using Marten.Internal.Sessions;
 using Marten.Storage;
 using Marten.Testing.Harness;
 using Shouldly;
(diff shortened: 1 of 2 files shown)


oskardudycz:
@haefele, thank you a lot for the contribution. Nicely tracked! I missed the custom projection while working on https://github.com/JasperFx/marten/pull/2497.

> Are there any other projections that would need this code-fix?

I think that we should be fine, as the rest of the projections either derive from `AggregationRuntime` or are mapped into while processing.

Would you like me to do a quick release with this fix, or there's no urgency?

PR #2781, opened 2023-11-13
Failing multistream testcase

jannikbryld:
Would expect Identities, many-to-one, to project the view i into the 2nd Apply method. If the multistream event is the first of the stream, then I have no way of creating a userId:notificationId id on the UserNotification.

@oskardudycz Indicated that Identities would automatically set the ID of the view. I might have understood the case differently, so this test case illustrates the problem.


jeremydmiller:

PR #2783, opened 2023-11-16
Support to write json with raw SQL to HttpContext

kwestground:
A typical use case occurs when you wish to write a custom query to summarize information and directly pass the response to the HttpResponse, without the need to map and serialize it through a DTO class.

### DIFF
--- a/src/Marten.AspNetCore/QueryableExtensions.cs
+++ b/src/Marten.AspNetCore/QueryableExtensions.cs
@@ -291,11 +291,12 @@ public static class QueryableExtensions
         string sql,
         HttpContext context,
         string contentType = "application/json",
-        int onFoundStatus = 200
+        int onFoundStatus = 200,
+        params object[] parameters
     )
     {
         var stream = new MemoryStream();
-        _ = await session.StreamJson<int>(stream, context.RequestAborted, sql).ConfigureAwait(false);
+        _ = await session.StreamJson<int>(stream, context.RequestAborted, sql, parameters).ConfigureAwait(false);
 
         context.Response.StatusCode = onFoundStatus;
         context.Response.ContentLength = stream.Length;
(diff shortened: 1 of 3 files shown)


src/Marten.AspNetCore/QueryableExtensions.cs
```diff
@@ -276,4 +276,33 @@ public static class QueryableExtensions
...
+    public static async Task WriteJson(
+        this IQuerySession session,
+        string sql,
+        HttpContext context,
+        string contentType = "application/json",
+        int onFoundStatus = 200,
+        params object[] parameters
+    )
+    {
+        var stream = new MemoryStream();
```
gfoidl, line 298: Not related to your PR:

We should find a way to avoid the temporary use of MemoryStream to avoid the allocations and the copying.

I once prototyped something but (unfortunately) didn't continue on that.
Is there interested to follow along with this?
If so I'd create a issue (and later when I'm not so busy I'll can look into that).

oskardudycz: @gfoidl that's fine by me; if you're open to contribution, then such performance optimisation would be more than welcome.

gfoidl: Sure.

Sometimes I wonder by myself what a query in a repo unveils: https://github.com/JasperFx/marten/discussions/2408
Which I had already forgotten...

Can that discussion be converted to an issue and assigned to me (so that I won't forget it anymore, as AFAIK discussions can't be assigned, etc.).
But note: at the moment I'm very busy, so I'll be able to work on that beginning next year (or maybe over the X-mas holidays...)

oskardudycz: @gfoidl I cannot convert discussion into an issue, but I created issue based on discussion https://github.com/JasperFx/marten/issues/2791 :)

oskardudycz (approved):
Nice one, thank you @kwestground!

PR #2789, opened 2023-11-21
Initial .NET 8 upgrade

oskardudycz:
PR contains the minimum set of .NET 8 related changes:
1. Added .NET 8 targets to projects.
2. Bumped packages to major versions.
3. Reconfigured CI to Run .NET 8 and supported PostgreSQL versions:

| CI              | .NET | Postgres  |        plv8        | Serializer | 
|-----------------|:----:|:---------:|:------------------:|:----------:|
| GitHub Actions  |  6   |   12.8    | :white_check_mark: |    STJ     | 
| GitHub Actions  |  6   | 15-alpine |        :x:         | Newtonsoft | 
| GitHub Actions  |  7   |   12.8    | :white_check_mark: |  JSON.NET  | 
| GitHub Actions  |  7   |  latest   |        :x:         |    STJ     | 
| Azure Pipelines |  8   |   12.8    | :white_check_mark: |  JSON.NET  | 
| Azure Pipelines |  8   |   12.8    | :white_check_mark: |    STJ     | 
| Azure Pipelines |  8   | 15-alpine |        :x:         |    STJ     | 
| Azure Pipelines |  8   |  latest   |        :x:         | Newtonsoft | 

Removed also benchmark pipeline, as it was unused.
4.  Added customisation for docker-compose and tests to allow easier using different PostgreSQL versions by running:
```
POSTGRES_IMAGE=postgres:15.3-alpine docker compose up
```
5. Adjusted code to recent changes in Npgsql:
-  **Fixed Event Metadata handling** - added the specific frame generation for Dictionary<string, object> to use serializer instead of the build-int Npgsql transformation to dynamic
- **Fixed error mapping for DateTime-related exceptions** - the error messages that we're using have changed.
6. Updated documentation adding PostgreSQL version support policy.
7.  Ignored `Can_query_by_hashset_contains test`, as @jeremydmiller handled that in the Linq branch.
8. Fixed ordering tests to have test data not vulnerable for different collation settings.


src/Marten.AspNetCore/Marten.AspNetCore.csproj
```diff
@@ -1,9 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
+        <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
         <Description>Helpers for Marten-backed AspNetCore applications</Description>
-        <VersionPrefix>6.4.0</VersionPrefix>
+        <VersionPrefix>7.0.0-alpha.1</VersionPrefix>
```
gfoidl, line 6: Have you considered using a `Directory.Build.props` where such common repeated properties can be defined once and are then re-used by MsBuild automatically?

Sou would just need to touch ~5 files instead of the 45 right now.

mysticmind: @gfoidl All the primary Marten libraries all use `Directory.Build.props`. Only the plugins related csproj have specific versions set. It is definitely 5 and not 45 :-)

gfoidl: I was just wondering to see the same / similar change to many csproj-files. 
So IMO `Directory.Build.props` could have more common properties, and together with [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management) such repeated changes could be minimized.

So nothing urgent or that like, just to save some work for @oskardudycz 😉.

> can have their own set of changes and related releases

Ah, OK. TIL they aren't versioned all together.

mysticmind: Also note that when we do a major version update, we set a starting version for these plugin/additional projects in alignment with the major version. After this point, as I said, it can go through it own life cycle of changes/releases.

gfoidl: I don't question that (was so in the first comment, but you told me the versioning strategy then).

So my point is*:
* set the target-frameworks in the Directory.Build.props, so an update goes only there and doesn't need to be done in all projects (each project still has to specify the property, but can use the variable defined in the Directory.Build.props)
* use central package management, so updates to packages go to one (central) place and are not spread to all csproj-files

\* this comment-thread is now in the wrong place, as the `VersionSuffix` is a you mention independ for each project

oskardudycz: @gfoidl, thank you for suggestions. I'll check the recent improvements around Central Package Management. The biggest challenge is that we try to minimise the project dependencies, and sometimes, we also need to have different package versions for different .NET versions. I'll check how much we can benefit from it. Good call. 👍

gfoidl: From my experience CPM kind of sucked in pre .NET 8 and had many troubles. Starting with .NET 8 (actually already starting with the previews) it's really good to use and doesn't have any problems.

For me the only point to keep in mind is that `PrivateAssets`, etc. still needs to be set in the individual csproj that use such a package.

oskardudycz: Yes, that instability was stopping me from trying that further, but thanks for letting me know that improved. I'll give it a try in this PR or (probably) a dedicated one :+1:

oskardudycz: @gfoidl I added a dedicated issue for that: https://github.com/JasperFx/marten/issues/2804. I'll try to tackle that in the follow up PR.

Issue #2791, opened 2023-11-22
Pooling of Streams in Marten.AspNetCore.QueryableExtensions


oskardudycz:
### Discussed in https://github.com/JasperFx/marten/discussions/2408

<div type='discussions-op-text'>

<sup>Originally posted by **gfoidl** December  1, 2022</sup>
For each request a [`new MemoryStream`](https://github.com/JasperFx/marten/blob/217cdbc835ad1086375db9a041f13c3f8821c2a6/src/Marten.AspNetCore/QueryableExtensions.cs#L23) is created, thus also the backing `byte[]`-buffer used by MS and if it needs to be resized another buffer with copying over the old contents, and so on.

As response header values can't be set once the body has started, any kind of temporary stream needs to be used -- as it's done now.
But instead of creating a MS every time there should be some kind of pooling, e.g. via [Microsoft.IO.RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream).

Is this in general something to look into, and is it OK to rely on that package?
Otherwise there could be some inbox-solution for pooled streams based on ArrayPool, etc.

I didn't run any benchmarks so far, just saw this when reading the code.

<sub>I can create a PR if needed.</sub>.</div>

oskardudycz:
@gfoidl volunteered https://github.com/JasperFx/marten/pull/2783#discussion_r1400391122 to work on that.

PR #2792, opened 2023-11-22
Avoid buffering the content before streaming to the response body

gfoidl:
Addressed the point outlined in https://github.com/JasperFx/marten/discussions/2408#discussioncomment-4321349 and contributes to https://github.com/JasperFx/marten/issues/2791 (but does not fix that issue[^1])

For array streaming the whole content was buffered in a memory stream, then copied to the ASP.NET Core's response stream.
This has some downsides:
* the content-length isn't needed, ASP.NET Core will produce a chunked output, which can be read by any client
* it needs copying to the memory stream to the response -- which can be resource intensive for large data
* for large data it's possible that the memory stream gets bigger than 85.000 bytes, so it's above the large object heap (LOH) threshold of the GC, thus that object will live until a Gen-2 collection happens

These points can be avoided by writing to the response stream directly.

[^1]: I'll try to address the pooling, etc. point later, as that's not so straight forward.

### DIFF
--- a/src/Marten.AspNetCore/QueryableExtensions.cs
+++ b/src/Marten.AspNetCore/QueryableExtensions.cs
@@ -59,15 +59,10 @@ public static class QueryableExtensions
         int onFoundStatus = 200
     )
     {
-        var stream = new MemoryStream();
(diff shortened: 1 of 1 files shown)


oskardudycz (approved):
Thanks, that's a nice improvement!

Issue #2777, opened 2023-11-12
QueryRawEventDataOnly with StreamIdentity string fails when Id is selected with InvalidCastException


gfoidl:
Repro for Marten version 6.3.0:
```c#
using Marten;
using Marten.Events;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;

HostApplicationBuilder appBuilder = Host.CreateApplicationBuilder(new HostApplicationBuilderSettings
{
    EnvironmentName = "Development"
});

appBuilder.Services.AddMarten(options =>
{
    options.Connection("host=10.0.0.20;username=root;password=root;database=test_db");

    options.Events.StreamIdentity = StreamIdentity.AsString;
})
.UseLightweightSessions();

IHost app = appBuilder.Build();
await app.StartAsync();

using IDocumentStore store = app.Services.GetRequiredService<IDocumentStore>();
await store.Advanced.Clean.CompletelyRemoveAllAsync();

await using (IDocumentSession session = store.LightweightSession())
{
    session.Events.Append("a", new DummyEvent("a"));

    await session.SaveChangesAsync();
}

await using (IQuerySession session = store.QuerySession())
{
    var ids = await session.Events.QueryRawEventDataOnly<DummyEvent>()
        .Select(d => d.Id)          // This causes the operation to fail
        .ToListAsync();
}

await app.StopAsync();

public record DummyEvent(string Id);
```

```
Unhandled exception. System.InvalidCastException: Can't cast database type uuid to String
   at Npgsql.Internal.TypeHandling.NpgsqlTypeHandler.ReadCustom[TAny](NpgsqlReadBuffer buf, Int32 len, Boolean async, FieldDescription fieldDescription)
   at Npgsql.NpgsqlDataReader.GetFieldValue[T](Int32 ordinal)
   at Npgsql.NpgsqlDataReader.GetFieldValueAsync[T](Int32 ordinal, CancellationToken cancellationToken)
   at Marten.Linq.SqlGeneration.ScalarStringSelectClause.ResolveAsync(DbDataReader reader, CancellationToken token)
   at Marten.Linq.QueryHandlers.ListQueryHandler`1.HandleAsync(DbDataReader reader, IMartenSession session, CancellationToken token)
   at Marten.Linq.MartenLinqQueryProvider.ExecuteHandlerAsync[T](IQueryHandler`1 handler, CancellationToken token)
[...]

PR #2819, opened 2023-11-27
Fix for event Linq querying if you happen to have an Event.Id member.…

jeremydmiller:
… Closes GH-2777

### DIFF
--- a/src/EventSourcingTests/Bugs/Bug_2777_querying_with_linq_against_events_and_event_has_id_property.cs
+++ b/src/EventSourcingTests/Bugs/Bug_2777_querying_with_linq_against_events_and_event_has_id_property.cs
@@ -1,0 +1,45 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Marten;
+using Marten.Events;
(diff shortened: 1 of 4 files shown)


src/EventSourcingTests/Bugs/Bug_2777_querying_with_linq_against_events_and_event_has_id_property.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            session.Events.Append("a", new DummyEvent("a"));
+
+            await session.SaveChangesAsync();
+        }
+
+        await using (var session = theStore.QuerySession())
+        {
+            var ids = await session.Events.QueryRawEventDataOnly<DummyEvent>()
+                .Select(d => d.Id)          // This causes the operation to fail
```
gfoidl, line 39:
```suggestion
                .Select(d => d.Id)
```
When it's fixed, then the comment is outdated.

jeremydmiller:
I'm not going to lose any sleep over that comment staying in:-)

PR #2823, opened 2023-11-28
Added support for IsNullOrEmpty and IsNullOrWhiteSpace string operators in Linq queries

### LINKED ISSUE #2815: Support string.IsNullOrEmpty() in Linq Where() clauses
As reported by @Hawxy 

Think it's "just":

* A new `ISqlFragment` for the is null or empty case
* Another `ISqlFragment` for the reverse NOT case so they can reverse to each other
* A new `IMethodCallParser`

oskardudycz:
- Added new `ISqlFragment`   for `IsNullOrEmpty` and ``IsNullOrWhiteSpace`
- Added new `IMethodCallParser` implementations accordingly,
- Added tests and extended current test suite with more pemutations

Fixes #2815

@Hawxy FYI

_Note: I recommend to check it with whitespace hidden: https://github.com/JasperFx/marten/pull/2823/files?diff=split&w=1_

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -261,7 +261,7 @@ public interface IMartenSessionLogger
     public void OnBeforeExecute(NpgsqlCommand command);
 }
 ```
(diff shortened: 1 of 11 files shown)


oskardudycz:
@jeremydmiller, you mentioned that:
> Another ISqlFragment for the reverse NOT case so they can reverse to each other

It seems to work without it and wrapping the SQL parts with `Not()` should I add explicit fragments or is it fine as it is?

PR #2824, opened 2023-11-28
Fixed EventProjection to respect global ordering

### LINKED ISSUE #2657: Event Projection should respect global order in events processing
Based on the [Discord discussion](https://discord.com/channels/1074998995086225460/1074999076896112661/1130518719857164328) with @migajek, we discovered that Event Projection it running everything by the stream, which may cause issues around the global ordering of event processing.

We should consider changing that behaviour as it may be misleading and causing unpredicted issues. We should ensure the processing also for Multi-Stream Projections and custom IProjections.

oskardudycz:
Based on the [Discord discussion](https://discord.com/channels/1074998995086225460/1074999076896112661/1130518719857164328) with @migajek, we discovered that Event Projection it running everything by the stream, which may cause issues around the global ordering of event processing.

This PR is changing that behaviour as it may be misleading and causing unpredicted issues. 

Fixes #2657

### DIFF
--- a/src/EventSourcingTests/Projections/EventProjections/EventProjectionOrderingTests.cs
+++ b/src/EventSourcingTests/Projections/EventProjections/EventProjectionOrderingTests.cs
@@ -1,0 +1,74 @@
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Marten.Events;
+using Marten.Events.Projections;
+using Marten.Testing.Harness;
+using Xunit;
+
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Funny, I was going to take this on this morning just to have something easy to start with and do it the exact same way you did it. We might come back a little later in the process and remove the "group it into streams, then ungroup it" work

oskardudycz:
@jeremydmiller, sorry for that 😅 I also added a test to reproduce the issue.

Issue #2304, opened 2022-07-25
NGramIndex on Inner Property does not work (Runtime Error)


OptimusPi:
@apereiratl here is an issue for what we briefly chatted about. 

Here is a simple Record I have: 

```
    public RentalProperty(Guid id, string name, PropertyDetails details, PropertySecurityInfo securityInfo)
    {
        Id = id;
        Name = name;
        Details = details;
        SecurityInfo = securityInfo;
    }
```

Here is a working NGramIndex, with the not-working index commented out

```
    public void Configure(IServiceProvider services, StoreOptions options)
    {
        // RentalProperty search indexes
        
        // Works great! Useful for whole-word matching.
        options.Schema.For<RentalProperty>().FullTextIndex(); 

        // Top-level string property works great! Useful for partial-match searching
        options.Schema.For<RentalProperty>().NgramIndex(x => x.Name); 
        
        // Inner Property NGramIndex does not work. There is a runtime error.
        // RentalProperty is a record, 
        // RentalProperty.Details is a record, 
        // RentalProperty.Details.MailingAddress is a record, 
        // RentalProperty.Details.MailingAddress.StreetAddress1 is a string.
        // BROKEN:
        // options.Schema.For<RentalProperty>().NgramIndex(x => x.Details.MailingAddress.StreetAddress1);
    }
```

Luckily, the NGramSearch still works fine. I just worry about not having that one field indexed properly, and speed concerns when we get more data in our database. 
```
  await using var session = await GetSession(cancellationToken);
  var rentalProperties = await session.Query<RentalProperty>()
      .Where(x => x.Name.NgramSearch(searchText) 
                  || x.Details.MailingAddress.StreetAddress1.NgramSearch(searchText)  // TODO - make sure there is a way to index this? 
                  || x.PlainTextSearch(searchText))
      .ToListAsync(cancellationToken);

  return rentalProperties;
```

So on to the error, when trying to NGramIndex on an inner property, this occurs:

```
[10:18:50 ERR] An unhandled exception has occurred while executing the request. <s:Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware>
Weasel.Postgresql.PostgresqlIdentifierInvalidException: Database identifier mt_doc_rentalproperty_idx_ngram_details as jsonb) ||   || cast(data ->> mailingaddress as jsonb) ||   || data ->> streetaddres
[...]

PR #2839, opened 2023-12-03
Fix NGram index on nested property throwing an exception

mysticmind:
- Fix NGram index logic for data config generation and indexname generation
- Add unit test for ngram index on nested property

Fixes GH-2304


oskardudycz (approved):
Looks good to me. :+1:

PR #2845, opened 2023-12-05
Allowing users to add LINQ order by operators with literal SQL. Close…

jeremydmiller:
…s GH-2843


docs/configuration/cli.md
gfoidl, line 1: @jeremydmiller I don't understand all the mechanics on how the docs are deployed, but to me it seems that this should be live-content and be not removed?

mysticmind: Docs always fetch the latest samples when deployed. So the requirement is to have the sample marker tags intact.

For PR reviews, it would be better to commit the docs changes as well so that reviewers can see. 

In this case, what Jeremy did is okay since he knows what would happen.

gfoidl: @mysticmind I'm aware that Jeremy knows what he does, but I'd like to learn and understand the mechanics that are going on.

My question is because [cli.md](https://github.com/JasperFx/marten/blob/master/docs/configuration/cli.md) is now empty, and when I hit "Suggest changes to this page" in [Command Line Tooling](https://martendb.io/configuration/cli.html) then I get the empty content too.

Please clarify where the content is coming from. Thanks in advance.

mysticmind: If you notice, I had checked earlier on phone and then deleted my message after taking a look from laptop. Looks like Jeremy did delete doc content which are not code samples.

mysticmind: As a follow on note, code samples are fetched afresh from the unit tests files and rest of the docs content are as is in the markdown files. If cli.md is empty then Jeremy did mess up that page.

gfoidl: Ah, got you. Thanks.
I replyd to the email notification I got, so didn't realize you deleted the message.

Issue #2762, opened 2023-10-26
Include documentation on DevOps including database migrations


erdtsieck:

mysticmind:
Moving this to discussion.

PR #2853, opened 2023-12-10
#2762 Include documentation on DevOps including database migrations

erdtsieck:
Initial documentation for Dev-Ops around Marten. One way to achieve it, but can be improved on later. Given as starting point


oskardudycz (approved):
Great stuff, @erdtsieck, thank you!

I think that the next will be adding a sample project and sample workflow in the repo. But that's for the other time! :)

PR #2855, opened 2023-12-12
Add support for NpgsqlDataSource

oskardudycz:
Added support for `NpgsqlDataSource`, the recommended way of handling connection management by the Npgsql team. This should also enable more scenarios, like integrating with the Aspire project, easier isolation testing etc.

Used `NpgsqlDataSource` throughout the codebase. It should be safe, as NpgsqlDataSource inferits currently the global config upon initialisation.

Added possibility to use registered `NpgsqlDataSource` in DI by calling `AddMarten().UseNpgsqlDataSource()` helper methods.

By default, `DocumentStore` will manage the lifetime of all data sources created by the factory as singletons. When injected from DI, only newly created tenants will be managed; the DI container will manage the main one.

Made `ITenancyConfiguration` in-store options lazy, as it's needed only in Runtime, and that enables easier and more flexible configuration.

I kept some Store Options as internal, but we can consider making them public depending on the usage patterns.

Currently left the wild permutations of the connection configuration but made them obsolete.

**Warning:** Bumped the Weasel version, which can cause some binding conflicts in Wolverine.


jeremydmiller:
@oskardudycz I say we pull this in anytime you'd like. We're not finalizing anything anyway, and we'll take the run through Wolverine soon too.

Issue #1641, opened 2020-12-04
Eliminate mt_append_event function


jeremydmiller:
Not just delete it from the code, but also do something in the migrations that detects that it exists and tries to delete it.

jeremydmiller:
This is done.

PR #2868, opened 2023-12-19
Update storage.md to remove `mt_append_event`

gfoidl:
As I see from https://github.com/JasperFx/marten/issues/1641 then `mt_append_event` is gone, so it should go from the docs too.

---

Just for my understanding: that function got removed to have correct handling of concurrent appends?


oskardudycz:
@gfoidl Thank you, it was removed, as we have more options for customisation our storage, so: 
- string or guide key,
- metadata or not, 
- tenancy, etc.

As we're doing code generation, keeping it flexible by performing regular SQL with different constraints was easier. Right @jeremydmiller?

PR #2871, opened 2023-12-20
Added Aspire config

oskardudycz:
....But I'm not touching that anymore as I wasn't able to install Aspire locally, plus reading https://anthonysimmon.com/exploring-microsoft-developer-control-plane-core-dotnet-aspire-dotnet-8, it's clear that Aspire in the current form is just demoware, and not something that I'd consider doing on prod.

Read more: https://event-driven.io/en/nay_to_aspire/


jeremydmiller:
Just didn't happen

Issue #2430, opened 2022-12-12
Primary key indexes in multitenant dbs are (probably) ordered incorrectly


elexisvenator:
Currently when multi tenancy is enabled for a database, indexes are created with the columns arranged with id first, tenant_id next, anything else after. As an example see the sql below:

```sql
CREATE UNIQUE INDEX pkey_mt_streams_id_tenant_id ON my_schema.mt_streams (id, tenant_id);
CREATE UNIQUE INDEX pk_mt_events_stream_and_version ON my_schema.mt_events (stream_id, tenant_id, version);

CREATE UNIQUE INDEX pkey_mt_doc_deadletterevent_id_tenant_id ON my_schema.mt_doc_deadletterevent (id, tenant_id);

-- single stream and multi stream aggregates
CREATE UNIQUE INDEX pkey_mt_doc_customprojection_id_tenant_id ON my_schema.mt_doc_customprojection (id, tenant_id);
```

This works for general event usage, and if this is the only use case then it is probably good - putting the high cardinality column first is (sometimes) ever so slightly more performant.  Where this causes problems is running queries to manage the tenant.  Tenant-specific queries range from cross-stream queries (statistics, analytics), BAU tasks (redactions, data cleanup, restreaming) and the really big one which is offboarding tenants.  Without these indexes starting by tenant the only options are to either do slow table scans or creating a duplicate of these indexes on every table in marten.

This is already an issue I have encountered using marten and for now the volumes of data we have is ok to table scan, but it very quickly wont be.

Maybe this could be a 6.0 breaking change? Worth discussing if its something to be concerned about or if its a matter of making the additional index creation documented (which i can help with).

Another thought is if support for partitioned tables will one day be added, the indexes would need to be flipped for that anyway.

jeremydmiller:
Let's knock this out in 7, because, yes, it's a breaking change. Does make sense though.

jeremydmiller:
So a few thoughts on this finally:

* I definitely don't think you want to reverse the order for normal usage because of how often the app in normal issue would be querying against the stream id
* Even if and when we add sharding, you won't be able to shard by tenant because of the high water detection. I think any sharding will have to happen by archived / not archived and maybe by sequence number, so sharding likely won't help for tenanting
* I think that a database per tenant model might be advantageous for these once in awhile needs
* I'd be more inclined to add additional indexes on just tenant id

elexisvenator:
Saw this yesterday and agree with your arguments.

[...]

PR #2886, opened 2024-01-03
Re-ordering indexes on event store tables to put tenant_id first. Clo…

jeremydmiller:
…ses GH-2430


oskardudycz (commented):
I'm not against such a change, but it'd be good to test query plans to see if this change indeed has a positive impact on the queries.

PR #2890, opened 2024-01-07
fix: Make MultiStreamProjection usable for AggregateTo<>

### LINKED ISSUE #2889: MultiStreamProjection is not supported with QueryAllRawEvents().AggregateTo<>()
As the title says, MultiStreamProjections are not supported, only SingleStreamProjection.

Blackclaws:
Adds support for MultiStreamProjection to AggregateTo<>

Open issue still to solve is that the Id of the resulting document is not set.

Closes #2889

### DIFF
--- a/src/EventSourcingTests/Bugs/Bug_2889_multi_stream_aggregate_to.cs
+++ b/src/EventSourcingTests/Bugs/Bug_2889_multi_stream_aggregate_to.cs
@@ -1,0 +1,87 @@
+using System.Linq;
+using System.Threading.Tasks;
+using EventSourcingTests.Projections.MultiStreamProjections.CustomGroupers;
+using Marten;
+using Marten.Events;
+using Marten.Events.Archiving;
+using Marten.Events.Projections;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+using Xunit.Abstractions;
+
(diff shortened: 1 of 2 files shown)


Blackclaws:
So see also my comments on discord. But essentially, the way QueryAllRawEvents.ToAggregate works is not really conducive to MultiDocument outputs as well as setting any sort of Id on the result. The reason being that there is no grouping performed on the events and there is only a single expected output, while a MultiStreamProjection might have two or more outputs (if using Identities) and also QueryAllRawEvents.ToAggregate does try to use all events that are in the collection.

jeremydmiller:
I think this is fine for what it is. It's not a usage that I'd ever be recommending, it's more of just a "if there's nothing else you can do, try this".

I don' think the grouping should have any effect on AggregateTo() in the first place.

Blackclaws:
> I think this is fine for what it is. It's not a usage that I'd ever be recommending, it's more of just a "if there's nothing else you can do, try this".
> 
> I don' think the grouping should have any effect on AggregateTo() in the first place.

You're probably right as its also consistent with how AggregateTo works for SingleStreamAggregations. I'll remove the WIP.

jeremydmiller:
This got addressed a little differently

Issue #2902, opened 2024-01-13
Add a testing helper package


erdtsieck:
For testing projections or just store operations.

I will build the initial package myself

jeremydmiller:
And automatically moving this to a discussion please.

PR #2903, opened 2024-01-13
#2902 Add a testing helper package

erdtsieck:


erdtsieck:
@jeremydmiller or @oskardudycz would be nice if one of you would look at this PR. Since I renamed theSession => TheSession and theStore => TheStore I will get merge conflicts every day 👎

src/Marten.TestHelpers/DaemonContextHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Threading.Tasks;
+using JasperFx.Core;
+using Marten.Events.Daemon;
+using Marten.Events.Daemon.HighWater;
+using Marten.Events.Projections;
+using Microsoft.Extensions.Logging;
+
+namespace Marten.TestHelpers;
+
+public class DaemonContextHelper(string schemaName, string connectionString, ILogger<IProjection> logger) : OneOffConfigurationsHelper(schemaName, connectionString)
```
oskardudycz, line 11: **SUGGESTION:** I don't like the `Helper` suffix. I would either name it `DaemonContext` or `ProjectionDaemonRunner`.

src/Marten.TestHelpers/Marten.TestHelpers.csproj
```diff
@@ -0,0 +1,33 @@
+<Project Sdk="Microsoft.NET.Sdk">
+     <PropertyGroup>
+        <Description>Helpers for writing tests for Marten</Description>
```
oskardudycz, line 3:
```suggestion
        <Description>Marten testing extensions to streamline your tests setup</Description>
```

src/Marten.TestHelpers/Marten.TestHelpers.csproj
```diff
@@ -0,0 +1,33 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
oskardudycz, line 1: **SUGGESTION:** I'd just call this NuGet `Marten.Testing`

oskardudycz (changes requested):
@erdtsieck looks great to me; just a naming fix, and I'm happy to pull it in!

src/Marten.TestHelpers/DaemonContextHelper.cs
erdtsieck: Will do

src/Marten.TestHelpers/Marten.TestHelpers.csproj
erdtsieck: Will do

oskardudycz (approved):
Looks good to me 👍

jeremydmiller:
@oskardudycz @erdtsieck Let's get this in fast if we're going to do it. The `theSession` / `TheSession` rename is going to be a nightmare for outstanding work as is. Or can we put the names back the way they were?

And hey, I think I've got to be the bad guy and ask if this actually provides any value. The async daemon helpers might be helpful, but I think I'd vote to leave all the other "this, that, or the other"Context classes alone and put them back where they were. I think those things are too specific to the Marten codebase to be usable for others. And the rename is going to cause harm.

How about we just focus on the async projection testing?

erdtsieck:
@jeremydmiller I think it is definitely beneficial to have a documented way to test your async projections and my solution to that is expose some stuff that is proven in the Marten codebase.

I can rename the session and store properties but that will conflict with property naming conventions so I will have to find a solution to that. Maybe make the store not a lazy evaluated thing or so. 

Before I do these thing, maybe a preferred design would be nice to discuss? The reasoning behind the current design is mostly to interface event projection testing stuff without a reference to xunit

oskardudycz:
If those helpers are useful, then I'm open to exposing them. As I mentioned on Discord, I wouldn't even mind to keep the base tests classes as we have with XUnit dependency. If we keep the package below version 1. We could see what's the reaction and then either put more there or make it obsolete at some point if we find a better solution.

I agree that the biggest value seems to be around setting up async tests and having basic setup and teardown. Sounds like many people are struggling around it when starting.

erdtsieck:
> If those helpers are useful, then I'm open to exposing them. As I mentioned on Discord, I wouldn't even mind to keep the base tests classes as we have with XUnit dependency. If we keep the package below version 1. We could see what's the reaction and then either put more there or make it obsolete at some point if we find a better solution.
> 
> I agree that the biggest value seems to be around setting up async tests and having basic setup and teardown. Sounds like many people are struggling around it when starting.

I will try to find a less invasive solution to that core problem in this or a new PR. @jeremydmiller you can go ahead and merge stuff, I will at least not do the big renaming thing ;-)

PR #2918, opened 2024-01-19
Overhaul of the execution pipeline

jeremydmiller:
* New "auto-closing connection lifetime" by default
* Async daemon is using NpgsqlBatch
* Ripped out IRetryPolicy
* Getting the operation processing ready for effective retries


docs/configuration/retries.md
```diff
@@ -1,117 +1,3 @@
...
-// Use DefaultRetryPolicy which handles Postgres's transient errors by default with sane defaults
-// We retry operations twice if they yield and NpgsqlException that is transient
-// Each error will cause sleep of N seconds where N is the current retry number
-c.RetryPolicy(DefaultRetryPolicy.Twice());
-```
-<sup><a href='https://github.com/JasperFx/marten/blob/master/src/Marten.Testing/Examples/RetryPolicyTests.cs#L106-L111' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_retrypolicy-samplepolicy-default' title='Start of snippet'>anchor</a></sup>
-<!-- endSnippet -->
-
-Also you could use the fantastic [Polly](https://www.nuget.org/packages/polly) library to easily build more resilient and expressive retry policies by implementing `IRetryPolicy`.
+Marten's previous, homegrown `IRetryPolicy` mechanism was completely replaced by  [Polly](https://www.nuget.org/packages/polly) in Marten V7.
```
oskardudycz, line 3: **SUGGESTION:** It'd be good to provide some samples on how to plug Polly policies to Marten.

src/Marten/StoreOptions.cs
```diff
@@ -81,61 +87,112 @@ public partial class StoreOptions: IReadOnlyStoreOptions, IMigrationLogger
...
+        ResiliencePipeline = builder.Build();
     }
 
-    internal IList<Type> CompiledQueryTypes { get; } = new List<Type>();
+    internal IList<Type> CompiledQueryTypes => _compiledQueryTypes;
+
+    /// <summary>
+    /// Polly policies for retries within Marten command execution
+    /// </summary>
+    public ResiliencePipeline ResiliencePipeline
```
oskardudycz, line 129: **QUESTION:** Shouldn't we have a single entrypoint for setting up the Resilience Policy? Isn't `ConfigurePolly`  enough?

src/Marten/StoreOptions.cs
```diff
@@ -267,11 +319,34 @@ public ITenancy Tenancy
     }
 
     private Lazy<ITenancy>? _tenancy;
+    private Func<string, NpgsqlDataSourceBuilder> _npgsqlDataSourceBuilderFactory = DefaultNpgsqlDataSourceBuilderFactory;
+    private INpgsqlDataSourceFactory _npgsqlDataSourceFactory = new DefaultNpgsqlDataSourceFactory();
+    private readonly IList<Type> _compiledQueryTypes = new List<Type>();
+    private ResiliencePipeline _resiliencePipeline;
+    private int _applyChangesLockId = 4004;
```
oskardudycz, line 326: **QUESTION:** Is there a specific reason for changing properties with defaults into backing one? I preferred the previous syntax, unless that's needed.

src/CoreTests/DefaultRetryPolicyTests.cs
```diff
@@ -1,134 +0,0 @@
...
-using Marten;
-using Marten.Testing.Harness;
-using Npgsql;
-using Shouldly;
-using Weasel.Postgresql;
-using Xunit;
-
-namespace CoreTests;
-
-public class DefaultRetryPolicyTests
```
oskardudycz, line 13: **SUGGESTION:** I think that we should still have tests ensuring that the Polly default policy is applied as expected.

oskardudycz:
The PR scope and type of changes is big, so pardon me if I didn't get it right. I assume that the only way to set a reusing connection now is to either:
- have it managed by Marten and use the connection from the document session in external calls,
- have it injected through SessionOptions with the transaction.

Is it correct?

I'd love to get rid of those injections with the Connection and lazy-loaded stuff, but I'm not sure if that won't be too big a change for some of the users and would make them stay on the older versions. On the other hand, that's more of a guess for now on how many people are using such a way.

We still need to do benchmarks if there's no performance degradation, as the opening connection can be quite costly, especially around the non-pooled usage. 

I think those changes are in the right direction, but it has to be precisely documented, as the auto-closing introduces more _magic_ and an implicit understanding of the stuff. My main fright is that this can be pretty nasty for the existing users, as such bugs may happen only in specific usage (so be seen as non-deterministic). Maybe we could introduce some "v6-compatibility mode" that'd set it up as close as possible to previous behaviour?

src/EventSourcingTests/append_events_with_optimistic_or_exclusive_locks.cs
```diff
@@ -52,13 +53,15 @@ public async Task append_optimistic_happy_path()
...
+
+            session1.Events.StartStream(streamId, new AEvent(), new BEvent());
+            await session1.SaveChangesAsync();
 
             // Fetch the expected version
-            await theSession.Events.AppendOptimistic(streamId, new CEvent(), new BEvent());
+            await session1.Events.AppendOptimistic(streamId, new CEvent(), new BEvent());
 
-            await using (var session = theStore.LightweightSession())
+            await using (var session = theStore.LightweightSession(new SessionOptions{Timeout = 1}))
```
oskardudycz, line 64: **QUESTION:** Out of curiosity, why is setting timeout needed here?

src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
```diff
@@ -119,7 +120,21 @@ private async Task<long> findCurrentMark(HighWaterStatistics statistics, Cancell
...
         _gapDetector.Start = statistics.SafeStartMark;
-        var current = await _runner.Query(_gapDetector, token).ConfigureAwait(false);
+        long? current;
+        try
+        {
+            current = await _runner.Query(_gapDetector, token).ConfigureAwait(false);
+        }
+        catch (InvalidOperationException e)
+        {
+            if (e.Message.Contains("An open data reader exists for this command"))
```
oskardudycz, line 130: **QUESTION:** Shouldn't that be a part of retry policy?

src/Marten/Internal/Sessions/AmbientTransactionLifetime.cs
```diff
@@ -26,20 +34,31 @@ public AmbientTransactionLifetime(SessionOptions options)
         {
             OwnsConnection = true;
         }
+
+        CommandTimeout = _options.Timeout ?? _connection?.CommandTimeout ?? 30;
```
oskardudycz, line 38: **SUGGESTION:** It might be worth putting `30`  value as some general const.

src/Marten/Internal/Sessions/AmbientTransactionLifetime.cs
```diff
@@ -49,90 +68,253 @@ public void Dispose()
...
+            _connection?.Dispose();
         }
     }
 
     public void Apply(NpgsqlCommand command)
     {
         BeginTransaction();
-        command.Connection = Connection;
+        command.CommandTimeout = CommandTimeout;
+        command.Connection = _connection;
```
oskardudycz, line 80: **QUESTION:** Shouldn't we use `Connection` here or ensure that it's not null?

src/Marten/Internal/Sessions/AmbientTransactionLifetime.cs
```diff
@@ -49,90 +68,253 @@ public void Dispose()
...
         command.CommandTimeout = CommandTimeout;
     }
 
     public async Task ApplyAsync(NpgsqlCommand command, CancellationToken token)
     {
         await BeginTransactionAsync(token).ConfigureAwait(false);
-        command.Connection = Connection;
+        command.Connection = _connection;
```
oskardudycz, line 87: **QUESTION:** Shouldn't we use `Connection` here or ensure that it's not null?

docs/configuration/retries.md
jeremydmiller: Yeah, I agree. Like I told y'all in Discord, I hadn't made any effort to document anything yet.

src/Marten/StoreOptions.cs
jeremydmiller: I don't have a strong opinion here. I thought some folks might want to push it in themselves. My guess is that very few people ever take advantage of this feature

src/Marten/StoreOptions.cs
jeremydmiller: Just to do the null check, no other reason

jeremydmiller:
@oskardudycz:

"I assume that the only way to set a reusing connection now is to either:" -- not quite. If you call `IDocumentSession.BeginTransaction()` or use the `IDocumentSession.Connection` in a raw way, it reverts to the old "sticky" connection handling. 


"I'd love to get rid of those injections with the Connection and lazy-loaded stuff, but I'm not sure if that won't be too big a change for some of the users and would make them stay on the older versions. On the other hand, that's more of a guess for now on how many people are using such a way."

- Ditto from me. I'd say to leave those alone and maybe encourage folks to move away from that over time in docs. One thing I wimped out on was changing the ambient transaction behavior. If you opt into ambient "EnlistInTransaction" behavior, it's still the old sticky connection code. And you do have to keep that connection pushing behavior to enable folks who want to use EF Core or Dapper and Marten in the same native db transaction. 

"We still need to do benchmarks if there's no performance degradation, as the opening connection can be quite costly, especially around the non-pooled usage."

Of course. I was working last night on battery power (and a cold). I'll run the benchmarks today and report back. I'm not buying that there's going to be any "performance degradation". The connection pooling is all underneath Npgsql itself, and I don't think this change will have any significant impact on performance due to connection pooling. Besides, even under the old model we didn't open the connection until we had to. Under heavy load, I'd expect it to be the opposite as Npgsql won't have to use so many connections. And also, look much closer at the change in behavior around Polly w/ the static lambdas vs how we were using IRetryPolicy before. What we were doing with IRetryPolicy was pretty gross in terms of object allocations. I would expect the new model to be faster.

"I think those changes are in the right direction, but it has to be precisely documented, as the auto-closing introduces more magic and an implicit understanding of the stuff. My main fright is that this can be pretty nasty for the existing users, as such bugs may happen only in specific usage (so be seen as non-deterministic). Maybe we could introduce some "v6-compatibility mode" that'd set it up as close as possible to previous behaviour?"

I'd thought about the v6-compatible switch. That's relatively easy to do, but I'd do that w/ a follow up PR. I think that's a good idea. The old connection lifetime mechanics are essentially still there, it's a state pattern switch here

src/Marten/Internal/Sessions/AutoClosingLifetime.cs
```diff
@@ -0,0 +1,393 @@
...
+    public void Dispose()
+    {
+
+    }
+
+
+    public int Execute(NpgsqlCommand cmd)
+    {
+        Logger.OnBeforeExecute(cmd);
+        using var conn = _database.CreateConnection();
```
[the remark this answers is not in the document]
jeremydmiller: I was going to agree with you, but changed my mind after thinking about it more. I'd recommend folks put the circuit breaker higher up in the stack. If folks were using Wolverine for example, I'd say you put the circuit breaker on the message processing level and do that with Wolverine itself.

Maybe an example in the docs though so you can opt into that yourself w/ some copy/paste code.

src/Marten/Internal/Sessions/AmbientTransactionLifetime.cs
jeremydmiller: The call to `BeginTransaction()` already makes sure the connection is open

src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
jeremydmiller: I think this is pretty specific to this code, and if we put it in the retry policy, then someone customized the resiliency pipeline, you'd lose the specific behavior. Putting Polly here might not be a bad idea, but I wouldn't make that a high priority

src/EventSourcingTests/append_events_with_optimistic_or_exclusive_locks.cs
jeremydmiller: The test hangs otherwise

jeremydmiller:
@oskardudycz *Another* problem for a day soon I guess, the `NpgsqlDataSource` mechanics do not work with Marten whatsoever. There's some differences in connection behavior with using the data source. You can't hide `NpgsqlDataSource` behind an `IConnectionFactory` like we do.

So we're still dead in the water w/ Project Aspire and using an underlying NpgsqlDataSource. I think maybe we could beat that after the IConnectionLifetime changes in this PR, but previously, that prevents Marten from releasing connections somehow and kaboom.

src/Marten/Internal/Sessions/AmbientTransactionLifetime.cs
jeremydmiller: Yes, and I also think we should move that default way down to maybe 5. 30 seems crazy excessive, but it's what we had before.

src/Marten/StoreOptions.cs
jeremydmiller: Changed.

Issue #1834, opened 2021-05-28
Optimized Patch operations *after* V4


jeremydmiller:
Putting this here for a post-4.0 world.

* Try to support as much as possible of the current Patch operations with native Postgresql JSON operators
* Multiple operations per command?

mysticmind:
Few old open issues related to Patch API, linking it here
https://github.com/JasperFx/marten/issues/699

PR #2923, opened 2024-01-22
Patch implementation using pure PL/pgSQL

mysticmind:
The following are changes included:
- Patch implementation using PL/pgSQL plus unit tests, fixes GH-1834
- Update build to include patching tests
- Set internals visible for PatchingTests project.

This implementation will work with Postgres 9.6 and above.


jeremydmiller:
It looks like a functional equivalent to the old Javascript version. Whenever we can beat Weasel into correctly determining the function body changes correctly, I'm ready to pull this in.

mysticmind:
> It looks like a functional equivalent to the old Javascript version. Whenever we can beat Weasel into correctly determining the function body changes correctly, I'm ready to pull this in.

I have retrofitted the functions as per the format where the delta is all fine with Weasel. That is what the unit tests are all green. So you are good to pull it in.

PR #2925, opened 2024-01-24
fix: Use count(*) instead of count(1) for performance reasons

Blackclaws:
When retrieving query statistics the use of count(*) instead of count(1) is faster for Postgres.

https://www.postgresql.org/message-id/flat/df354b76-a645-73fc-9645-4508e5291d71%40gmx.net


mysticmind (approved):
Looks good to me.

Issue #2943, opened 2024-02-03
Code generation for aggregates sometimes fails


erdtsieck:
This happens 99% of the time in my codebase on Marten 7.0 beta 5 when I use Wolverine on multiple types of aggregates in my integration test since these cause parallel code generation in Marten for aggregates.

You can reproduce it like this in Marten itself:

```cs
options.Projections.LiveStreamAggregation<Something>();
var store = new DocumentStore(options);

Parallel.For(1, 100, _ =>
{
    Parallel.ForEach(store.Events.As<ICodeFileCollection>().BuildFiles().OfType<IProjectionSource>(), projection =>
    {
        projection.Build(store);
    });
});
```

```
System.ArgumentNullException
Value cannot be null. (Parameter 'type')
   at System.Activator.CreateInstance(Type type, BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes)
   at System.Activator.CreateInstance(Type type, Object[] args)
   at Marten.Events.Aggregation.GeneratedAggregateProjectionBase`1.BuildRuntime(DocumentStore store) in C:\projecten\marten-fork\src\Marten\Events\Aggregation\GeneratedAggregateProjectionBase.CodeGen.cs:line 177
   at Marten.Events.Aggregation.GeneratedAggregateProjectionBase`1.buildProjectionObject(DocumentStore store) in C:\projecten\marten-fork\src\Marten\Events\Aggregation\GeneratedAggregateProjectionBase.CodeGen.cs:line 139
   at Marten.Events.Projections.GeneratedProjection.Marten.Events.Projections.IProjectionSource.Build(DocumentStore store) in C:\projecten\marten-fork\src\Marten\Events\Projections\GeneratedProjection.cs:line 65
   at EventSourcingTests.Projections.CodeGeneration.ProjectionCodeGenerationTests.<>c__DisplayClass1_0.<LiveStreamAggregation_GeneratesCodeFiles>b__1(IProjectionSource projection) in C:\projecten\marten-fork\src\EventSourcingTests\Projections\CodeGeneration\ProjectionCodeGenerationTests.cs:line 61
   at System.Threading.Tasks.Parallel.<>c__DisplayClass44_0`2.<PartitionerForEachWorker>b__1(IEnumerator& partitionState, Int32 timeout, Boolean& replicationDelegateYieldedBeforeCompletion)
--- End of stack trace from previous location ---
[...]

PR #2944, opened 2024-02-03
#2943 Do not fail aggregate code generation on parallel execution

erdtsieck:


erdtsieck:
@jeremydmiller I think it failed on a flaky test. So, maybe retry the build?

Issue #2952, opened 2024-02-08
New IProjectionCoordinator service to allow users to start/stop/pause projections programmatically


jeremydmiller:
@oskardudycz @Hawxy  ^^^

Oskar & I were discussing this a bit this week. There will be a new service at the singleton scope in the container that will give you access to alter the execution of different projections.

First though, let's discuss the possible use cases?

* Tell Marten to start rebuilding a projection, i.e., pause it, reset the progress to 0, tear down what's there, go
* Tell Marten to stop a particular projection
* What else?

jeremydmiller:
It's already possible. Anything fancier can go into Critter Stack Pro

Issue #1715, opened 2021-01-23
Async Projection Sharding


jeremydmiller:
Even in the prior async daemon, we executed each individual projection in its own independent track. In V4, let's try to parallelize projection processing within a single projection wherever possible. The V4 async daemon code is already somewhat built around this idea, but there's only one shard per projection so far.

See the `AsyncProjectionShards()` method below. Today these methods always return a single shard.

```
    internal interface IProjectionSource
    {
        string ProjectionName { get; }

        IProjection Build(DocumentStore store);

        IReadOnlyList<IAsyncProjectionShard> AsyncProjectionShards(IDocumentStore store, ITenancy tenancy);
    }
```

Some ideas:

* The obvious way is to use the tenant id in multi-tenanted situations. Maybe you say there's one shard for a certain tenant or range of tenants, and other shards for different tenants. That would allow us to spin up multiple `ProjectionAgent` for different logical shards of the same projection
* For aggregations by the stream, maybe create a synthetic way based on the stream id to assign the entire shard to one shard or another. Could do that randomly when a stream is created, could do it based on the stream id. In this case, I'd vote for us to record the new stream "shard name" in a new column in the streams table to make it easy to query events for one shard at a time.
* Some kind of after the fact tagging/sharding strategy? Maybe utilize the projection dependency (#1714) to first assign streams or events to a shard, then secondly to build the actual projection.

PR #2954, opened 2024-02-09
Async Daemon Overhaul for V7 -- and also a new Dynamic Tenant Database recipe

jeremydmiller:
It's big, we can talk a lot more about this one. 

**I'll write docs and migration guides if & after we agree to take this in**

I shamelessly kicked the can on [user initiated rebuilds to this related issue](https://github.com/JasperFx/marten/issues/2952).

Highlights:

* There's some somewhat related work in here about a new dynamic discovery of tenant databases that originally spawned this
* The old `ShardAgent` was completely replaced by a new `SubscriptionAgent`
* Tries to process events in bigger batches, especially effective for projections that pick sparsely through the greater events table
* Improved "Rebuild" to detect the completion -- or error condition -- more reliably
* If targeting a single database, Marten possibly runs projections on separate nodes
* If targeting multiple databases, Marten will run all of the projections for a single database on the same node, but may run different database projections from different nodes
* The hot/cold failover is still using advisory locks
* Serialization error tracking, and apply errors can both be configured to be skipped or not. Default (@oskardudycz 's idea) is that both types of errors are skipped in normal async daemon usage, but not skipped in rebuilds where you have immediate feedback
* The custom error handling we had in Marten 4-6 was deleted. Polly is in place for transient errors now, and there are just plain old try/catch loops around any exceptions that escape Polly and automatically pause the projection that threw the exception -- where Marten will occasionally try to restart the projection


src/Marten/Events/Daemon/HighWater/HighWaterAgent.cs
```diff
@@ -52,15 +52,15 @@ public async Task Start()
         _tracker.Publish(
             new ShardState(ShardState.HighWaterMark, _current.CurrentMark) { Action = ShardAction.Started });
 
-        _loop = Task.Factory.StartNew(DetectChanges, _token,
+        _loop = Task.Factory.StartNew(detectChanges, _token,
             TaskCreationOptions.LongRunning | TaskCreationOptions.AttachedToParent, TaskScheduler.Default);
```
Hawxy, line 56: FYI this might not be doing what you expect. `StartNew` has no support for async functions, so as soon as the first `await` is hit in `detectChanges`, the thread created from the `LongRunning` flag will be thrown away.

src/Marten/Events/Daemon/Coordination/MultiTenantedProjectionDistributor.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        foreach (var advisoryLock in _locks)
+        {
+            await advisoryLock.DisposeAsync().ConfigureAwait(false);
+        }
+    }
+
+    public async ValueTask<IReadOnlyList<IProjectionSet>> BuildDistributionAsync()
+    {
+        var random = new Random();
```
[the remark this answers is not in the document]
jeremydmiller: I didn't know about that. It's a minor allocation saving, but it *might* help spawn a more random distribution of values, no pun intended. Doing that right now.

src/Marten/Events/Daemon/HighWater/HighWaterAgent.cs
jeremydmiller: I'm shifting that to Task.Run() right now.

jeremydmiller: And using Task.Run() makes the daemon hang, so yes, that was done at least somewhat purposely

src/Marten/Events/Daemon/ProjectionDaemon.cs
```diff
@@ -1,53 +1,52 @@
...
-    private readonly HighWaterAgent _highWater;
-    private readonly ILogger _logger;
     private readonly DocumentStore _store;
-    private CancellationTokenSource _cancellation;
-    private INodeCoordinator? _coordinator;
-
-    private bool _isStopping;
-    private readonly RetryBlock<DeadLetterEvent> _deadLetterBlock;
+    private readonly IAgentFactory _factory;
+    private readonly List<ISubscriptionAgent> _active = new();
```
oskardudycz, line 22: **CONCERN:** Shouldn't we use here `ImHashMap`? That'd potentially limit the scenarios with running multithreaded access when starting/stopping/detecting the shard agents in the methods below (Start, Stop, etc.). This can be important in terms of pausing and restarting active daemon shards.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
```diff
@@ -0,0 +1,252 @@
...
+    {
+    }
+}
+
+public class ProjectionCoordinator : IProjectionCoordinator
+{
+    private readonly StoreOptions _options;
+    private readonly ILogger<ProjectionCoordinator> _logger;
+
+    private readonly ConcurrentDictionary<IMartenDatabase, IProjectionDaemon> _daemons = new();
```
oskardudycz, line 26: **QUESTION:** Is there any reason not to use `ImHashMap` here?

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
```diff
@@ -0,0 +1,252 @@
...
+            return daemon;
+        }
+
+        daemon = database.StartProjectionDaemon(Store, _logger);
+        _daemons[database] = daemon;
+
+        return daemon;
+    }
+
+    public async ValueTask<IProjectionDaemon> DaemonForDatabase(string databaseIdentifier)
```
oskardudycz, line 72: **QUESTION:** I don't see that used anywhere. Is it a future concept?

Also minor, but if it's needed, then it may be worth adding private method  `ValueTask<IProjectionDaemon> DaemonForDatabase(MartenDatabase database)` unifying the handling in both `DaemonForMainDatabase` methods.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
```diff
@@ -0,0 +1,252 @@
...
+        _logger = logger;
+        _resilience = store.Options.ResiliencePipeline;
+
+        Store = store;
+    }
+
+    public IProjectionDaemon DaemonForMainDatabase()
+    {
+        var database = (MartenDatabase)Store.Tenancy.Default.Database;
+        if (_daemons.TryGetValue(database, out var daemon))
```
oskardudycz, line 61: **SUGGESTION:** I think that it'd be safer to use here `GetOrAdd`. We're not doing here async operations, so it's safer to use as we won't have race conditions between TryGetValue and assignments calls.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
```diff
@@ -0,0 +1,252 @@
...
+
+    public DocumentStore Store { get; }
+
+    public IProjectionDistributor Distributor { get; }
+
+    internal record DaemonShardName(IProjectionDaemon Daemon, ShardName Name);
+
+    public Task StartAsync(CancellationToken cancellationToken)
+    {
+        _runner = Task.Run(() => executeAsync(_cancellation.Token), _cancellation.Token);
```
oskardudycz, line 94: **CONCERN:** Why are we adding a `Task.Run` wrapper here if executeAsync already returns a task?

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
```diff
@@ -0,0 +1,252 @@
...
+#endif
+
+        try
+        {
+#pragma warning disable VSTHRD003
+            await _runner.ConfigureAwait(false);
+#pragma warning restore VSTHRD003
+        }
+        catch (Exception)
+        {
```
oskardudycz, line 114: **CONCERN:** Shouldn't we add at least a log entry here?

src/Marten/Events/Daemon/Coordination/SingleTenantProjectionDistributor.cs
```diff
@@ -0,0 +1,86 @@
...
+        var database = _store.Storage.Database;
+        var projectionShards = _store.Options.Projections.AllShards();
+
+        var random = new Random();
+
+        IReadOnlyList<IProjectionSet> sets = projectionShards.Select(shard =>
+        {
+            // Make deterministic for each projection name
+            var lockId =
+                Math.Abs($"{_store.Options.EventGraph.DatabaseSchemaName}:{shard.Name.Identity}"
```
oskardudycz, line 39: **QUESTION:** Will (currently or in the future) shard name be different than `All`? Of what I see currently, even though there's a constructor allowing to pass different shard name, it's not used anywhere besides the default constructor with `IProjectionSource`.

src/Marten/Events/Daemon/Coordination/SingleTenantProjectionDistributor.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        _locks = new(db => new AdvisoryLock(db, logger));
+    }
+
+    public ValueTask<IReadOnlyList<IProjectionSet>> BuildDistributionAsync()
+    {
+        var database = _store.Storage.Database;
+        var projectionShards = _store.Options.Projections.AllShards();
+
+        var random = new Random();
```
oskardudycz, line 33: **SUGGESTION:** Same suggestion as @gfoidl about `Random.Shared`

src/Marten/Events/Daemon/Coordination/SingleTenantProjectionDistributor.cs
```diff
@@ -0,0 +1,86 @@
...
+    public Task ReleaseLockAsync(IProjectionSet set)
+    {
+        return _locks[set.Database].ReleaseLockAsync(set.LockId);
+    }
+
+    public async Task ReleaseAllLocks()
+    {
+        foreach (var @lock in _locks)
+        {
+            await @lock.DisposeAsync().ConfigureAwait(false);
```
oskardudycz, line 81: **CONCERN:** Shouldn't we use `SafeDispose` here? If one disposal fails for some reason, others won't be disposed, and locks won't be released.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
```diff
@@ -0,0 +1,252 @@
...
+            try
+            {
+                var sets = await Distributor
+                    .BuildDistributionAsync().ConfigureAwait(false);
+
+                foreach (var set in sets)
+                {
+                    // Is it already running here?
+                    if (Distributor.HasLock(set))
```
oskardudycz, line 160: **CONCERN:** Do we need this if on HasLock? I think that just calling `TryAttainLock` should be enough; if we don't get it, then we're good.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
```diff
@@ -0,0 +1,252 @@
...
+                    await Task.Delay(_options.Projections.AgentPauseTime, stoppingToken).ConfigureAwait(false);
+                }
+                else
+                {
+                    await Task.Delay(_options.Projections.LeadershipPollingTime.Milliseconds(), stoppingToken).ConfigureAwait(false);
+                }
+            }
+            catch (TaskCanceledException)
+            {
+                // just get out of here, this signals a graceful shutdown attempt
```
oskardudycz, line 199: **CONCERN:** Could we add logging here?

src/Marten/Events/Daemon/Internals/SubscriptionAgent.cs
```diff
@@ -0,0 +1,298 @@
...
+        try
+        {
+#if NET8_0_OR_GREATER
+            await _cancellation.CancelAsync().ConfigureAwait(false);
+#else
+            _cancellation.Cancel();
+#endif
+            await _execution.HardStopAsync().ConfigureAwait(false);
+            PausedTime = DateTimeOffset.UtcNow;
+            Status = AgentStatus.Paused;
```
oskardudycz, line 67: **QUESTION:** What does the paused status mean in this context? Will it also be used in the future for the pausing and stopping agent? Currently, I see that's only paused when there was an error in the processing. Then it'll be restarted in ProjectionCoordinator, but if the issue is non-transient, then it will always fail again. 

Am I missing something? Maybe it's worth having differentiation between paused and errored?

src/Marten/Events/Daemon/Internals/SubscriptionAgent.cs
```diff
@@ -0,0 +1,298 @@
...
+        }
+        catch (Exception e)
+        {
+            _logger.LogError(e, "Error trying to pause subscription agent {Name}", ProjectionShardIdentity);
+        }
+    }
+
+    long ISubscriptionAgent.Position => LastCommitted;
+
+    public AgentStatus Status { get; private set; } = AgentStatus.Running;
```
oskardudycz, line 85: **CONCERN:** Shouldn't default status be paused?

src/Marten/Events/Daemon/Internals/SubscriptionAgent.cs
```diff
@@ -0,0 +1,298 @@
...
+
+    public async Task StopAndDrainAsync(CancellationToken token)
```
oskardudycz, line 87: **QUESTION:** Shouldn't this method change Agent Status to paused?

oskardudycz (commented):
I did what I could to review the PR, adding some notes. I hope they will be helpful. The scope of changes is so big, and the renames are not making it easier to find what's new and what just moved place. I feel like I barely scratched the surface...

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
oskardudycz: See also https://gist.github.com/davidfowl/3dac8f7b3d141ae87abf770d5781feed#file-concurrentdictionaryextensions-cs-L53.

src/Marten/Events/Daemon/ProjectionDaemon.cs
jeremydmiller: Good catch. See https://github.com/JasperFx/marten/issues/2967 for follow up work. Yeah, I think you're right about the potential for concurrent access. The `ImHashMap` might be appropriate (or ImList instead), but I'd still lock *write* access to it regardless. In this new model, we're scrapping and rebuilding the `SubscriptionAgent` completely instead of just pausing/restarting. Made for a little bit simpler model

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
jeremydmiller: Yes, I purposely used ConcurrentDictionary here. We're not accessing it often enough for the perf to matter, but the big thing was to never end up with an orphaned object instance of an IProjectionDaemon. We could use ImHashMap, but we'd need to add locking around writes to it. Or whatever the built in wrapper around it it. I just thought it was easier to use ConcurrentDictionary here.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
jeremydmiller: Everything around IProjectionCoordinator is "future use" pending whatever ends up happening here: https://github.com/JasperFx/marten/issues/2952. The mechanics for accessing the main database and a named database isn't actually quite the same here, so I wouldn't worry about unifying them.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
jeremydmiller: At this point I'm more comfortable with ImHashMap + a lock and I'm switching that now like you suggested before.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
jeremydmiller: Scratch what I said, switching to ImHashMap & a lock now.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
jeremydmiller: Nope, you were right. Good suggestion. New code will have this:

```csharp
    private IProjectionDaemon findDaemonForDatabase(MartenDatabase database)
    {
        if (_daemons.TryFind(database, out var daemon))
        {
            return daemon;
        }

        lock (_daemonLock)
        {
            if (_daemons.TryFind(database, out daemon))
            {
                return daemon;
            }

            daemon = database.StartProjectionDaemon(Store, _logger);
            _daemons = _daemons.AddOrUpdate(database, daemon);
        }

        return daemon;
    }
```

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
jeremydmiller: Did so. Not sure it's valuable, but why not just in case because I've been wrong about that before. Did have it ignore TaskCanceledException though.

src/Marten/Events/Daemon/Coordination/SingleTenantProjectionDistributor.cs
jeremydmiller: Sigh. This is a bit of future proofing that got into V4. It's from this: https://github.com/JasperFx/marten/issues/1715

We ran out of time and energy to do this in V4. At the time I though that we would provide "sharding" within some projections, and that's what that was meant for. Like being able to say things like arbitrarily dividing up streams into sets (based on a number using modulo operators maybe?), then running the same projection, but segmented in to multiple shards. Or being able to say "run this tenant in one shard, this other tenant in another shard, and lastly, all other tenants in this third shard"

I don't think we'll make any attempt to do anything w/ this in V7, but its' something I'd like to consider in the future. 

At this point, we enforce that every projection has a unique projection name. So every projection will have just the one "shard" for "All", but the hash will at least be unique. I incorporated the shard name into the hash as a future proofing thing. To sum that up, **every single shard will get a unique hash because it's built from the projection name, shard name, and schema name too**.

The inclusion of the schema name was to try to proof this against folks using separate stores in the same DB.

src/Marten/Events/Daemon/Coordination/SingleTenantProjectionDistributor.cs
jeremydmiller: Good catch. Just missed it before.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
jeremydmiller: I wasn't sure that the call through to try to attain the lock would be repeatable to be honest. This is specifically for the case of "we own this projection (shard), but it's been paused, so try to restart it"

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
jeremydmiller: I honestly don't want to, because this is just indicative of the system being gracefully shutdown. You can get these exceptions even when there's nothing that's actually wrong. I didn't want to log this, because it makes people freak out even though it's really nothing.

src/Marten/Events/Daemon/Internals/SubscriptionAgent.cs
jeremydmiller: "Paused" absolutely just means "temporarily stopped because of an error" here.

I think I'd like to talk this through with you over Zoom. This and https://github.com/JasperFx/marten/issues/2952 and what monitoring might be for this stuff. Also how it's going to be different in "Critter Stack Pro".

Let's track this work here: https://github.com/JasperFx/marten/issues/2968

src/Marten/Events/Daemon/Internals/SubscriptionAgent.cs
jeremydmiller: That's actually correct in this iteration. To avoid having a lot of mutable state like the action blocks and what not inside of `SubscriptionAgent`, I opted to say that `SubscriptionAgent` and its internals are completely replaced when it's "restarted". So the `SubscriptionAgent` is started any time it's built. That's a difference between the old ShardAgent and this.

I didn't like the mutable state like this in Wolverine sending agents, and wanted the more immutable state this time to simplify the internals of SubscriptionAgent.

src/Marten/Events/Daemon/Internals/SubscriptionAgent.cs
jeremydmiller: I went with `Stopped` becuase this is a result of a graceful shutdown. Good catch though.

jeremydmiller:
> I did what I could to review the PR, adding some notes. I hope they will be helpful. The scope of changes is so big, and the renames are not making it easier to find what's new and what just moved place. I feel like I barely scratched the surface...

That was a lot! I spawned two new issues #2968 and #2967. Plus #2952 is also a follow up item.

I think I've taken care of or answered the other comments. I'm going to try to do something about #2967 before pushing a new beta.

src/Marten/Events/Daemon/Internals/SubscriptionAgent.cs
```diff
@@ -0,0 +1,302 @@
...
+
+#if NET8_0_OR_GREATER
+            await _cancellation.CancelAsync().ConfigureAwait(false);
+#else
+            _cancellation.Cancel();
+#endif
+
+            await _execution.StopAndDrainAsync(token).ConfigureAwait(false);
+        }
+        catch (TaskCanceledException)
```
gfoidl, line 103:
```suggestion
        catch (OperationCanceledException)
```
Because of https://github.com/dotnet/runtime/issues/83406

src/Marten/Events/Daemon/ProjectionDaemon.cs
```diff
@@ -56,549 +56,312 @@ await using (session.ConfigureAwait(false))
...
 
-            var parameters = new ActionParameters(agent, async () =>
+    public async Task StopAsync(string shardName, Exception ex = null)
+    {
+        if (_agents.TryFind(shardName, out var agent))
+        {
+            await _semaphore.WaitAsync(_cancellation.Token).ConfigureAwait(false);
+            try
             {
+                var cancellation = new CancellationTokenSource();
```
gfoidl, line 189: Should this be a [linked cancellation token source](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.createlinkedtokensource?view=net-8.0)?
I.e. if cancellation happens externally (via the passed in token) or through the timeout (5 seconds) both will cancel the action.

The same on other places here.

src/Marten/Events/Daemon/Coordination/ProjectionCoordinator.cs
oskardudycz: I understand, but why do we need `Task.Run` wrapper instead of just running `executeAsync`?

src/Marten/Events/Daemon/Coordination/SingleTenantProjectionDistributor.cs
oskardudycz: Ok, understood; as we now have a single shard per projection, this will work. I was wondering if you want to include some additional sharding here, e.g., per tenant. Then it wouldn't be unique.

src/Marten/Events/Daemon/Internals/SubscriptionAgent.cs
oskardudycz: Ok, I'm curious how that will work. I'm fine with moving that to other dedicated issues.

src/Marten/Events/Daemon/ProjectionDaemon.cs
jeremydmiller: Yeah, I think there was a comment about that in at least one place. I'll make this one today

oskardudycz (approved):
I'm fine with letting that in. It looks good, although the scope of change is big, and I might not notice some stuff. I tried to grasp the general concept and issues, but it's non-trivial.

Issue #2962, opened 2024-02-15
Add some kind of "BeforeCommit" hook to IChangeListener


jeremydmiller:
More details to follow...

PR #2970, opened 2024-02-19
Got a BeforeCommitAsync() as part of async listeners. Closes GH-2962

jeremydmiller:


oskardudycz (approved):
It looks good to me; we'll need to cover that in the breaking changes notes as some people may be surprised by the change of `AfterCommit` behaviour.

Issue #2941, opened 2024-02-01
Using Explicit Code for Marten live aggregate projections


erdtsieck:
At the moment if you write explicit code, such as https://jeremydmiller.com/2024/01/31/using-explicit-code-for-marten-projections/?replytocom=48346#respond for an aggregate and you change `opts.Projections.LiveStreamAggregation<Gesprek>();`  into `opts.Projections.Add(new GesprekProjection(), ProjectionLifecycle.Live);`

You get `Unhandled exception. System.ArgumentOutOfRangeException: Live cannot be used for IProjection (Parameter 'lifecycle')`

For the reasons in the article and to prevent code generation bugs, it would be nice to be able to do this

PR #3002, opened 2024-02-29
#2941 Tests for: Using Explicit Code for Marten live aggregate projections

erdtsieck:
@jeremydmiller  I am not sure if I can do the actual implementation, but I think having a test upfront helps?


jeremydmiller:
@erdtsieck It certainly does!

jeremydmiller:
This did get done finally.

PR #3008, opened 2024-03-02
move BeforeSaveChanges before ProcessEvents

borgez:


borgez:
https://github.com/JasperFx/marten/discussions/3004

borgez:
@jeremydmiller everything is ok? maybe set plan to merge?

jeremydmiller:
@borgez It's a breaking change in behavior. I can't take this in. What was the purpose of this? Why did you need this?

borgez:
@jeremydmiller 

I want to add general data like uid, etc... to the event, and expected behavior: projections handle this changes in "inline" mode.

The current behavior doesn't seem quite as expected because the changes are applied to projections before the hook.
Maybe then add another hook for this case like BeforeApplyChanges, but it doesn't seem very good?

The test describes my case:
```
      public override void BeforeSaveChanges(IDocumentSession session)
      {
          var streams = session.PendingChanges.Streams();
          foreach (var stream in streams)
          {
              foreach (var e in stream.Events)
              {
                  if (e.Data is EnrichedUserEvent eueb)
                  {
                      eueb.Actor = "some user from identity";
                  }
              }
          }
      }
```

borgez:
@jeremydmiller maybe add hook BeforeEventProcessing?

borgez:
@jeremydmiller revorked

PR #3015, opened 2024-03-06
Fix non awaited tasks in WaitForNonStaleProjectionDataAsync()

Xzelsius:
`Task.Factory.StartNew()` will not await the `WaitForNonStaleProjectionDataAsync()` inside.
So the extension method on the `IDocumentStore` will return "completed" before the actual task on the DB is done.

### DIFF
--- a/src/Marten/Events/AsyncProjectionTestingExtensions.cs
+++ b/src/Marten/Events/AsyncProjectionTestingExtensions.cs
@@ -51,7 +51,7 @@ public static class TestingExtensions
 
     public static async Task WaitForNonStaleProjectionDataAsync(this IMartenDatabase database, TimeSpan timeout)
     {
-        var cancellationSource = new CancellationTokenSource();
+        using var cancellationSource = new CancellationTokenSource();
         cancellationSource.CancelAfter(timeout);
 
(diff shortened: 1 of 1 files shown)


src/Marten/Events/AsyncProjectionTestingExtensions.cs
```diff
@@ -26,7 +26,7 @@ public static async Task WaitForNonStaleProjectionDataAsync(this IDocumentStore
         else
         {
             var tasks = databases
-                .Select(db => Task.Factory.StartNew(() => db.WaitForNonStaleProjectionDataAsync(timeout))).ToArray();
+                .Select(db => db.WaitForNonStaleProjectionDataAsync(timeout)).ToArray();
```
gfoidl, line 29: I'd drop the `ToArray` here*.

Currently `Task.WhenAll` will also create an array when an `IEnumerable<Task>` is given, but maybe in future the .NET implementation comes up with a solution that avoids that allocation, than Marten can benefit from this for free.

\* I see that this was there before

Xzelsius: Thanks, wasn't aware of this.
Remove it :)

gfoidl (approved):

PR #3022, opened 2024-03-07
docs(getting-started.md): use new paket-cli syntax

GregOnNet:
The previous command used paket add nuget ...
nuget is the default and can be omitted.
The Paket CLI also gives a warning: "Please use the new syntax: 'nuget' is the default argument and should be omitted."

image


mysticmind:
@GregOnNet Thanks for the PR!

Issue #3031, opened 2024-03-09
False "Document type object is not configured as soft deleted" when using MaybeDeleted in generic linq query


JonnyWideFoot:
These use cases worked fine in MartenDB version 6.4.1. See inline comments for each test pass/fail status.

The use case from our code base is the ArchiveStateExtensions.WithArchivedState extension method, and I have tried to provide a minimal reproducible example in JonTests.AddWhere() below.

```csharp
using System;
using System.Linq;
using Marten.Linq.SoftDeletes;
using Marten.Metadata;
using Marten.Testing.Harness;
using Shouldly;

namespace LinqTests.Bugs;

public class JonTests : BugIntegrationContext
{
    public static IQueryable<T> AddWhere<T>(IQueryable<T> q)
    {
        return q.Where(x => x.MaybeDeleted());
    }

    public static IQueryable<Entity> AddWhereNonGeneric(IQueryable<Entity> q)
    {
        return q.Where(x => x.MaybeDeleted());
    }

    // Fails with "System.InvalidOperationException":
    // "Document type object is not configured as soft deleted"
    [Fact]
    public void test_any_maybe_deleted_in_generic_method()
    {
        var entity = new Entity { Id = Guid.NewGuid() };

        using (var sessionX = theStore.LightweightSession())
        {
            sessionX.Store(entity, entity);
            sessionX.SaveChanges();
        }

        using var session = theStore.QuerySession();
        var q = session.Query<Entity>();
        var w = AddWhere(q);
        var result = w.Any();

        result.ShouldBeTrue();
    }

    // Works fine
    [Fact]
    public void test_any_maybe_deleted_in_non_generic_method()
    {
        var entity = new Entity { Id = Guid.NewGuid() };

        using (var sessionX = theStore.LightweightSession())
        {
            sessionX.Store(entity, entity);
            sessionX.SaveChanges();
        }

        using var session = theStore.QuerySession();
        var q = session.Query<Entity>();
        var w = AddWhereNonGeneric(q);
        var result = w.Any();

        result.ShouldBeTrue();
    }

    // Fails with "System.InvalidOperationException":
    // "Document type object is not configured as soft deleted"
    [Fact]
    public void test_any_maybe_deleted_in_generic_extension_method()
    {
        var entity = new Entity { Id = Guid.NewGuid() };

        using (var sessionX = theStore.LightweightSession())
        {
            sessionX.Store(entity, entity);
            sessionX.SaveChanges();
        }

[...]

PR #3034, opened 2024-03-10
Add generic context support for SoftDelete extension methods

JonnyWideFoot:
See issue: https://github.com/JasperFx/marten/issues/3031

### DIFF
--- a/src/Marten/Linq/SoftDeletes/SoftDeletedExtensions.cs
+++ b/src/Marten/Linq/SoftDeletes/SoftDeletedExtensions.cs
@@ -8,10 +8,8 @@ public static class SoftDeletedExtensions
     ///     The search results should include all documents, whether
     ///     soft-deleted or not
     /// </summary>
-    /// <typeparam name="T"></typeparam>
-    /// <param name="doc"></param>
     /// <returns></returns>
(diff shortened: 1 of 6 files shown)


src/Marten/Linq/SoftDeletes/SoftDeletedExtensions.cs
```diff
@@ -8,10 +8,8 @@ public static class SoftDeletedExtensions
     ///     The search results should include all documents, whether
     ///     soft-deleted or not
     /// </summary>
-    /// <typeparam name="T"></typeparam>
-    /// <param name="doc"></param>
     /// <returns></returns>
-    public static bool MaybeDeleted(this object doc)
+    public static bool MaybeDeleted<T>(this T doc)
```
gfoidl, line 12: Noting that this is a binary breaking change, so updates w/o re-compilation won't work.

Adding new overloads with the generic signature wouldn't be breaking change, and the C# compiler has to bind to the generic overload (as it's more specific than the `object` overload).

jeremydmiller:
Ugh, I didn't see this fast enough. What @gfoidl said is unfortunately true. I've got a PR coming that unwraps the Expression to slide around this problem.

PR #3042, opened 2024-03-11
Advanced SQL query

e-tobi:
AdvancedSqlQueryAsync<T, ...>(sql, ...) allows to perform more complex SQL queries than would be possible with Query<T>(sql). It allows to query documents, scalars and deserializable JSON objects or any combination of these returned as a tuple.
This also allows to query for related documents. Metadata mapping, dirty tracking etc. for documents is fully supported. The full SQL must be provided (no "magic" SQL building) so the user is responsible to provide all the columns needed by Marten in the correct order.

Currently result tuples with up to 3 elements are possible (a single result type will not be returned as a tuple). This could be easily extend, but then I would move all the overloads into a partial class. For the same reason I have not created an extra overload for the async methods without a CancellationToken.

I've tried to implement this as least "invasive" as possible. Only IQuerySession and its implementation are extended, the magic happens in the new AdvancedQueryHandler class.

Future improvments might be:
- More result types
- Instead of mapping columns by index, map them by name (would require larger changes, especially to the generated code)

### DIFF
--- a/docs/documents/querying/advanced-sql.md
+++ b/docs/documents/querying/advanced-sql.md
@@ -4,9 +4,9 @@ Besides Linq queries or simple raw SQL queries via `session.Query<T>("where...")
 With this method Marten does not try to add any missing parts to the SQL query, instead you have to provide the whole query string yourself.
 
 Marten just makes some assumptions on how the schema of the SQl query result must look like, in order to be able to map the query result to documents, scalars or other JSON serializable types.
-With `AdvancedSqlQueryAsync` it is even possible to return multiple documents, objects and scalars as a tuple.
+With `AdvancedSqlQueryAsync` / `AdvancedSqlQuery` it is even possible to return multiple documents, objects and scalars as a tuple. Currently up to three result types can be queried for.
 
-The following rules must be followed when doing queries with `AdvancedSqlQueryAsync`:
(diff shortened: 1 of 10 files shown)


jeremydmiller:
I'm going to put this into 7.4 later this week if that's okay. I'm rushing out a bug fix release first

docs/documents/querying/advanced-sql.md
```diff
@@ -0,0 +1,138 @@
...
+4. `mt_version` - only when versioning is enabled
+5. `mt_last_modified` - only if this metadata is enabled 
+6. `mt_created_at` - only if this metadata is enabled
+7. `correlation_id` - only if this metadata is enabled
+8. `causation_id` - only if this metadata is enabled
+9. `last_modified_by` - only if this metadata is enabled
+10. `mt_deleted` - only if this metadata is enabled
+11. `mt_deleted_at` - only if this metadata is enabled
+
+You can always check the the correct result column order, by inspecting the command text created from a Linq query: `var commandText = session.Query<T>().ToCommand().CommandText;`
```
pascalberger, line 31:
```suggestion
You can always check the correct result column order, by inspecting the command text created from a Linq query: `var commandText = session.Query<T>().ToCommand().CommandText;`
```

e-tobi: Thanks!

jeremydmiller:
This will go into 7.5 shortly. I didn't mean to keep this lingering so long.

PR #3055, opened 2024-03-17
Fix for codegen issue w/ live aggregator in heavy parallelization usage

jeremydmiller:


erdtsieck (approved):

src/Marten/Events/Aggregation/GeneratedAggregateProjectionBase.CodeGen.cs
```diff
@@ -18,11 +18,19 @@ namespace Marten.Events.Aggregation;
 
 public abstract partial class GeneratedAggregateProjectionBase<T>
 {
+    private readonly object _compilationLock = new();
```
erdtsieck, line 21: Shouldn't this one be static since code gen and compilation is not tied to this object instance per se?

PR #3063, opened 2024-03-18
Case where an include for a parent relationship from a child is no longer working.

corstian:


jeremydmiller:
@corstian Easy money, fix will be in Marten 7.4.0 shortly

Issue #3071, opened 2024-03-20
New resiliency configuration via `.ConfigurePolly()` does not allow to extend default marten policies


a-shtifanov-laya:
As discussed on discord, previously (before v7) it was possible to extend marten default resilience policy.
v7 introduced resilient policy configuration with Polly (a breaking change), and there is no possibility right now to extend default marten configuration - only overwrite it.

a-shtifanov-laya:
..pr is comming

jeremydmiller:
Closed by #3072

PR #3072, opened 2024-03-20
Allow to extend Resiliency Policies via `.ExtendPolly()`

a-shtifanov-laya:
#3071

### DIFF
--- a/docs/configuration/retries.md
+++ b/docs/configuration/retries.md
@@ -1,7 +1,7 @@
 # Resiliency Policies
 
 ::: info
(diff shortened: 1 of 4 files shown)


a-shtifanov-laya:
i didn't find any tests for the original `.ConfigurePolly()` method, so i'm not sure if i need add any here as well

jeremydmiller:
I'll put a smoke test on top of this, no worries on that part. Thanks for updating the docs. I'll get this into Marten 7.4 maybe later today

Issue #3078, opened 2024-03-21
Providing a version for aggregating that is too high should aggregate to null


erdtsieck:
```cs
// Given
var streamId = Guid.NewGuid();
var options = new StoreOptions();
options.Connection(ConnectionSource.ConnectionString);
options.Projections.LiveStreamAggregation<PolledAggregate>();

// When
var store = new DocumentStore(options);
var session = store.LightweightSession();
session.Events.Append(streamId, new PollingShouldFail(streamId));
await session.SaveChangesAsync();

// Then
var aggregate = await session.Events.AggregateStreamAsync<PolledAggregate>(streamId, 2);
aggregate.ShouldBeNull();
```

PR #3079, opened 2024-03-21
#3078 Providing a version for aggregating that is too high should aggregate to null

erdtsieck:


jeremydmiller:
@erdtsieck I'm not sure I agree with this one

erdtsieck:
> @erdtsieck I'm not sure I agree with this one

@jeremydmiller Well, I do not expect you to. My intent is to be able to do polling whether an aggregate has arrived at an expected point, as @oskardudycz did in his example: https://github.com/oskardudycz/EventSourcing.NetCore/blob/main/Sample/Helpdesk.Wolverine/Helpdesk.Api/Incidents/ResolutionBatch/GettingIncidentResolutionBatch/GetIncidentsBatchResolution.cs

But you could of course just do the aggregate.version check yourself instead of using the API

Or another API method?

jeremydmiller:
There's the FetchStreamStateAsync() API you could use for that that would be much, much lighter weight. Not sure what Oskar was up to there.

jeremydmiller:
I think I'd rather have it throw an exception rather than do that, and I think I disagree w/ that too because it's filtering the events and might not even know what the current stream version is

Issue #3080, opened 2024-03-22
`WaitForNonStaleProjectionDataAsync` does not fetch projection progress correctly


baltie:
We've been trying to use the new `IDocumentStore.WaitForNonStaleProjectionDataAsync()` test extension method in our integration tests to wait for the async projections to stabilize before moving on in the test, using code that depends on the projections being up to date.

However, it seems `WaitForNonStaleProjectionDataAsync` doesn't work correctly yet. Reading the source code, the projection progress is retrieved only once, and then it will keep checking that same result over and over until timeout. See https://github.com/JasperFx/marten/blob/61de7fcbe5450d42e401580fa76bb915a2e49430/src/Marten/Events/AsyncProjectionTestingExtensions.cs#L67-L72

Here, `AllProjectionProgress` returns an `IReadOnlyList<ShardState>` and fetching the projection progress will not be executed again. All our tests are thus cancelled after the timeout. I've verified in the database that the async projections have run, but it is not picked up by `WaitForNonStaleProjectionDataAsync`, and I strongly suspect it is due to the above code snippet.

baltie:
See #2624. There's the additional problem of `WaitForNonStaleProjectionDataAsync` not waiting for all async projections to appear in `mt_event_progression` before proceeding. This only occurs when starting with an empty database, but that is the case when we run our integration tests - we let Marten create everything.

jeremydmiller:
Pfft, this is the 3rd or 4th attempt to improve this thing. I have a reproduction project from a JasperFx client that I'm about to use to try to improve this. This is the next thing up for Marten at least.

a-shtifanov-laya:
i'm actually preparing a PR for this 😃

jeremydmiller:
Closed by https://github.com/JasperFx/marten/commit/96a9e2032d9fc44fafdaacedae6d399625cc6bc1

daveHylde:
Sorry to bring this back to light again, but I'm facing these issues still, also after upgrading to v7.18.0.
The `WaitForNonStaleProjectionDataAsync` call works for a single test in an `IClassFixture`, but the rest of the tests gets the timeout excetion. All pass when run one at a time.
@baltie can you confirm that you don't have this issue anymore?

daveHylde:
I've also noticed that this only is an issue when initializing tests with `ResetAllData`.

jeremydmiller:
[...]

PR #3081, opened 2024-03-23
Correction for `WaitForNonStaleProjectionDataAsync` method

a-shtifanov-laya:
GH-3080

### DIFF
--- a/src/Marten.AsyncDaemon.Testing/event_projections_end_to_end.cs
+++ b/src/Marten.AsyncDaemon.Testing/event_projections_end_to_end.cs
@@ -127,6 +127,7 @@ public class event_projections_end_to_end : DaemonContext
         var events = (await session.Events.QueryAllRawEvents().ToListAsync());
         var travels = events.OfType<Event<Travel>>().ToDictionary(x => x.Id);
 
+        distances.Count.ShouldBe(travels.Count);
         foreach (var distance in distances)
         {
             if (travels.TryGetValue(distance.Id, out var travel))
--- a/src/Marten/Events/AsyncProjectionTestingExtensions.cs
+++ b/src/Marten/Events/AsyncProjectionTestingExtensions.cs
@@ -1,8 +1,11 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using JasperFx.Core;
+using Marten.Events.Daemon;
+using Marten.Events.Projections;
 using Marten.Storage;
 
 namespace Marten.Events;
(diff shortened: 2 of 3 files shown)


src/Marten.AsyncDaemon.Testing/event_projections_end_to_end.cs
```diff
@@ -127,6 +127,7 @@ private async Task CheckExpectedResults(IDocumentSession session)
         var events = (await session.Events.QueryAllRawEvents().ToListAsync());
         var travels = events.OfType<Event<Travel>>().ToDictionary(x => x.Id);
 
+        distances.Count.ShouldBe(travels.Count);
```
a-shtifanov-laya, line 130: looks like this is why the tests have passed.. `distances` collection was always empty (because `WaitForNonStaleProjectionDataAsync` method didn't await for shards to start), so iteration 
`foreach (var distance in distances)`  below was not executed and nothing was asserted

src/Marten/Events/AsyncProjectionTestingExtensions.cs
```diff
@@ -13,62 +16,76 @@ public static class TestingExtensions
...
-    /// <param name="timeout"></param>
     public static async Task WaitForNonStaleProjectionDataAsync(this IDocumentStore store, string tenantIdOrDatabaseName, TimeSpan timeout)
     {
         // Assuming there's only one database in this usage
         var database = await store.Storage.FindOrCreateDatabase(tenantIdOrDatabaseName).ConfigureAwait(false);
 
         if (store.Storage is DefaultTenancy)
-
-        await WaitForNonStaleProjectionDataAsync(database, timeout).ConfigureAwait(false);
+            await WaitForNonStaleProjectionDataAsync(database, store.AsyncProjectionsCount(), timeout).ConfigureAwait(false);
```
a-shtifanov-laya, line 46: not sure, if this code is valid.. my Rider shows 
`Suspicious type check: there is no type in the solution that is inherited from both 'Marten.IMartenStorage' and 'Marten.Storage.DefaultTenancy'`

image

src/Marten/Events/AsyncProjectionTestingExtensions.cs
```diff
@@ -13,62 +16,76 @@ public static class TestingExtensions
...
+    /// </summary>
+    /// <param name="projectionsCount">Will be awaited till all shards have been started before checking if they've caught up with the sequence number</param>
+    /// <exception cref="TimeoutException"></exception>
+    public static async Task WaitForNonStaleProjectionDataAsync(this IMartenDatabase database, int projectionsCount, TimeSpan timeout)
     {
         using var cancellationSource = new CancellationTokenSource();
         cancellationSource.CancelAfter(timeout);
 
-        var initial = await database.FetchEventStoreStatistics(cancellationSource.Token).ConfigureAwait(false);
-        while (initial.EventSequenceNumber == 0 && !cancellationSource.IsCancellationRequested)
```
a-shtifanov-laya, line 57: `initial` was evaluated once and never updated within `while` loop. 
never happened to me that no events were present, but i thought it is still better to correct it

src/Marten/Events/AsyncProjectionTestingExtensions.cs
```diff
@@ -13,62 +16,76 @@ public static class TestingExtensions
...
-        var projections = await database.AllProjectionProgress(cancellationSource.Token).ConfigureAwait(false);
-        while (!projections.All(x => x.Sequence >= initial.EventSequenceNumber) &&
-               !cancellationSource.IsCancellationRequested)
+        IReadOnlyList<ShardState> projections;
+        do
         {
+            projections = await database.AllProjectionProgress(cancellationSource.Token).ConfigureAwait(false);
+            if ((projections.Count >= projectionsCount && projections.All(x => x.Sequence >= initial.EventSequenceNumber))
+                || cancellationSource.IsCancellationRequested)
+                break;
```
a-shtifanov-laya, line 80: here are the main problems that i had
1. `projections` were evaluated only once and never updated
2. on _cold_ start `AllProjectionProgress()` returns an empty collection, so `projections.All(...)` evaluation returns false positive result.. that is why i had to explicitly check if all configured async projection shards are running
@baltie this is exactly what you're experiencing

src/Marten/Events/AsyncProjectionTestingExtensions.cs
```diff
@@ -77,4 +94,8 @@ public static async Task WaitForNonStaleProjectionDataAsync(this IMartenDatabase
                 $"The projections timed out before reaching the initial sequence of {initial.EventSequenceNumber}");
         }
     }
+
+    private static int AsyncProjectionsCount(this IDocumentStore store)
+        => store.Options.Events.Projections().Count(p => p.Lifecycle == ProjectionLifecycle.Async)
+           + new[] { ShardState.HighWaterMark }.Length;
```
a-shtifanov-laya, line 100: not 100% sure if it is the best way to get **all** AsyncProjectionCount.. 
at least, i tried to explicitly reference `ShardState.HighWaterMark` and not just add `+1`

a-shtifanov-laya (commented):
some explanation comments

jeremydmiller:
@a-shtifanov-laya Let me look at this closer. I have a "fix" for the same issue that I didn't push Friday, but otherwise ready to go. I had reproduction steps from a JasperFx client on a test they were using. My version is:

```csharp
    /// <summary>
    /// Wait for any running async daemons to catch up to the latest event sequence at the time
    /// this method is invoked for all projections. This method is meant to aid in automated testing
    /// </summary>
    /// <param name="store"></param>
    /// <param name="timeout"></param>
    public static Task WaitForNonStaleProjectionDataAsync(this IDocumentStore store, TimeSpan timeout)
    {
        if (store.As<DocumentStore>().Tenancy is DefaultTenancy)
        {
            return store.Storage.Database.WaitForNonStaleProjectionDataAsync(timeout);
        }

        throw new InvalidOperationException(
            "If using multi-tenancy through any kind of separate databases per tenant, please specify a tenant id or database name");
    }


    /// <summary>
    /// Wait for any running async daemon for a specific tenant id or database name to catch up to the latest event sequence at the time
    /// this method is invoked for all projections. This method is meant to aid in automated testing
    /// </summary>
    /// <param name="store"></param>
    /// <param name="tenantIdOrDatabaseName">Either a tenant id or the name of a database within the system</param>
    /// <param name="timeout"></param>
    public static async Task WaitForNonStaleProjectionDataAsync(this IDocumentStore store, string tenantIdOrDatabaseName, TimeSpan timeout)
    {
        // Assuming there's only one database in this usage
        var database = await store.Storage.FindOrCreateDatabase(tenantIdOrDatabaseName).ConfigureAwait(false);
        await WaitForNonStaleProjectionDataAsync(database, timeout).ConfigureAwait(false);
    }

    public static async Task WaitForNonStaleProjectionDataAsync(this IMartenDatabase database, TimeSpan timeout)
    {
        var all = database.As<MartenDatabase>().Options.Projections.AllShards()
            .ToDictionary(s => s.Name.Identity, s => 0L);

        using var cancellationSource = new CancellationTokenSource();
        cancellationSource.CancelAfter(timeout);

        var initial = await database.FetchEventStoreStatistics(cancellationSource.Token).ConfigureAwait(false);
        while (initial.EventSequenceNumber == 0 && !cancellationSource.IsCancellationRequested)
        {
            await Task.Delay(100.Milliseconds(), cancellationSource.Token).ConfigureAwait(false);
        }

        if (initial.EventSequenceNumber == 0)
        {
            throw new TimeoutException("No projection or event activity was detected within the timeout span");
        }

        var completed = async () =>
        {
            var projections = await database.AllProjectionProgress(cancellationSource.Token).ConfigureAwait(false);
            foreach (var state in projections)
            {
                all[state.ShardName] = state.Sequence;
            }

            return all.Values.All(x => x >= initial.EventSequenceNumber);
        };

        while (!cancellationSource.IsCancellationRequested)
        {
            if (await completed().ConfigureAwait(true))
            {
                return;
            }
            await Task.Delay(25.Milliseconds(), cancellationSource.Token).ConfigureAwait(false);
        }

        if (cancellationSource.IsCancellationRequested)
        {
            throw new TimeoutException(
                $"The projections timed out before reaching the initial sequence of {initial.EventSequenceNumber}");
        }
    }
```

jeremydmiller:
I think your version would work too fwiw

jeremydmiller:
Nah, I'll use the test that I have and your version. Your version is cleaner. I am going to make it warn to specify a tenant id if you're using multi-tenancy through separate dbs

Thank you for tackling this!

baltie:
Thanks guys! Looking forward to getting rid of my semi-functional workarounds when this is released. 😀

Leh2:
I encountered the same issue. Additionally, I observed that when a timeout occurs within a method utilizing the cancellation token (etc `Task.Delay(..)`), it can lead to a `TaskCanceledException`. As a result, we won't throw the `TimeoutException`.

I can implement a fix, provided no one else is currently making significant changes to this method 😄

src/Marten/Events/AsyncProjectionTestingExtensions.cs
Leh2: @a-shtifanov-laya, I don't believe this is necessary. FetchEventStoreStatistics retrieves the maximum sequence from the event table, eliminating the need for any waiting here.

PR #3092, opened 2024-03-26
Switch to Polly.Core & Bump version

tomap:
It seems Marten was already using the v8 stuff from Polly.Core, but still depending on Polly (legacy): https://www.pollydocs.org/migration-v8.html

I changed the dependency and bumped to latest: https://github.com/App-vNext/Polly/releases


mysticmind (approved):
@tomap Looks good to me. Also approved to kick-off the CI runs.

PR #3102, opened 2024-03-31
Add OpenTelemetry event tracing to connections

SeanFarrow:
This PR add event tracing to IConnectionLifetime.

@jeremydmiller  I have a few questions:
How do we want to test the EventConnectionLifetime and MartenTracing classes?
When do we want to update the documentation?
Do we want to wait to merge this until all the other OTel stuff is done?

Thanks,
Sean.


jeremydmiller:
Honestly, just smoke tests. We'll have to rig up Prometheus or Signoz or something in a test harness to really test it manually.

The decorator around the connection lifetime, does that even need all the if/then checks? Shouldn't it be, only apply the decorator if that option is selected in the first place? You want to absolutely minimize the runtime steps because everything is in the hot path there.

For event metrics, I assumed we'd use an IDocumentSessionListener anyway

SeanFarrow:
Yep, I’ll fix that, I wrote the thing bottom up, so the additional checks are not needed.
Also, the LINQ tests are currently failing on CI. I haven’t investigated this as yet, do you want me to raise an issue for this?
Let me have a look at tests later in the week, we might be able to use some of the fakes Microsoft recently added.
Thanks,
Sean.

src/Marten/Internal/OpenTelemetry/MartenTracing.cs
```diff
@@ -0,0 +1,25 @@
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Marten.Internal.OpenTelemetry;
+
+internal static class MartenTracing
+{
+    // See https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/messaging/ for more information
+    public static string MartenTenantId = "MartenTentantId";
+    public static string MartenCorelationId = "MartenCorelationId";
```
gfoidl, line 10: Can these be `const` are `static readonly`?
When they should be setable, then make them a property.

src/Marten/Internal/OpenTelemetry/MartenTracing.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    internal static ActivitySource ActivitySource { get; } = new(
+        "Marten",
+        typeof(MartenTracing).Assembly.GetName().Version!.ToString());
+
+    public static Activity? StartConnectionActivity(Activity? parentActivity =null, IEnumerable<KeyValuePair<string, object?>>? tags =null)
```
gfoidl, line 16: Nit
```suggestion
    public static Activity? StartConnectionActivity(Activity? parentActivity =null, IEnumerable<KeyValuePair<string, object?>>? tags = null)
```

src/Marten/Internal/OpenTelemetry/MartenTracing.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        return StartActivity("connection", parentActivity, tags);
+    }
+
+    public static Activity StartActivity(string spanName, Activity? parentActivity =null, IEnumerable<KeyValuePair<string, object?>>? tags =null, ActivityKind activityKind =ActivityKind.Internal)
```
gfoidl, line 21:
```suggestion
    public static Activity StartActivity(string spanName, Activity? parentActivity = null, IEnumerable<KeyValuePair<string, object?>>? tags = null, ActivityKind activityKind = ActivityKind.Internal)
```

src/Marten/Internal/OpenTelemetry/MartenTracing.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        return ActivitySource.CreateActivity(spanName, activityKind, parentActivity?.ParentId, tags);
```
gfoidl, line 23: Should this be
```suggestion
        return ActivitySource.StartActivity(spanName, activityKind, parentActivity?.ParentId, tags);
```
?

Otherwise an explicit `Start` needs to be done on the activity.

Or this method should be called `CreateActivity` too, else it's very misleading.

src/Marten/Internal/Sessions/EventTracingConnectionLifetime.cs
```diff
@@ -0,0 +1,286 @@
...
+
+internal class EventTracingConnectionLifetime :
+    IConnectionLifetime
+{
+    private const string MartenCommandExecutionStarted = "MartenCommandExecutionStarted";
+    private const string MartenCommandExecutionFailed = "MartenCommandExecutionFailed";
+    private const string MartenBatchExecutionStarted = "MartenBatchExecutionStarted";
+    private const string MartenBatchExecutionFailed = "MartenBatchExecutionFailed";
+    private const string MartenBatchPagesExecutionStarted = "MartenBatchPagesExecutionStarted";
+    private const string MartenBatchPagesExecutionFailed = "MartenBatchPagesExecutionFailed";
```
gfoidl, line 21: According [OTel semantics](https://opentelemetry.io/docs/specs/semconv/general/events/) these should have different names which we should follow.
```suggestion
    private const string MartenCommandExecutionStarted = "marten.command.execution.started";
    private const string MartenCommandExecutionFailed = "marten.command.execution.failed";
    private const string MartenBatchExecutionStarted = "marten.batch.execution.started";
    private const string MartenBatchExecutionFailed = "marten.batch.execution.failed";
    private const string MartenBatchPagesExecutionStarted = "marten.batch.pages.execution.started";
    private const string MartenBatchPagesExecutionFailed = "marten.batch.pages.execution.failed";
```

src/Marten/Internal/Sessions/EventTracingConnectionLifetime.cs
```diff
@@ -0,0 +1,286 @@
...
+
+        try
+        {
+            return _innerConnectionLifetime.Execute(cmd);
+        }
+        catch (Exception e)
+        {
+            if (_openTelemetryOptions.TrackConnectionEvents)
+            {
+                _databaseActivity.AddEvent(new ActivityEvent(MartenCommandExecutionFailed, DateTimeOffset.UtcNow, new ActivityTagsCollection(new KeyValuePair<string, object>[] {new KeyValuePair<string, object>("ExceptionType", e.GetType()) })));
```
gfoidl, line 78: There's an extension method `RecordException` (from OTel) that can be used here.

src/Marten/Internal/Sessions/EventTracingConnectionLifetime.cs
```diff
@@ -0,0 +1,286 @@
...
+
+        try
+        {
+            return _innerConnectionLifetime.ExecuteAsync(command, token);
+        }
+        catch (Exception e)
+        {
+            if (_openTelemetryOptions.TrackConnectionEvents)
+            {
+                _databaseActivity.AddEvent(new ActivityEvent(MartenCommandExecutionFailed, DateTimeOffset.UtcNow,
```
gfoidl, line 100: Same here for `RecordException`.
If that can't be used, refactor here and above to a helper-method.

src/Marten/Internal/Sessions/EventTracingConnectionLifetime.cs
```diff
@@ -0,0 +1,286 @@
...
+            _databaseActivity.AddEvent(new ActivityEvent("Database command execution started"));
+        }
+
+        try
+        {
+            return _innerConnectionLifetime.ExecuteReader(command);
+        }
+        catch (Exception e)
+        {
+            if (_openTelemetryOptions.TrackConnectionEvents)
```
gfoidl, line 124: The same. I'll stop to comment on these, but there are some more below.

src/Marten/Internal/Sessions/EventTracingConnectionLifetime.cs
```diff
@@ -0,0 +1,286 @@
...
+
+            throw;
+        }
+    }
+
+    public Task<DbDataReader> ExecuteReaderAsync(NpgsqlCommand command, CancellationToken token = default)
+    {
+        if (_openTelemetryOptions.TrackConnectionEvents)
+        {
+            _databaseActivity.AddEvent(new ActivityEvent("Database command execution started"));
```
gfoidl, line 141: Event name should be a constant too, and have a name following OTel semantics (see first comment).

src/Marten/Services/OpenTelemetryOptions.cs
```diff
@@ -0,0 +1,10 @@
+#nullable enable
+namespace Marten.Services;
+
+public sealed class OpenTelemetryOptions
+{
+    /// <summary>
+    /// Used to track OpenTelemetry events for a connection, for example when a command or data reader has been executed. This defaults to false.
+    /// </summary>
+    public bool TrackConnectionEvents { get; set; } = false;
```
gfoidl, line 9:
```suggestion
    public bool TrackConnectionEvents { get; set; }
```
`false` is the default by C# language specs, so can be omitted.

src/Marten/Services/SessionOptions.cs
```diff
@@ -93,6 +101,23 @@ internal IConnectionLifetime Initialize(DocumentStore store, CommandRunnerMode m
             throw new DefaultTenantUsageDisabledException();
         }
 
+        var innerConnectionLifetime =GetInnerConnectionLifetime(store, mode);
```
gfoidl, line 104:
```suggestion
        var innerConnectionLifetime = GetInnerConnectionLifetime(store, mode);
```

src/Marten/Services/SessionOptions.cs
```diff
@@ -93,6 +101,23 @@ internal IConnectionLifetime Initialize(DocumentStore store, CommandRunnerMode m
...
+
+        var corelationId = Activity.Current is not null ? Activity.Current.RootId : Guid.NewGuid().ToString();
```
gfoidl, line 106: When there's no activity that is recored (`Current = null`), is it OK to just create a new id?
I don't know what's best here.

jeremydmiller:
@SeanFarrow *One* combination of the LInq tests will fail periodically on CI, but always succeed on a subsequent run. I've never been able to reproduce it locally.

src/Marten/Services/SessionOptions.cs
SeanFarrow: As far as I understand things, if the current activity is null, this means you are the first activity in the chain, so creating a new activity/id is the right thing to do.

gfoidl: > if the current activity is null, this means you are the first activity in the chain

Being the first one in the chain means that there's no parent, but `Current` is set an `Activity`.

When there's no listeneres interested in an activity or no listeners are registered, then `ActivitySource.StartActivity` will return `null`, hence `Activity.Current` will be `null` too.

So when there's no activity I don't know if just creating a new (random) Id is correct.
Maybe in that case it should just be empty?

SeanFarrow: @jeremydmiller 
So should we set the current activity if it is null? or not start an activity at all?

src/Marten/Internal/Sessions/EventTracingConnectionLifetime.cs
Hawxy: That would require taking a hard dependency on OTEL. This implementation is also missing the stack trace.

src/Marten/Internal/Sessions/EventTracingConnectionLifetime.cs
```diff
@@ -0,0 +1,286 @@
...
+                _databaseActivity.AddEvent(new ActivityEvent(MartenCommandExecutionFailed, DateTimeOffset.UtcNow, new ActivityTagsCollection(new KeyValuePair<string, object>[] {new KeyValuePair<string, object>("ExceptionType", e.GetType()) })));
+            }
+
+            throw;
+        }
+    }
+
+    public Task<int> ExecuteAsync(NpgsqlCommand command, CancellationToken token = new CancellationToken())
+    {
+        if (_openTelemetryOptions.TrackConnectionEvents)
```
Hawxy, line 87: Why is this checked here if it's already checked when initializing this object?

src/Marten/Services/SessionOptions.cs
```diff
@@ -93,6 +101,23 @@ internal IConnectionLifetime Initialize(DocumentStore store, CommandRunnerMode m
...
+        var tags = new Dictionary<string, object>
```
Hawxy, line 107: Prefer `TagsList` when dealing with OTEL tags. This should also be lazy initialized only if `EventTracingConnectionLifetime` is being used.

src/Marten/Services/SessionOptions.cs
```diff
@@ -93,6 +101,23 @@ internal IConnectionLifetime Initialize(DocumentStore store, CommandRunnerMode m
...
+        {
+            { MartenCorelationId, corelationId },
+            { MartenTenantId, Tenant.TenantId }
+        };
+
+        return OpenTelemetryOptions.TrackConnectionEvents
+            ? new EventTracingConnectionLifetime(innerConnectionLifetime, OpenTelemetryOptions,
+                StartConnectionActivity(Activity.Current, tags))
+            : innerConnectionLifetime;
```
Hawxy, line 116: Ideally we only allocate the event tracing lifetime if there's any active listeners:
```suggestion
        return (OpenTelemetryOptions.TrackConnectionEvents && ActivitySource.HasListeners())
            ? new EventTracingConnectionLifetime(innerConnectionLifetime, OpenTelemetryOptions,
                StartConnectionActivity(Activity.Current, tags))
            : innerConnectionLifetime;
```

SeanFarrow:
OK, I think I've fixed everything, can people please check they are happy. I'll then look at tests.
@jeremydmiller  Do you have any opposition to using Testcontainers in a separate project to run prometheus?

src/Marten/Internal/Sessions/EventTracingConnectionLifetime.cs
```diff
@@ -0,0 +1,204 @@
...
+        var tagsToAdd = new ActivityTagsCollection(new[] { new KeyValuePair<string, object?>(MartenCommandFailedExceptionType, exceptionToRecord.GetType()) });
+        _databaseActivity?.AddEvent(new ActivityEvent(eventName, DateTimeOffset.UtcNow, tagsToAdd));
+    }
+
+    private void RecordExceptions(AggregateException exceptionsToRecord, string eventName)
+    {
+        var innerExceptionTypes = exceptionsToRecord.InnerExceptions.Select(t => t.GetType());
+        var tagsToAdd = new ActivityTagsCollection(new[] { new KeyValuePair<string, object?>(MartenCommandFailedExceptionTypes, innerExceptionTypes) });
+        _databaseActivity?.AddEvent(new ActivityEvent(eventName, DateTimeOffset.UtcNow, tagsToAdd));
+    }
```
Hawxy, line 203: I think it'd be more useful to align with OTEL standards in regards to exception reporting. Consider vendoring `RecordException` from OTEL.NET: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs#L81 into `MartenTracing`

AggregateExceptions can have the inner exceptions be individually recorded then.

SeanFarrow: That makes complete sense. The only question I have is whether we want to take a dependency on the OTel API NuGet package. From what I can see that doesn't depend on any other OTel packages, just the DiagnosticSource package. It would probably make sense if we are going to invest in OTel as a way of observing Marten. @jeremydmiller, what do you think?

SeanFarrow: @jeremydmiller  That was my main concern. There are stable versions of the API package, [version 1.8 was just released](https://www.nuget.org/packages/OpenTelemetry.Api).

I'll bring the RecordException method from this package in to Marten later today/tomorrow .

SeanFarrow: @Hawxy  I'll just bring that method and any required constants in then.
Should we be calling Stop on the activity we start in the dispose method?

gfoidl: >  Should we be calling Stop on the activity we start in the dispose method?

When the activity is disposed, it stops automatically / internally. So just make sure that the activity gets disposed (used within `using`, etc.).

jeremydmiller: I'm fine with the dependency then. Thanks for all the work on this!

SeanFarrow: @jeremydmiller  OK, I’ll pull in the OpenTelemetry.API package at the latest stable version, assuming that’s what you meant?
Do we want any automated tests in a separate test project, I think I can automate the Prometheus stuff using Docker and TestContainers.

gfoidl: Unit tests can be done with the `ActivityListener` from .NET.

OTel setup for exporting is IMO out of scope for Marten, and documented on their side. So I'd just go w/ some basic units tests for activities / tracing, and don't see a need to mess around with Prometheus (actually I think Jaeger would be better for tracing, as Prometheus isn't made for this).

Hawxy: https://github.com/JasperFx/alba/blob/master/src/Alba.Testing/ActivityTests.cs Alba has a basic test for OTEL as an example.

jeremydmiller:
@SeanFarrow Anything new on this? Is there anything you're blocked on?

SeanFarrow:
@jeremydmiller  I'm just adding tests. Things gone a bit mad at work as I've just switched teams.
I'll push what I've got later today. It's up to you whether you pick this last bit up.

I'll start the event/other stuff at the end of next week when things calm down.
Thanks,
Sean.

SeanFarrow:
Update: I should be code complete by tomorrow evening UK time. The only thing to then do is the docs.

Thanks,
Sean.

jeremydmiller:
@SeanFarrow You'd done a ton man, I'll take you up on some docs, but I'm good to pull this in sooner rather than later. I can easily deal w/ whatever CI hiccups are happening.

SeanFarrow:
@jeremydmiller  I'm done and happy.
Please check you are happy with the test for the EventTracingConnectionLifetime class.
Other than that it's just docs and the CI issues.
I should be able to help with the event document listener in a couple of weeks if you don't get to it first.

Thanks.
Sean

jeremydmiller:
> @jeremydmiller I'm done and happy. Please check you are happy with the test for the EventTracingConnectionLifetime class. Other than that it's just docs and the CI issues. I should be able to help with the event document listener in a couple of weeks if you don't get to it first.
> 
> Thanks. Sean

Cool, and thanks Sean! I'll try to pull this in in the next couple days. Might be until next week before 7.9 gets out with this though.

Issue #3083, opened 2024-03-24
Duplicate Type Generation


felixkmh:
I ran into a memory issue that ultimately comes from duplicate code generation for document types. Because certain events were processed in parallel in batches of size **n**, when the first batch was processed, code was generated for each type up to **n** times and loaded into the AppDomain, causing the application to consume multiple gigabytes of memory in some cases.

I could fix this behavior by registering the types upfront and setting the code generation mode to `Auto`. But my suggestion would be to make `ProviderGraph.cs` thread-safe such that code generation is guaranteed to only occur once per type.

jeremydmiller:
What version of Marten? We've fought this one several times already

felixkmh:
Oh, sorry for not mentioning that. It first occured on 6.0.4, but also 6.4.1 did not seem to fix this issue.

jeremydmiller:
Okay. All I'm going to do is the exact same thing, so if you're interested, "I take pull requests"

felixkmh:
I'll take you up on that. I would like to check whether it would be feasable to have one lock per type though. If not, I'll just go with the safer option.

jeremydmiller:
Stick the lock if anything in GeneratedProjection.CodeGen, not in ProviderGraph? that's roughly how we did the same kind of locking for Wolverine for similar issues

felixkmh:
My understanding was that multiple threads can walk over `if (_storage.TryFind(documentType, out var stored)) { ... }`  for the same `documentType` here:
```csharp
public DocumentProvider<T> StorageFor<T>() where T : notnull
{
        var documentType = typeof(T);

        if (_storage.TryFind(documentType, out var stored))
        {
            return stored.As<DocumentProvider<T>>();
        }
        // Generate code (possibly time intesive) and then update _storage
}
```
and enter the code generation part (i.e. `builder.InitializeSynchronously(rules, _options, null);`) if they run in parallel, because `_storage` is only updated after the code generation. That's why I considered to lock in there. Am I on the wrong track here?

felixkmh:
I put together a script to illustrate my idea:

<details>
<summary>Script</summary>

```csharp
using System.Collections.Concurrent;
using System.Threading;
using System.Diagnostics;

ConcurrentDictionary<Type, object> _storage = new();
Dictionary<Type, object> typeLockObjects = new();

public object GetTypeLock(Type type)
{
    lock(typeLockObjects)
    {
        // return existing lock object or generate a new one if none exists
[...]

PR #3109, opened 2024-04-02
Locking in ProviderGraph to prevent duplicate code generation

felixkmh:
- [x] Implemented fix as discussed in #3083 
  - Added single lock object to synchronize code generation
- [x] Added unit test to cover this fix
- [x] Confirm that the project builds
- [x] All automated tests succeed

### DIFF
--- a/src/CoreTests/Bugs/Bug_3083_concurrent_type_generation.cs
+++ b/src/CoreTests/Bugs/Bug_3083_concurrent_type_generation.cs
@@ -3,6 +3,7 @@ using System.Threading.Tasks;
 using Marten;
 using Marten.Internal;
 using Marten.Internal.Storage;
+using Marten.Schema.BulkLoading;
 using Marten.Testing.Harness;
 using Shouldly;
 using Xunit;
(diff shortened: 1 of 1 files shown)


src/Marten/Internal/ProviderGraph.cs
```diff
@@ -35,70 +39,78 @@ public ProviderGraph(StoreOptions options)
             return stored.As<DocumentProvider<T>>();
         }
 
-        if (documentType == typeof(IEvent))
+        lock (_storageLock)
```
jeremydmiller, line 42: Could you extract a method for anything happening within that lock? Just for code tidy-ness.

src/Marten/Internal/ProviderGraph.cs
```diff
@@ -23,7 +24,10 @@ public ProviderGraph(StoreOptions options)
 
     public void Append<T>(DocumentProvider<T> provider) where T : notnull
     {
-        _storage = _storage.AddOrUpdate(typeof(T), provider);
+        lock (_storageLock)
```
jeremydmiller, line 27: Don't do the lock here, it's unnecessary w/ the immutable hash map

jeremydmiller:
@felixkmh Thanks for taking this on!

src/Marten/Internal/ProviderGraph.cs
felixkmh: I'm not sure how `ImHashMap`s work exactly, but the reason I also locked it is because I thought that if two threads enter this method concurrently, then they call `AddOrUpdate` on the same instance. Then they each would create a new map based on that same map instance, resulting in two separate instances that do not contain the item the other thread tried to add. The assignment is atomic, so whoever assigns to `_storage` last "wins".

felixkmh: Maybe I'll just add that as another test case and see if that can happen. Although tests of that nature aren't the most reliable.

felixkmh: Seems that it is at least possible:
image
I can't speak to how likely that is in practice.

PR #3112, opened 2024-04-03
Support NpgsqlMultiHostDataSource

Hawxy:
Requires https://github.com/JasperFx/weasel/pull/126

This is a draft for the time being to see if the design is suitable. I've tried to achieve this with the least amount of code changes possible. Docs yet to be written.
Some considerations:
- `NpgsqlMultiHostDataSource` registers itself in the DI container as a normal data source, so marten won't explicitly know it's a multi host source until it tries to fetch a connection from it. If marten has no defaults for multi host usage then the data source will return _any_ connection type and potentially cause unexpected errors as it tries to write against a read replica. For this reason, I've baked in some sane fallbacks for marten & weasel. 
- If a `NpgsqlMultiHostDataSource` is present we won't target the replica by default, it's up to the user to weigh up the potential pitfalls of replica lag.
- I'm only targeting user queries for this functionality. All internal queries including daemon operation will remain pointed at the primary node for reliability. (although for distributed projections/rebuilds there might be performance benefits to using the replicas)
- Does Wolverine have support for a user-supplied data source yet? The fix for `Wolverine.Postgres` is just to get users to only supply a datasource capable of creating primary node connections `dataSource.WithTargetSession(TargetSessionAttributes.Primary);`
- This implementation only supports the `AutoClosingLifetime`. I'm unsure if it's practical to support the others just yet.

### DIFF
--- a/docs/configuration/hostbuilder.md
+++ b/docs/configuration/hostbuilder.md
@@ -129,7 +129,7 @@ services.AddMarten()
     .UseLightweightSessions()
     .UseNpgsqlDataSource();
 ```
-<sup><a href='https://github.com/JasperFx/marten/blob/master/src/CoreTests/MartenServiceCollectionExtensionsTests.cs#L291-L299' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_using_usenpgsqldatasource' title='Start of snippet'>anchor</a></sup>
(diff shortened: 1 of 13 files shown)


jeremydmiller:
@Hawxy Sorry this has taken so long. I'm good with that as is. I can do the work to pull over the Weasel changes any time, or I can make sure you have all the rights if you wanna do that

Issue #3158, opened 2024-04-24
Tombstone process is generating multiple errors due to implementation issue 


svrx:
**Context**
Recently upgraded to 6.4.1 and noticed that error rate has increased significantly for the same workloads.
Upon investigation found that most of the errors aren't representative of a functional issue but rather originated by an issue in the tombstone process triggered when a concurrency version error occurs, which requires tombstones to be created.

The issue seem to have been introduced by the changes  cb232d79cd29765cd0b0a3ae17189487d8c081f8 and  39288c2c7b03c23fcd7d900c518017ed414ba15a.

**Issue**
1. When tombstones are posted a new `LightWeightSession()` is created with default isolation level of `ReadCommited`, which as per implementation **doesn't** start a transaction underneath.
2. On calling `ExecuteBatch()` for the tombstone batch, on the new session, the command is executed and committed  automatically 
3. `ExecuteBatch()` assumes the session has a transaction and calls `_connection.Commit()` getting immediately an exception `InvalidOperationException("Trying to commit a transaction that was never started")` 
4. The applicable retry policy may retry the `Commit()`  an additional number of times which subsequently posts the batch to the `_tombstones` (`RetryBlock<>`) which tries another 3 times deferred in time.

```c#
// marten\src\Marten\Events\EventGraph.Processing.cs
internal void PostTombstones(UpdateBatch tombstoneBatch)
{
    try
    {
        using var session = (DocumentSessionBase)_store.LightweightSession();
        session.ExecuteBatch(tombstoneBatch);
    }
 ...

// marten\src\Marten\Internal\Sessions\DocumentSessionBase.SaveChanges.cs
internal void ExecuteBatch(IUpdateBatch batch)
{
    try
    {
        batch.ApplyChanges(this);
        _retryPolicy.Execute(_connection.Commit);
    }
...
``` 

In all this try and retry process, none of the retries has any chance of success, since the same issue is present, although on step 2 the desired effect has already been applied.

Even though has been identified in 6.4.1, it looks to still be present in the latest version of master.

jeremydmiller:
@svrx "Even though has been identified in 6.4.1, it looks to still be present in the latest version of master."

I'm going to contest that statement and tell you to just move up to Marten 7. The entire execution pipeline is significantly different in Marten 7, and the issues you're calling out here would not exist there.

I'll accept a pull request to address this in the 6.* branch if you'd like, but otherwise my advice is to move to 7. 

[...]

PR #3160, opened 2024-04-24
Ensures transaction is initialized when executing a batch avoiding unnecessary errors and churn.

svrx:
As discussed in #3158.


jeremydmiller:
Looks good to me, I'll get it turned around by maybe lunch tomorrow

PR #3164, opened 2024-04-27
Support for query.Include mapped by any property

elexisvenator:
Currently `query.Include()` only supports including related documents by their identity.  This PR extends the functionality to support joining between any two values on the documents.

To support this, the syntax for creating an include has changed slightly. Now the `Include()` method only requires the callback/list/dictionary that the included document will be handled by. This is then followed by an `On()` method (named after sql `JOIN ... ON ...`) which allows you to specify a idSource, an optional idMapping, and an optional filter.

- `Expression<Func<T, object>> idSource`: When using an overload with no `idMapping`, this behaves exactly like current Includes and lets you specify the property to map as the related documents identity.
- `Expression<Func<T, TId?>> idSource`: When using an overload that includes `idMapping`, this allows you to specify and property on the queried document. The type should be a value type but apart from that anything can be mapped.
- `Expression<Func<TInclude, TId?>> idMapping`: Specify what property on the related document should be used for mapping. It is recommended that this field be indexed.  The type of the field must match the type of the property mapped on the queried document.
- `Expression<Func<TInclude, bool>> filter`: Filtering, no changes here.

The fluent approach also removes the need for specifying types in some cases, particularly around dictionaries.

Some examples:

```csharp
// old
query.Include<User>(a => a.RefId, x => included = x);
query.Include(a => a.RefId, includeList);
query.Include(a => a.RefId, includeDict);

// new
query.Include<RefDoc>(x => included = x).On(x => a.RefName, r => r.Name);
query.Include(includeList).On(x => a.RefName, r => r.Name);
query.Include(includeDict).On(x => a.RefName, r => r.Name);
```

The old methods of `.Include(idSource, callback [, filter])` all still exist and behave the same, meaning this is not a breaking change. (should these be marked obsolete?).

The actual code for the new functionality is pretty small, and there are a lot of tests to cover it.

Other changes:

- Added support for Include filters to batched queries.
- Because one-to-many joins are now possible, added a version of `Include` that takes a `Dictionary<TKey, IList<TInclude>>`. This will populate the dictionary with lists of all matching documents for the given key.
- Added to the Include documentation, and also fixed some incorrect parts of the current doco.
- Added a missing `;` to the end of the create-temp-table sql. This wasn't a bug unless you tried to manually execute the command text.
- Fixed bug where the session's TenantId was not always added to the CommandBuilder. This caused the `.ToCommand()` method to sometimes fail in multitenanted queries. 
- Changed some of the include tests to use `IntegrationContext` instead of `OneOffConfigurationContext` which sped up their execution.


src/Marten/Linq/MartenLinqQueryable.cs
```diff
@@ -222,49 +238,14 @@ public Task<int> StreamJsonArray(Stream destination, CancellationToken token)
...
-
+    
     public NpgsqlCommand ToPreviewCommand(FetchType fetchType)
     {
         var parser = new LinqQueryParser(MartenProvider, Session, Expression);
 
         var command = new NpgsqlCommand();
 
-        var sql = new CommandBuilder(command);
+        var sql = new CommandBuilder(command) {TenantId = Session.TenantId};
```
elexisvenator, line 248: Fix for errors when calling `ToCommand()` on multi-tenant queries.

src/Marten/Linq/Includes/TemporaryTableStatement.cs
```diff
@@ -30,6 +29,7 @@ protected override void configure(ICommandBuilder sql)
     {
         sql.Append("drop table if exists ");
         sql.Append(ExportName);
+        sql.Append("; ");
```
jeremydmiller, line 32: You no longer need the semicolon between commands with the newer NpgsqlBatch usage.

PR #3191, opened 2024-05-10
Add failing test for not increasing version

e-tobi:
Not sure if this is a bug, but shouldn't the version increase with each saved change to a document?


jeremydmiller:
@e-tobi Long story short, use UpdateRevision() here. You're going to get a better exception message explaining that now. Not a recommended usage to repeatedly call Store() on the same document in the same session.

jeremydmiller:
This is going to get changed to warn about this. Use UpdateRevision()

Issue #3198, opened 2024-05-14
Referencing nullable types in OData LINQ queries results in an InvalidOperationException being thrown


iress-ljm:
Attempting to perform a query using Marten Queryable and ASP.NET OData 8 and seeing the following exception when a nullable type is referenced in the OData $filter: 

```
System.InvalidOperationException: variable '$it' of type 'Api.Model.Order' referenced from scope '', but it is not defined
   at System.Linq.Expressions.Compiler.VariableBinder.Reference(ParameterExpression node, VariableStorageKind storage)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitParameter(ParameterExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitMember(MemberExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitBinary(BinaryExpression node)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitUnary(UnaryExpression node)
   at System.Linq.Expressions.ExpressionVisitor.Visit(ReadOnlyCollection`1 nodes)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitLambda[T](Expression`1 node)
   at System.Linq.Expressions.Compiler.LambdaCompiler.Compile(LambdaExpression lambda)
   at System.Linq.Expressions.Expression`1.Compile()
   at FastExpressionCompiler.ExpressionCompiler.CompileSys[TDelegate](Expression`1 lambdaExpr) in /_/src/FastExpressionCompiler/FastExpressionCompiler.cs:line 171
   at FastExpressionCompiler.ExpressionCompiler.CompileFast[R](Expression`1 lambdaExpr, Boolean ifFastFailedReturnNull, CompilerFlags flags) in /_/src/FastExpressionCompiler/FastExpressionCompiler.cs:line 193
   at Marten.Linq.Parsing.LinqInternalExtensions.ReduceToConstant(Expression expression)
   at Marten.Linq.Parsing.SimpleExpression..ctor(IQueryableMemberCollection queryableMembers, Expression expression)
   at Marten.Linq.Parsing.WhereClauseParser.VisitBinary(BinaryExpression node)
   at Marten.Linq.Parsing.WhereClauseParser.Visit(Expression node)
   at Marten.Linq.SqlGeneration.Statement.ParseWhereClause(IReadOnlyList`1 wheres, IMartenSession session, IQueryableMemberCollection collection, IDocumentStorage storage)
[...]

PR #3201, opened 2024-05-14
GH3198: add example odata API to recreate #3198

iress-ljm:
* Adds an example OData API which exposes an IMartenQueryable to enable queries to execute against the database
* Adds a docker-compose launch project which runs compose up on the API and postgres


jeremydmiller:
@iress-ljm I'm happy to take this as is if you want, but I'll remove your new docker compose file and use the existing Marten file for the same. I'll add Alba specs before going any farther to the actual fix

jeremydmiller:
Got this with a rebase, thanks @iress-ljm!

PR #3207, opened 2024-05-15
Allow initial System.Text.Json.JsonSerializerOptions to be a copy

### LINKED ISSUE #3206: Allow initial System.Text.Json.JsonSerializerOptions to be a copy from existing options
For most of our applications we are used to configuring a single `JsonSerializerOptions` instance that is reused everywhere. For consistency we would also like to use the same configuration for Marten, but currently it is not able to provide a `JsonSerializerOptions` instance that is used to configure the serializer, instead it is only possible to provide a configuration function.

My suggestion would be to introduce a new `SystemTextJsonSerializer` constructor that accepts a `JsonSerializerOptions` instance that is passed as argument to the constructors for `_clean`, `_options`, `_optionsDeserialize` and `_withTypes`. This ensures that the settings configured provided by this base `JsonSerializerOptions` are copied to the initial versions of the 4 internal options.

I'm not entirely sure if this is the cleanest solution, but it does the job of enabling us to provide a pre-configured `JsonSerializerOptions` instance.

CptWesley:
Resolves #3206

### DIFF
--- a/docs/configuration/storeoptions.md
+++ b/docs/configuration/storeoptions.md
@@ -328,5 +328,5 @@ var store = DocumentStore.For(_ =>
     _.NameDataLength = 100;
 });
 ```
-<sup><a href='https://github.com/JasperFx/marten/blob/master/src/CoreTests/StoreOptionsTests.cs#L313-L322' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_setting-name-data-length' title='Start of snippet'>anchor</a></sup>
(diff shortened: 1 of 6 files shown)


jeremydmiller:
@CptWesley Got it, looks fine when the build goes green. This will be in 7.13. I don't have any set time for that offhand

CptWesley:
The failing test seems flaky, it passes locally most of the time, but fails occasionally. Not sure if that was a known issue.

jeremydmiller:
Known aggravation. Only fails on CI, and very rarely outside of that one specific combo in the CI matrix

PR #3213, opened 2024-05-16
TeardownDataOnRebuild = false prevents deleting existing projections

AlexZeitler:
It's a repro for this discussion on Discord.

### DIFF
--- a/src/DaemonTests/multistream_data_teardown_on_rebuild.cs
+++ b/src/DaemonTests/multistream_data_teardown_on_rebuild.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Marten.Events.Projections;
@@ -21,6 +20,19 @@ public class ImplementationB: Base
 {
 }
 
+public class Base2
+{
+    public string Id { get; set; }
+}
+
+public class ImplementationA2: Base2
+{
+}
+
+public class ImplementationB2: Base2
+{
+}
+
 public class SomethingHappened
 {
     public Guid Id { get; set; }
@@ -31,7 +43,7 @@ public class ImplementationAProjection: MultiStreamProjection<ImplementationA, s
     public ImplementationAProjection()
     {
         Identity<SomethingHappened>(e => $"a-{e.Id}");
-        TeardownDataOnRebuild = false;
+        Options.TeardownDataOnRebuild = false;
     }
 
     public static ImplementationA Create(
@@ -44,7 +56,7 @@ public class ImplementationBProjection: MultiStreamProjection<ImplementationB, s
     public ImplementationBProjection()
     {
         Identity<SomethingHappened>(e => $"b-{e.Id}");
-        TeardownDataOnRebuild = false;
+        Options.TeardownDataOnRebuild = false;
     }
 
     public static ImplementationB Create(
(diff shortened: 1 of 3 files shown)


AlexZeitler:
When commenting out [this line](https://github.com/JasperFx/marten/blob/master/src/Marten/Events/Daemon/ProjectionDaemon.Rebuilding.cs#L117), the test succeeds.

AlexZeitler:
Would it make sense to add `TeardownDataOnRebuild` to `AsyncOptions` and make the call to `teardownExistingProjectionProgress` ([here](https://github.com/JasperFx/marten/blob/2f0c5b40786743de62ed6ccc157612307748fcf8/src/Marten/Events/Daemon/ProjectionDaemon.Rebuilding.cs#L117)) conditional based on the new property?

I've added it locally and set the option when adding the projection and my test then succeeded.

But it also broke some other tests.

AlexZeitler:
Digging into the history, I noticed that `TeardownDataOnRebuild` started to be available starting with Commit 3a9ba4e435d38dd366e3a93f54d9d0c6f50f2e72, but it already broke with the next one f7642cdfe5d0fd1399ac9ff7a3245e809e7b8ef8.

The test I ran is similar to the one in this PR but it's using `MultiStreamAggregation` (before it has been refactored to `MultiStreamProjection`.

```csharp
using System;
using System.Threading;
using System.Threading.Tasks;
using Marten.Events.Projections;
using Marten.Testing.Harness;
using Xunit;

namespace Marten.AsyncDaemon.Testing;

public class Base
{
    public string Id { get; set; }
}

public class ImplementationA: Base
{
}

public class ImplementationB: Base
{
}

public class SomethingHappened
{
    public Guid Id { get; set; }
}

public class ImplementationAProjection: MultiStreamAggregation<ImplementationA, string>
{
    public ImplementationAProjection()
    {
        Identity<SomethingHappened>(e => $"a-{e.Id}");
        TeardownDataOnRebuild = false;
    }

    public static ImplementationA Create(
        SomethingHappened e
    ) => new() { Id = $"a-{e.Id}" };
}

public class ImplementationBProjection: MultiStreamAggregation<ImplementationB, string>
{
    public ImplementationBProjection()
    {
        Identity<SomethingHappened>(e => $"b-{e.Id}");
        TeardownDataOnRebuild = false;
    }

    public static ImplementationB Create(
        SomethingHappened e
    ) => new() { Id = $"b-{e.Id}" };
}

public class MultiStreamDataTeardownOnRebuildTests: OneOffConfigurationsContext
{
    [Fact]
    public async Task Adheres_to_disabled_teardown_on_rebuild()
    {
        StoreOptions(
            opts =>
            {
                opts.Schema.For<Base>()
                    .AddSubClass<ImplementationA>()
                    .AddSubClass<ImplementationB>();

                opts.Projections.Add<ImplementationAProjection>(ProjectionLifecycle.Inline);

                opts.Projections.Add<ImplementationBProjection>(ProjectionLifecycle.Inline);
            }
        );

        var commonId = Guid.NewGuid();

        using var daemon = await theStore.BuildProjectionDaemonAsync();
        await using var session = theStore.LightweightSession();

        session.Events.StartStream(
            commonId,
            new SomethingHappened() { Id = commonId }
        );

        var implementationA = await session.LoadAsync<ImplementationA>($"a-{commonId}");
        var implementationB = await session.LoadAsync<ImplementationB>($"b-{commonId}");

        implementationA.ShouldNotBeNull();
        implementationB.ShouldNotBeNull();

        await daemon.RebuildProjection<ImplementationAProjection>(CancellationToken.None);

        var implementationBAfterRebuildOfA = await session.LoadAsync<ImplementationB>($"b-{commonId}");

        implementationBAfterRebuildOfA.ShouldNotBeNull();
    }
}
``

AlexZeitler:
After a few random checks in the history up to the current status, I am tempted to say that it has never worked again apart from the first commit.

AlexZeitler:
Turns out, that the failing tests also failed without my changes.

Before applying my changes:
image

After applying my changes:
image

AlexZeitler:
As can be seen in the tests, the fix doesn't use the existing `TeardownDataOnRebuild` property from `ProjectionBase` [but](https://github.com/JasperFx/marten/pull/3213/files#diff-8375c9e67cd19659d7237cd252b419a6729f8906d0a68f289fae4dfd3882c80aR46) `Options.TeardownDataOnRebuild` from `AsyncOptions` on `IProjectionSource`.

Having `TeardownDataOnRebuild` in `AsyncOptions` it can be set either in the Projection implementation or [in the registration](https://github.com/JasperFx/marten/pull/3213/files#diff-8375c9e67cd19659d7237cd252b419a6729f8906d0a68f289fae4dfd3882c80aR153).

Issue #3257, opened 2024-06-11
Overriding Identity Types + Overriding Property Types in queries.


nkosi23:
Another hello from the F# world :)

In F# it is common to wrap identity types in Single Case discriminated unions such as:

```fsharp
type OrderId = Id of Guid
```

The goal and culture are to encode the domain and business rules in the type system as much as possible to leverage the power of compile-time checks to design and evolve models safely. Many folks are also starting to do it in C# (as part of DDD), but it is much more common in F# due to the ease and conciseness of designing such types.

The above type is not exactly a Guid, it is instantiated like so: `OrderId.Id(Guid.NewGuid())`. The most commonly used F# to JSON serializer unwraps such types and stores them as plain strings. Therefore the problem is not with the persistence format, but with the .NET type constraints introduced by Marten (ie. trying to use such a type as a primary key throws an exception, as explained in the documentation)

In this context, it would be very useful to be able to instruct Marten to override the type of primary keys, but also of properties in queries. 

First of all by adding the following overload to the identity function to have something like the below to tell Marten that it should consider that the primary key has the provided generic type in the database:

```csharp
storeOptions.Schema.For<Order>()
        .Identity<Order,Guid>(fun x -> x.Id)
```

Then another thing that could be useful is a Where<T1,T2> overload on IMartenQueryable so that we can do something like the below, for example when querying an Invoice document having an OrderId property (OrderId cannot be converted to Guid by default. Moreover, even if such a cast could be translated to SQL by the Linq provider, it would be quite cumbersome to define casts explicitly on all such types - especially since they are supposed to be one-liners):

```csharp
documentSession.Query<Invoice>()
      .Where<Invoice,Guid>(x => x.OrderId, x => x == orderId)
```

The first lambda would be used to select the property, while the second lambda would be the one actually being turned into SQL using the type of the second generic parameter. 

I think this would be useful beyond F#, as it would give users full flexibility to decouple the JSON representation of data from their representation in .NET. It would empower users to model their domain in the way making the most sense for them from a business perspective and use custom JsonConverters for serialization. These two methods would enable quite a lot of use cases.

What are your thoughts on this idea?

haefele:
Seems very similar to the Typed-ID work from jeremy the last days.
See [these discord messages](https://discord.com/channels/1074998995086225460/1074999076896112661/1249762911396560927) and [this branch](https://github.com/JasperFx/marten/tree/value-types).

But could also be different situation because of F#.

[...]

PR #3260, opened 2024-06-13
[WIP] Overriding Identity Types + Overriding Property Types in queries

nkosi23:
Implements the feature request in #3257 (work in progress ~need to do the same for queries~)
Feedback would be very welcome


jeremydmiller:
@nkosi23 Hey man, thanks for trying this, but I'm really not enthusiastic about this approach at all.

PR #3269, opened 2024-06-18
Expose Revision column in MetadataConfig to allow custom mapping

dystopiandev:


jeremydmiller:
@dystopiandev Hey, at a bare minimum, this is going to need a test of this actually being used

dystopiandev:
> @dystopiandev Hey, at a bare minimum, this is going to need a test of this actually being used

@jeremydmiller  Does c8a21ee suffice?

dystopiandev:
@jeremydmiller It is done

PR #3279, opened 2024-06-21
F# discriminated union support for strongly-typed identifiers

nkosi23:
Implements #3280


jeremydmiller:
@nkosi23 Can you reverse the rename of `ValueTypeInfo` to `StrongTypedIdInfo`? The earlier name was to reflect that that was used within LINQ and possibly independently of the strong typed ids

PR #3297, opened 2024-07-08
fix: Tenanted bulk loading should consider tenant id as part of document identity

elexisvenator:
This fixes 2 issues:

- A massive performance issue where not joining or updating on tenant id meant that no index could be used
- The bug where if the same Id exists in multiple tenants then an update would happen instead of an insert.

Also has a small tweak to calling binay import to use the async method where appropriate.

### DIFF
--- a/src/DocumentDbTests/Writing/bulk_loading.cs
+++ b/src/DocumentDbTests/Writing/bulk_loading.cs
@@ -263,6 +263,33 @@ public class bulk_loading_Tests : OneOffConfigurationsContext, IAsyncLifetime
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


elexisvenator:
I also have an extension of this branch that changes the logic to use a single `insert .. on conflict do update` to make the merge quicker.

elexisvenator:
Another bug that I haven't addressed as its more programmer error: There are no checks in place to ensure that a tenant is specified when the bulk imported document uses conjoined tenancy.  Tenancy defaults to default but given how low level this op is it's probably worth requiring the use of default tenant to be explicit.

PR #3301, opened 2024-07-09
Use `insert .. on conflict` for bulk import operations

elexisvenator:
Updates the behaviour of `BulkInsertMode.OverwriteExisting` to use `insert ... on conflict do update`. This is one less db statement than before and a less prone to constraint issues.
Also updates the behaviour of `BulkInsertMode.IgnoreDuplicates` to use `insert ... on conflict do nothing`. 

(Wasn't going to create this so soon but ran into an issue where spamming some seed data resulted in unique constraint violations when using `OverwriteExisting`) 

Tests pass but I had to delete the checked in generated code files. I am not sure what the correct way to regenerate these files is - can this be added to the doco somewhere so I can fix the issue myself next time 🙇 .


jeremydmiller:
I'm going to pull this down locally and try it first. Can't see any reason why this would fail offhand

elexisvenator:
Hi @jeremydmiller,

I have given this another pass. Bulk inserting now supports conditional overwrites - when in `BulkLoadMode.OverwriteExisting` it is now possible to pass in a condition that must pass before a document is overwritten. This is very useful for bulk loading from external sources as it allows for safe out-of-order imports as long as documents have some sort of versioning column. Previously the only way to do this was to first query all the documents you wanted to import and drop the ones that were out of date. This way should be a lot faster.

jeremydmiller:
@elexisvenator I'm sorry we didnt' act on this one earlier, but it's way past stale.

PR #3308, opened 2024-07-14
Add event projection failure scenario when using "Update"

Hawxy:
Unsure if this is related to community reported issues, but whilst building out a potential reproduction I ran into this. `ops.Store` works fine but `ops.Update` fails. The observations I've made is that the operations & batch execution appear in the correct order. The last operation in a batch ends up throwing a `Marten.Exceptions.NonExistentDocumentException`


jeremydmiller:
@Hawxy I'd rather beat that with documentation that strongly suggests always using Store() inside of projection code.

Issue #3307, opened 2024-07-13
Bulk events performance experiments


elexisvenator:
This is the result of a series of tests I have been doing on the performance of bulk operations with events.

## The tests:

THe branch I am running these experiments on is https://github.com/JasperFx/marten/compare/master...elexisvenator:marten:bulk-event-experiments

I am testing the following scenarios:

1. Insert a single stream with 1000 events
2. Update a single stream with 1000 events
3. Insert 1000 streams with 1 event each
4. Update 1000 streams with 1 event each

For all of these tests, there are 2 inline projections - 1 that uses the event inserted, and 1 that doesn't. Additionally multi tenancy is on, the stream identity is a string, and quick append is enabled.  With this configuration **scenarios 3 and 4 are an exact match for my real-world use cases relating to bulk import/update**.

From a benchmarking perspective, these tests are not at all scientific. I am using them to try to identify low hanging fruit for performance improvements, but as the obvious fixes are dealt with these tests will need to be turned into proper benchmarks. For this reason, no specific durations are mentioned below.

## The findings

* As is fairly obvious, 1000 events in 1 stream is way more performant than 1 event in 1000 streams.
* Updates are slightly slower than inserts, but the difference is tiny compared to the previous points performance difference.
* Pretty sure the function `mt_quick_append_events` function will not compile if single-tenant mode is enabled. It doesn't affect these tests but it's something to consider.
* Inline projections will fetch the current aggregate to apply events to. One improvement is to skip this for `ISingleStreamSlicer` when it is known that the stream is a new stream. This is done with d2f84428789bcc054264266457b0921e9c68d00d and is a significant performance boost to scenario 3, bringing its performance almost on par with scenario 1.
* When using quick append, **revision numbers on inline projections are (very) broken**. This is because:
  * Revision numbers are based on the version of the last handled event. When updating a stream with quick append, this number is always 1 (unless using optimistic appends or an `expectedVersion` is provided). This causes the inline projection's revision to be set to 1
  * If an inline projection's revision is less than what already exists in the database, then an insert silently fails 😱 
  * My first thought to fix this (which I noted in in the commit linked above) is to infer the new version from the inline projection's current version. This should be possible as the information is already fetch, just not readily available in the current code.  I have since realized that this will not work due to a) it requires the projection version to remain consistent with the stream, with no tolerance for drift, and b) the projection version already drifts by design whenever the new events in a stream are not ones processed by the projection (because the projection processing is skipped).
[...]

PR #3310, opened 2024-07-16
bug: Inline projections don't have correct version when appending to an existing stream (quickappend)

elexisvenator:
Found as part of https://github.com/JasperFx/marten/issues/3307, Inline projections take their revision from the version of the last event appended to a stream, but for quickAppend this is not a valid version.

This is the tests I was using for performance, one of them is failing due to this issue, so the rest can be removed if you dont care about them. The size of the streams / iterations are scaled way back to make the tests fast.


jeremydmiller:
Got it with some merge/rebase clumsiness. @elexisvenator Thanks for doing all the work to find this issue and to repro it!

PR #3316, opened 2024-07-17
Change IList -> List in Public API

Hawxy:
`IList` is a bit of a miserable type as it's missing `.AddRange`. One of least useful interfaces to expose and I try to avoid it.


jeremydmiller:
Pfft, thumbs up from me. I've stopped using `IList<T>` in my code after we got the typed target new() stuff anyway. And ditto to the annoyance

PR #3329, opened 2024-07-22
Eliminate Npgsql value boxing within codegen

Hawxy:
Requires https://github.com/JasperFx/weasel/pull/136


jeremydmiller:
@Hawxy I got this with a local rebase

PR #3333, opened 2024-07-25
Do not advance reader on callbacks from appending events or inserting…

jeremydmiller:
… a new stream


jeremydmiller:
This is obsolete

PR #3343, opened 2024-07-28
Fix issues with deletions and updates in dirty tracked session

e-tobi:


src/Marten/Internal/Sessions/DocumentSessionBase.Deletes.cs
```diff
@@ -26,6 +27,7 @@ public abstract partial class DocumentSessionBase
         _workTracker.Add(deletion);
 
         documentStorage.Eject(this, entity);
+        ChangeTrackers.RemoveAll(x => ReferenceEquals(entity, x.Document));
```
jeremydmiller, line 30: You've got to do that by document identity and not just the reference equals

e-tobi: Why? Eject does the same:

https://github.com/JasperFx/marten/blob/master/src/Marten/Internal/Sessions/DocumentSessionBase.cs#L256

And there is this method, which I think isn't used anymore:

https://github.com/JasperFx/marten/blob/master/src/Marten/Internal/Sessions/DirtyCheckingDocumentSession.cs#L56

jeremydmiller: Because what if someone passes the session around and passes a different object with the same id in to delete it *which people have absolutely done before even though it sounds weird*. What if the document is immutable -- which people absolutely do as well -- so they might have modified the document and hence created a new document that they then passed into the Delete() method.

e-tobi: Ok, I get your point. But this should then be done for Eject as well, shouldn't it?

jeremydmiller: DocumentStorage.Eject() does it by identity

jeremydmiller:
@e-tobi I have a magic window of time today to push a Marten release. I'll take it from here. Thanks for doing this!

Issue #3351, opened 2024-08-01
Include does not work after migrating to v7.26.1 from v6


bhosale-ajay:
Two entities

```
User 
{
 public string? Id { get; set; }
 public string? Name { get; set; }
}
UserInfo
{
    public string? Id { get; set; }
    public string SomeThing { get; set; }
}
```

Both are built out of projections

```
var userInfos = new Dictionary<string, UserInfo>();
var queryable = _session.Query<User>()
    .Where() // some where conditions   
    .Include(userInfos).On(x => x.Id!) // This syntax got changed in V7
    .ForAllTokens()
    .ForSomeTokens()
   .OrderBy() // Some ordering
var list = await queryable
                .ToPagedListAsync(paging.CurrentIndex, paging.PageSize, ct);
```

The last line throws below error.

```
Marten.Exceptions.MartenCommandException: Marten Command Failure:$
42703: column d.id does not exist

POSITION: 72
 ---> Npgsql.PostgresException (0x80004005): 42703: column d.id does not exist

POSITION: 72
   at Npgsql.Internal.NpgsqlConnector.ReadMessageLong(Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
   at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
   at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
   at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(Boolean async, CommandBehavior behavior, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(Boolean async, CommandBehavior behavior, CancellationToken cancellationToken)
   at Marten.Internal.Sessions.AutoClosingLifetime.ExecuteReaderAsync(NpgsqlBatch batch, CancellationToken token)
   at Marten.Internal.Sessions.AutoClosingLifetime.ExecuteReaderAsync(NpgsqlBatch batch, CancellationToken token)
[...]

PR #3356, opened 2024-08-05
To debug the issue with include

bhosale-ajay:
This branch adds a new project and try to replicate issue with Include and ToPagedListAsync.

Original Issue : https://github.com/JasperFx/marten/issues/3351


jeremydmiller:
How is this any different than the unit test that's already in there? And passing?

jeremydmiller:
Oh, and could you still use the tests and not a console app? Tests help pin down a problem, extra console apps are just extra weight and noise in the solution

bhosale-ajay:
This is not for merging into code base, but to demonstrate that there is difference between SQL statements generated when running from tests and when used outside tests.
Please switch to branch and run it, so to check if you can replicate it.

Issue #3375, opened 2024-08-21
Polly retries not working when getting connection failures. 


Erwinvandervalk:
Hi guys,

I've experienced some issues with using MartenDB in Azure (postgresdb flexible server). When a maintenance event happens, I'm getting a lot of transient errors in the logs that seem like they are not being picked up by the retry policy. 

I've tried to simulate this issue below in a minimal repro. The code just starts to poll martendb in the background. Then, if I pause / stop my database (it's running locally in docker, so can pause / stop it from docker desktop), then the retry policy doesn't seem to be used. I can clearly see that the query is not retried. 

``` c#
// See https://aka.ms/new-console-template for more information

using System.ComponentModel;
using Marten;
using Marten.Exceptions;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Microsoft.Extensions.Logging;
using Npgsql;
using Polly;
using Polly.Retry;
using Serilog;
using Weasel.Core;

var host = CreateHostBuilder(args).Build();
await host.RunAsync();


static IHostBuilder CreateHostBuilder(string[] args) =>
    Host.CreateDefaultBuilder(args)
        .ConfigureServices((builder, services) =>
        {
            
            services.AddSingleton<IHostedService, TimedHostedService>();
            // This is the absolute, simplest way to integrate Marten into your
// .NET application with Marten's default configuration
            services.AddMarten(options =>
            {
                // Establish the connection string to your Marten database
                options.Connection("User ID=postgres;Password=postgres;Host=localhost;Port=5432;Database=postgres;Timeout=4;Command Timeout=4");

                // Specify that we want to use STJ as our serializer
                options.UseSystemTextJsonForSerialization();

                options.Schema.For<TestDocument>();
                options.ConfigurePolly(builder =>
                {
                    builder
                        .AddRetry(new RetryStrategyOptions
                        {
                            ShouldHandle = new PredicateBuilder()
                                .Handle<NpgsqlException>()
                                .Handle<MartenCommandException>(),
                            Delay = TimeSpan.FromMilliseconds(100),
                            MaxRetryAttempts = 3,
                            UseJitter = true,
                            OnRetry = args =>
                            {
                                Console.WriteLine($"Retry Attempt Number : {args.AttemptNumber} after {args.Duration.TotalMilliseconds} ms.");
                                return default;
                            }
                        })
                        .Build();
                });
[...]

PR #3384, opened 2024-09-02
fix await with polly

Erwinvandervalk:
This PR is related to this issue: https://github.com/JasperFx/marten/issues/3375. Please see my issue for a repro of the issue.

I was having issues that polly wasn't retrying, even though it appeared that the retry policy was being used. 

So I did some tracing. It turned out that the policy returned successfully. Which is weird, because my query was failing. 

I also noticed that the catch block in the linq provider (https://github.com/JasperFx/marten/blob/master/src/Marten/Linq/MartenLinqQueryProvider.cs#L131) wasn't being hit. 

this gave me the idea that we weren't awaiting in the right place. Turns out that was the case.

The resilience pipeline wasn't using async, which meant that it was successfully building a Task, not awaiting said task.

Could you point me to a good place where I can add a good unit test for this?

Thanks!

### DIFF
--- a/src/Marten/Internal/Sessions/QuerySession.Execution.cs
+++ b/src/Marten/Internal/Sessions/QuerySession.Execution.cs
@@ -29,7 +29,7 @@ public partial class QuerySession
     public Task<int> ExecuteAsync(NpgsqlCommand command, CancellationToken token = new())
     {
         RequestCount++;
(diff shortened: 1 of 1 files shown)


jeremydmiller:
@Erwinvandervalk I thought I'd tried that and it didn't work. I've pulled this down and I'm working through the tests. Mostly good so far, and I adjusted one test so far that was broken just from retries actually happening. Fingers crossed, and thanks for taking this on!

Erwinvandervalk:
Happy to be able to contribute. Marten is great :) 

Please let me know if I can assist further on this.

PR #3393, opened 2024-09-04
Fix to actually set the currect timestamp to utc now when using quick…

jeremydmiller:
… append


jeremydmiller:
Going to fix the async void issue in a different push

PR #3397, opened 2024-09-05
Created a failing test for a bug with inline lifecycle

PauloSRH:
Created a failing test for a bug with inline lifecycle and using FetchForWriting


jeremydmiller:
Got it, fix will be in 7.26.6 shortly. Thanks!

jeremydmiller:
Got it with a rebase. The fix is in 7.26.6

PR #3403, opened 2024-09-07
Allow id to have a foreign key reference

e-tobi:
This only requires to NOT create a duplicate field, because there already is an id column.

I think this is very usable when having a 1:1 relationship between different document types.


e-tobi:
If someone already did use a foreign key on a non-default Id property (like e.g. in this test: https://github.com/JasperFx/marten/blob/a2ac808fa690caa7005b8720417fe2756d4c5456/src/DocumentDbTests/ForeignKeys/configuring_foreign_key_fields.cs#L50 ) then now the non-default-id property is not duplicated anymore, but the id column is used instead. Should be no difference, but this might be considered a breaking change?

PR #3404, opened 2024-09-08
dealing with FATAL:  role "root" does not exist

murlakatam:
The error is inside postgres containers. (marten pipeline logs)

The reason is some commands are called with "root" role (default container user) which doesn't exist 
image
image

image


murlakatam:
https://dev.azure.com/jasperfx-marten/e086f0a4-d373-436d-8db8-87b2b6f78cfc/_apis/build/builds/3496/logs/105 no longer has 
FATAL: role "root" does not exist

jeremydmiller:
@murlakatam The best PRs sometimes are fixes for issues you didn't even know exist. Thank you!

PR #3432, opened 2024-09-23
docs: Clarify warning about metadata in inline projections

thomaslevesque:
This limitation no longer applies, according to https://github.com/JasperFx/marten/discussions/3431#discussioncomment-10726662

### DIFF
--- a/docs/events/projections/index.md
+++ b/docs/events/projections/index.md
@@ -91,6 +91,8 @@ There is also a matching asynchronous `AggregateStreamAsync()` mechanism as well
 
 ## Inline Projections
 
+_First off, be aware that some event metadata (`IEvent.Version` and `IEvent.Sequence`) is not available during the execution of inline projections when using the ["Quick" append mode](/events/appending). If you need to use this metadata in your projections, please use asynchronous or live projections, or use the "Rich" append mode._
+
 If you would prefer that the projected aggregate document be updated _inline_ with the events being appended, you simply need to register the aggregation type in the `StoreOptions` upfront when you build up your document store like this:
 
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@thomaslevesque Actually needs more context rather than removing that. That is long, long obsolete if you use the `Rich Append` mode (the default), but goes back to being true if opting into the faster "quick append" mode. 

https://www.youtube.com/watch?v=CQeezCdF4mk

thomaslevesque:
@jeremydmiller ah, I see, thanks.
So, if I understand [this](https://martendb.io/events/appending.html#rich-vs-quick-appends) correctly, the `IEvent.Version` and `IEvent.Sequence` aren't available in inline projections when using "Quick" append mode, but the rest of the metadata is available?

jeremydmiller:
@thomaslevesque Correct. All about compromises. The "rich append" mode was put in place for Marten 4 because people wanted to use the event sequence and stream version data in inline projections, but it turns out that comes with a big perf cost, potentially causes a lot of concurrency issues, and makes the async daemon more prone to skipping events. So back came the <=3.0 "quick append" behavior. 

Except just to make it more complicated, the new quick append method can at least use the correct stream version within aggregates stored in the database along the way. The event sequence is only assigned at the time the events are persisted though.

thomaslevesque:
OK, thanks for the explanation. I'll try to reword the warning to make things clearer

jeremydmiller:
@thomaslevesque Thank you for taking this on!

thomaslevesque:
@jeremydmiller is that better?

mysticmind:
@thomaslevesque the changes look fine.

Issue #3439, opened 2024-09-24
!.Any() regression


AlexSikilinda:
It seems like Marten is generating an incorrect SQL for scenarios when we want to check if a nested collection is empty, e.g.:

```
session.Query<MyDocument>().Where(doc => !doc.NestedCollection.Any());
```
will produce something like 

```
  select d.id, d.data, d.mt_version 
  from public.mt_doc_mydocument as d 
  where d.data @? '$ ? (@.NestedCollection[*] == null || @.NestedCollection[*].size() == 0)' 
```

If the collection is empty (`"NestedCollection":[]`), the second part will be always false which is caused by applying `[*]` to the collection, so 
`@.NestedCollection[*].size() == 0` is `false`, but `@.NestedCollection.size() == 0` seems to work fine.

I believe the issue was introduced with this change in **CollectionIsEmpty**:
![image](https://github.com/user-attachments/assets/741dcaa6-1bcf-4e65-917f-0fb23fd588bf)

jeremydmiller:
@AlexSikilinda Since you're already deep into that, would you mind just doing the pull request to address it?

jeremydmiller:
Closed by https://github.com/JasperFx/marten/commit/c82ea9e480871610ad6bf8f65c7bb7ae8b5b56d5

PR #3446, opened 2024-09-25
Fix IsEmpty and Any for child collections of objects

AlexSikilinda:
Fix for https://github.com/JasperFx/marten/issues/3439

### DIFF
--- a/src/Marten/Linq/SqlGeneration/Filters/CollectionIsEmpty.cs
+++ b/src/Marten/Linq/SqlGeneration/Filters/CollectionIsEmpty.cs
@@ -16,7 +16,7 @@ internal class CollectionIsEmpty: IReversibleWhereFragment
     {
         _member = member;
         var jsonPath = member.WriteJsonPath();
-        _text = $"d.data @? '$ ? (@.{jsonPath} == null || @.{jsonPath}.size() == 0)'";
+        _text = $"d.data @? '$ ? (@.{member.MemberName} == null || @.{member.MemberName}.size() == 0)'";
     }
 
     public ISqlFragment Reverse()
@@ -28,5 +28,4 @@ internal class CollectionIsEmpty: IReversibleWhereFragment
     {
         builder.Append(_text);
     }
-
 }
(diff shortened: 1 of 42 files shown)


src/Marten/Linq/SqlGeneration/Filters/CollectionIsEmpty.cs
```diff
@@ -16,7 +16,7 @@ public CollectionIsEmpty(ICollectionMember member)
     {
         _member = member;
         var jsonPath = member.WriteJsonPath();
-        _text = $"d.data @? '$ ? (@.{jsonPath} == null || @.{jsonPath}.size() == 0)'";
+        _text = $"d.data @? '$ ? (@.{member.MemberName} == null || @.{member.MemberName}.size() == 0)'";
```
AlexSikilinda, line 19: `member.MemberName`, as I understand, will not work with nested collection of inner objects (e.g. Target -> Inner -> Children), but it didn't work with `jsonPath` either.

Another option would be to remove `[*]` at the end of `jsonPath` with Substring, but it seesm it will behave exactly like `.MemberName`, unless I'm missing something.

jeremydmiller: Hey, I'm going to take a longer look at this in the next release with some more test cases -- but still this week!

I've got to get a smaller 7.28.1 out today, but I'll try to roll this into 7.29 later this week.

Thank you for taking this on!

AlexSikilinda: Hi @jeremydmiller, did you have chance to look at this yet? Wondering if anything else needed from me

jeremydmiller: I can take it from here. Line 19 above doesn't work though, because the JSON can't be considered to just be teh member name. Some folks will be using Pascal casing like you're assuming, others will use camelCasing that would blow that up. Plus we have to be able to handle deeper nested properties. You kind of fell down the OSS trap of tactically fixing your particular issue, but not seeing the whole picture that I have to worry about. I've got it from here, and thanks for doing what you've done so far as is

AlexSikilinda: Thank you, Jeremy! Definitely not insisting the fix is 100% valid, my contribution here is really the unit test, so thanks again for handling it. Nested properties collections didn't work with the previous implementation, so the reasoning was "let's at least fix the basic scenario", didn't really thought about camelCase vs Pascal case, definitely a valid concern.

Issue #3514, opened 2024-10-25
When Using Archived Stream Partitioning Tombstone Stream creation fails on insert 


ericgreenmix:
NPGSQL Exception -`there is no unique or exclusion constraint matching the ON CONFLICT specification`

`is_archived` needs to be factored into the `ON CONFLICT` pk fields of the insert statement in `EstablishTombstoneStream`.

PR incoming

jeremydmiller:
Closed by https://github.com/JasperFx/marten/pull/3515

PR #3515, opened 2024-10-25
Fix for #3514

ericgreenmix:
Adds the `is_archived` column to the insert columns and sets it in the `ON CONFLICT` PK's if `UseArchivedStreamPartitioning`.

Ref #3514


jeremydmiller:
@ericgreenmix Good catch, and thank you!

Issue #3510, opened 2024-10-24
Subscription registered using AddSubscriptionWithServices does not filter by IncludeType in the constructor


dedels:
If a subscription is registered with: 
```
 .AddSubscriptionWithServices<MySubscription>(ServiceLifetime.Scoped, _ => {})
```

and in MySubscription the constructor uses: 
```
this.IncludeType<MyEvent>();
```

The subscription does not filter on IncludeType in this case.

If we update the call to AddSubscriptionWithServices and use the configure method to use IncludeType - this does filter events.  However it still invokes ProcessEventsAsync with EventRange containing 0 events.

PR #3516, opened 2024-10-25
Adding a subscription based on subscription base via IoC copies the e…

jeremydmiller:
…vent filters from the inner subscription. Closes GH-3510


src/DaemonTests/Subscriptions/subscriptions_end_to_end.cs
dedels, line 1: This test doesn't actually run data via the subscription.  The belief is that the injection screwed up the subscription?

PR #3599, opened 2024-12-15
Fixes #3597: InvalidCastException in FetchForWriting

### LINKED ISSUE #3597: FetchForWriting throws exception when passing expected version with UseIdentityMapForAggregates = true.
There is a bug with the version of FetchForWriting which takes a Guid identifier and an expected starting version when configured to use the identity map for aggregates.

The exception:
```
Unhandled exception. System.InvalidCastException: Unable to cast object of type 'Marten.Generated.DocumentStorage.RevisionedAggregateProvider771832005' to type 'Marten.Internal.Storage.IDocumentStorage`2[RevisionedAggregate,System.Guid]'.
```

I believe the issue is in `Marten.Events.Fetching.FetchingInlinedPlan<TDoc, TId>`.

specifically this method:
`public async Task<IEventStream<TDoc>> FetchForWriting(DocumentSessionBase session, TId id, long expectedStartingVersion, CancellationToken cancellation = default)`

I cloned to repo and ran the relevant tests, but none failed. I then discovered that there isn't a test for FetchForWriting when using the expected version and while configured to use the identity map for aggregates. I wrote a test for this case, and it does indeed fail. I believe the issue can be resolved by modifying this particular FetchForWriting from:
```csharp
public async Task<IEventStream<TDoc>> FetchForWriting(DocumentSessionBase session, TId id,
    long expectedStartingVersion, CancellationToken cancellation = default)
{
    IDocumentStorage<TDoc, TId> storage = null;
    if (session.Options.Events.UseIdentityMapForAggregates)
    {
        storage = (IDocumentStorage<TDoc, TId>)session.Options.Providers.StorageFor<TDoc>();
        // Opt into the identity map mechanics for this aggregate type just in case
        // you're using a lightweight session
        session.UseIdentityMapFor<TDoc>();
    }
    else
    {
        storage = session.StorageFor<TDoc, TId>();
    }
    // ... 
}
```
to:
```csharp
public async Task<IEventStream<TDoc>> FetchForWriting(DocumentSessionBase session, TId id,
    long expectedStartingVersion, CancellationToken cancellation = default)
{
    IDocumentStorage<TDoc, TId> storage = null;
    if (session.Options.Events.UseIdentityMapForAggregates)
    {
        // Replaced hard cast. This overload now more closely mirrors the other overload in this class.
        storage = session.Options.ResolveCorrectedDocumentStorage<TDoc, TId>(DocumentTracking.IdentityOnly);
        // Opt into the identity map mechanics for this aggregate type just in case
        // you're using a lightweight session
        session.UseIdentityMapFor<TDoc>();
    }
    else
    {
        storage = session.StorageFor<TDoc, TId>();
    }
    // ...
}
```

This change fixes the issue, and it doesn't cause any of the event sourcing tests to fail.

Here is a minimal project which demonstrates the bug:
```csharp
using Marten;
using Marten.Events.Projections;
using Marten.Metadata;

WebApplicationBuilder builder = WebApplication.CreateBuilder(args);

builder.Services.AddMarten(
           opts => {
               opts.Connection("Host=localhost;Port=5432;Database=postgres;Username=postgres;Password=postgres");
               opts.Events.UseIdentityMapForAggregates = true;
               opts.Projections.Snapshot<RevisionedAggregate>(SnapshotLifecycle.Inline);
           })
       .UseLightweightSessions();

WebApplication app = builder.Build();

using (IDocumentStore store = app.Services.GetRequiredService<IDocumentStore>())
{
    await using IDocumentSession session = store.LightweightSession();

    Guid aggregateId = Guid.NewGuid();

    session.Events.StartStream(aggregateId, new AggregateCreated(aggregateId));

    // InvalidCastException thrown here
    var stream = await session.Events.FetchForWriting<RevisionedAggregate>(aggregateId, 1);

    // Never reached
    Console.WriteLine(stream.Aggregate.Version);
}

app.Run();

public record AggregateCreated(Guid Id);

public class RevisionedAggregate : IRevisioned {
    public Guid Id { get; set; }
    public int Version { get; set; }

    public static RevisionedAggregate Create(AggregateCreated created) =>
        new() {
            Id = created.Id
        };
}
```

I hope this level of detail is helpful. I've not contributed before, so I'm not sure how much information I should give.

Strythio:
**Fixes #3597: InvalidCastException in FetchForWriting**

This PR resolves #3597, where the `FetchForWriting<T>` (the overload accepting an expected version) would throw an `InvalidCastException` under the following conditions:
- The aggregate is Inline.
- `UseIdentityMapForAggregates = true`.

### Changes Made
- Fixed the bug in `FetchInlinedPlan.cs`.
- Added tests specifically for the bug.

### Testing
- [x] Initial tests run successfully.
- [x] Added tests for the bug.
- [x] Verified that all tests (new and existing) pass.

### Notes
- I only wrote tests for the `Guid` identifier case. Many of the existing tests target both `Guid` and `string` identifier aggregates. While my tests should cover the scenario effectively, I am happy to add tests for the `string` variants if needed. Let me know if this is required.
- I should note that many of the tests in the `StressTests` project failed, but they fail for me with or without my changes.

### Feedback
This is my first pull request, and I am eager to learn! If any changes are needed, I will address them promptly. Any additional feedback on the implementation or testing strategy is also welcome.

### DIFF
--- a/src/EventSourcingTests/FetchForWriting/fetching_inline_aggregates_for_writing.cs
+++ b/src/EventSourcingTests/FetchForWriting/fetching_inline_aggregates_for_writing.cs
@@ -375,10 +375,89 @@ public class fetching_inline_aggregates_for_writing : OneOffConfigurationsContex
     }
 
 
+    [Fact]
+    public async Task fetch_existing_stream_for_writing_Guid_identifier_with_expected_version_using_identity_map()
+    {
+        StoreOptions(
+            opts =>
+            {
+                opts.Events.UseIdentityMapForAggregates = true;
+                opts.Projections.Snapshot<SimpleAggregate>(SnapshotLifecycle.Inline);
+            });
+
+        var streamId = Guid.NewGuid();
+
+        theSession.Events.StartStream<SimpleAggregate>(streamId, new AEvent(), new BEvent(), new BEvent(), new BEvent(),
+            new CEvent(), new CEvent());
+        await theSession.SaveChangesAsync();
+
+        var stream = await theSession.Events.FetchForWriting<SimpleAggregate>(streamId, 6);
+        stream.Aggregate.ShouldNotBeNull();
+        stream.CurrentVersion.ShouldBe(6);
+
+        stream.AppendOne(new EEvent());
+        await theSession.SaveChangesAsync();
+
+        var latest = await theSession.Events.FetchLatest<SimpleAggregate>(streamId);
+        latest.Version.ShouldBe(7);
+    }
+    [Fact]
+    public async Task fetch_existing_stream_for_writing_Guid_identifier_with_expected_version_using_identity_map_immediate_sad_path()
+    {
+        StoreOptions(
+            opts =>
+            {
+                opts.Events.UseIdentityMapForAggregates = true;
+                opts.Projections.Snapshot<SimpleAggregate>(SnapshotLifecycle.Inline);
+            });
 
 
+        var streamId = Guid.NewGuid();
+
+        theSession.Events.StartStream<SimpleAggregate>(streamId, new AEvent(), new BEvent(), new BEvent(), new BEvent(),
+            new CEvent(), new CEvent());
+        await theSession.SaveChangesAsync();
+
+        await Should.ThrowAsync<ConcurrencyException>(async () =>
+        {
+            var stream = await theSession.Events.FetchForWriting<SimpleAggregate>(streamId, 5);
+        });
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@Strythio Thank you very much for doing this! The CI failures aren't related to your work (we have constant problems w/ blinking tests". This will be in 7.34.1 immediately

Strythio:
Awesome! Glad I could help.

PR #3604, opened 2024-12-20
Support named parameters everywhere you can pass parameters in

elexisvenator:
Requires https://github.com/JasperFx/weasel/pull/159, ~https://github.com/JasperFx/weasel/pull/162~ + related package update. Tests will not pass unless the latest package is there.

Some places supported named parameters (like `.Query()`) but this was not consistent. This PR adds support for the remaining places that I could think of (~`MatchesSql`~, `QueueSqlCommand`), adds tests for these changes as well as the changes in Weasel.   I have also updated the documentation to mention named parameters wherever parameters are already discussed.

The tl:dr of the functional changes:

- Passing in a single parameter that is an anonymous object will be treated as a bucket of named parameters.
- Passing in a single parameter that is a `IDictionary<string, T>`, where T is anything, will be treated as a dictionary of named parameters.
- The exception is compiled queries, which will continue to throw if named parameters are passed in. This is existing behaviour.

This should not break the public API for anyone, as previously passing in a dictionary or an anonymous object would have resulted in Npgsql throwing an error.

Update:
This does _not_ support `MatchesSql` due to the mixing of named and positional parameters. The error from npgsql for this is pretty clear, and the documentation mentions that this shouldn't be done.


elexisvenator:
closed because named params are not the way to go with npgsql

PR #3605, opened 2024-12-20
Handling Shared Identity Across Multi-Stream Projections

erdtsieck:


docs/events/projections/multi-stream-projections.md
```diff
@@ -362,6 +362,60 @@ public class UserFeatureTogglesProjection: MultiStreamProjection<UserFeatureTogg
...
+```cs
+public class ProjectEventGrouper : IAggregateGrouper<Guid>
+{
+    public async Task Group(IQuerySession session, IEnumerable<IEvent> events, ITenantSliceGroup<Guid> grouping)
+    {
+        foreach (var @event in userEvents)
+        {
+            if(@event.Data is TaskCompleted taskCompleted)
+            {
+                var mapping = await session.Query<UserJoinedProject>()
```
erdtsieck, line 409: hmm, I  found that you should use a raw query for this. I do not know how I feel about that.

PR #3616, opened 2024-12-31
Fix some issues with PostgreSQL not running with an english locale

e-tobi:
Instead of relying on exception messages, use SqlState, because exception messages might be language dependent.

### DIFF
--- a/src/CoreTests/adding_custom_schema_objects.cs
+++ b/src/CoreTests/adding_custom_schema_objects.cs
@@ -126,7 +126,7 @@ public class adding_custom_schema_objects: OneOffConfigurationsContext
         martenException.InnerException.ShouldNotBeNull();
         martenException.InnerException.As<PostgresException>().ShouldSatisfyAllConditions(
             e => e.SqlState.ShouldBe(PostgresErrorCodes.UndefinedFunction),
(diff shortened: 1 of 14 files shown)


rud (approved):
That's pretty elegant. 

Would it be worth proving this with integration test or is that going too far?

rud:
@e-tobi the tests are failing, may just be a flaky test. Could you poke at it?

e-tobi:
> That's pretty elegant.
> 
> Would it be worth proving this with integration test or is that going too far?

Such an integration test would require the PostgreSQL server to be started with a different locale. Unfortunately the tests are currently designed to use a fixed external PostgreSQL server. It surely can be done (I'm actually doing this in the unit tests of my application that uses Marten) but I'm not sure, Jeremy would like this.

e-tobi:
> @e-tobi the tests are failing, may just be a flaky test. Could you poke at it?

I'm pretty sure the tests are not failing because of the changes in this PR.
I see failing tests also in the workflow run of the last release: https://github.com/JasperFx/marten/actions/runs/12773892842/job/35606812488#step:20:47

jeremydmiller:
The failing tests are unrelated (and most likely my fault for not catching it before)

PR #3622, opened 2025-01-10
Add overload that allows a custom parameter placeholder.

elexisvenator:
The `char placeholder` param is always before the `string sql` param. This is to ensure that the overload is backwards compatible.,

Updated all methods that accept parameters except for:
- Sync methods that are marked obsolete
- `MatchesJsonPath` - This should probably be marked as obsolete as the placeholder overload does the same thing but better.

Todo:
- [x] Tests. 
- [x] Documentation

### DIFF
--- a/src/Marten/IAdvancedSql.cs
+++ b/src/Marten/IAdvancedSql.cs
@@ -1,7 +1,9 @@
 #nullable enable
+using System;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using Marten.Internal.Sessions;
 
 namespace Marten;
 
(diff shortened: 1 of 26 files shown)


src/Marten/IAdvancedSql.cs
```diff
@@ -62,6 +114,7 @@ public interface IAdvancedSql
     /// <param name="sql"></param>
     /// <param name="parameters"></param>
     /// <returns></returns>
+    [Obsolete(QuerySession.SynchronousRemoval)]
     IReadOnlyList<T> Query<T>(string sql, params object[] parameters);
```
elexisvenator, line 118: A couple of spots where the obsolete warning for synchronous db methods was missing.

elexisvenator:
There will be some small merge conflicts between this and #3604. Some of the wording in the documentation will need to change. There should be no issues with the code itself.

elexisvenator:
failing tests are related to stronglyTypedIds and don't appear to be related to these code changes

PR #3627, opened 2025-01-14
Fix EventSourcingIntro sample

rud:
Output was inconsistent before, so want that to be fixed.

### DIFF
--- a/src/samples/EventSourcingIntro/EventSourcingIntro.csproj
+++ b/src/samples/EventSourcingIntro/EventSourcingIntro.csproj
@@ -2,7 +2,7 @@
 
     <PropertyGroup>
         <OutputType>Exe</OutputType>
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@rud Thanks for doing this, but where in the world is this code? I'll have to admit I didn't even know it was there

rud:
Ooh, now that I have you @jeremydmiller, would the `TreatWarningsAsErrors` setting be something that might be useful to enable elsewhere in the project? 

In this particular case it absolutely pinpointed the bug, maybe there are other similar cases elsewhere that could be uncovered with gradually rolling it out?

jeremydmiller:
@rud If you'd asked me that yesterday, I'd have said I don't have time for that noise -- but then a bug got into Wolverine that Rider was trying to flag. So, maybe?

Issue #3636, opened 2025-01-24
Explicit live aggregations should support string stream identities


erdtsieck:
```
System.InvalidOperationException
Id/id values cannot be null or empty
   at Marten.Generated.DocumentStorage.LightweightSimpleAggregateAsStringDocumentStorage1909523420.AssignIdentity(SimpleAggregateAsString document, String tenantId, IMartenDatabase database)
   at Marten.Internal.Storage.LightweightDocumentStorage`2.Store(IMartenSession session, T document) in C:\Projecten\marten\src\Marten\Internal\Storage\LightweightDocumentStorage.cs:line 19
   at Marten.Internal.Sessions.DocumentSessionBase.storeEntity[T](T entity, IDocumentStorage`1 storage) in C:\Projecten\marten\src\Marten\Internal\Sessions\DocumentSessionBase.cs:line 392
   at Marten.Internal.Sessions.DocumentSessionBase.store[T](IEnumerable`1 entities) in C:\Projecten\marten\src\Marten\Internal\Sessions\DocumentSessionBase.cs:line 370
   at Marten.Internal.Sessions.DocumentSessionBase.Store[T](T[] entities) in C:\Projecten\marten\src\Marten\Internal\Sessions\DocumentSessionBase.cs:line 88
   at Marten.Events.Aggregation.CustomProjection`2.ApplyChangesAsync(DocumentSessionBase session, EventSlice`2 slice, CancellationToken cancellation, ProjectionLifecycle lifecycle)
   at Marten.Events.Aggregation.CustomProjection`2.Marten.Events.Projections.ILiveAggregator<TDoc>.BuildAsync(IReadOnlyList`1 events, IQuerySession session, TDoc snapshot, CancellationToken cancellation) in C:\Projecten\marten\src\Marten\Events\Aggregation\CustomProjection.cs:line 379
   at Marten.Events.QueryEventStore.AggregateStreamAsync[T](String streamKey, Int64 version, Nullable`1 timestamp, T state, Int64 fromVersion, CancellationToken token)
   at EventSourcingTests.Projections.using_explicit_code_for_live_aggregation.using_a_custom_projection_for_live_aggregation_that_has_string_as_id() in C:\Projecten\marten\src\EventSourcingTests\Projections\using_explicit_code_for_live_aggregation.cs:line 51
[...]

PR #3637, opened 2025-01-24
#3636 Explicit live aggregations should support string stream identities

erdtsieck:


src/Marten/Events/Aggregation/CustomProjection.cs
```diff
@@ -376,9 +376,15 @@ async ValueTask<TDoc> ILiveAggregator<TDoc>.BuildAsync(IReadOnlyList<IEvent> eve
         var documentSessionBase = session as DocumentSessionBase ?? (DocumentSessionBase)session.DocumentStore.LightweightSession();
 
         var slice = new EventSlice<TDoc, TId>(default, session, events);
-        await ApplyChangesAsync(documentSessionBase, slice, cancellation).ConfigureAwait(false);
-
-        ApplyMetadata(slice.Aggregate, events.Last());
+        if (Lifecycle == ProjectionLifecycle.Live)
+        {
+            slice.Aggregate = await BuildAsync(session, slice.Aggregate, slice.Events()).ConfigureAwait(false);
+            ApplyMetadata(slice.Aggregate, events.Last());
```
erdtsieck, line 382: ApplyMetadata also happens in ApplyChangesAsync, so no need to do that again in the else

Issue #3644, opened 2025-01-31
`Marten.StoreOptions.AutoRegister` method throws Object reference not set to an instance of an object.


pomianowski:
I have an application that can be simplified like this.

 - **MyCompany.Orders** - all source code
 - **MyCompany.Orders.Worker** - background tasks, RabbitMQ etc.
 - **MyCompany.Orders.Api.Grpc** - gRPC server for external actors

Both the worker and api use the same Marten code. Both make projections, both have the same documents. To simplify it, I configured the application as shown below, so that the generated code is in **MyCompany.Orders**:

```csharp
        _ = services
            .AddMarten(options =>
            {
#if DEBUG
                // When in development mode, store generated code in the same assembly as the UnitOfWork
                // https://martendb.io/configuration/prebuilding
                options.SetApplicationProject(typeof(UnitOfWork).Assembly);
#else
                options.SetApplicationProject(typeof(UnitOfWork).Assembly, "/");
                options.GeneratedCodeMode = TypeLoadMode.Auto;
#endif
            })
            .AddAsyncDaemon(DaemonMode.HotCold)
#if DEBUG
            .OptimizeArtifactWorkflow(TypeLoadMode.Static)
#endif
            .UseLightweightSessions();
```

it more or less works, but it does not look elegant.

So I tried it this way:

```csharp
        _ = services
            .AddMarten(options =>
            {
#if DEBUG
                // When in development mode, store generated code in the same assembly as the UnitOfWork
                // https://martendb.io/configuration/prebuilding
                options.SetApplicationProject(typeof(UnitOfWork).Assembly);
#else
                options.AutoRegister(scanner =>
                {
                    scanner.Assembly(typeof(UnitOfWork).Assembly);
                });
                options.GeneratedCodeMode = TypeLoadMode.Auto;
#endif
            })
            .AddAsyncDaemon(DaemonMode.HotCold)
#if DEBUG
            .OptimizeArtifactWorkflow(TypeLoadMode.Static)
#endif
            .UseLightweightSessions();
```

but in `StoreOptions`, the:
`var assemblies = _assemblies.Union([options.ApplicationAssembly]).ToArray();`
creates array of the provided `typeof(UnitOfWork).Assembly` assembly and `null`, because ApplicationAssembly is not set currently.

In my opinion, this can be solved in two ways, either add Assembly, to scan immediately, or omit it from the union and let the application work as before, but without adding application assembly if it has not yet been defined.

```csharp
[...]

PR #3645, opened 2025-01-31
Ignore `ApplicationAssembly` when scanning compiled types using `AutoRegister`

pomianowski:
Fixes null reference exception when `ApplicationAssembly` is not defined yet, #3644


jeremydmiller:
@pomianowski I'll take a better look at this today before getting it in. I wonder if there's a timing issue here.

PR #3663, opened 2025-02-07
ProjectionWrapper honoring settings of projections based on EventFilterable

netijoe:
When ProjectionWrapper wraps a projection that inherits from `EventFilterable` (e.g. when inheriting from `ProjectionBase`) it should use it's settings regarding eventtypes and archived events instead of ignoring it.


jeremydmiller:
I'll deal with the compilation failures. Nothing related to your work. Something to do w/ Vogen source generator code. Good times.

PR #3667, opened 2025-02-11
Examples of using DocumentListener with OpenTelemetry

STRUDSO:
- CorrelationListener
- Events listener

### DIFF
--- a/src/DocumentDbTests/DocumentDbTests.csproj
+++ b/src/DocumentDbTests/DocumentDbTests.csproj
@@ -19,6 +19,8 @@
         <PackageReference Include="Jil" Version="3.0.0-alpha2" />
         <PackageReference Include="Lamar" Version="14.0.1" />
         <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
+        <PackageReference Include="OpenTelemetry" Version="1.8.1" />
+        <PackageReference Include="OpenTelemetry.Exporter.InMemory" Version="1.8.1" />
         <PackageReference Include="xunit.runner.visualstudio" Version="2.8.2">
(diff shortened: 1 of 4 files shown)


jeremydmiller:
@STRUDSO Can you start by explaining why all of Marten's existing Otel support doesn't cover whatever you want here? 
https://martendb.io/otel.html#open-telemetry-and-metrics

STRUDSO:
Let me spend a little more time with this, and yes some of it probably belongs in Samples.

At first i didn't realize the work in:
https://github.com/JasperFx/marten/blob/a0b339a68330d4771def55bdd34aea52f8f04706/src/Marten/ISessionFactory.cs#L46

Spent some time on understanding traceparent, trace_id and parent_id in the context of this:
https://www.w3.org/TR/trace-context/#version-format 
And how it manifests itself in dotnet's activity class.

In SessionFactoryBase it has:
````csharp
        documentSession.CorrelationId = Activity.Current?.RootId;
````
RootId = TraceId in my understanding, so all good, don't know anything about tracing pre opentelemetry, so maybe this is how it should be. But seems like opentelemetry calls it traceid. i guess this is what you would us to find "all" of the spans in jaeger eg.

````csharp
        documentSession.CausationId = Activity.Current?.ParentId;
````
ParentId i don't fully understand. my understanding is:

````csharp
using var activity1 = source.Start
using var activity2 = source.Start
// activity2.ParentId == activity.Id
using var session = store.OpenSession()
session.Store(doc)

````
then doc.CausationId would point to the Id of activity1.
Shouldn't this be activity2.Id?

jeremydmiller:
Hey, sorry this never got in, but it's too far past now

PR #3708, opened 2025-03-10
Feature/bug n gram not public schema

STRUDSO:
Seems like there is an issue with the  mt_grams_vector not being applied or "used" wrongly when using ngram and having entities defined in a different schema.

This is the raw exception:
Npgsql.PostgresException
42883: function not_public.mt_grams_vector(text) does not exist

If i change the "default" schema on the documentstore it seems to create the objects.

### DIFF
--- a/src/LinqTests/Bugs/Bug_3708_NGram_in_other_schema.cs
+++ b/src/LinqTests/Bugs/Bug_3708_NGram_in_other_schema.cs
@@ -8,7 +8,7 @@ using Weasel.Core;
 
 namespace LinqTests.Bugs;
 
-public class Bug_xxx_NGram_in_other_schema : BugIntegrationContext
+public class Bug_3708_NGram_in_other_schema : BugIntegrationContext
 {
     private string AnotherSchema = "not_public";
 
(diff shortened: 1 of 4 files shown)


STRUDSO:
@jeremydmiller guess the intend is that these "system" functions are only in schema as specified by the store config?

So maybe this fix is correct :)

jeremydmiller:
@STRUDSO   So, "yes" to your question above. The full text search stuff was built through community contributions, and it's quite (very likely) possible that I didn't do a good enough job watching it coming in or have really gotten familiar with it since. 

I'll get this in soon. The CI build failures are unrelated and yet another "get around to it" issue for me

mysticmind:
@jeremydmiller This looks fine to pull it in. I kicked started the CI earlier today after a brief look at the changes.

STRUDSO:
@mysticmind @jeremydmiller 

Guys, we have tested this out, and we see that having the "functions" mt_grams_vector, eg in the a different schema from the tables hurts performance.

So maybe hold off on merging this.

STRUDSO:
FAST:
image
SLOW:
image

mysticmind:
I have added a unit tests with few tweaks to the original unit tests what you provided on running a different schema by creating a new store. Everything works fine as expected and there is no issue as such. See https://github.com/JasperFx/marten/commit/969bcfd5365d32a0cfab6b2bafa15d51139f91aa. Unit tests are all passing.

I am closing this.

PR #3713, opened 2025-03-13
Applying unaccent to ngram searches for better multilingual experience

jakobt:
This will give less surprises for people trying to search in non english texts.

### DIFF
--- a/docs/documents/full-text.md
+++ b/docs/documents/full-text.md
@@ -307,7 +307,7 @@ var posts = session.Query<BlogPost>()
 
 ## Partial text search in a multi-word text (NGram search)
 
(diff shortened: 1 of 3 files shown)


baltie:
Unfortunately, this gave plenty of surprise to us who does not have the unaccent extension enabled on the postgresql server...

jakobt:
Hi @baltie ,

Sorry about that! But what problems did you have since  7.39.1 should enable the extension but still have a problem that it insists on reapplying a procedure as part of migrations?

That is addressed in https://github.com/JasperFx/marten/pull/3721

baltie:
We didn't have any problems in our integration tests as those use a database where the user has the right to enable extensions. In our database in Azure, the user did not have such access. But I guess the PR you point to will help somewhat.

Regardless, we've requested the extension to be enabled and we will upgrade when it is there. So not a big problem for us. 😄

PR #3721, opened 2025-03-14
Make sure unaccent extension is enabled and plays nice with migration

jakobt:


mysticmind (approved):
Looks good to me 🚀

haefele:
How about the "CREATE EXTENSION IF NOT EXISTS unaccent;" in the sql? Will it work with the migrations that way?

jakobt:
> How about the "CREATE EXTENSION IF NOT EXISTS unaccent;" in the sql? Will it work with the migrations that way?

This was removed and being replaced by adding the extension as an ISchemaObject that Marten automatically manages and generates code for.

PR #3725, opened 2025-03-18
Fix null check in mt_jsonb_fix_null_parent.sql

marcin-brzozowski:
Fixing null check in the `mt_jsonb_fix_null_parent.sql` function as described in #3722 , which will allow to create full hierarchy of nested keys.


mysticmind (approved):
Looks good to me 👍

PR #3738, opened 2025-03-28
docs(Ngram search): Added doc on how to search on multiple properties

WilvanBil:
Based on the lovely feedback I had in the Discord regarding Ngram search, I have added an example in the documentation.


mysticmind:
@WilvanBil I will take care of converting this into a linked sample from code into docs.

PR #3782, opened 2025-05-02
Improved select parsing

lukedukeus:
Throws exceptions if it is unable to parse a select projection.
Before, it would either not run the `Select` but still return the query without the projection, or return an empty queryable. Now, it throws an exception if it cannot parse.

### DIFF
--- a/src/Marten/Linq/Parsing/SimpleExpression.cs
+++ b/src/Marten/Linq/Parsing/SimpleExpression.cs
@@ -1,6 +1,7 @@
 #nullable disable
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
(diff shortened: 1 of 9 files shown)


src/Marten/Linq/Parsing/SimpleExpression.cs
```diff
@@ -308,13 +309,51 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+                            break;
+                        case bool ic:
+                            ignoreCase = ic;
+                            break;
+                        case CultureInfo ci:
+                            culture = ci;
+                            break;
+                        case CompareOptions co:
+                            compareOptions = co;
+                            break;
```
jakobt, line 343: Would probably make sense to have some default in this case statement also resulting in the BadLinqExpression if people would try to use the overloads with substrings etc. like this one https://learn.microsoft.com/en-us/dotnet/api/system.string.compare?view=net-9.0#system-string-compare(system-string-system-int32-system-string-system-int32-system-int32)

lukedukeus: @jakobt added in 156df24cd7dbb7e258fb92f053f64573c46d0477

jeremydmiller:

PR #3818, opened 2025-06-02
Update docs of UseOptimizedProjectionRebuilds

schadowfax:


jeremydmiller:
@schadowfax I'll update this with less obnoxious verbiage tomorrow.

Just so we're clear, if you ever gripe about Marten again on a public forum without making any real good faith effort to work with us on whatever your very specific and also quite rare issue is, I'm permanently banning you from both Discord and GitHub.

PR #3820, opened 2025-06-03
updated cli commands

JeffryGonzalez:
I added some updated use for the command line stuff, and removed the reference to `marten.commandline` nuget.


JeffryGonzalez:
I'm wondering if the "Outside the CLI" section is still needed? I can't imagine anyone using Marten and not using the CLI, and the examples here are pretty vague and could be confusing.

Issue #3778, opened 2025-04-28
Schema name ending with `d` results in issues with generated index definitions


tomap:
Hello, 

I'm using Marten 3.40.0 + .Net 9 but I believe the issue has been there for some time now (3.x + .Net 8)

```
Marten.Exceptions.MartenSchemaException: DDL Execution for 'All Configured Changes' Failed!

DO $$
BEGIN    IF NOT EXISTS(
        SELECT schema_name
          FROM information_schema.schemata
          WHERE schema_name = 'pprd'
      )
    THEN
      EXECUTE 'CREATE SCHEMA pprd';
    END IF;

END
$$;


DROP TABLE IF EXISTS pprd.mt_doc_docwrapricreference CASCADE;
CREATE TABLE pprd.mt_doc_docwrapricreference (
    id                  varchar                     NOT NULL,
    data                jsonb                       NOT NULL,
    mt_last_modified    timestamp with time zone    NULL DEFAULT (transaction_timestamp()),
    mt_version          uuid                        NOT NULL DEFAULT (md5(random()::text || clock_timestamp()::text)::uuid),
    mt_dotnet_type      varchar                     NULL,
CONSTRAINT pkey_mt_doc_docwrapricreference_id PRIMARY KEY (id)
);

CREATE INDEX mt_doc_docwrapricreference_idx_document_date ON pprd.mt_doc_docwrapricreference USING btree ((pprmt_immutable_timestamp(data -> 'Document' ->> 'Date')));

CREATE INDEX mt_doc_docwrapricreference_idx_document_ric_key ON pprd.mt_doc_docwrapricreference USING btree ((data -> 'Document' ->> 'RicKey'));

CREATE INDEX mt_doc_docwrapricreference_idx_expiration ON pprd.mt_doc_docwrapricreference USING btree ((pprmt_immutable_timestamptz(data ->> 'Expiration')));

CREATE OR REPLACE FUNCTION pprd.mt_upsert_docwrapricreference(doc JSONB, docDotNetType varchar, docId varchar, docVersion uuid) RETURNS UUID LANGUAGE plpgsql SECURITY INVOKER AS $function$
DECLARE
  final_version uuid;
BEGIN
INSERT INTO pprd.mt_doc_docwrapricreference ("data", "mt_dotnet_type", "id", "mt_version", mt_last_modified) VALUES (doc, docDotNetType, docId, docVersion, transaction_timestamp())
  ON CONFLICT (id)
[...]

PR #3824, opened 2025-06-04
Fix blind replace of `d.` in TypedLocator causing issues when schema name ends with `d`

mysticmind:
- Fix blind replace of d.in TypedLocator causing issues when schema name ends with d
- Fixes GH-3778

### DIFF
--- a/src/DocumentDbTests/Bugs/Bug_3778_schema_name_issue.cs
+++ b/src/DocumentDbTests/Bugs/Bug_3778_schema_name_issue.cs
@@ -1,0 +1,30 @@
+using System;
+using System.Threading.Tasks;
+using JasperFx;
+using Marten.Testing;
+using Marten.Testing.Harness;
+using Xunit;
+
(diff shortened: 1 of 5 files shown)


mysticmind:
The CI failure `DocumentDbTests.Bugs.Bug_2169_concurrency_error.there_should_be_no_error [FAIL]` is unrelated to this PR.

aare-arnold:
If the index contains a reference to the additional column (for example, a column used as a foreign key), the SQL becomes invalid because ``d.`` no longer gets replaced.

output prior this PR:
``CREATE UNIQUE INDEX mt_doc_user_accounts_uidx_provider_nameuser_id ON public.mt_doc_user_accounts USING btree ((data ->> 'providerName'), (user_id));``

and from now on:
``CREATE UNIQUE INDEX mt_doc_user_accounts_uidx_provider_nameuser_id ON public.mt_doc_user_accounts USING btree ((data ->> 'providerName'), (d.user_id));``

Issue #3823, opened 2025-06-04
Masking Rule Support for Record Types


kopfrechner:
Hi, 

first, thanks for the beautiful libraries of the Critter Stack 👍 

Here's my Issue with solution proposal:

I wanted to mask some of my events using the `AddMaskingRuleForProtectedInformation<T>(Action<T>)` method as described here
https://martendb.io/events/protection.html#snippet-sample_defining_masking_rules

However, i use records for events, so with an `Action` the record can't be mutated in the lambda:
```cs
options.Events.AddMaskingRuleForProtectedInformation<User>(x => x with { FullName = "****" });
```

My suggestion is to offer an overload
```cs
public void AddMaskingRuleForProtectedInformation<T>(Func<T, T> func)
{
    ArgumentNullException.ThrowIfNull(action);
    _maskers.Add(new FuncMasker<T>(func));
}
```

And of corse a separate masker
```cs
internal class FuncMasker<T> : IMasker where T : notnull
{
    private readonly Func<T,T> _masking;

    public FuncMasker(Func<T,T> masking)
    {
        _masking = masking;
    }

    public bool TryMask(IEvent @event)
    {
        if (@event is IEvent<T> e)
        {
            e.Data = _masking(e.Data); // Not sure if this is assignment to e.Data is going to work...
            return true;
        }

        return false;
    }
}
```

Here's the relevant code

https://github.com/JasperFx/marten/blob/40d94f8bf5d5a38d7f54fed24263e04a5348cc0d/src/Marten/Events/EventGraph.Masking.cs#L18

If this would fit in your design, i can open a pull request.

Best Regards, 
Christoph

jeremydmiller:
Doh, yeah, what you proposed makes perfect sense to me. Feel free to do that PR and we'll make that part of 8.1

PR #3831, opened 2025-06-05
fix: #3823 fix tests fix interface

kopfrechner:
* Fix Test since we need to check the actual data instead of the reference now.
* Expose new functionality via Interface.


kopfrechner:
Test failed, on it...
https://github.com/JasperFx/marten/actions/runs/15467303942/job/43542324002?pr=3831#step:13:48

kopfrechner:
This is a follow up to #3825. I screwed it up a little bit since 
* the new test failed
* the data replacement of the record type event was not correct
* the new functionality was not exposed

Sorry for the inconvenience, this is the fix for it.

Best Regards, 
Christoph 

FIY @jeremydmiller

PR #3835, opened 2025-06-11
Introduced `RegisterValueType<TValueType>()` method

HakamFostok:
- Introduced `RegisterValueType<TValueType>()` method to allow custom value type registration using generics
which is just a wrapper for the `RegisterValueType(Type type)` method

- Use ArgumentNullException.ThrowIfNull instead of throw manually


Hawxy (approved):

HakamFostok:
Thanks for approving. I have a question about the process.
As my PR has been approved, why has it not been merged yet?
Is there a merge window before releasing a new version, and all the merges happen at that window?
Thank you

PR #3848, opened 2025-06-16
Count length properties performance

HakamFostok:
Replace `Count()` LINQ method with `Count` or `Length` properties for performance improvments


HakamFostok:
@jeremydmiller Absolutely, I really did not understand why git did that, but anyway it is on me, let me fix this and make a clean PR. Thank you for letting me know

PR #3860, opened 2025-07-02
Check if tenant id is not empty before adding parameter to query

rikharink:
also clean-up unused synchronous version of DeleteSingleEventStreamAsync


jeremydmiller:
@rikharink I think it should probably assert and throw an exception if the tenant id is empty in this case

PR #3862, opened 2025-07-07
Reference JasperFx for CommandLine

alfkonee:
updated documentation for CLI for v8 Marten

### DIFF
--- a/docs/configuration/cli.md
+++ b/docs/configuration/cli.md
@@ -1,15 +1,16 @@
 # Command Line Tooling
 
 ::: warning
-The usage of Marten.CommandLine shown in this document is only valid for applications bootstrapped with the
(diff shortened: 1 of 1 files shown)


jeremydmiller:
@alfkonee Hey, thanks for this, but it's not quite right. JasperFx is a transitive dependency of Marten 8, so there's nothing else to install. I'll get this in and tweak it a bit. I know I'm behind on docs, but something magically comes up every time I start to work on updating it

jeremydmiller:
@alfkonee 

> Ok so does that mean that mean that I for v8 to access the CLI tools I just need to install Marten itself??

Yes. I'll get this in and make the little corrections right now. Thanks again!

Part of the V8 release was consolidating the underlying pieces to make CLI setup easier *from scratch*

PR #3899, opened 2025-08-07
Avoid using 'Enumerable.Any()' extension method

HakamFostok:
Performance improvement

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1860


jeremydmiller:
Huh. Used to be the other way around

Issue #3896, opened 2025-08-06
Support duplicating/indexing fields on subclass documents in hierarchies


erdtsieck:
**Issue:** Marten currently does not allow marking fields that exist only on subclass types in a document hierarchy as duplicated or indexed fields in the schema configuration. According to the documentation, only properties on the top-level (base) document can be made searchable/indexed. This means properties present exclusively in subclass implementations cannot be duplicated via `StoreOptions.Schema.For<T>()` (attempts to cast in the lambda, e.g. `.Duplicate(x => ((ISubType)x).Field)`, are unreliable and fail for documents of other subtypes).

**Impact:** In scenarios with document hierarchies (using `AddSubClassHierarchy` or an interface as aggregate root), there is no straightforward way to optimize queries on subclass-specific fields. Developers must resort to hacks or forego indexing, which leads to slower queries or more complex domain models.

**Proposal:** Enhance Marten to support duplicated fields (and indexes) for subclass properties. For example, an API could allow specifying a generic subtype for duplication: `.Schema.For<BaseType>().Duplicate<SubType>(x => x.SubProperty)`. Marten would then create a column for that property and populate it for documents of that subtype (null for others). Internally, this requires the codegen to handle conditional writing (checking the document type or interface and writing the value if available). The solution should also guard against naming collisions (different subclasses having same-named properties) by requiring an explicit subtype-specific call.

jeremydmiller:
Closed by GH-3896

PR #3910, opened 2025-08-11
3896 support duplicating and indexing fields on subclass documents in hierarchies

erdtsieck:
This change addresses #3896 , where duplicating fields from subclasses (`Duplicate<TSub>`) resulted in unconditional casts in the generated `Upsert`, `Insert`, and `Update` operations.
When saving instances of the base type, this caused an `InvalidCastException`.

Changes include:

* Updated `UpsertArgument.GenerateCodeToSetDbParameterValue` to perform safe type checks before casting to the subclass.
* Applied the same safe access pattern in `GenerateBulkWriterCodeAsync` to avoid runtime cast errors.
* Extended use of `AccessorString` with guards for subtype casting.
* Added XML documentation for both `Duplicate<TSub>` overloads in `DocumentMappingExpression` and `DocumentMapping<T>` to clarify API intent.

Impact:

* Subclass fields can now be safely marked as duplicated fields without causing runtime cast exceptions.
* Generated storage code works correctly for mixed collections of base type and subclass instances.
* API documentation now clearly describes the usage and parameters of `Duplicate<TSub>`.

### DIFF
--- a/docs/configuration/cli.md
+++ b/docs/configuration/cli.md
@@ -29,7 +29,7 @@ scenarios.
 To use the expanded command line options to a .NET application, add this last line of code shown below to your `Program.cs`:
 
 <!-- snippet: sample_using_WebApplication_1 -->
-<a id='snippet-sample_using_webapplication_1'></a>
+<a id='snippet-sample_using_WebApplication_1'></a>
 ```cs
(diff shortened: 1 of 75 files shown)


erdtsieck:
image
Document tests succeed :-) 

Except for some test I think is probably flaky anyway: 
image

PR #3935, opened 2025-08-29
Fix issue where existing aggregates could not be rebuilt

alexanderjasper:
In case of code changes to an aggregate, it may be necesary to rebuild the stored aggregate documents even though they are up to date version-wise.

This is not currently happening because the upsert function only updates data if the version number has strictly increased, which is naturally not the case when rebuilding an existing up-to-date aggregate, since in that case the version number is unchanged.

This PR ensures that the data is updated even when the version number is unchanged.

### DIFF
--- a/src/EventSourcingTests/Aggregation/rebuilding_a_single_stream_projection.cs
+++ b/src/EventSourcingTests/Aggregation/rebuilding_a_single_stream_projection.cs
@@ -2,7 +2,7 @@ using System;
 using System.Threading.Tasks;
 using EventSourcingTests.FetchForWriting;
 using JasperFx.Events;
-using Marten.Events;
+using Marten.Events.Projections;
 using Marten.Testing.Harness;
 using Shouldly;
 using Xunit;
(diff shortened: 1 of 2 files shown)


alexanderjasper:
> @alexanderjasper I'll take the test, but under absolutely no circumstances are we changing that `UpsertFunction` code. There's an option to force the "overwrite" we can use in the rebuild step now that I see what might be going on. Now I'll have to look at the existing tests for that functionality and see why they are passing.

Sure, I understand. That seems like a good solution. I would still expect `session.Store(document)` to overwrite regardless of version numbers. But then again, one should probably never mess with projection documents manually like that.

Thanks for looking into this issue. We are overcoming this by calling `session.Delete` before `store.Advanced.RebuildSingleStreamAsync` which causes the document to be re-added correctly. But we will be happy to remove that step again once there is a fix for this.

jeremydmiller:
"I would still expect session.Store(document) to overwrite regardless of version numbers." -- That's opposite of what it *should* do!

jeremydmiller:
@alexanderjasper Hey, thanks for doing this. I took your test, but "fixed" it a little differently. I'll have this pushed by EOD

PR #3996, opened 2025-11-04
Fixed logger for connections inside EventTracingConnectionLifetime

JTeeuwissen:
Currently, the `EventTracingConnectionLifetime` has it's own `IMartenSessionLogger Logger` property, which is set in it's constructor. This however, does not set the logger of the `InnerConnectionLifetime`. Which causes logs to be written to the default `NulloLogger`. This PR replaces the Logger property from `EventTracingConnectionLifetime` with a proxy for the Logger property from `InnerConnectionLifetime`.


jeremydmiller:
Nice catch!

Issue #4012, opened 2025-11-13
Using unaccent breaks when using Ancillary Marten Stores


erdtsieck:
```
Marten.Exceptions.MartenCommandException: Marten Command Failure:$

Marten.Exceptions.MartenCommandException
Marten Command Failure:$
$
$
42883: function unaccent(unknown) does not exist

POSITION: 8
   at JasperFx.Core.Exceptions.ExceptionTransformExtensions.TransformAndThrow(IEnumerable`1 transforms, Exception ex) in /_/src/JasperFx/Core/Exceptions/IExceptionTransform.cs:line 170
   at JasperFx.Core.Exceptions.ExceptionTransforms.TransformAndThrow(Exception ex) in /_/src/JasperFx/Core/Exceptions/IExceptionTransform.cs:line 84
   at Marten.Exceptions.MartenExceptionTransformer.WrapAndThrow(Exception exception) in C:\Projecten\marten\src\Marten\Exceptions\MartenExceptionTransformer.cs:line 62
   at Marten.Linq.MartenLinqQueryProvider.ExecuteHandlerAsync[T](IQueryHandler`1 handler, CancellationToken token) in C:\Projecten\marten\src\Marten\Linq\MartenLinqQueryProvider.cs:line 141
   at Marten.Internal.Sessions.QuerySession.QueryAsync[T](Char placeholder, String sql, CancellationToken token, Object[] parameters) in C:\Projecten\marten\src\Marten\Internal\Sessions\QuerySession.Querying.cs:line 51
   at StressTests.using_multiple_document_stores_in_same_host.use_unaccent_in_ancillary_store() in C:\Projecten\marten\src\StressTests\using_multiple_document_stores_in_same_host.cs:line 135
   at StressTests.using_multiple_document_stores_in_same_host.use_unaccent_in_ancillary_store() in C:\Projecten\marten\src\StressTests\using_multiple_document_stores_in_same_host.cs:line 136
   at Xunit.Sdk.TestInvoker`1.<>c__DisplayClass47_0.<<InvokeTestMethodAsync>b__1>d.MoveNext() in /_/src/xunit.execution/Sdk/Frameworks/Runners/TestInvoker.cs:line 259
--- End of stack trace from previous location ---
   at Xunit.Sdk.ExecutionTimer.AggregateAsync(Func`1 asyncAction) in /_/src/xunit.execution/Sdk/Frameworks/ExecutionTimer.cs:line 48
[...]

PR #4013, opened 2025-11-13
#4012 Using unaccent breaks when using Ancillary Marten Stores

erdtsieck:


erdtsieck:
It does work if you run `await store.Storage.ApplyAllConfiguredChangesToDatabaseAsync();` at the start of your tests!

PR #4039, opened 2025-11-29
fix:linq collections contain in C# 14

StellaAlexis:
When updating my project to .Net 10 and C# 14 I noticed that the linq collection expressions around contains were broken. 
References on the change in C#14 can be found[here](https://github.com/dotnet/roslyn/blob/4a48be83fb038db194d76f8fee430679fec2d46f/docs/compilers/CSharp/Compiler%20Breaking%20Changes%20-%20DotNet%2010.md#spant-and-readonlyspant-overloads-are-applicable-in-more-scenarios-in-c-14-and-newer)

To validate the issue I changed the TargetFramework and LangVersion of LinqTests.csproj. That resulted in a bunch of the existing tests failing right away. 
Suggesting a fix similar to how EnumerableContains is written. After the change all Linq tests are passing again.

### DIFF
--- a/src/LinqTests/LinqTests.csproj
+++ b/src/LinqTests/LinqTests.csproj
@@ -1,8 +1,17 @@
 <Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <LangVersion>14</LangVersion>
         <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
         <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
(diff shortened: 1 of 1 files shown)


StellaAlexis:
Noticed the failing tests. Assume that is because of the hardcoded .net 10 framework. Managed to also run the tests without it so removed that part.

1sabelle:
We're also affected by this. Temporary solution for us is to pin language version to C# 13 just as seems to have been the prior solution here. Nice to see a fix.

jeremydmiller:
Thanks @Hawxy !

PR #4072, opened 2026-01-15
Restrict read replica usage to IQuerySession

### LINKED ISSUE #4070: Bug: FetchForWriting use standby connection when ReadSessionPreference = PreferStandby
### Marten Version
- 8.17.0

### PostgreSQL Version
- PostgreSQL 16.8

### Deployment / Topology
PostgreSQL is deployed locally via Docker Compose with:
- Primary (write) instance on port 55432
- Read replica on port 55433

Marten is configured using a multi-host data source with a read replica and:
- ReadSessionPreference set to PreferStandby

```csharp
services.AddMultiHostNpgsqlDataSource(
    "Host=localhost:55432,localhost:55433;Username=postgres;Password=postgres;Database=workmgmt");

services.AddMarten(options =>
{
    options.Events.StreamIdentity = StreamIdentity.AsGuid;
    options.Events.AppendMode = EventAppendMode.Quick;
    options.AutoCreateSchemaObjects = AutoCreate.All;

    // --- Projections Configuration ---
    /* ... INLINE ... */

    // --- Schema Configuration ---
    /* ... */

    options.Advanced.MultiHostSettings.ReadSessionPreference =
        TargetSessionAttributes.PreferStandby;
})
.UseNpgsqlDataSource()
.UseLightweightSessions()
.ApplyAllDatabaseChangesOnStartup();

```

---

### Actual Behavior

When ReadSessionPreference is set to PreferStandby:

- All IQuerySession reads behave correctly and are routed to the read replica
- IDocumentSession.FetchForWriting performs read operations against the read replica
- AppendMany correctly writes to the primary, but the reads that occur before the write do not

Observed via OpenTelemetry and Npgsql instrumentation during a FetchForWriting → AppendMany command:

1. Initial connection checks occur against the primary
2. Subsequent SELECT statements executed as part of FetchForWriting are routed to the read replica
3. Final INSERT / UPDATE operations are executed against the primary

This results in FetchForWriting performing its internal reads against the standby while writes are sent to the primary.

For comparison, a query executed using only IQuerySession behaves as expected:
- Initial connect against the primary
- Query execution routed to the read replica

---

### Impact

This breaks strong consistency during command handling.

FetchForWriting is a critical concurrency and consistency path. Because read replicas may lag behind the primary:
- Commands may read stale aggregate or projection state
- Business decisions may be made using outdated data
- Concurrency handling can become unreliable
- In some cases, failures may occur if the replica has not yet received the latest writes

---

### Expected Behavior

When ReadSessionPreference is set to Standby, PreferStandby, or Readonly:

- Only IQuerySession should route queries to the read replica
- All IDocumentSession operations should execute entirely against the primary
- FetchForWriting should always guarantee strongly consistent reads and never use a standby connection

---

### Minimal Reproduction

1. Set up a PostgreSQL primary and read replica (for example via Docker)
2. Configure Marten with a multi-host data source and ReadSessionPreference = PreferStandby
3. Execute a minimal FetchForWriting followed by AppendMany
4. Observe query routing via logging or OpenTelemetry

---

### Workaround

A workaround was implemented locally by:
- Configuring Marten to use the primary data source only
- Manually overriding IQuerySession to route read-only queries to the replica

This workaround functions but is hacky and bypasses Marten’s built-in multi-host routing behavior.

```csharp
// Override IQuerySession to use read data source when read replica is configured
if (dbOptions.HasReadReplica)
{
    services.RemoveAll<IQuerySession>();
    services.AddScoped<IQuerySession>(sp =>
    {
        var readDataSource =
            sp.GetRequiredKeyedService<NpgsqlDataSource>(DataSourceKeys.Read);

        var store = sp.GetRequiredService<IDocumentStore>();
        var conn = readDataSource.OpenConnection();

        return store.QuerySession(SessionOptions.ForConnection(conn));
    });
}
```

Hawxy:
Closes #4070

### DIFF
--- a/docs/configuration/hostbuilder.md
+++ b/docs/configuration/hostbuilder.md
@@ -186,11 +186,9 @@ services.AddMarten(x =>
 <!-- endSnippet -->
 
 ::: warning
-Marten will only use your read node preference with user queries (using IQuerySession) that are using a Marten-managed lifetime. 
+Marten will only use your read node preference with queries run with `IQuerySession` that are using a Marten-managed connection lifetime. 
 
-Internal queries, including the async daemon, will always use your primary node for reliability.
-
-Ensure your replication delay is acceptable as you risk returning outdated queries.
+Queries executed via `IDocumentSession`, alongside internal queries (ie async daemon), will always use your primary node to ensure write-side consistency.
 :::
 
 ## Composite Configuration with ConfigureMarten()
--- a/src/Marten/Internal/Sessions/AutoClosingLifetime.cs
+++ b/src/Marten/Internal/Sessions/AutoClosingLifetime.cs
@@ -20,6 +20,7 @@ internal class AutoClosingLifetime: ConnectionLifetimeBase, IConnectionLifetime,
 {
     private readonly SessionOptions _options;
     private readonly IMartenDatabase _database;
+    private readonly ConnectionUsage _readerConnectionUsage;
 
     public AutoClosingLifetime(SessionOptions options, StoreOptions storeOptions)
     {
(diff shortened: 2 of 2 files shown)


jeremydmiller:
Thank you @Hawxy!

PR #4112, opened 2026-02-04
Fix F# option date index member selection

nkosi23:
### Motivation
- Fix a regression where F# option date/time fields were resolved as generic value types and emitted `CAST(... as timestamp)`/`timestamptz` in computed index DDL instead of using the `mt_immutable_*` functions, which caused Postgres errors about non-immutable functions in index expressions.
- Ensure the computed index test exercises F# option value type registration so the regression is covered.
### Description
- In `src/Marten/StoreOptions.MemberFactory.cs` added a guard in `ValueTypeMemberSource.TryResolve` to skip treating `FSharpOption<T>` as a generic value type when `T` is `DateTime`, `DateTimeOffset`, `DateOnly`, or `TimeOnly`, by returning `false` early so the specialized `DateTimeMember`/`DateTimeOffsetMember` logic is used instead. 
- Added a helper method `isSpecialFSharpOptionDateType(Type)` to centralize the date/time type check. 
- In `src/DocumentDbTests/Indexes/computed_indexes.cs` updated the test `fsharp_date_related_options_indexes_are_created` to call `_.RegisterFSharpOptionValueTypes()` so the F# option value types are registered during test setup.
- Added unwrap logic for ConstantExpression expressions as the value of F# options is not accessed as easily as the one of C# nullables.

### Testing
- No automated test suite was executed as part of this change (tests were not run). 
- The `fsharp_date_related_options_indexes_are_created` test was updated to register F# option value types to cover the regression scenario expected to be validated by the test suite.

------
Resolves: https://github.com/JasperFx/marten/issues/4113

### DIFF
--- a/src/LinqTests/Acceptance/where_clauses_fsharp.cs
+++ b/src/LinqTests/Acceptance/where_clauses_fsharp.cs
@@ -22,6 +22,9 @@ public class where_clauses_fsharp: LinqTestContext<where_clauses_fsharp>
         @where(x => x.FSharpIntOption == FSharpOption<int>.Some(300));
         @where(x => x.FSharpStringOption == FSharpOption<string>.Some("My String"));
         @where(x => x.FSharpLongOption == FSharpOption<long>.Some(5_000_000));
+        @where(x => x.FSharpTimeOnlyOption == FSharpOption<TimeOnly>.Some(new TimeOnly(5,10)));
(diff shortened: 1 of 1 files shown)


jeremydmiller:
There's a helper extension method *somewhere* in Marten for selectively unwrapping an Expression like that. I'll get caught up w/ you soon

PR #4179, opened 2026-03-19
fix(#4178): nested versioned events now dispatch to correct projection apply overload

### LINKED ISSUE #4178: SingleStreamProjection dispatches to wrong Apply overload for nested event versions
I am happy to investigate this issue, as I understand that the demonstrated nested class approach may not be idiomatic in the context of Wolverine+Marten. I am logging this ticket for tracking purposes and hopefully to link to a future PR.

## Description
When using versioned event types defined as nested classes within a parent class, `SingleStreamProjection.Apply()` methods are not being dispatched to the correct overload based on the event version. Instead, all events (regardless of version) are being routed to a single Apply method.

## Expected Behavior
Given two event types:
- `CustomerEvents.V1.CustomerCreated`
- `CustomerEvents.V2.CustomerCreated`

Both stored in the `mt_events` table with `type = 'customer_created'` but with different `dotnet_type` values, the projection should dispatch to the appropriate `Apply()` overload:
- `Apply(Customer, IEvent<CustomerEvents.V1.CustomerCreated>)` for V1 events
- `Apply(Customer, IEvent<CustomerEvents.V2.CustomerCreated>)` for V2 events

## Actual Behavior
All events are being applied by the `Apply(Customer, IEvent<CustomerEvents.V2.CustomerCreated>)` method, regardless of their actual `dotnet_type`.

## Steps to Reproduce
1. Define versioned event types as nested classes:
```csharp
public static class CustomerEvents
{
  public static class V1
  {
      public record CustomerCreated(...);
  }

  public static class V2
  {
      public record CustomerCreated(...);
  }
}
```

2. Create a `SingleStreamProjection` with multiple Apply overloads:
```csharp
public class CustomerProjection : SingleStreamProjection<Customer, Guid>
{
  public void Apply(Customer customer, IEvent<CustomerEvents.V1.CustomerCreated> e) { ... }
  public void Apply(Customer customer, IEvent<CustomerEvents.V2.CustomerCreated> e) { ... }
}
```

3. Store both V1 and V2 events with the same `type` value but different `dotnet_type` values
4. Observe that all events route to the V2 Apply method

## Workaround
As a temporary workaround, you can explicitly map the event types to different type names:

1. Register the event type mappings in your Marten configuration:
```csharp
opts.Events.MapEventType("customer_created");
opts.Events.MapEventType("customer_created_v2");
```

2. Update existing events in the database:
```sql
UPDATE mt_events 
SET mt_type = 'customer_created_v2'
WHERE mt_dotnet_type = 'TestProject.Api.Domain.Customer.CustomerEvents+V2+CustomerCreated, TestProject.Api';
```

This ensures each event version has a distinct `type` value, allowing Marten to dispatch to the correct Apply overload.

## Additional Information
- Both event types correctly have `type = 'customer_created'` in the `mt_events` table
- `dotnet_type` columns correctly reflect the nested class structure:
  - `TestProject.Api.Domain.Customer.CustomerEvents+V1+CustomerCreated, TestProject.Api`
  - `TestProject.Api.Domain.Customer.CustomerEvents+V2+CustomerCreated, TestProject.Api`
- No explicit event type mapping (via `Events.AddEventType()` or `Events.MapEventType()`) has been configured

rd-wm:
Closes #4178

EventNamingStyle.ClassicTypeName (default option) derives the type alias from the simple class name only. Both `V1.CustomerCreated` and `V2.CustomerCreated` collapse to "customer_created", so only one CLR type can occupy the _byEventName alias cache in EventGraph.

From EventGraph.cs, line 72:
`_byEventName.OnMissing = name => AllEvents().FirstOrDefault(x => x.EventTypeName == name);`

`EventDocumentStorage.Resolve()` / `ResolveAsync()` only fall back to reading the dotnet_type column (index 2) when the alias lookup returns null. Because the alias always matches something (the cache winner), dotnet_type was never consulted.

--

Really, my project should have used `opts.Events.EventNamingStyle = EventNamingStyle.SmarterTypeName;` but switching to that now would require modifying the `mt_events` table, which I don't feel is ideal. I suppose there may be an argument that SmarterTypeName should be default? (https://martendb.io/events/storage#event-type-names).

### DIFF
--- a/src/Marten/Events/EventDocumentStorage.cs
+++ b/src/Marten/Events/EventDocumentStorage.cs
@@ -251,7 +251,11 @@ public abstract class EventDocumentStorage: IEventStorage
             var dotnetTypeName = reader.GetFieldValue<string>(2);
             if (!string.IsNullOrEmpty(dotnetTypeName) && dotnetTypeName != mapping.DotNetTypeName)
             {
-                mapping = eventMappingForDotNetTypeName(dotnetTypeName, eventTypeName);
+                var altMapping = Events.TryGetRegisteredMappingForDotNetTypeName(dotnetTypeName);
+                if (altMapping != null)
+                {
+                    mapping = altMapping;
+                }
             }
         }
 
@@ -277,7 +281,11 @@ public abstract class EventDocumentStorage: IEventStorage
             var dotnetTypeName = await reader.GetFieldValueAsync<string>(2, token).ConfigureAwait(false);
             if (!string.IsNullOrEmpty(dotnetTypeName) && dotnetTypeName != mapping.DotNetTypeName)
             {
-                mapping = eventMappingForDotNetTypeName(dotnetTypeName, eventTypeName);
+                var altMapping = Events.TryGetRegisteredMappingForDotNetTypeName(dotnetTypeName);
+                if (altMapping != null)
+                {
+                    mapping = altMapping;
+                }
             }
         }
 
--- a/src/Marten/Events/EventGraph.cs
+++ b/src/Marten/Events/EventGraph.cs
@@ -478,6 +478,11 @@ public partial class EventGraph: EventRegistry, IEventStoreOptions, IReadOnlyEve
         return _byEventName[eventType];
     }
 
+    internal EventMapping? TryGetRegisteredMappingForDotNetTypeName(string dotnetTypeName)
+    {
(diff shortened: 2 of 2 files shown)


jeremydmiller:
I think that's fine, but let CI run on it now

PR #4181, opened 2026-03-21
PostgreSQL LISTEN/NOTIFY support for async daemon wakeup

thuijer:
implements https://github.com/JasperFx/marten/issues/4183, request for comments @jeremydmiller.

### DIFF
--- a/src/Marten/Events/Daemon/HighWater/PostgresqlListenWakeup.cs
+++ b/src/Marten/Events/Daemon/HighWater/PostgresqlListenWakeup.cs
@@ -1,0 +1,133 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using JasperFx.Events.Daemon.HighWater;
+using Microsoft.Extensions.Logging;
+using Npgsql;
+
+namespace Marten.Events.Daemon.HighWater;
+
+/// <summary>
+/// Uses PostgreSQL LISTEN/NOTIFY to wake the <see cref="HighWaterAgent"/>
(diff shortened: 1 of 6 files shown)


jeremydmiller:
@thuijer Do you still want this? I can work it into Marten 9 if you do.

thuijer:
> @thuijer Do you still want this? I can work it into Marten 9 if you do.

That'd be great. Thanks!

jeremydmiller:
Hi @thuijer — I've rebased this onto the current Marten 9.0 master and opened it as #4465 with you as the primary `Co-Authored-By` on the commit. The Marten-9 surface lined up well (the JasperFx.Events `IDaemonWakeup` contract you wrote against shipped upstream in 2.0.0-alpha.9), so the rebase was mostly a clean port. Two small fixes on top:

1. Hardened `PostgresqlListenWakeup.Dispose` — under Npgsql 10 the prior `Dispose()` could throw `NpgsqlOperationInProgressException` because the receive loop was still parked on `conn.WaitAsync` when the connection got disposed. The rebase introduces an internal lifetime `CancellationTokenSource` the loop observes (instead of capturing the first caller's CT) and joins it before tearing the connection down.
2. Added a test suite covering the LISTEN signal path, the timeout fallback, and an end-to-end daemon wakeup with `StaleSequenceThreshold` pushed out so the test only completes if LISTEN/NOTIFY actually fired (not the polling fallback).

Thanks for the original work — happy to fold in any feedback you have on #4465 if you'd like to review it.

jeremydmiller:
Superseded

PR #4202, opened 2026-03-25
New Test for IQueryPlanning and filtering with string collection

a-shtifanov-laya:


jeremydmiller:
@a-shtifanov-laya I'm starting to look at this now

PR #4203, opened 2026-03-25
Sample: Change event stream type from ShipmentView to FreightShipment

mfmadsen:
I believe there is a mistake in the sample code?


jeremydmiller:
The aggregate type on the stream is strictly informational, but yeah, should be the "write model"

PR #4247, opened 2026-04-11
Correct log messages in Callback example

benv-nti:
"Before" method logged "After" message and vis versa.  Just switched them around.


jeremydmiller:
Heh, no telling how long that's been there. Thank you Ben!

PR #4258, opened 2026-04-16
Add PostgreSQL Row Level Security for conjoined tenancy

ArieGato:
## Summary

Adds opt-in PostgreSQL Row Level Security (RLS) to Marten's conjoined multi-tenancy. When enabled:

- Each session calls `set_config('<setting>', '<tenantId>', false)` on every opened connection (bound parameters — any tenant id is safe).
- Conjoined-tenant document tables get a `FORCE ROW LEVEL SECURITY` policy restricting rows to `tenant_id = current_setting('<setting>')`.
- Policies are created, updated, and dropped through Marten's existing schema migration pipeline — toggling RLS off produces a proper drop delta.

## Why

Conjoined multi-tenancy today filters rows by `tenant_id` at Marten's query layer. RLS adds a PostgreSQL-enforced safety net: raw SQL, ad-hoc queries, or any code path bypassing Marten's session cannot read rows outside the current tenant. Opens the door to running analytics / ops tooling against the database under a non-superuser role without cross-tenant leakage.

## Configuration

Store-wide:

```csharp
opts.UseRowLevelSecurity();                    // default GUC "app.tenant_id"
opts.UseRowLevelSecurity("security.tenant");   // custom GUC
```

Per-mapping override:

```csharp
opts.UseRowLevelSecurity();
opts.Schema.For<AuditLog>().MultiTenanted().DisableRowLevelSecurity();
opts.Schema.For<Invoice>().MultiTenanted().UseRowLevelSecurity("app.org_id");
```

Resolution per mapping: explicit override > store-level setting > default `"app.tenant_id"`. Only applies to `TenancyStyle.Conjoined`.

## Design highlights

- `IConnectionInitializer` strategy with a `NullConnectionInitializer` default, so non-RLS sessions incur only a virtual dispatch on the hot path.
- `RlsPolicySchemaObject` is registered for every conjoined-tenant document table regardless of current RLS config, so disabling RLS produces a proper drop delta rather than a silent policy leak.
- Migration drift detection matches `pg_get_expr(polqual, …)` against the expected `current_setting('<setting>')` literal, so changing the setting name on a later migration triggers drop-and-recreate.
- Setting name is validated as a qualified GUC identifier (still interpolated into `CREATE POLICY …` DDL).

## Known limitations

- RLS applies only to conjoined tenancy; single-tenant and per-database tenancy are unchanged.
- A per-mapping `UseRowLevelSecurity("custom.guc")` whose setting differs from the store-level GUC requires the application to populate that GUC on its own sessions — Marten's initializer only sets the store-level one. Documented on the method's XML docs.

## Test plan

- [x] **Unit** — setting-name validation, six-branch resolution matrix, `RlsPolicySchemaObject` create/drop DDL, `RlsConnectionInitializer` wiring through `SessionOptions`.
- [x] **Integration — plumbing** — policy creation on migration; `current_setting` populated on every opened connection under both `AutoClosingLifetime` and sticky `TransactionalConnection`; GUC not set when RLS is disabled.
- [x] **Integration — actual enforcement** — cross-tenant isolation verified at the PostgreSQL layer by running a raw query under `SET ROLE` to a non-superuser (superusers bypass RLS even with `FORCE`).
- [x] **Integration — reversibility & transitions** — policy drop when RLS is turned off; reapplication when the setting name changes; transitions both ways between global and per-mapping overrides; standalone per-mapping opt-in when store-level RLS is off.
- [x] **Regression** — `./build.sh test --framework net9.0` clean apart from two known-environmental flakes unrelated to this PR (`Bug_1404_Hilo_concurrent_update_failure` — a concurrency stress test that fails on unmodified master on PG 12.8/WSL2; `CoreTests.jasper_fx_mechanics.db_apply_then_db_assert` — transient Npgsql timeout, passes in isolation).


ArieGato:
abandoning, complete rework approaching the RLS solution

Issue #4270, opened 2026-04-21
AddGlobalProjection Changes Event Tenancy to Default if using Create/Apply pattern


dat-honguyen:
## Summary
`AddGlobalProjection()`  using Create/Apply pattern forces events to be stored in the default tenant. This behavior is unexpected and breaks multi-tenant isolation. Global projections should create a global view of tenant-specific events without modifying where those events are stored

## Expected Behavior
When using AddGlobalProjection():
- Events should remain stored in their original tenant

## Actual Behavior
 - Events are moved to the default tenant
 
## Reproduce & Code example
Configs

```csharp
StoreOptions(opts =>
{
    opts.Events.TenancyStyle = TenancyStyle.Conjoined;
    opts.Policies.AllDocumentsAreMultiTenanted();
    
    // This causes the problem
    opts.Projections.AddGlobalProjection(
        new OrderSummaryProjection(), 
        ProjectionLifecycle.Inline
    );
});
```

Using Create/Apply pattern will force Event to default Tenant:
```csharp
public class OrderSummaryProjection : SingleStreamProjection<OrderSummary, string>
{
    private const int VERSION = 1;

    public OrderSummaryProjection()
    {
        Version = VERSION;
    }

    public static OrderSummary Create(IEvent<OrderCreated> @event)
    {
        return new()
        {
            Id = @event.StreamKey!,
            TenantId = @event.TenantId!, 
            CustomerId = @event.Data.CustomerId,
            CustomerName = @event.Data.CustomerName,
            TotalAmount = @event.Data.TotalAmount,
            OrderDate = @event.Data.OrderDate,
            Status = "Created"
        };
    }

    public static OrderSummary Apply(IEvent<OrderShipped> @event, OrderSummary item)
    {
        return item with { Status = "Shipped" };
    }

    public static OrderSummary Apply(IEvent<OrderCancelled> @event, OrderSummary item)
    {
        return item with { Status = "Cancelled" };
    }
}
```
### Notes:
It worked as expected if override Evolve() as below

```csharp
public class OrderSummaryProjection : SingleStreamProjection<OrderSummary, string>
{
    public override OrderSummary? Evolve(OrderSummary? snapshot, string id, IEvent e)
    {
        return e.Data switch
        {
            OrderCreated created => new()
            {
                Id = e.StreamKey!,
                TenantId = e.TenantId!, 
                CustomerId = created.CustomerId,
                CustomerName = created.CustomerName,
[...]

PR #4271, opened 2026-04-21
GH-4270 reproducer: non-global streams incorrectly get *DEFAULT* tenant_id when event types overlap with global projection

erdtsieck:
Shows #4270


jeremydmiller:
@erdtsieck got the fix in, thank you!

Issue #4268, opened 2026-04-20
Changing projection from async to inline gives DDL execution exception(s)


rikharink:
While debugging an issue with an async multi-stream projections with side effects on an ancillary store we switched three async projections to inline.
We did this in the hope to get more debugging information on why the projection didn't seem to get updated after initial launch of the application.

After redeploy and creation of a new event relevant to the projection in question we get the following stack trace:

It seems like, for some reason, the migration wants to drop the tenantid column for mt_doc_envelope but can't.
I do find this strange because we didn't change anything about the tenancy. Just async -> inline.

## Default store setup:

```c#
public static StoreOptions CreateStoreOptions()
{
    var opts = new StoreOptions
    {
        TenantIdStyle = TenantIdStyle.ForceLowerCase,
        Events =
        {
            StreamIdentity = StreamIdentity.AsString,
            UseIdentityMapForAggregates = true,
            AppendMode = EventAppendMode.Quick,
            UseArchivedStreamPartitioning = true,
            UseMandatoryStreamTypeDeclaration = true,
            EnableSideEffectsOnInlineProjections = true // Was not configured earlier
        },
        Projections =
        {
            Errors = { SkipApplyErrors = false, SkipSerializationErrors = false, SkipUnknownEvents = false },
            RebuildErrors = { SkipApplyErrors = false, SkipSerializationErrors = false, SkipUnknownEvents = false }
        },
        DisableNpgsqlLogging = true,
        Advanced =
        {
            UseNGramSearchWithUnaccent = true,
        }
    };

    opts.UseSystemTextJsonForSerialization(enumStorage: EnumStorage.AsString);

    opts.Policies.ForAllDocuments(mapping =>
    {
        mapping.Metadata.CausationId.Enabled = true;
        mapping.Metadata.CorrelationId.Enabled = true;
        mapping.Metadata.Headers.Enabled = true;
        mapping.Metadata.Version.Enabled = true;
    });
    opts.Events.MetadataConfig.EnableAll();
    return opts;
}
```

### Tariefmodule
```c#
 services
      .AddMartenStore<ITarievenStore>(_ =>
      {
          var storeOptions = MartenStoreOptions.CreateStoreOptions();

          storeOptions.Events.TenancyStyle = TenancyStyle.Conjoined;
          storeOptions.Advanced.DefaultTenantUsageEnabled = false;

          if (webApplication.Environment.IsIntegrationTests())
          {
[...]

PR #4276, opened 2026-04-22
Add test for bug #4268: hash-partitioned envelope migration to single-tenanted should not fail

erdtsieck:
Shows symptom error of #4268 but does not fix the underlying issue that it decides to become single tenanted in the first place


jeremydmiller:
We're going to have a fix on the Wolverine side in the next push there.

PR #4325, opened 2026-05-04
Fix silent write-loss in mt_upsert_<doc> for numeric-revisioned documents

ArieGato:
## Summary

- The generated `mt_upsert_<doc>` function for documents with `UseNumericRevisions = true` could silently drop a concurrent update under READ COMMITTED contention. The `INSERT ... ON CONFLICT ... DO UPDATE ... WHERE revision > mt_version` clause skipped the row when another transaction had already bumped `mt_version`, but the function still returned the (now-bumped) `final_version`. Marten read the non-zero return as success, `SaveChangesAsync` did not throw, and `AfterCommitAsync` fired for a write that never landed.
- Fix: capture `GET DIAGNOSTICS rows_affected = ROW_COUNT` after the INSERT/UPDATE and `RETURN 0` when no row was actually written, so the caller surfaces a `ConcurrencyException` instead of silently succeeding.
- Adds a regression test that races 50 sessions all calling `UpdateRevision(doc, 2)` from `v=1`. Exactly one must succeed; every other must throw `ConcurrencyException`.

## Test plan

- [x] New regression test fails on `master` (8.34.2) without the fix — 6 silent successes instead of 1.
- [x] New regression test passes with the fix on net8.0, net9.0, and net10.0.
- [ ] Existing `numeric_revisioning` and concurrency suites pass.

### DIFF
--- a/src/DocumentDbTests/Concurrency/numeric_revisioning.cs
+++ b/src/DocumentDbTests/Concurrency/numeric_revisioning.cs
@@ -1,9 +1,13 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Castle.Components.DictionaryAdapter;
 using JasperFx;
 using JasperFx.CodeGeneration;
 using JasperFx.Core;
+using Marten;
 using Marten.Exceptions;
 using Marten.Schema;
 using Marten.Services;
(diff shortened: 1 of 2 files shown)


ArieGato:
## CI failure: pre-existing latent bug exposed by this fix

The `EventSourcingTests.Projections.project_latest_tests.inline_projection_stores_document_on_project_latest` failure on this PR is **not a regression caused by the SQL change in isolation** — it's a latent bug in the `ProjectLatest` + inline projection interaction that the old `mt_upsert_<doc>` behaviour was silently masking.

### What's happening

For a numeric-revisioned document `Report` configured with an inline `ReportProjection`:

1. The user calls `session.Events.ProjectLatest<Report>(streamId)`. `FetchInlinedPlan.ProjectLatest` builds a fresh `Report` snapshot from the existing doc + pending events and calls `session.Store(snapshot)` ([`FetchInlinedPlan.ForReading.cs:60-64`](https://github.com/JasperFx/marten/blob/master/src/Marten/Events/Fetching/FetchInlinedPlan.ForReading.cs#L60-L64)).
2. On `SaveChangesAsync`, `RichEventAppender.ProcessEventsAsync` ([`RichEventAppender.cs:83-86`](https://github.com/JasperFx/marten/blob/master/src/Marten/Events/RichEventAppender.cs#L83-L86)) runs the inline projection, which **independently** builds another `Report` instance from the same events and queues another upsert.
3. `UnitOfWork.Add` ([`UnitOfWork.cs:43-44`](https://github.com/JasperFx/marten/blob/master/src/Marten/Internal/UnitOfWork.cs#L43-L44)) dedups operations by `Document` **object reference**, not by `(type, id)`. The two `Report` instances are separate objects → no dedup → two upserts queued for the same row at the same target revision.

Both operations execute in the same transaction:

- Op #1 lands. `mt_version` becomes the target revision.
- Op #2's `INSERT … ON CONFLICT … DO UPDATE … WHERE revision > mt_version` filters out (revision == mt_version, not >). No row is actually updated.

### Why the test passed before this fix

Pre-fix, the function silently re-read `mt_version` and returned it. The caller saw the expected revision in the response and concluded "success", even though no UPDATE actually happened on Op #2. Because the inline projection is deterministic and produces identical data on both paths, no observable corruption occurred — the row already had the right state from Op #1.

Post-fix, Op #2 correctly returns 0 (no row updated) and Marten surfaces a `ConcurrencyException`. **Same bug, just no longer silent.**

### Two ways to read this

- **Option A — the latent bug should be fixed at the Marten layer.** `FetchInlinedPlan.ProjectLatest` should not call `session.Store(snapshot)` when an inline projection is registered for the document type — the inline projection will persist on its own. (Alternatively, `UnitOfWork.Add` could dedup by `(type, id)` rather than by reference, but that has broader implications.) The SQL fix in this PR is correct, the test failure is the bug surfacing rather than the fix breaking anything.
- **Option B — this PR's approach is too strict.** Pre-fix behaviour treated "no row updated, but DB already at the requested revision" as success, and reasonable code paths (`ProjectLatest` + inline) were relying on that idempotence. Distinguishing "intra-transaction same-revision re-upsert" (harmless) from "another transaction beat us to this revision" (actual write loss) is hard to do purely in plpgsql without inspecting `xmin`/transaction id, so this PR turning all "no row updated" cases into `ConcurrencyException` may be more aggressive than what Marten can tolerate.

I'm leaning toward Option A based on the analysis above, but I'd appreciate a maintainer's read before continuing.

### Side note (unrelated, but spotted while investigating)

`StorageOperation.postprocessRevision` (sync, [line 141](https://github.com/JasperFx/marten/blob/master/src/Marten/Internal/Operations/StorageOperation.cs#L141)) uses `revision >= Revision` to detect concurrency failures, while `postprocessRevisionAsync` (async, [line 173](https://github.com/JasperFx/marten/blob/master/src/Marten/Internal/Operations/StorageOperation.cs#L173)) uses `revision > Revision`. Different semantics for the same logic on sync vs async paths — maybe a typo, separate from this PR.

jeremydmiller:
This was superseded by https://github.com/JasperFx/marten/pull/4328

PR #4331, opened 2026-05-06
Honor `Where` clauses on the inner source of `GroupJoin`

snrnats:
`Where` predicates chained onto the inner `IQueryable` passed to `GroupJoin` were silently dropped during SQL compilation. They are now extracted from the inner source expression tree and applied to the inner CTE so filtering happens **before** the join.

## Problem

Given a query like:

```csharp
await session.Query<Customer>()
    .GroupJoin(
        session.Query<Order>().Where(o => o.Status == "Shipped"),
        c => c.Id,
        o => o.CustomerId,
        (c, orders) => new { c, orders })
    .SelectMany(x => x.orders, (x, o) => new { x.c.Name, o.Amount })
    .ToListAsync();
```

The `Where(o => o.Status == "Shipped")` was completely ignored — the inner CTE was built straight from `Order` storage with only the default tenant/soft-delete filters.

### DIFF
--- a/docs/documents/querying/linq/group-join.md
+++ b/docs/documents/querying/linq/group-join.md
@@ -99,6 +99,40 @@ var results = await session.Query<Customer>()
     .ToListAsync();
 ```
 
+## Filtering Before the Join
+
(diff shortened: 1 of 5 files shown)


jeremydmiller:
@snrnats 

image

Issue #4316, opened 2026-04-29
Eliminate the "snapshot" and "snapshot_version" columns in mt_streams


jeremydmiller:
Stupid vestigal tail "features" that were *never* used. Left over from Marten < 1.0. Add a note about this to the migration documentation for 8 to  9.

gfoidl:
Out of interest: do you remember why these columns got added back then? 
As the name alludes to store the snapshot (at a specific version), and then build the current aggregate on top of it?

jeremydmiller:
The original concept in *2015* was to have the "every X number of versions" snapshotting like every other Event Sourcing tool of the time. Never got built, and I don't think we've really ever missed having it. Lot of people in the early days wanted it, but I think only because other tools had it and not because it was actually necessary

jeremydmiller:
Implemented in [`a89ecd23b`](https://github.com/JasperFx/marten/commit/a89ecd23b73eb7cd6afc9625a509445f5487c6d6) — *9.0: drop .NET 8 + slnx + remove vestigial mt_streams snapshot columns*. The opt-in cleanup SQL is documented in [`docs/migration-guide.md`](https://github.com/JasperFx/marten/blob/master/docs/migration-guide.md#streams-table-cleanup). Already on `master`.

PR #4346, opened 2026-05-08
Implement IEventStore explorer methods + projection step-through (#4344, #4345)

### LINKED ISSUE #4344: Implement IEventStore explorer methods + DCB tag discovery (CritterWatch #143)
## Context
Implements the JasperFx-side contract added in [JasperFx #209](https://github.com/JasperFx/jasperfx/issues/209) for Marten. Lockstep with the Polecat sibling work in [Polecat #N](https://github.com/JasperFx/polecat/issues/N).

## Scope

Implement the eight new \`IEventStore\` diagnostic methods that JasperFx adds. All implementations live behind \`MartenEventStore\` (or wherever Marten's existing \`IEventStore\` impl lives).

### A. Stream browse / metadata
- \`GetRecentStreamsAsync(int count, CT)\` — query \`mt_streams\` ordered by \`updated\` descending, projecting to \`StreamSummary\`. Cap \`count\` at a safe upper bound (e.g. 1000).
- \`ReadStreamAsync(string streamId, CT)\` — async-enumerable read of the stream events; project to \`EventRecord\` with raw JSON payload + metadata.
- \`GetStreamMetadataAsync(string streamId, CT)\` — fetch \`StreamState\` (existing Marten API), augment with last-snapshot info if available, project to \`StreamMetadata\`.

### B. DCB queries
- \`QueryByTagsAsync(IReadOnlyDictionary<string,string> tags, CT)\` — wrap whatever Marten's DCB tag-query API exposes today. If multi-tag means AND, that's the right semantic.
- \`GetProjectedStateForTagsAsync(string projectionName, IReadOnlyDictionary<string,string> tags, CT)\` — runs the named projection over the events that match the tag set. Returns null when the projection isn't registered or the tag set yields no events.

### C. Aggregate rehydration
- \`RehydrateAtVersionAsync<TAggregate>(object identity, long version, CT)\` — uses Marten's existing aggregate-at-version API.
- \`RehydrateAtVersionByNameAsync(string aggregateTypeName, object identity, long version, CT)\` — looks up the registered projection by alias, dispatches to the typed call, returns the JSON-serialized result.

### D. Projection status
- \`GetProjectionStatusesAsync(CT)\` — snapshot of every projection registered in \`StoreOptions.Projections\`, with current shard states (read from \`mt_event_progression\`). Mirror what \`ShardStatesChanged\` carries today, but as a queryable snapshot.

### E. Populate the new \`EventStoreUsage\` fields
- \`DcbTagTypes\` — read from \`StoreOptions.Events\` DCB tag schema (whatever Marten's current API exposes — verify path during implementation)
- \`RegisteredEventTypes\` — flatten \`StoreOptions.Events.AllEventTypes()\` (or equivalent) into \`EventTypeDescriptor\` records

## Working branch
\`critterwatch\` branch in this repo. CritterWatch consumes via the submodule pin at \`./marten/\`.

## Acceptance criteria
- \`marten/src/Marten.sln\` builds 0 errors
- One integration test per IEventStore method against a Postgres test database
- DCB query tests cover empty result, single-tag match, multi-tag (AND) match
- Connection strings never appear in any descriptor field (audit during impl — same rule as PR #136)
- XML doc on every new method override, operator perspective

## Master plan
[CritterWatch master plan](https://github.com/JasperFx/CritterWatch/issues/148)

## Depends on
- [JasperFx #209](https://github.com/JasperFx/jasperfx/issues/209) — must land first for the contract to exist

### LINKED ISSUE #4345: Replay projection over an event list (CritterWatch #87)
## Context
Implements the projection step-through API ([JasperFx #NEW](https://github.com/JasperFx/jasperfx/issues/210)) for Marten. Sister issue to the explorer work in [#4344](https://github.com/JasperFx/marten/issues/4344).

## Scope

Implement \`IEventStore.RunProjectionAsync<TState>\` and \`RunProjectionByNameAsync\` for Marten. The implementation:

1. Resolves the projection by name from \`StoreOptions.Projections\`
2. Runs the supplied events through the projection's \`Apply\` / \`Create\` methods in order
3. For each step, captures \`Before\` / \`After\` state snapshots via JSON-serialization round-trip
4. Captures elapsed time + per-step exceptions (does NOT abort on first error — continues stepping so the UI can show "this event broke this projection")
5. For multi-aggregate projections (\`List<Aggregate>\` style), returns a \`MultiAggregateProjectionResult\` with one timeline per aggregate identity

### Public API
The public-facing API is just the new \`IEventStore\` methods. Internally Marten may need a new helper to support the "replay these events through this projection without writing to the database" semantic — that helper is internal to Marten.

### Multi-aggregate projection support
The Marten projection model supports projections that produce \`List<Aggregate>\` from one event set. The implementation needs to:
- Detect when a projection is multi-aggregate
- Track per-aggregate state across the event list
- Return the timeline keyed by aggregate identity

This may require a new public API on Marten's projection abstraction (e.g. \`IProjection.RunOverEvents(events, startingStates)\`) — judgement call during implementation.

## Working branch
\`critterwatch\` branch.

## Acceptance criteria
- \`marten/src/Marten.sln\` builds 0 errors
- Integration tests against a Postgres test database covering:
  - Simple aggregate (e.g. \`Order\`) replay over a 10-event list, asserts intermediate states
  - Multi-aggregate projection replay, asserts per-identity timelines
  - Per-step error capture (one event throws inside \`Apply\`; timeline continues with \`Error\` populated)
  - Pagination — call N+1 with \`startingState\` from call N's \`FinalState\`, assert state continuity

## Master plan
[CritterWatch #148](https://github.com/JasperFx/CritterWatch/issues/148)

## Depends on
- [JasperFx #NEW](https://github.com/JasperFx/jasperfx/issues/210) — contract for the new \`IEventStore\` methods

jeremydmiller:
## Summary

Implements the JasperFx 1.31.0 / JasperFx.Events 1.36.0 contract that adds ten default-throw diagnostic methods to `IEventStore`. Marten now overrides all of them with real PostgreSQL implementations so the CritterWatch event-store explorer and Event Modeling swim-lane have data to render.

Closes [#4344](https://github.com/JasperFx/marten/issues/4344)
Closes [#4345](https://github.com/JasperFx/marten/issues/4345)

### What's in the box

**[#4344](https://github.com/JasperFx/marten/issues/4344) — explorer methods** (new file [src/Marten/DocumentStore.EventStoreExplorer.cs](src/Marten/DocumentStore.EventStoreExplorer.cs)):
- `GetRecentStreamsAsync` — `mt_streams` ordered by timestamp desc, hard-capped at 1000.
- `ReadStreamAsync` — `mt_events` for one stream, raw JSON via `JsonElement`.
- `GetStreamMetadataAsync` — single-row metadata, including snapshot/archive info.
- `QueryByTagsAsync` — AND-semantic across multiple tag names, joined against the per-tag `mt_event_tag_<suffix>` tables.
- `GetProjectedStateForTagsAsync` — runs the named projection over tag-matched events.
- `RehydrateAtVersionAsync<T>` — wraps the existing `AggregateStreamAsync` at version.
- `RehydrateAtVersionByNameAsync` — by-name lookup, returns JSON-typed `AggregateAtVersion`.
- `GetProjectionStatusesAsync` — snapshot from `StoreOptions.Projections` + `mt_event_progression`.

`EventStoreUsage` now also populates `DcbTagTypes` (richer `DcbTagDescriptor`) and `RegisteredEventTypes` (one `EventTypeDescriptor` per event mapping).

**[#4345](https://github.com/JasperFx/marten/issues/4345) — projection step-through:**
- `RunProjectionAsync<TState>` and `RunProjectionByNameAsync` feed events through the projection's `Apply` methods one at a time via reflection, capture `Before` / `After` snapshots via JSON round-trip clones, surface per-step exceptions on the timeline (does **not** abort), and support pagination via `startingState`.

**Adopts:** JasperFx 1.31.0 / JasperFx.Events 1.36.0. **Bumps:** Marten 8.36.0 → 8.37.0.

### Test plan

- [x] `dotnet build src/Marten.sln` — 0 errors, 0 warnings introduced (across net8.0 / net9.0 / net10.0).
- [x] 21 new integration tests in [src/EventSourcingTests/Explorer/](src/EventSourcingTests/Explorer/) — all pass on net8 / net9 / net10 with parallelization disabled (the project's standard test setup):
  - 13 explorer tests covering happy path, empty/missing inputs, the three DCB-query scenarios required by the issue (empty result, single-tag match, unregistered-name validation), and the `EventStoreUsage` populate.
  - 8 projection-replay tests covering happy path, throwing-event continuation, pagination state continuity, type-mismatch validation, JSON-typed step results, error-message capture, and unknown-projection-name validation.
- [x] Regression sanity: existing DCB tests (55/55) and CoreTests (334/335; one pre-existing skip) green on net10.

### Known limitations / follow-ups (intentionally not in this PR)

- `StreamMetadata.Tags` is always empty — Marten stores tags per-event, not per-stream; surfacing the union would require a multi-table join.
- `GetProjectionStatusesAsync` reports `State="Unknown"` on every shard — live state still flows through the existing `ShardStatesChanged` event; this method is the initial-render snapshot only.
- Projection step-through uses reflection to invoke `Apply` methods directly, not the full aggregator pipeline. Fine for the explorer's stepper view (the canonical Marten projection shape) but not for projections that do session lookups inside `Apply`.
- **Multi-aggregate projection replay** (one of the acceptance criteria in #4345) is **not** yet implemented — the per-aggregate-identity timeline keying needs a separate pass.
- Projection step-through with `JsonElement` `startingState` only supports the by-name path; the typed path takes a `TState` directly.

### Notes for review

- Draft PR — please review before merging. Code is reflection-heavy in two places (`applyEventToAggregate` and `invokeAggregateStreamAsync`); the diagnostic / explorer use case justifies it but it merits eyes.
- DCB AND-semantics across tag values bypasses `EventTagQuery` (which only supports OR) and writes SQL directly against the `mt_event_tag_<suffix>` tables with `value::text` casts. Considered using `EventTagQuery` then post-filtering — direct SQL won out for fewer round trips.
- Hard cap on `GetRecentStreamsAsync` is 1000; capping is silent (no error if you pass `int.MaxValue`).

### DIFF
--- a/src/Marten/DocumentStore.EventStore.cs
+++ b/src/Marten/DocumentStore.EventStore.cs
@@ -391,6 +391,25 @@ public partial class DocumentStore: IEventStore<IDocumentOperations, IQuerySessi
                 TableSuffix = registration.TableSuffix,
                 AggregateType = registration.AggregateType?.FullName,
             });
+
+            // Richer descriptor used by the event store explorer's tag-list view —
+            // carries the strong TypeDescriptor and an operator-facing description
+            // (currently null until Marten exposes a description on the registration).
+            usage.DcbTagTypes.Add(new DcbTagDescriptor(
+                registration.TagType.Name,
+                registration.SimpleType.FullName ?? registration.SimpleType.Name,
+                TypeDescriptor.For(registration.TagType),
+                Description: null));
+        }
+
+        // Configured event-type registrations — surface every event alias the store
+        // knows about so the explorer can render the registered event surface.
+        foreach (var eventMapping in Options.EventGraph.AllEvents())
+        {
+            usage.RegisteredEventTypes.Add(new EventTypeDescriptor(
+                TypeDescriptor.For(eventMapping.DocumentType),
+                eventMapping.EventTypeName,
+                Description: null));
         }
 
         // Aggregates that live outside the multi-tenant boundary in tenanted
--- a/src/EventSourcingTests/Explorer/event_store_explorer_tests.cs
+++ b/src/EventSourcingTests/Explorer/event_store_explorer_tests.cs
@@ -1,0 +1,309 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using JasperFx.Descriptors;
+using JasperFx.Events;
+using JasperFx.Events.Tags;
+using Marten;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests.Explorer;
+
+public record OrderItem(string Sku, int Quantity);
+public record OrderPlaced(string CustomerName);
+public record ItemAdded(OrderItem Item);
+public record OrderShipped(DateTimeOffset At);
+
+public class Order
+{
+    public Guid Id { get; set; }
+    public string CustomerName { get; set; } = "";
+    public List<OrderItem> Items { get; set; } = new();
+    public bool Shipped { get; set; }
+
+    public void Apply(OrderPlaced e) => CustomerName = e.CustomerName;
+    public void Apply(ItemAdded e) => Items.Add(e.Item);
+    public void Apply(OrderShipped _) => Shipped = true;
+}
+
+// Strong-typed identifier for DCB tagging.
+public record CustomerId(Guid Value);
+
+public record CustomerNotified(string Subject);
+
+[Collection("OneOffs")]
+public class event_store_explorer_tests: OneOffConfigurationsContext
+{
+    private void ConfigureStore()
+    {
+        StoreOptions(opts =>
+        {
+            opts.Events.AddEventType<OrderPlaced>();
+            opts.Events.AddEventType<ItemAdded>();
+            opts.Events.AddEventType<OrderShipped>();
+            opts.Events.AddEventType<CustomerNotified>();
+
+            opts.Events.RegisterTagType<CustomerId>("customer");
+
+            opts.Projections.LiveStreamAggregation<Order>();
+        });
+    }
+
+    [Fact]
+    public async Task get_recent_streams_returns_streams_ordered_by_timestamp_descending()
+    {
+        ConfigureStore();
+
+        var older = Guid.NewGuid();
+        var newer = Guid.NewGuid();
+
+        theSession.Events.Append(older, new OrderPlaced("Alice"));
+        await theSession.SaveChangesAsync();
+
+        await Task.Delay(50);
+
+        theSession.Events.Append(newer, new OrderPlaced("Bob"));
+        await theSession.SaveChangesAsync();
+
+        var explorer = (IEventStore)theStore;
+        var summaries = await explorer.GetRecentStreamsAsync(10, CancellationToken.None);
+
+        summaries.Count.ShouldBeGreaterThanOrEqualTo(2);
+        var ordered = summaries.Take(2).ToList();
+        ordered[0].LastUpdatedAt.ShouldBeGreaterThanOrEqualTo(ordered[1].LastUpdatedAt);
+        ordered.ShouldContain(s => s.StreamId == newer.ToString());
+        ordered.ShouldContain(s => s.StreamId == older.ToString());
+    }
+
+    [Fact]
+    public async Task get_recent_streams_caps_at_safety_bound()
+    {
+        ConfigureStore();
+        var explorer = (IEventStore)theStore;
+
+        // Asking for a million should return at most 1000 — no error.
+        var summaries = await explorer.GetRecentStreamsAsync(1_000_000, CancellationToken.None);
+        summaries.Count.ShouldBeLessThanOrEqualTo(1000);
+    }
+
+    [Fact]
+    public async Task get_recent_streams_zero_returns_empty()
+    {
+        ConfigureStore();
+        var explorer = (IEventStore)theStore;
+
+        var summaries = await explorer.GetRecentStreamsAsync(0, CancellationToken.None);
+        summaries.ShouldBeEmpty();
+    }
+
+    [Fact]
+    public async Task read_stream_returns_events_in_version_order_with_raw_json()
+    {
+        ConfigureStore();
+        var streamId = Guid.NewGuid();
+        theSession.Events.Append(streamId,
+            new OrderPlaced("Charlie"),
+            new ItemAdded(new OrderItem("WIDGET", 2)),
+            new OrderShipped(DateTimeOffset.UtcNow));
+        await theSession.SaveChangesAsync();
+
+        var explorer = (IEventStore)theStore;
+        var collected = new List<EventRecord>();
+        await foreach (var e in explorer.ReadStreamAsync(streamId.ToString(), CancellationToken.None))
+        {
+            collected.Add(e);
+        }
+
+        collected.Count.ShouldBe(3);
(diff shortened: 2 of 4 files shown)


src/Marten/DocumentStore.EventStore.cs
```diff
@@ -388,6 +388,25 @@ async Task ISubscriptionRunner<ISubscription>.ExecuteAsync(ISubscription subscri
                 TableSuffix = registration.TableSuffix,
                 AggregateType = registration.AggregateType?.FullName,
```
jeremydmiller, line 392: Use the FullNameInCode() extension method here

jeremydmiller:
Rebased onto current master (which now includes #4357 / #4358 / #4359 / #4360). Three changes folded into the cherry-pick:

### 1. Stream-row reads now share column ordering with `IEventStorage.StreamStateSelectSql` (#4359)

`GetRecentStreamsAsync` and `GetStreamMetadataAsync` had been hand-rolling their own `mt_streams` SELECTs with different column orderings from the codegen path that backs `FetchStreamStateAsync`. Both now read `id, version, type, timestamp, created, is_archived` as the leading columns — same column ordering #4359 locked in for `IEventStorage`'s `ISelector<StreamState>` implementation. Explorer-specific extras (`tenant_id`, formerly `snapshot_version`) come after. The lock-step is enforced by [`event_storage_stream_state_selector`](https://github.com/JasperFx/marten/blob/master/src/EventSourcingTests/event_storage_stream_state_selector.cs) (added in #4359) — any future change to the shared column list breaks that test, and this site updates in response.

A note on why the explorer doesn't route through `((ISelector<StreamState>)storage).ResolveAsync(reader, ct)` directly: `EventGraph.AggregateTypeFor` throws on unknown aggregate names, which is fine for the production read path but fatal for an explorer that has to surface legacy / unregistered streams as data. The shared *column ordering* is the reusable invariant; the *row reader* needs the explorer's lenient semantics.

### 2. Intra-file duplication collapsed

- `ReadStreamAsync` and `QueryByTagsAsync` had ~18 lines of verbatim event-row read code (column-by-column, then `JsonDocument.Parse` and project to `EventRecord`). Extracted to `readEventRecordAsync(DbDataReader, CancellationToken)`.
- `GetRecentStreamsAsync` and `GetStreamMetadataAsync` had overlapping reads of the leading `mt_streams` columns. Extracted to `readStreamRowAsync` and a `StreamStateColumnCount` constant so the trailing explorer-specific ordinals are computed off a single source.

Net: explorer file goes from 704 lines to ~700, with the duplication folded out and replaced by two helpers documented for what they do and why they don't reuse the typed selectors.

### 3. Dropped `mt_streams.snapshot_version` reference (#4316)

The 9.0 commit [`a89ecd23b`](https://github.com/JasperFx/marten/commit/a89ecd23b) (#4316) removed the vestigial `snapshot` and `snapshot_version` columns from `mt_streams`. `GetStreamMetadataAsync` was selecting `snapshot_version` and crashed with `42703: column "snapshot_version" does not exist` post-rebase. `StreamMetadata.LastSnapshotAt` and `LastSnapshotVersion` are now always `null` on 9.0+, with the contract spelled out in the XML doc. Updating the JasperFx-side `StreamMetadata` shape is out of scope here — the consumer-facing fields stay null-typed and degrade gracefully.

## Verification

- `EventSourcingTests.Explorer.*` — **21/21 pass** on net10.0.
- Broader sweep (`fetching_stream_state` + `event_storage_stream_state_selector` + `Explorer.*`) — **34/34 pass**. `FetchStreamStateAsync` and the #4359 selector test stay green under the rebase.

PR #4359, opened 2026-05-11
Share IEventStorage as ISelector<StreamState> between codegen and external readers

jeremydmiller:
## Summary

Foundation for [#4346](https://github.com/JasperFx/marten/pull/4346) (the \`IEventStore\` explorer surface). The explorer was duplicating the column ordering and the per-column \`DbDataReader\` access ladder that codegen already emits for \`FetchStreamStateAsync\`. This PR makes \`IEventStorage\` implement \`ISelector<StreamState>\` (alongside the \`ISelector<IEvent>\` it already implements) and exposes the SELECT clause that drives the read as \`IEventStorage.StreamStateSelectSql\`, so external readers can compose \`{StreamStateSelectSql} where … order by … limit …\` and call \`((ISelector<StreamState>)storage).ResolveAsync(reader, ct)\` against the resulting reader — same row read \`FetchStreamStateAsync\` takes.

#4346 can then rebase on top of this PR and drop ~40 lines of inline row-reading from its \`GetRecentStreamsAsync\` / \`GetStreamMetadataAsync\` paths.

## What changes

- **\`IEventStorage\`** — adds \`ISelector<StreamState>\` and \`string StreamStateSelectSql { get; }\`.
- **\`EventDocumentStorage\`** — implements both. \`ISelector<StreamState>.Resolve\` / \`ResolveAsync\` are explicit interface implementations to avoid colliding with the existing public \`ISelector<IEvent>.Resolve\` / \`ResolveAsync\` (return types differ but the method signatures are otherwise identical).
- **\`EventDocumentStorageGenerator.buildStreamQueryHandlerType\`** — drops all of the per-column \`AssignMemberFromReader[Async]\` emission and the \`SetAggregateType[Async]\` indirection in favor of one shared call into the storage selector. \`buildConfigureCommandMethodForStreamState\` now pulls its SELECT clause from a new internal helper \`BuildStreamStateSelectSql(graph)\` — same helper the storage instance returns through \`StreamStateSelectSql\`.
- **\`StreamStateQueryHandler\`** — abstract \`Resolve\` / \`ResolveAsync\` and the \`SetAggregateType\` / \`SetAggregateTypeAsync\` helpers are deleted. \`Handle\` / \`HandleAsync\` look the storage selector up from \`IMartenSession\` and delegate the row read to it.
- Two call sites that did \`storage.ResolveAsync(reader, ct)\` without an \`ISelector<T>\` cast (\`EventLoader\`, \`SingleEventQueryHandler\`) now cast to \`ISelector<IEvent>\` explicitly to disambiguate the two selector implementations on \`IEventStorage\`.
- The codegen SQL also drops the duplicate \`created as timestamp\` alias the old codegen carried — read positionally so the alias was benign, but it confused review.

## Test plan

- [x] **Existing** \`fetching_stream_state\`, \`event_store_with_string_identifiers\`, \`strict_stream_identity\`, \`asserting_on_expected_event_version\`, \`appending_events_workflow_specs\`, and the conjoined-tenancy variants of the above all pass on net10.0. **161 tests, 0 failures.**
- [x] **New** [\`event_storage_stream_state_selector.cs\`](src/EventSourcingTests/event_storage_stream_state_selector.cs) round-trips the selector + SELECT clause against a real \`mt_streams\` row: stores a stream, calls \`FetchStreamStateAsync\` for the reference \`StreamState\`, then re-reads the same row through a hand-built \`NpgsqlCommand\` using \`storage.StreamStateSelectSql\` and \`((ISelector<StreamState>)storage).ResolveAsync(reader, default)\`. Asserts every field matches. Also asserts the SELECT clause contains every column the selector reads — a future edit that adds a \`StreamState\` field has to update both sides or this test fails loudly.

## Followups

- [ ] Rebase #4346 onto this branch; the explorer's \`GetRecentStreamsAsync\` / \`GetStreamMetadataAsync\` shrink to one selector call each.
- [ ] [JasperFx/polecat#57](https://github.com/JasperFx/polecat/issues/57) tracks the parallel consolidation in Polecat. (Polecat doesn't use Roslyn codegen — the equivalent there is just hand-written, but the same selector + SELECT-fragment abstraction applies.)

### DIFF
--- a/src/EventSourcingTests/event_storage_stream_state_selector.cs
+++ b/src/EventSourcingTests/event_storage_stream_state_selector.cs
@@ -1,0 +1,93 @@
+using System;
+using System.Threading.Tasks;
+using Marten.Events;
+using Marten.Internal.Sessions;
+using Marten.Linq.Selectors;
+using Marten.Testing.Harness;
+using Npgsql;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests;
+
+/// <summary>
+/// Coverage for the shared <see cref="ISelector{StreamState}"/> implementation on
+/// <see cref="IEventStorage"/>. Both Marten's own <c>FetchStreamStateAsync</c> codegen
+/// and external consumers (e.g. an event store explorer surface) read
+/// <c>mt_streams</c> through this selector + the matching <c>StreamStateSelectSql</c>
+/// — this test fixture exercises the path with a hand-built command to prove the
+/// abstraction is reachable and the column ordering is in sync with the SELECT
+/// fragment.
+/// </summary>
+public class event_storage_stream_state_selector: IntegrationContext
+{
+    public event_storage_stream_state_selector(DefaultStoreFixture fixture) : base(fixture)
+    {
+    }
+
+    [Fact]
+    public async Task selector_round_trips_a_real_stream_row()
+    {
+        var streamId = Guid.NewGuid();
+
+        await using (var session = theStore.LightweightSession())
+        {
+            session.Events.StartStream<Quest>(streamId,
+                new QuestStarted { Name = "selector smoke test" });
+            await session.SaveChangesAsync();
+        }
+
+        // The reference point: what FetchStreamStateAsync — which goes through the
+        // codegen'd QueryForStream handler — returns for the same stream.
+        StreamState expected;
+        await using (var session = theStore.LightweightSession())
+        {
+            expected = await session.Events.FetchStreamStateAsync(streamId);
+        }
+
+        expected.ShouldNotBeNull();
+
+        // Now drive the SAME row read manually through ISelector<StreamState>.
+        await using var query = (DocumentSessionBase)theStore.LightweightSession();
+        var storage = query.EventStorage();
+        var selector = (ISelector<StreamState>)storage;
+
+        var cmd = new NpgsqlCommand($"{storage.StreamStateSelectSql} where id = @id");
+        cmd.Parameters.AddWithValue("id", streamId);
+
+        await using var reader = await query.ExecuteReaderAsync(cmd);
+        (await reader.ReadAsync()).ShouldBeTrue();
+
+        var actual = await selector.ResolveAsync(reader, default);
+
+        actual.ShouldNotBeNull();
+        actual.Id.ShouldBe(expected.Id);
+        actual.Version.ShouldBe(expected.Version);
+        actual.Created.ShouldBe(expected.Created);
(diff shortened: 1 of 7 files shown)


jeremydmiller:
**Triage of the CI failures on this PR — both are pre-existing on master, not caused by these changes.**

Two distinct failures showed up in the CI run:

### 1. \`TestCodeGen\` — \`ArgumentNullException: Value cannot be null. (Parameter 'type')\`

Stack trace points at \`DocumentProviderBuilder.BuildProvider<T>()\` line 83 (\`Activator.CreateInstance(_providerType!, _mapping)\`), reached via \`CompiledQueryCollection.BuildFiles\` → \`QueryCompiler.BuildQueryPlan\` → \`StorageFor<T>\`. \`_providerType\` is \`null\`.

**Verified pre-existing on master:**

- Checked out plain \`origin/master\` (bed5d0648) in a temp worktree and ran \`dotnet run --project src/CommandLineRunner -- codegen write\`. **Same exact \`ArgumentNullException\`.**
- The master \`Build & Test - NET10 - PGLatest - System.Text.Json\` workflow has been failing this target on every run since at least 2026-05-10 23:13Z. Last green run was 2026-05-09 21:51Z (\`25612695142\`).

This is a master-level regression that this PR doesn't touch and shouldn't try to fix in scope.

### 2. \`TestEventSourcing\` — \`assert_on_max_event_id_on_event_stream_append\` with \`PostgresException 23505\`

\`Npgsql.PostgresException : 23505: duplicate key value violates unique constraint \"pg_namespace_nspname_index\"\` — that's two concurrent test classes racing on \`CREATE SCHEMA\`. The \`pg_namespace_nspname_index\` is Postgres's unique index on schema names; both transactions tried to claim the same name before either committed.

**Verified pre-existing flake:**

- The master \`Build & Test - NET10 - PGLatest - System.Text.Json - Event Sourcing\` workflow was green on the most recent master run (\`25672247615\`).
- None of this PR's changes touch schema-creation paths (\`Database.EnsureStorageExistsAsync\`, \`StoreOptions.AutoCreate\`, etc.) — the diff is scoped to \`IEventStorage\` + \`EventDocumentStorage\` + the codegen that generates the per-permutation \`StreamStateQueryHandler\` subclass.

This is a known-shape transient CI race. Suggest re-running the job.

### Action

No fix to push on this PR — fixing either of the above is out of scope for the consolidation change. I'd recommend re-running just the EventSourcing job to clear the schema race; the \`TestCodeGen\` failure is the same problem master is currently seeing and needs a separate triage.

PR #4447, opened 2026-05-17
[#4444] Port UseVersionFromMatchingStream to closed-shape SQL

### LINKED ISSUE #4444: Closed-shape: port UseVersionFromMatchingStream to inline UPSERT SQL
Surfaced while landing the W3 closed-shape `DocumentStorage<TDoc, TId>` work (#4404, PR #4438).

The codegen `UpsertFunction` had a multi-statement SQL function body that, when `mapping.UseVersionFromMatchingStream == true`, read `current_version` from `mt_streams.version` (instead of the doc table's own `mt_version`) and set `revision = current_version` before the INSERT — so the document's stored revision tracks the stream's actual version after a quick-append.

The closed-shape `BuildUpsertSql` emits a single-expression `CASE WHEN ? = 0 THEN <table>.mt_version + 1 ELSE ? END`. That handles the explicit-revision case fine but doesn't replicate the stream-version-source branch, so inline snapshots + `EventAppendMode.Quick` end up with `doc.Version` auto-incrementing by 1 per save (1, 2, 3, …) instead of matching the stream's final event version (5, 6, …).

### Affected tests (14, all green under the old codegen path)

- `EventSourcingTests.Aggregation.setting_version_number_on_aggregate.set_version_on_aggregate_with_explicit_Version_attribute`
- `EventSourcingTests.Aggregation.using_custom_aggregate_with_soft_deletes_and_update_only_events.return_correct_data_after_restarts`
- `EventSourcingTests.Bugs.Bug_3310_inline_projections_with_quick_append.start_and_append_events_to_same_stream`
- `EventSourcingTests.Bugs.Bug_3310_inline_projections_with_quick_append.update_1_stream_with_many_events` (both `appendWithExpectedVersion: True/False`)
- `EventSourcingTests.Bugs.Bug_3946_tenancy_with_for_tenant_and_projection_issues.when_projection_is_multi_stream_then_tenant_is_passed_to_projection_document` (both `useRebuild: True/False`)
- `EventSourcingTests.Dcb.hstore_auto_discover_tag_types_from_projections.auto_discovered_tag_type_works_for_fetch_for_writing_in_hstore_mode`
- `EventSourcingTests.Dcb.hstore_auto_discover_tag_types_from_projections.auto_discovered_tag_type_works_for_querying_in_hstore_mode`
- `EventSourcingTests.FetchForWriting.fetching_inline_aggregates_for_writing.fetch_existing_stream_for_writing_Guid_identifier_with_expected_version_using_identity_map`
- `EventSourcingTests.FetchForWriting.fetching_inline_aggregates_for_writing.revision_is_updated_after_quick_appending_with_IRevisioned`
- `EventSourcingTests.FetchForWriting.fetching_inline_aggregates_for_writing.revision_is_updated_after_quick_appending_with_IRevisioned_in_batch`
- `EventSourcingTests.FetchForWriting.fetching_inline_aggregates_for_writing.revision_is_updated_after_quick_appending_with_custom_mapped_version`
- `EventSourcingTests.FetchForWriting.fetching_inline_aggregates_for_writing.revision_is_updated_after_quick_appending_with_custom_mapped_version_in_batch`

### What needs to happen

Reference: the codegen branch is in [`src/Marten/Storage/UpsertFunction.cs`](https://github.com/JasperFx/marten/blob/master/src/Marten/Storage/UpsertFunction.cs#L83-L103) (the `_versionSourceTable = mt_streams` / `_versionColumnName = version` branch).

Sketch for the closed-shape `BuildUpsertOrOverwriteSql` (under `Marten/Internal/ClosedShape/DocumentStorageDescriptorBuilder.cs`):

- When `mapping.UseVersionFromMatchingStream`, the revision binder's value-expression in the INSERT VALUES list switches to a correlated subquery:
  `(CASE WHEN ? = 0 THEN COALESCE((SELECT version FROM {eventsSchema}.mt_streams WHERE id = excluded.id [AND tenant_id = excluded.tenant_id]), 1) ELSE ? END)`
- The `ON CONFLICT DO UPDATE SET mt_version = …` branch needs the same subquery so updates also track the stream.
- The matching `_currentVersionGuardCondition` for explicit-revision concurrency is `current_version > revision` (looser than the non-stream case of `>=`), so the WHERE-clause concurrency guard also needs to switch when `UseVersionFromMatchingStream` is set.
- `mapping.StoreOptions.Events.DatabaseSchemaName` provides the events schema name; the streams table is always `mt_streams` in that schema.

### Notes

- The `StoreProjection` callsite (`DocumentSessionBase.ProjectionStorage.cs:154`) already passes `r.Revision = lastEvent.Version` for `SingleStream` and `lastEvent.Sequence` for `MultiStream` projections, but under `EventAppendMode.Quick` the inline projection runs before the events are flushed so `lastEvent.Version` arrives as 0 — that's the case the stream-version-source branch was designed to handle.
- All 14 tests share this root cause; the remaining DocumentDbTests / LinqTests are green on the branch.

jeremydmiller:
## Summary

Closes #4444.

Inline single-stream projections under `EventAppendMode.Quick` arrive at `StoreProjection` with `lastEvent.Version = 0`, so the closed-shape upsert receives `Revision = 0` and the existing `CASE WHEN ? = 0 THEN 1 ELSE ? END` expression in `BuildCoreColumns` made the stored revision auto-increment 1, 2, 3, … instead of tracking the stream's actual version (5, 6, …). The codegen `UpsertFunction` handled this with a plpgsql branch keyed off `mapping.UseVersionFromMatchingStream` that read `current_version` from `{events_schema}.mt_streams.version`; this PR ports that branch into the single-statement closed-shape SQL.

## Approach

- **INSERT VALUES** — auto branch becomes
  ```sql
  CASE WHEN ? = 0
       THEN COALESCE((select version from <events>.mt_streams where id = ? [and tenant_id = ?]), 1)
       ELSE ?
  END
  ```
  Subquery id (+ tenant_id under conjoined tenancy) consumes extra `?` slots because one VALUES expression can't reference another at the same statement level.
- **ON CONFLICT DO UPDATE SET mt_version** — same subquery, but id/tenant_id come from the existing row (`{table}.id` / `{table}.tenant_id`) — no extra slots on the SET side.
- **ON CONFLICT WHERE** — `? = 0 OR COALESCE(<stream-version>, 0) <= ?`, mirroring the codegen's looser `>` guard for the stream-source case (reject only when the stream is strictly ahead of the caller's revision).
- **`BuildUpdateSql`** — same SET / WHERE rewrites for the standalone UPDATE path. WHERE / SET reference the existing row, so no extra slots on UPDATE.

Plumbed via a new `DocumentStorageDescriptor.UseVersionFromMatchingStream` flag (only set when `concurrencyMode == Numeric`, since the codegen `UpsertFunction` only honored the flag under `RevisionArgument`). `ClosedShapeInsertOperation`, `ClosedShapeUpsertOperation`, and `ClosedShapeOverwriteOperation` all consume the new slot count off that flag.

## Verification

`MARTEN_USE_CLOSED_SHAPE_STORAGE=true dotnet test src/EventSourcingTests/EventSourcingTests.csproj -f net9.0 --filter "FullyQualifiedName~Bug_3310_inline_projections_with_quick_append|FullyQualifiedName~fetching_inline_aggregates_for_writing|FullyQualifiedName~setting_version_number_on_aggregate|FullyQualifiedName~Bug_3946_tenancy_with_for_tenant|FullyQualifiedName~hstore_auto_discover_tag_types_from_projections|FullyQualifiedName~using_custom_aggregate_with_soft_deletes_and_update_only_events|FullyQualifiedName~explicit_code_for_aggregation_logic"`

→ 50 passed, 0 failed (4 skipped — see below). The codegen lane is identical (50 passed, 0 failed).

## Re-tagged tests

The issue called out 14 tests tagged with `Skip = "Closed-shape upsert needs UseVersionFromMatchingStream port. Tracked at #4444."` 9 of them genuinely fail on the UseVersionFromMatchingStream gap and now pass under both lanes after this PR.

The remaining 5 fail on **both** master's codegen and closed-shape lanes today — confirmed by running each against master with the closed-shape flag unset — so they aren't gated on this port:

- `using_custom_aggregate_with_soft_deletes_and_update_only_events.return_correct_data_after_restarts`
- `Bug_3946_tenancy_with_for_tenant_and_projection_issues.when_projection_is_multi_stream_then_tenant_is_passed_to_projection_document(useRebuild: True | False)`
- `hstore_auto_discover_tag_types_from_projections.auto_discovered_tag_type_works_for_querying_in_hstore_mode`
- `hstore_auto_discover_tag_types_from_projections.auto_discovered_tag_type_works_for_fetch_for_writing_in_hstore_mode`

Re-tagged each with a more accurate `[Skip = "Pre-existing master failure (codegen + closed-shape) — …"]` note so they don't carry the misleading #4444 reference into the next sweep. Worth filing separate triage tickets for these (`uuid = bigint` select-side bug for Bug_3946; soft-delete restart + hstore tag-based concurrency for the others).

## Test plan
- [x] Closed-shape lane green for the un-skipped #4444 tests
- [x] Codegen lane unchanged
- [ ] CI matrix runs both lanes

### DIFF
--- a/src/EventSourcingTests/Aggregation/explicit_code_for_aggregation_logic.cs
+++ b/src/EventSourcingTests/Aggregation/explicit_code_for_aggregation_logic.cs
@@ -467,7 +467,7 @@ public class using_custom_aggregate_with_soft_deletes_and_update_only_events: On
         aggregate.Deleted.ShouldBeTrue();
     }
 
-    [Fact(Skip = "Closed-shape upsert needs UseVersionFromMatchingStream port. Tracked at https://github.com/JasperFx/marten/issues/4444.")]
+    [Fact(Skip = "Pre-existing master failure (codegen + closed-shape) — soft-delete + restart flow throws ConcurrencyException. Not caused by #4444's UseVersionFromMatchingStream gap. Needs separate triage.")]
     public async Task return_correct_data_after_restarts()
     {
         var streamId = Guid.NewGuid();
--- a/src/EventSourcingTests/Aggregation/setting_version_number_on_aggregate.cs
+++ b/src/EventSourcingTests/Aggregation/setting_version_number_on_aggregate.cs
@@ -117,7 +117,7 @@ public class setting_version_number_on_aggregate : OneOffConfigurationsContext
         }
     }
 
-    [Fact(Skip = "Closed-shape upsert needs UseVersionFromMatchingStream port. Tracked at https://github.com/JasperFx/marten/issues/4444.")]
+    [Fact]
     public async Task set_version_on_aggregate_with_explicit_Version_attribute()
     {
         StoreOptions(opts => opts.Projections.Snapshot<MyAggregateWithDifferentVersionProperty>(SnapshotLifecycle.Inline));
--- a/src/EventSourcingTests/Bugs/Bug_3310_inline_projections_with_quick_append.cs
+++ b/src/EventSourcingTests/Bugs/Bug_3310_inline_projections_with_quick_append.cs
@@ -40,7 +40,7 @@ public class Bug_3310_inline_projections_with_quick_append : BugIntegrationConte
         });
     }
 
-    [Fact(Skip = "Closed-shape upsert needs UseVersionFromMatchingStream port. Tracked at https://github.com/JasperFx/marten/issues/4444.")]
+    [Fact]
     public async Task start_and_append_events_to_same_stream()
     {
         await using var session = theStore.LightweightSession(tenant);
@@ -80,7 +80,7 @@ public class Bug_3310_inline_projections_with_quick_append : BugIntegrationConte
         }
     }
 
-    [Theory(Skip = "Closed-shape upsert needs UseVersionFromMatchingStream port. Tracked at https://github.com/JasperFx/marten/issues/4444.")]
+    [Theory]
     [InlineData(true)]
     [InlineData(false)]
     public async Task update_1_stream_with_many_events(bool appendWithExpectedVersion)
--- a/src/EventSourcingTests/Bugs/Bug_3946_tenancy_with_for_tenant_and_projection_issues.cs
+++ b/src/EventSourcingTests/Bugs/Bug_3946_tenancy_with_for_tenant_and_projection_issues.cs
@@ -18,7 +18,7 @@ namespace EventSourcingTests.Bugs;
 
 public class Bug_3946_tenancy_with_for_tenant_and_projection_issues : BugIntegrationContext
 {
-    [Theory(Skip = "Closed-shape upsert needs UseVersionFromMatchingStream port. Tracked at https://github.com/JasperFx/marten/issues/4444.")]
+    [Theory(Skip = "Pre-existing master failure (codegen + closed-shape) — selecting Entity throws \"Reading as System.String is not supported for fields having DataTypeName 'bigint'\". MultiStreamProjection + conjoined tenancy select-path bug, unrelated to #4444's UseVersionFromMatchingStream gap. Needs separate triage.")]
     [InlineData(true)]
     [InlineData(false)]
     public async Task when_projection_is_multi_stream_then_tenant_is_passed_to_projection_document(bool useRebuild)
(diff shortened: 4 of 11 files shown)


jeremydmiller:
The `Daemon` workflow failure on `DaemonTests.Aggregations.build_aggregate_projection.simple_scenario` here is a **deterministic master bug** (same test fails on master CI today, unrelated to the changes in this PR). Root-cause + fix at #4453: `DocumentTable.SelectColumns` extracts `VersionColumn` but not `RevisionColumn`, so Numeric-mode + conjoined-tenancy aggregates emit a scrambled SELECT projection and the read binders pair with the wrong columns. Once #4453 lands, merge master in here and the Daemon job will go green.

PR #4450, opened 2026-05-17
[#4441] Add Marten regression test for daemon catch-up + IMessageBatch contract

### LINKED ISSUE #4441: ForceAllMartenDaemonActivityToCatchUpAsync hangs for AsyncDaemon projections without RaiseSideEffects (regression between 9.0.0-alpha.1 and 9.0.0-alpha.2)
## Summary

`Marten.Events.TestingExtensions.ForceAllMartenDaemonActivityToCatchUpAsync(...)` (the static helper that test fixtures use to drive an async daemon to "catch up immediately to the latest changes") **hangs indefinitely** when applied against an `AsyncDaemon`-registered daemon running a projection that does NOT override `RaiseSideEffects`. The call starts but never returns within the test's 5-second budget.

The same helper applied against a projection that DOES override `RaiseSideEffects` returns promptly. The same helper applied to a Wolverine-managed event subscription distribution (no `AddAsyncDaemon(Solo)` involved) also returns promptly. The failure is specific to the combination: **`AddAsyncDaemon(Solo)` daemon + projection without side effects**.

This is the third Marten regression I've found in the alpha.1 → alpha.2 window today, all from the same Wolverine test-suite sweep:

- [JasperFx/marten#4439](https://github.com/JasperFx/marten/issues/4439) — `FetchForWriting<T>` + `SnapshotLifecycle.Inline`: in-memory aggregate mutations now persisted alongside appended events.
- [JasperFx/marten#4440](https://github.com/JasperFx/marten/issues/4440) — Composite projection `ShardNames` N → 1 collapse.
- This issue.

All three sit in the **Marten 9.0.0-alpha.1 → 9.0.0-alpha.2** lift window (with JasperFx.Events alpha.7 → alpha.8 landing in the same Wolverine bump commit).

## Impact

Any test fixture (Marten's own integration tests, Wolverine's, downstream consumers') that combines an `AddAsyncDaemon` registration with a projection that doesn't override `RaiseSideEffects` and uses `ForceAllMartenDaemonActivityToCatchUpAsync` to drive deterministic catch-up will hang and time out. The hang is silent — no exception, just a `TaskCanceledException` from the test's outer cancellation token after the timeout fires.

The most likely user-visible failure mode in production: a deployed app that calls `ForceAllMartenDaemonActivityToCatchUpAsync` from a maintenance / replay path will hang the request rather than completing the catch-up.

## Where I see it

Wolverine 6.0 `MartenTests.TestHelpers` namespace, 4 deterministic failures (verified 3/3 consistent):

| Test | Setup | Projection | Result |
|---|---|---|---|
| [`catch_up_and_then_do_nothing.with_main_store`](https://github.com/JasperFx/wolverine/blob/main/src/Persistence/MartenTests/TestHelpers/catch_up_and_then_do_nothing.cs#L62) | `AddAsyncDaemon(Solo)` + `LetterCountsProjection` (no `RaiseSideEffects`) | Without side-effects | **Hangs, 5s timeout** |
| `catch_up_and_then_do_nothing.with_ancillary_store` | same | same | **Hangs, 5s timeout** |
| `catch_up_then_restart.with_main_store` | same shape, `CatchUpMode.AndResumeNormally` | same | **Hangs, 5s timeout** |
| `catch_up_then_restart.with_ancillary_store` | same | same | **Hangs, 5s timeout** |

Adjacent passing tests in the same fixture suite isolate the variable:

| Test | Setup | Projection | Result |
|---|---|---|---|
| [`second_stage_waiting.with_main_store`](https://github.com/JasperFx/wolverine/blob/main/src/Persistence/MartenTests/TestHelpers/second_stage_waiting.cs#L65) | **Same** `AddAsyncDaemon(Solo)` setup | `LetterCountsProjectionWithSideEffects` (overrides `RaiseSideEffects`) | ✅ Passes |
| [`catch_up_when_using_wolverine_distribution.with_main_store`](https://github.com/JasperFx/wolverine/blob/main/src/Persistence/MartenTests/TestHelpers/catch_up_when_using_wolverine_distribution.cs#L54) | **No** `AddAsyncDaemon` (Wolverine-managed event subscription distribution instead) | `LetterCountsProjection` (no side-effects) | ✅ Passes |

The two passing variants isolate the failure precisely: it's the **combination** of `AddAsyncDaemon` + projection-without-`RaiseSideEffects`.

## Stack trace at the timeout

```
System.TimeoutException : This TrackedSession timed out before all activity completed.
Activity detected:

| Message Id                             | Message Type                                                         | Time (ms) | Event              |
| 08deb413-993b-8c22-baa4-8b734d750000   | CatchUp:Marten:DaemonActivity:MartenTests.TestHelpers.ILetterStore   |        186| ExecutionStarted   |

  at Wolverine.Tracking.TrackedSession.ExecuteAndTrackAsync()
```

`ExecutionStarted` fires for the catch-up wrapper envelope, but the corresponding `ExecutionFinished` (which would run **after** `await runtime.Services.ForceAllMartenDaemonActivityToCatchUpAsync(cancellation, mode)` returns) never fires. So the `ForceAllMartenDaemonActivityToCatchUpAsync` call is the one hanging.

## Minimal Marten-only repro

```csharp
public class LetterCounts
{
    public Guid Id { get; set; }
    public int ACount { get; set; }
}

public record AEvent;

// Side-effect-free projection — overrides nothing else
public class LetterCountsProjection : SingleStreamProjection<LetterCounts, Guid>
{
    public override LetterCounts Evolve(LetterCounts? snapshot, Guid id, IEvent e)
    {
        snapshot ??= new LetterCounts { Id = id };
        if (e.Data is AEvent) snapshot.ACount++;
        return snapshot;
    }
}

[Fact]
public async Task force_catch_up_hangs_for_async_daemon_without_side_effects()
{
    using var host = await Host.CreateDefaultBuilder()
        .ConfigureServices(s =>
        {
            s.AddMarten(m =>
            {
                m.Connection(connStr);
                m.Projections.Add<LetterCountsProjection>(ProjectionLifecycle.Async);
            }).AddAsyncDaemon(DaemonMode.Solo);
        })
        .StartAsync();

    var store = host.Services.GetRequiredService<IDocumentStore>();
    using (var session = store.LightweightSession())
    {
        session.Events.StartStream<LetterCounts>(new AEvent(), new AEvent(), new AEvent());
    }

    using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(10));

    // Expected: returns within seconds, returns an empty IList<Exception>
    // Actual:   never returns within 10s; TaskCanceledException
    var exceptions = await host.ForceAllMartenDaemonActivityToCatchUpAsync(cts.Token);
    exceptions.ShouldBeEmpty();
}
```

If the projection above is swapped for one that overrides `RaiseSideEffects` (regardless of what that override DOES — even an empty implementation that just returns `default`), the call completes promptly. So the bug is gated on `RaiseSideEffects` being absent on the projection.

## Version delta

Same window as JasperFx/marten#4439 and #4440:

| Wolverine commit | Marten | JasperFx.Events | Tests |
|---|---|---|---|
| [`754b4076a`](https://github.com/JasperFx/wolverine/commit/754b4076a) | 8.35.x → 9.0.0-alpha.1 | (unchanged) | passing |
| [`7ecfaaf39`](https://github.com/JasperFx/wolverine/commit/7ecfaaf39) | 9.0.0-alpha.1 → 9.0.0-alpha.2 | alpha.7 → alpha.8 | **4 failures hang** |

The `RaiseSideEffects` coupling hypothesis: the dedupe-pillar `IMessageOutbox` / `IMessageBatch` plumbing landed in JasperFx.Events 2.0-alpha.8 (the same lift that introduced the [`IProjectionCoordinator<T>` ambiguity](https://github.com/JasperFx/wolverine/issues/2806) and the [composite-projection shard collapse](https://github.com/JasperFx/marten/issues/4440)). If `ForceAllMartenDaemonActivityToCatchUpAsync`'s "have I caught up?" condition is now coupled to `IMessageBatch.AfterCommitAsync` firing, and projections that don't override `RaiseSideEffects` never create a batch (so `AfterCommitAsync` never fires), the wait condition is unsatisfiable.

## Expected behavior

`ForceAllMartenDaemonActivityToCatchUpAsync` returns within bounded time regardless of whether the daemon's projections override `RaiseSideEffects`. The "are we caught up?" condition tracks high-water-mark vs daemon position, not message-outbox batch completion. Projections without side-effects should be treated as "no batch to wait for" rather than "wait forever for the batch that never comes."

## Repository for reference

JasperFx/wolverine `main` at commit `543e328ec` — `dotnet test src/Persistence/MartenTests/MartenTests.csproj -f net9.0 --filter "FullyQualifiedName~MartenTests.TestHelpers"` reproduces the 4 failures deterministically (12 tests, 4 fail, 8 pass).

Cross-references:
- JasperFx/marten#4439 — `FetchForWriting<T>` regression in the same alpha.2 lift.
- JasperFx/marten#4440 — composite projection `ShardNames` collapse in the same alpha.2 lift.

jeremydmiller:
## Summary

#4441 reproduces against Wolverine 6.0's `MartenToWolverineMessageBatch` + `MessageContext.FlushOutgoingMessagesAsync` plumbing — **not** against Marten itself.

I narrowed the Marten side first with a Marten-only repro that exactly mirrors the issue's scenario: `AddAsyncDaemon(DaemonMode.Solo)` + `SingleStreamProjection<LetterCounts, Guid>` (no `RaiseSideEffects` override) + `host.ForceAllMartenDaemonActivityToCatchUpAsync(...)`. Against the default `NulloMessageOutbox`, the call returns within seconds 3/3.

I then dropped in a tracing `IMessageOutbox` and confirmed Marten's catch-up path correctly:

1. `DocumentStore.StartProjectionBatchAsync` runs `ProjectionUpdateBatch.SpinUpMessageBatchAsync` in `Continuous` mode → calls `IMessageOutbox.CreateBatch`.
2. The resulting `IMessageBatch` is added as a listener.
3. After the batch executes, both `IMessageBatch.BeforeCommitAsync` and `IMessageBatch.AfterCommitAsync` fire.
4. `ForceAllMartenDaemonActivityToCatchUpAsync` returns and `host.Services` settles cleanly.

So the issue's hypothesis "the wait condition is coupled to `IMessageBatch.AfterCommitAsync` firing, and projections that don't override `RaiseSideEffects` never create a batch" is **wrong**: `SpinUpMessageBatchAsync` always creates a batch in Continuous mode, regardless of `RaiseSideEffects`. The actual hang is downstream of `IMessageBatch.AfterCommitAsync` — inside Wolverine's `Context.FlushOutgoingMessagesAsync()` — and lives in Wolverine, not Marten.

This PR adds two regression tests so a future Marten-side change can't silently break the contract that downstream outbox implementations depend on:

- `force_catch_up_returns_for_async_daemon_without_side_effects` — end-to-end against the default `NulloMessageOutbox`, asserts catch-up returns no exceptions within a 15s budget.
- `force_catch_up_invokes_message_batch_lifecycle_with_custom_outbox` — installs a recording `IMessageOutbox` and asserts both `BeforeCommit` and `AfterCommit` are invoked during catch-up. Pins the listener-hook contract so a regression surfaces here rather than via a downstream Wolverine hang weeks later.

## Verification

```bash
dotnet test src/EventSourcingTests/EventSourcingTests.csproj \
  --filter "FullyQualifiedName~Bug_4441_force_catch_up_with_outbox" -f net9.0
```

→ 2/2 passed, 1s.

## Wolverine follow-up

The actual hang fix lives in Wolverine. Will file a separate Wolverine issue with the narrowed root cause pointing at `MartenToWolverineMessageBatch.AfterCommitAsync` → `MessageContext.FlushOutgoingMessagesAsync()` under `AddAsyncDaemon(Solo)` + `Durability.Solo`. The Marten team can close #4441 with this PR (or, if you prefer it stays open as a tracking issue, leave it linked).

### DIFF
--- a/src/EventSourcingTests/Bugs/Bug_4441_force_catch_up_with_outbox.cs
+++ b/src/EventSourcingTests/Bugs/Bug_4441_force_catch_up_with_outbox.cs
@@ -1,0 +1,163 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using JasperFx.Events.Daemon;
+using JasperFx.Events.Projections;
+using Marten;
+using Marten.Events;
+using Marten.Events.Aggregation;
+using Marten.Events.Projections;
+using Marten.Internal.Sessions;
+using Marten.Services;
+using Marten.Testing.Harness;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests.Bugs;
+
+// Regression guard for https://github.com/JasperFx/marten/issues/4441.
+//
+// The original bug fired against Wolverine's MartenToWolverineMessageBatch +
+// MessageContext.FlushOutgoingMessagesAsync (when paired with
+// AddAsyncDaemon(Solo) + a SingleStreamProjection without a RaiseSideEffects
+// override). The Marten-side flow — daemon CatchUpAsync → buildBatchAsync →
+// ProjectionUpdateBatch.SpinUpMessageBatchAsync → IMessageOutbox.CreateBatch →
+// listener Before/AfterCommit — is shared by every IMessageOutbox
+// implementation, so the tests below pin the Marten contract: a non-blocking
+// IMessageBatch must catch up cleanly and the listener hooks must fire even
+// when the projection raises no side effects.
+
+public class Bug_4441_force_catch_up_with_outbox
+{
+    public class LetterCounts
+    {
+        public Guid Id { get; set; }
+        public int ACount { get; set; }
+    }
+
+    public record AEvent;
+
+    // Plain SingleStreamProjection — no RaiseSideEffects override (the exact
+    // shape that triggered the original Wolverine hang).
+    public class LetterCountsProjection: SingleStreamProjection<LetterCounts, Guid>
+    {
+        public override LetterCounts Evolve(LetterCounts? snapshot, Guid id, IEvent e)
+        {
+            snapshot ??= new LetterCounts { Id = id };
+            if (e.Data is AEvent) snapshot.ACount++;
+            return snapshot;
+        }
+    }
+
+    [Fact(Timeout = 30000)]
+    public async Task force_catch_up_returns_for_async_daemon_without_side_effects()
+    {
+        using var host = await Host.CreateDefaultBuilder()
+            .ConfigureServices(s =>
+            {
+                s.AddMarten(m =>
+                {
+                    m.Connection(ConnectionSource.ConnectionString);
+                    m.DatabaseSchemaName = "bug4441_default";
+                    m.Projections.Add<LetterCountsProjection>(ProjectionLifecycle.Async);
+                }).AddAsyncDaemon(DaemonMode.Solo);
+            })
+            .StartAsync();
+
+        var store = host.Services.GetRequiredService<IDocumentStore>();
+        await store.Advanced.Clean.CompletelyRemoveAllAsync();
+
+        using (var session = store.LightweightSession())
+        {
+            session.Events.StartStream<LetterCounts>(new AEvent(), new AEvent(), new AEvent());
+            await session.SaveChangesAsync();
+        }
+
+        using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(15));
+        var exceptions = await host.ForceAllMartenDaemonActivityToCatchUpAsync(cts.Token);
+        exceptions.ShouldBeEmpty();
+    }
+
+    [Fact(Timeout = 30000)]
+    public async Task force_catch_up_invokes_message_batch_lifecycle_with_custom_outbox()
+    {
+        var outbox = new RecordingOutbox();
+
+        using var host = await Host.CreateDefaultBuilder()
+            .ConfigureServices(s =>
+            {
+                s.AddMarten(m =>
+                {
+                    m.Connection(ConnectionSource.ConnectionString);
+                    m.DatabaseSchemaName = "bug4441_outbox";
+                    m.Events.MessageOutbox = outbox;
+                    m.Projections.Add<LetterCountsProjection>(ProjectionLifecycle.Async);
+                }).AddAsyncDaemon(DaemonMode.Solo);
+            })
+            .StartAsync();
+
+        var store = host.Services.GetRequiredService<IDocumentStore>();
+        await store.Advanced.Clean.CompletelyRemoveAllAsync();
+
+        using (var session = store.LightweightSession())
+        {
+            session.Events.StartStream<LetterCounts>(new AEvent(), new AEvent());
+            await session.SaveChangesAsync();
+        }
+
+        using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(15));
+        var exceptions = await host.ForceAllMartenDaemonActivityToCatchUpAsync(cts.Token);
+        exceptions.ShouldBeEmpty();
+
+        // Pins the Marten contract that downstream outbox implementations rely on:
+        // SpinUpMessageBatchAsync runs in Continuous mode → CreateBatch fires →
+        // BeforeCommit/AfterCommit on the resulting IMessageBatch fire during
+        // catch-up even when the projection raises no side effects. If any
+        // listener step is silently skipped, downstream outboxes like
+        // Wolverine's MartenToWolverineMessageBatch lose their flush hook.
+        outbox.CreateBatchCount.ShouldBeGreaterThan(0);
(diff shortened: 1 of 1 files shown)


jeremydmiller:
The `Daemon` workflow failure on `DaemonTests.Aggregations.build_aggregate_projection.simple_scenario` here is a **deterministic master bug** (same test fails on master CI today, unrelated to the changes in this PR). Root-cause + fix at #4453: `DocumentTable.SelectColumns` extracts `VersionColumn` but not `RevisionColumn`, so Numeric-mode + conjoined-tenancy aggregates emit a scrambled SELECT projection and the read binders pair with the wrong columns. Once #4453 lands, merge master in here and the Daemon job will go green.

PR #4465, opened 2026-05-18
[#4183] PostgreSQL LISTEN/NOTIFY support for async daemon wakeup (supersedes #4181)

### LINKED ISSUE #4183: feat: PostgreSQL LISTEN/NOTIFY for near-instant async projection updates
## Problem

The async daemon polls PostgreSQL at configurable intervals (250ms fast, 1s slow) to detect new events via the high water mark. This creates a latency floor between when events are appended and when projections process them. For use cases where sub-second projection updates matter, this polling delay is unnecessary overhead.

## Proposed solution

Use PostgreSQL's built-in LISTEN/NOTIFY mechanism to wake the async daemon immediately when new events are appended, instead of waiting for the next poll cycle.

The implementation spans two repositories:

### 1. JasperFx.Events — `IDaemonWakeup` extension point

PR: https://github.com/JasperFx/jasperfx/pull/164

Introduces an `IDaemonWakeup` interface that abstracts the delay between polling cycles in the `HighWaterAgent`. The default `TaskDelayDaemonWakeup` preserves existing behavior (`Task.Delay`). Custom implementations can use external signals to wake the agent immediately.

### 2. Marten — PostgreSQL LISTEN/NOTIFY implementation

PR: https://github.com/JasperFx/marten/pull/4172

Implements `PostgresqlListenWakeup` (an `IDaemonWakeup`) that:
- Opens a dedicated connection that LISTENs on the `mt_events_appended` channel
- Adds a `NotifyEventAppendedOperation` that calls `pg_notify('mt_events_appended', '')` in the same transaction as the event append
- Uses a semaphore to signal the high water agent instantly upon notification
- Falls back to the configured polling interval if no notification arrives (safety net)

Enabled via:
```csharp
opts.Events.UseListenNotifyForEventAppends = true;
```

## Considerations

- Requires the event writer and the async daemon to connect to the same PostgreSQL instance (LISTEN/NOTIFY does not work across read replicas)
- Requires one additional long-lived PostgreSQL connection per daemon node
- Not compatible with PgBouncer in transaction pooling mode (LISTEN requires session-level state)
- Polling remains as a fallback, so the feature is additive and backward-compatible

jeremydmiller:
## Summary

Closes #4183. **Supersedes #4181 by @thuijer** — rebased onto Marten 9.0 master with a few fixes the original CI didn't catch. All credit for the design and original implementation goes to @thuijer; the rebase keeps his commit shape and adds him as the primary `Co-Authored-By` on the single rebase commit.

The async daemon polls Postgres at the configured interval (250ms fast / 1s slow) to spot new appends via the high-water mark. For event sourcing workloads where projections need to catch up *now*, that polling floor is unnecessary latency. This PR adds an opt-in flag that turns the wait between polls into a PostgreSQL LISTEN — when a new event is appended (in the same transaction) we `pg_notify`, the daemon's wakeup semaphore is released, and the high-water agent loops immediately. Polling remains as the safety net for missed signals.

```csharp
services.AddMarten(opts =>
{
    opts.Events.UseListenNotifyForEventAppends = true;
});
```

## Wiring

- `Marten.Events.Daemon.HighWater.PostgresqlListenWakeup` (implements `JasperFx.Events.Daemon.HighWater.IDaemonWakeup`) — opens a dedicated connection, runs `LISTEN mt_events_appended`, and dispatches Npgsql notifications onto a `SemaphoreSlim` the high-water agent awaits.
- `Marten.Events.Operations.NotifyEventAppendedOperation` — queued by `EventGraph.ProcessEventsAsync` alongside the event-append operations when the flag is on; emits `select pg_notify('mt_events_appended', '')` in the same transaction so the signal fires only on a committed append.
- `MartenDatabase.StartProjectionDaemon` — when the flag is set, replaces `DaemonSettings.Wakeup` with the LISTEN-driven wakeup before constructing the projection daemon.
- New `IEventStoreOptions.UseListenNotifyForEventAppends` knob.

## Rebase notes vs #4181

- **Rebased on top of Marten 9.0**. The original PR was authored against pre-9.0; in 9.0 the upstream `IDaemonWakeup` abstraction has already shipped in `JasperFx.Events` (2.0.0-alpha.9), so the wakeup just plugs into it.
- **Hardened `PostgresqlListenWakeup.Dispose` against `NpgsqlOperationInProgressException`.** The original disposed the connection while the receive loop was still parked on `conn.WaitAsync`, which Npgsql 10 rejects with `'Waiting'` state. The rebase introduces an internal lifetime `CancellationTokenSource` that the receive loop observes (instead of capturing the first caller's CT — which would tear the listener down the moment the first WaitAsync returns), and Dispose cancels it + joins the loop before tearing the connection down.
- Cleaned up `_signal.Wait(0, token)` → `await _signal.WaitAsync(0, token)` in the drain loop for consistency.

## Test plan

Three new tests in `EventSourcingTests.Daemon.postgres_listen_notify_wakeup_tests`:

- `wait_async_returns_quickly_when_a_pg_notify_arrives` — pins the wiring: the wakeup signals well under the timeout when an external `pg_notify` fires (uses a separate Npgsql connection — indistinguishable from the production `NotifyEventAppendedOperation` flow from the wakeup's point of view).
- `wait_async_returns_when_timeout_elapses_without_a_notification` — sanity check on the safety-net fallback (no NOTIFY → timeout elapses → WaitAsync returns).
- `daemon_wakes_up_through_pg_notify_when_flag_is_on` — end-to-end: real Marten host, real `AddAsyncDaemon`, real append. `StaleSequenceThreshold` is bumped to 30s so the assertion (`projection.Total == 7`) only succeeds if LISTEN/NOTIFY actually fired — if the test relied on the polling fallback, the test timeout (15s) would fire first.

Broader runs:

- [x] Build clean across net9.0 + net10.0 (0 errors)
- [x] All three wakeup tests pass 3-for-3 back-to-back locally
- [x] `appending_events_workflow_specs` + `Bugs.Bug_44*` smoke — 69/69 pass
- [x] `EventSourcingTests.Daemon` + `Bug_4441` — 7/7 pass
- [ ] CI matrix

## Considerations (carried over from #4183)

- Requires the writer host and the daemon host to connect to the same PostgreSQL instance. LISTEN/NOTIFY doesn't cross read replicas.
- Burns one extra long-lived connection per daemon node for the LISTEN.
- Incompatible with PgBouncer's transaction-pooling mode (LISTEN needs session-level connection state).
- Off by default. Polling stays as both the off-mode behavior and the on-mode safety net.

### DIFF
--- a/src/Marten/Events/IEventStoreOptions.cs
+++ b/src/Marten/Events/IEventStoreOptions.cs
@@ -149,6 +149,14 @@ namespace Marten.Events
         public bool EnableEventSkippingInProjectionsOrSubscriptions { get; set; }
 
         /// <summary>
+        /// When enabled, uses PostgreSQL LISTEN/NOTIFY to wake the async projection daemon
+        /// immediately when new events are appended, instead of relying solely on polling.
+        /// This provides near-instant projection updates while still falling back to polling
+        /// as a safety net. Default is false.
+        /// </summary>
+        public bool UseListenNotifyForEventAppends { get; set; }
+
+        /// <summary>
         ///     Directs the schema migration functionality to ignore the presence of the named index
         ///     on the event-store tables (<c>mt_events</c>, <c>mt_streams</c>, <c>mt_event_progression</c>).
         ///     Use this when an external mechanism (e.g. a custom <c>IFeatureSchema</c>) declares an index
--- a/src/EventSourcingTests/Daemon/postgres_listen_notify_wakeup_tests.cs
+++ b/src/EventSourcingTests/Daemon/postgres_listen_notify_wakeup_tests.cs
@@ -1,0 +1,186 @@
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using JasperFx.Events.Daemon;
+using JasperFx.Events.Projections;
+using Marten;
+using Marten.Events.Aggregation;
+using Marten.Events.Daemon.HighWater;
+using Marten.Testing.Harness;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging.Abstractions;
+using Npgsql;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests.Daemon;
+
+// Covers the IDaemonWakeup contract implementation in
+// Marten.Events.Daemon.HighWater.PostgresqlListenWakeup.
+//
+// The wakeup opens a dedicated NpgsqlConnection that LISTENs on its
+// channel, plus a Task.Run loop that drives Npgsql's connection.WaitAsync
+// to dispatch incoming NOTIFYs. WaitAsync() then awaits a SemaphoreSlim
+// the dispatcher releases.
+//
+// The tests fire a `pg_notify(...)` from a SEPARATE connection (the
+// production flow runs pg_notify inside the event-append transaction;
+// from the wakeup's point of view it's indistinguishable from any other
+// LISTEN/NOTIFY signal). We then assert the wakeup unblocks well before
+// the wait timeout — i.e. the listener wired the signal up correctly.
+public class postgres_listen_notify_wakeup_tests
+{
+    [Fact]
+    public async Task wait_async_returns_quickly_when_a_pg_notify_arrives()
+    {
+        var dataSource = NpgsqlDataSource.Create(ConnectionSource.ConnectionString);
+        try
+        {
+            using var wakeup = new PostgresqlListenWakeup(
+                dataSource,
+                NullLogger.Instance,
+                channel: "marten_test_listen_notify_wakeup");
+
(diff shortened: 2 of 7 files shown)


src/Marten/Events/IEventStoreOptions.cs
```diff
@@ -148,6 +148,14 @@ public interface IEventStoreOptions
         /// </summary>
         public bool EnableEventSkippingInProjectionsOrSubscriptions { get; set; }
 
+        /// <summary>
+        /// When enabled, uses PostgreSQL LISTEN/NOTIFY to wake the async projection daemon
+        /// immediately when new events are appended, instead of relying solely on polling.
+        /// This provides near-instant projection updates while still falling back to polling
+        /// as a safety net. Default is false.
+        /// </summary>
+        public bool UseListenNotifyForEventAppends { get; set; }
```
thomas-px, line 157: I'd add the deployment constraints for this to work correctly:
* writer and daemon must hit the same Postgres instance (no read replicas)
* incompatible with PgBouncer transaction pooling (LISTEN needs session state)
* burns one extra long-lived connection per daemon node

Issue #4516, opened 2026-05-20
Reduce ProjectionCoordinator to a ProjectionCoordinatorBase subclass (consume jasperfx#326)


jeremydmiller:
## Context
JasperFx/jasperfx#326 (PR JasperFx/jasperfx#336) lifted the projection-coordinator execute loop + agent helpers + lifecycle into `JasperFx.Events.Daemon.ProjectionCoordinatorBase`. This is the Marten-side consumption.

## What to do
- Reduce `Marten.Events.Daemon.Coordination.ProjectionCoordinator` to a subclass of `JasperFx.Events.Daemon.ProjectionCoordinatorBase`.
- Supply via the base ctor: `BuildDistributor(store)`, the store's `ResiliencePipeline`, `Events.TimeProvider`, and the three settings (`LeadershipPollingTime` as a TimeSpan, `AgentPauseTime`, `HealthCheckPollingTime`).
- Implement the abstract seams:
  - `ResolveDaemon(IProjectionSet)` -> `findDaemonForDatabase((MartenDatabase)set.Database)` (keep the `ImHashMap` + double-checked-lock cache in the subclass).
  - `ResolvedDaemons()` -> snapshot of the cache values.
  - `DaemonForMainDatabase` / `DaemonForDatabase` / `AllDaemonsAsync` -> existing Marten implementations.
- Delete the lifted-into-base members: `executeAsync`, `startAgentsIfNecessaryAsync`, `stopAgentsIfNecessaryAsync`, `tryStartAgent`, the Start/Pause/Resume/Stop bodies, and the `DaemonShardName` record.
- Marten's behavior is unchanged (the base adopted Marten's resilient agent-start + lifecycle model verbatim).

## Acceptance
- [ ] `ProjectionCoordinator` reduced to subclass + distributor/daemon-resolver wiring
- [ ] Hot-cold lease-handover / fast-restart daemon tests pass on Postgres
- [ ] Pin bumped to the JasperFx.Events rc that contains #326

## Refs
- jasperfx#326 / PR jasperfx#336 — the lift
- Parent dedupe pillar: jasperfx#214

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4535, opened 2026-05-20
Cut 9.0.0-rc.2: fix ProjectionCoordinator construction when async daemon is Disabled

jeremydmiller:
## Problem

The #4516 dedupe lifted the coordinator leadership/agent loop into `JasperFx.Events.Daemon.ProjectionCoordinatorBase`, whose ctor now rejects a null distributor:

```csharp
Distributor = distributor ?? throw new ArgumentNullException(nameof(distributor));
```

But Marten's `ProjectionCoordinator.BuildDistributor` returns **`null`** in the `default:` case — i.e. for `DaemonMode.Disabled`. So constructing a `ProjectionCoordinator` in Disabled mode throws:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'distributor')
  at JasperFx.Events.Daemon.ProjectionCoordinatorBase..ctor(IProjectionDistributor distributor, …)
  at Marten.Events.Daemon.Coordination.ProjectionCoordinator..ctor(DocumentStore store, ILogger logger)
```

A `ProjectionCoordinator` can legitimately be **constructed (but never started)** while async projections are Disabled. Concretely, Wolverine's ancillary-store integration resolves `IProjectionCoordinator<T>` for stores whose subscription distribution Wolverine manages itself (`AncillaryWolverineOptionsMartenExtensions`), newing up Marten's `ProjectionCoordinator<T>` directly. Pre-dedupe this was a null assignment that constructed fine.

## Fix

`BuildDistributor` returns a no-op `NulloProjectionDistributor` for the `Disabled` case instead of `null`. This restores the pre-dedupe "constructible while disabled" behavior without weakening the lifted base ctor's non-null contract. Every member of the no-op distributor is benign (empty distribution, no locks held), so it is also safe if the coordinator is ever started in Disabled mode.

## Validation

- New fast unit test `constructing_projection_coordinator_when_async_disabled` — constructs the coordinator with the default (Disabled) async mode; previously threw, now passes.
- Reproduced and verified the fix end-to-end against the regressing Wolverine test `MartenTests.AncillaryStores.bootstrapping_ancillary_marten_stores_with_wolverine.projection_coordinator_from_marten` using a locally-packed Marten — fails on `9.0.0-rc.1`, passes with this change.

## Why it matters now

This is the only blocker for the **Wolverine 6.0.0-rc.1** release (the rest of the critter-stack rc bump is green). Needs a Marten 9.0.0-rc.2 with this fix before Wolverine can re-pin and publish.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
     <PropertyGroup>
-        <Version>9.0.0-rc.1</Version>
+        <Version>9.0.0-rc.2</Version>
         <LangVersion>13.0</LangVersion>
         <Authors>Jeremy D. Miller;Babu Annamalai;Jaedyn Tonee</Authors>
         <PackageIconUrl>https://martendb.io/logo.png</PackageIconUrl>
--- a/src/CoreTests/constructing_projection_coordinator_when_async_disabled.cs
+++ b/src/CoreTests/constructing_projection_coordinator_when_async_disabled.cs
@@ -1,0 +1,32 @@
+using Marten;
+using Marten.Events.Daemon.Coordination;
+using Marten.Testing.Harness;
+using Microsoft.Extensions.Logging.Abstractions;
+using Shouldly;
+using Xunit;
+
+namespace CoreTests;
+
+public class constructing_projection_coordinator_when_async_disabled
+{
+    // Regression: a ProjectionCoordinator can be *constructed* (but never started) while the
+    // async daemon is Disabled — e.g. Wolverine's ancillary-store integration directly news up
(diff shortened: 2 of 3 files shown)


jeremydmiller:
Added the `Cut 9.0.0-rc.2` version bump on top of the fix, so merging this PR cuts the rc.2 release that Wolverine 6.0.0-rc.1 needs. Dependency pins are unchanged from rc.1.

jeremydmiller:
Filed JasperFx/jasperfx#352 to track the upstream root cause (the lifted `ProjectionCoordinatorBase` ctor's strict non-null distributor guard). Polecat is not affected yet — it hasn't migrated to `ProjectionCoordinatorBase`. Once the upstream guard is relaxed, the `NulloProjectionDistributor` workaround in this PR can be removed.

Issue #4591, opened 2026-05-31
DCB: AssertDcbConsistency is a non-locking check — truly-concurrent appends sharing a tag both commit (no DcbConcurrencyException)


Baggs:
**Environment:** Marten 9.3.4 · `DcbStorageMode.HStore` · Conjoined tenancy · default `EventAppendMode.QuickWithServerTimestamps` · default session isolation (READ COMMITTED) · .NET 10 · PostgreSQL 18.

## Summary

Two **truly-concurrent** `FetchForWritingByTags<TBoundary>(query)` → `AppendOne` → `SaveChangesAsync()` flows whose appended events share the query's tag value can **both commit** — neither throws `DcbConcurrencyException` — so two events the DCB boundary should serialize both persist. **Sequential** conflicts (the second writer begins after the first commits) *are* caught correctly; only genuinely-overlapping writers slip through. (`TBoundary` is an identity-less `[BoundaryAggregate]` in my case, but the mechanism below is independent of that.)

## Root cause (as far as I can tell)

`AssertDcbConsistency.ConfigureCommand` (HStore branch) emits a plain `select exists(select 1 from mt_events e where e.seq_id > :lastSeen and <hstore tag predicate> [and tenant_id = …])`. `FetchForWritingByTagsHandler.HandleAsync` captures `lastSeenSequence = events.Max(e => e.Sequence)` and unconditionally `QueueOperation(assertion)`. The check is a **non-locking `SELECT`** executed as a *separate statement* from the `INSERT`s, and the `SaveChanges` transaction opens at the session's default isolation (`TransactionalConnection.BeginTransaction(_options.IsolationLevel)` → READ COMMITTED). There's no `SERIALIZABLE` and no advisory lock anywhere under `src/Marten/Events/Dcb/`.

So two concurrent transactions both run their `exists` check before either commits, both see no conflict, and both insert — a check-then-act race. (Stream-based optimistic concurrency avoids this via the `(stream_id, version)` unique constraint; the tag-based DCB path has no equivalent serializing constraint.)

## Expected vs actual

- **Expected:** exactly one of the concurrent appends commits; the other throws `DcbConcurrencyException`.
- **Actual:** both commit.

## Reproduction

[...]

PR #4594, opened 2026-06-01
Research #4591: reproduction + analysis of the DCB tag-concurrency race

jeremydmiller:
Research PR for [#4591](https://github.com/JasperFx/marten/issues/4591). **No fix here** — per direction, advisory locks and `SERIALIZABLE` are off the table, and reproduction was the first step before committing to a design.

## Reproduction — confirmed

`Bug_4591_truly_concurrent_dcb_tag_appends_both_commit.cs` reproduces the bug. The test is `[Fact(Skip = …)]` on master so CI stays green; unskip locally to observe the race. Flip to a passing `[Fact]` in the same PR that ships the fix.

**Test shape:**
- Reporter's environment exactly: `DcbStorageMode.HStore` + `TenancyStyle.Conjoined` + `LightweightSession` (default isolation: `READ COMMITTED`).
- Two tag types for the boundary aggregate. Each racer uses its **own `StudentId`** (`WithTag(studentId, courseId)` → `EventBoundary.RouteEventByTags` picks the first tag's `AggregateType` for the streamId, so distinct `StudentId` = distinct stream — **bypasses the `(stream_id, version)` unique constraint** that would otherwise catch same-stream concurrent writers).
- The shared **`CourseId`** is the actual DCB boundary tag the check races on.
- `TaskCompletionSource` barrier — every racer fetches first so they all capture the same `lastSeenSequence`, then races into `SaveChangesAsync` simultaneously.

**Observed against master (Marten 9.3.4):** 8 racers → **6 commits + 2 throws**. Contract says 1 commit + 7 throws. ✅ Bug reproduced.

## Root cause — confirmed

`AssertDcbConsistency.ConfigureCommand` (HStore branch, `src/Marten/Events/Dcb/AssertDcbConsistency.cs:33`) emits:

```sql
SELECT EXISTS (
    SELECT 1 FROM {schema}.mt_events e
    WHERE e.seq_id > :lastSeen
      AND <hstore tag predicate>
      [AND e.tenant_id = :tenant]
    LIMIT 1
)
```

Run as a **separate non-locking statement** before the INSERTs, with the transaction at `READ COMMITTED`. Two concurrent transactions both run the EXISTS check before either commits, both see no conflict, both insert. Same fundamental shape as a check-then-act race; mechanism is independent of the boundary aggregate type.

## Fix-option analysis (assuming the user's constraints: no advisory locks, no `SERIALIZABLE`)

The fundamental issue: at `READ COMMITTED`, two concurrent transactions need to be **serialized on the shared tag** without any of the standard tools that do that. Available serializing primitives are:

| Tool | Available? | Notes |
| --- | --- | --- |
| `SERIALIZABLE` isolation (SSI) | ❌ excluded | Would catch the predicate-read-vs-write conflict natively |
| `pg_advisory_xact_lock` | ❌ excluded | Cheapest, mechanically correct |
| Row-level locks on existing rows | ⚠️ partial | `SELECT … FOR UPDATE` only locks existing rows; a tag value with no prior events has nothing to lock |
| Unique-constraint races | ✅ usable | Mirrors how stream-version concurrency works — needs a side schema |

The reporter's third suggestion (`INSERT … SELECT … WHERE NOT EXISTS`) **does not actually serialize at `READ COMMITTED`** — both transactions' statements still snapshot at statement-start and both see no conflict. That option only works under `REPEATABLE READ` / `SERIALIZABLE`. Worth ruling out explicitly.

### Recommended direction: **DCB-version side table with optimistic concurrency**

Mirror how stream-version concurrency works on `mt_events.(stream_id, version)`. Add a side table — schema sketch:

```sql
create table mt_dcb_tag_version (
    tag_table    varchar      not null,   -- e.g. "course"
    tag_value    text         not null,   -- the stringified tag value (or per-type column)
    tenant_id    varchar      not null default '*DEFAULT*',
    version      bigint       not null default 0,
    primary key (tag_table, tag_value, tenant_id)
);
```

- `FetchForWritingByTags<T>(query)` SELECTs the matching `(tag_table, tag_value, tenant_id) → version` rows (insert-or-fetch to handle the "no row yet" case; the upsert itself takes a row-level lock for the duration of the fetching transaction).
- `SaveChangesAsync` emits, in the same batch as the event INSERTs, one `UPDATE mt_dcb_tag_version SET version = version + 1 WHERE (tag_table, tag_value, tenant_id, version) = (...captured...)` per matching tag.
- 0 rows affected on any of those UPDATEs → throw `DcbConcurrencyException`.
- The `UPDATE … WHERE version = $captured` is the optimistic-concurrency check; the row-level lock the UPDATE takes is what serializes truly-concurrent transactions on the same tag.

**Why this beats the alternatives:**
- No advisory lock, no `SERIALIZABLE`. ✅
- Reuses the optimistic-concurrency idiom Marten already exposes for streams — same mental model for users.
- Tag-table-level table is additive — existing rows / queries don't need to migrate.
- Cost is one `UPDATE` per distinct tag-table in the boundary per save.

**Cost / scope:**
- One new schema object (table + indexes).
- Migration story: backfill `version = 0` for every distinct existing tag-table/tag-value combination on store activation, or initialize lazily on first `FetchForWritingByTags` (whichever fits the rest of Marten's migration ergonomics).
- Plumbing inside `FetchForWritingByTagsHandler` + `AssertDcbConsistency` (or its replacement) to capture and re-check versions.

### Alternative directions considered + set aside

- **Sentinel-row insert with `ON CONFLICT DO UPDATE … RETURNING`** on a `mt_dcb_locks(tag_value)` table. Mechanically equivalent to a row-level advisory lock — same correctness, more surface area, no clear advantage over the side-table-with-version approach above.
- **`SELECT … FOR UPDATE` of matching tag rows before the EXISTS check.** Doesn't lock the gap — concurrent INSERTs into the same tag value aren't blocked.
- **Documenting `SERIALIZABLE` as the contract** (the issue's third option). Excluded.

## Recommended next step

Confirm the DCB-version-side-table direction (or propose another) before I cut a fix PR. Once the design is settled, the fix PR can:

1. Add the schema object + Weasel feature.
2. Wire `FetchForWritingByTagsHandler` to capture versions and stage the version UPDATEs.
3. Replace / supplement `AssertDcbConsistency` so the boundary check is the row-level UPDATE.
4. Un-skip the reproduction test and flip its assertion comment — it becomes a passing regression.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -13,13 +13,13 @@
     <PackageVersion Include="FluentAssertions" Version="6.12.0" />
     <PackageVersion Include="FSharp.Core" Version="9.0.100" />
     <PackageVersion Include="FSharp.SystemTextJson" Version="1.3.13" />
-    <PackageVersion Include="JasperFx" Version="2.2.1" />
-    <PackageVersion Include="JasperFx.Events" Version="2.2.1" />
-    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.2.1">
+    <PackageVersion Include="JasperFx" Version="2.4.0" />
+    <PackageVersion Include="JasperFx.Events" Version="2.4.0" />
+    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.4.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageVersion>
-    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.2.1" />
+    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.4.0" />
     <PackageVersion Include="Jil" Version="3.0.0-alpha2" />
     <PackageVersion Include="Lamar" Version="7.1.1" />
     <PackageVersion Include="Lamar.Microsoft.DependencyInjection" Version="15.0.0" />
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -141,6 +141,7 @@ const config: UserConfig<DefaultTheme.Config> = {
             { text: 'Json Serialization', link: '/configuration/json' },
             { text: 'Resiliency Policies', link: '/configuration/retries' },
             { text: 'Command Line Tooling', link: '/configuration/cli' },
+            { text: 'JasperFx Commands in the Aspire Dashboard', link: '/configuration/aspire-commands' },
             { text: 'Optimized Development Workflow', link: '/configuration/optimized_artifact_workflow' },
             { text: 'Publishing with Native AOT', link: '/configuration/aot-publishing' },
             { text: 'Multi-Tenancy with Database per Tenant', link: '/configuration/multitenancy' },
--- a/docs/configuration/aspire-commands.md
+++ b/docs/configuration/aspire-commands.md
@@ -1,0 +1,129 @@
+# JasperFx Commands in the Aspire Dashboard <Badge type="tip" text="9.4" />
+
+The optional **`JasperFx.Aspire`** package adds Marten's command-line verbs as
+clickable **custom commands** on each resource tile in the
+[.NET Aspire dashboard](https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/dashboard).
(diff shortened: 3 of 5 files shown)


jeremydmiller:
Superseded by #4591

Issue #4617, opened 2026-06-03
Expansive test battery for per-tenant event-store partitioning (UseTenantPartitionedEvents) — new test project + shared string/guid fixtures


jeremydmiller:
## Goal &amp; scope

`opts.Events.UseTenantPartitionedEvents = true` list-partitions **both** `mt_events` and `mt_streams` by `tenant_id`, creates a per-tenant event sequence `mt_events_sequence_{suffix}`, requires `TenancyStyle.Conjoined`, and is **QuickAppend-only** (`EventAppendMode.Quick` / `QuickWithServerTimestamps`) — incompatible with `Rich` append and with `UseArchivedStreamPartitioning`. It must compose with every Marten multi-tenancy database model: single-DB conjoined, `MasterTableTenancy` (db-per-tenant), `SingleServerMultiTenancy` (auto-create db-per-tenant), and sharded pooling (`MultiTenantedWithShardedDatabases` + auto-assign / `UseSmallestDatabaseAssignment`).

This issue gathers **every** test for this option — existing and net-new — into a single dedicated project built on **one shared `DocumentStore` per stream-identity flavor**, covering every event-sourcing API **including the async daemon**, plus regression tests for every known failure mode.

Grounded against Marten 9.5.0 / JasperFx.Events 2.8.0 (with #4596/#4598/#4605/#4606/#4607/#4611 merged). Config guards live in `StoreOptions.Validate()` (`src/Marten/StoreOptions.cs:783-833`); the flag is `EventGraph.UseTenantPartitionedEvents` (`EventGraph.cs:261`).

### The behavioral facts that shape every test

- **`seq_id` (`IEvent.Sequence`) is per-tenant, NOT global.** Each tenant draws from `mt_events_sequence_{suffix}` (`QuickAppendEventFunction.cs:128-148`). Two tenants both get `seq_id = 1` for their first event. Any assertion of globally-unique/globally-ordered `seq_id` is wrong. `IEvent.Version` (per-stream, from `mt_streams`) is unaffected — tests must assert the two **independently**.
[...]

PR #4641, opened 2026-06-03
#4617 deferred 8/N — pin AutoCreate matrix under partitioning

jeremydmiller:
Continues the #4617 deferred-items series — coverage PR #4631 left as follow-ups in its description. This is **deferred PR 8/N**.

## What this pins

5 tests in `src/TenantPartitionedEventsTests/Admin/autocreate_matrix_under_partitioning.cs` covering the `AutoCreate` × `UseTenantPartitionedEvents` matrix.

### `permissive_autocreate_values_let_AddMartenManagedTenantsAsync_create_partitions` (Theory: `All` / `CreateOrUpdate` / `CreateOnly`)

Standard on-boarding flow: virgin schema → `AddMartenManagedTenantsAsync` → append → read. All three permissive `AutoCreate` values let the full flow succeed end-to-end without the user having to pre-create anything.

### `AutoCreate_None_against_virgin_schema_creates_partitions_but_append_fails_42883`

The surprising-but-observed **asymmetric contract**:

- `AddMartenManagedTenantsAsync` DOES run even under `None` — the admin call is exempt from the AutoCreate gate and creates the partition parents + per-tenant partition tables.
- But the surrounding `mt_quick_append_events` plpgsql function is NOT created (that path goes through the regular feature-schema-apply which IS gated). The next append errors with `42883: function ... does not exist`.

Pin captures this partial half-create state so a future fix — either (a) extending the admin call to install the function too, or (b) tightening the admin gate so the partition tables aren't half-installed without their function — flips the assertion intentionally.

### `AutoCreate_None_works_after_a_primer_store_creates_the_schema`

Typical CLI-apply-then-runtime-None deployment: primer store with `AutoCreate.All` lays down the full schema, runtime store with `AutoCreate.None` against the same schema runs the full `AddMartenManagedTenantsAsync` + append + read flow. Confirms `None` is production-usable when the schema is provisioned out-of-band.

## Local verification

```
Passed!  - Failed:     0, Passed:     5, Skipped:     0, Total:     5
```

## Tracking

- Bug: #4617 (section 3e — admin/schema concerns under partitioning)
- Parent (closed): #4631 (initial wave of TenantPartitionedEventsTests coverage)
- Series so far: #4634, #4635, #4636, #4637, #4638, #4639, #4640 (raw IProjection), **this PR**.

### DIFF
--- a/src/TenantPartitionedEventsTests/Admin/autocreate_matrix_under_partitioning.cs
+++ b/src/TenantPartitionedEventsTests/Admin/autocreate_matrix_under_partitioning.cs
@@ -1,0 +1,185 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using JasperFx;
+using JasperFx.Events;
+using JasperFx.MultiTenancy;
+using Marten;
+using Marten.Events;
+using Marten.Storage;
+using Marten.Testing.Harness;
+using Npgsql;
+using Shouldly;
+using Weasel.Postgresql;
+using Xunit;
+
+namespace TenantPartitionedEventsTests.Admin;
+
+/// <summary>
+/// #4617 section 3e deferred — pin <see cref="AutoCreate"/> × <c>UseTenantPartitionedEvents</c>.
+///
+/// <para>
+/// AutoCreate values that allow implicit schema mutation (All, CreateOrUpdate,
+/// CreateOnly) all let <c>AddMartenManagedTenantsAsync</c> succeed: the parent
+/// mt_events / mt_streams tables are created on first storage touch, and the
+/// admin call creates the per-tenant partition tables.
+/// </para>
+///
+/// <para>
+/// AutoCreate.None: the observed contract is incomplete. The admin call
+/// <c>AddMartenManagedTenantsAsync</c> is EXEMPT from the gate for the per-
+/// tenant sequence DDL it owns, but the partition-table-migrate path
+/// (<see cref="Weasel.Postgresql.Tables.Partitioning.ManagedListPartitions"/>'s
+/// <c>MigrateAsync</c>) does NOT find the events tables to migrate when there's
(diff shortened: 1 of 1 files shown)


jeremydmiller:
**Attempted fix, reverted — fix is non-trivial, deferred.**

Tried the obvious in-band fix: have `AddMartenManagedTenantsAsync` imperatively install `mt_quick_append_events` + `mt_archive_stream` next to the existing `PerTenantEventSequences.EnsureSequencesAsync` bypass. Two problems surfaced:

1. **`mt_streams` parent is also missing under `None`**, not just the functions — the partition-migrate path in `ManagedListPartitions` doesn't pick the events tables up for a virgin schema. So the failure shape is actually `42P01: mt_streams does not exist`, hit before the function-resolution error. The test now accepts either `42P01` OR `42883` since both shapes are the same half-installed state.

2. **Imperative function install conflicts with `CreateOnly`'s diff-tightness.** Once the function is pre-installed by the admin call, the subsequent `EnsureStorageExistsAsync(typeof(IEvent))` in `SaveChangesAsync` computes a diff that flags the function as `Update` (presumably body comparison differs at the byte level from what Weasel's diff expects), and `CreateOnly` rejects all `Update` deltas via `SchemaMigration.AssertPatchingIsValid`. The whole CreateOnly theory case fails.

Two viable proper fixes that need design discussion:

- (a) Apply the events feature's full schema in the admin call via a per-feature `ApplyAllConfiguredChangesToDatabaseAsync`-style call with a scoped `CreateOrUpdate` override. No such per-feature API exists today.
- (b) Tighten the admin gate so partition tables don't get half-installed without their function — i.e., reject the admin call under `None` with a clear "pre-create your schema first" error rather than partially succeeding.

The pin in this PR is updated to capture the half-installed state honestly with a broader error-shape match so a future fix flips the assertion intentionally.

Force-pushed commit: `939d6c1cb`.

Issue #4685, opened 2026-06-08
Phase 4-bis: BulkWriter (binary COPY) for INSERT-only rebuild batches


jeremydmiller:
### Parent

Child of [CritterWatch/CritterWatch#208](https://github.com/JasperFx/CritterWatch/issues/208) — the master tracking issue for rebuild optimization work. Composes with Phase 4 (deferred flush via identity-map accumulation) and Phase E (measurement #4684).

### Background

Rebuild has a property continuous catch-up doesn't: **post-TRUNCATE, the entire rebuild write path is INSERT-only**. There's no need for UPSERT machinery, no `ON CONFLICT` path, no concurrency checking — rebuild is authoritative by definition.

Binary COPY (via Npgsql's `BeginBinaryImport`) is typically **3-10× faster than per-row INSERT** for bulk loads in PostgreSQL. On a 20M-event rebuild producing 5M aggregate rows, this could be 4-8 minutes of wall-clock saving on the write side alone.

Marten already has the COPY machinery: `IDocumentStore.BulkInsertAsync` uses `BeginBinaryImport` with the document binder pipeline. The rebuild path should reuse those binders.

### Scope

Add a BulkWriter-based flush path to `ProjectionUpdateBatch` that activates when **all operations in the batch are inserts** (which rebuild always satisfies post-TRUNCATE). Fall back to per-row UPSERT otherwise (continuous mode).

#### Concrete shape

```
Per (tenant, projection) rebuild:
  1. TRUNCATE the projection's rows for this tenant
     - Partition-aware: TRUNCATE the per-tenant partition when UseTenantPartitionedEvents is on (cheap DDL)
     - Conjoined-only: DELETE FROM mt_doc_xxx WHERE tenant_id = $1 (slower but still cheaper than per-row INSERT)
  2. Replay events through normal batch machinery (same EvolveAsync, same aggregate cache)
  3. At batch flush:
     - If batch is INSERT-only (true for rebuild post-truncate):
         open BulkWriter via BeginBinaryImport within the open transaction,
         stream all rows, end COPY
     - Otherwise: fall back to per-row UPSERT (continuous mode)
  4. Progression row UPDATE at batch end (single statement; optional COPY-into-temp optimization deferred to a separate phase if measurement shows it dominates)
```

[...]

Issue #4667, opened 2026-06-05
Async daemon: eliminate session-shared dictionary access from projection work (VersionTracker, ItemMap, ChangeTrackers)


jeremydmiller:
### Background

`DocumentSession` is **by-contract not thread-safe**. The one exception is the async projection daemon: a single `DocumentSessionBase` is shared across parallel slice handlers during projection processing.

PR #4658 closed the most visible site — `DocumentSessionBase.StoreProjection()` now calls `_storage.OverwriteProjected(doc, tenantId)`, a new `IDocumentStorage` method that constructs the operation with `null` trackers and structurally bypasses `session.Versions.VersionsFor<T>()` / `RevisionsFor<T>()`. But that fixed *one* site. A deep audit (summarized below) confirmed several remaining call sites that still touch `session.Versions`, `session.ItemMap`, and `session.ChangeTrackers` from inside the parallel block.

We're getting more reports of concurrency violations. This issue is the structural plan to make it **impossible** for projection work to touch those dictionaries — not "thread-safe via locks" but "the relevant methods don't take a session reference, so there's nothing to race on."

### Confirmed race shape

`JasperFx.Events/Daemon/AggregationRunner.cs:59` creates `new Block<EventSliceExecution>(10, ...)` — 10 parallel `Task.Run` workers per `(range, tenant)`, all closing over:

- one `batch` from `BuildBatchAsync` line 49
- one `operations` (the Marten `DocumentSessionBase`) from `batch.SessionForTenant(group.TenantId)` line 139
- one `storage` (the Marten `IProjectionStorage` wrapper) from `operations.FetchProjectionStorageAsync(...)` line 144

All 10 workers concurrently invoke `ApplyChangesAsync`, which routes into Marten's `ProjectionStorage<TDoc, TId>` and from there into `_storage.{Upsert, Insert, Update, Overwrite, LoadAsync, LoadManyAsync}(..., _session, ...)`. **Composite projections multiply this**: `ExecutionStage.cs:11-27` runs N child executions concurrently against a shared `ActiveBatch`, each of which then opens its own 10-wide `Block`. Worst case: `N × 10` concurrent ops against one batch.

#### A. Write paths from `ProjectionStorage<TDoc, TId>` still touching shared dicts

`src/Marten/Internal/Sessions/DocumentSessionBase.ProjectionStorage.cs`:

| Method | Line | Race path |
|---|---|---|
[...]

PR #4699, opened 2026-06-09
#4685 PR 2 — rebuild Insert-only routing seam + proof of the Phase-4 prerequisite

jeremydmiller:
Part of #4685 (Phase 4-bis: BulkWriter binary `COPY` for INSERT-only rebuild batches). Follows #4685 PR 1 (the `BatchFlushMode` classifier seam).

## What this does

1. **Adds the `ExecutionMode` seam.** `DocumentSessionBase.ExecutionMode` (virtual, default `Continuous`), overridden by `ProjectionDocumentSession` to return its per-shard `Mode`, and delegated by `NestedTenantSession` to its parent (cross-tenant projection writes share the parent's batch). This is the rebuild-vs-continuous signal PR 3's BulkWriter dispatch needs.

2. **Routes projection-storage writes to `InsertProjected` when rebuilding.** The real aggregate-persistence seam is `ProjectionStorage<TDoc,TId>` (`Store` / `Store(id,tenant)` / `StoreProjection`), **not** `DocumentSessionBase.store<T>` — the daemon uses the session-free `*Projected` operations (the #4667 path). When rebuilding, those route to `InsertProjected` instead of `UpsertProjected`/`OverwriteProjected`, so a rebuild batch classifies as `BatchFlushMode.InsertOnly`.

3. **Gates the routing behind an EXPERIMENTAL, default-off flag** (`ProjectionOptions.RebuildWithInsertOnly`, internal). **Zero behavior change to any existing rebuild.**

## Why it's gated off — the Phase-4 prerequisite (proven, not asserted)

The issue's "rebuild is INSERT-only post-TRUNCATE" premise only holds if **each aggregate id is written exactly once** across the whole rebuild. Marten's rebuild does **not** guarantee that today: it flushes **page-by-page** (one `ProjectionUpdateBatch` per `EventPage`, `GroupedProjectionExecution.processRangeAsync`), and the aggregate cache spans pages — so a stream whose events span multiple pages has its aggregate re-stored once per page. With UPSERT that's fine (later pages overwrite). With INSERT, the second page throws `23505 / DocumentAlreadyExistsException`.

`Bug_4685_rebuild_insert_only_multipage` pins this empirically with a 30-event single stream at `BatchSize=10` (3 pages):

- `default_upsert_routing_rebuilds_a_multipage_stream_correctly` → **passes** (Count == 30; default routing unchanged).
- `insert_only_routing_collides_on_a_multipage_stream_until_phase_4` → with the flag on, the rebuild cannot reach Count == 30 (duplicate-key).

The single-flush guarantee that makes INSERT-only safe is **CritterWatch#208 Phase 4** (deferred flush via identity-map accumulation), which must land first. When it does, the second test flips to expect success and the flag can be promoted to default-on (and PR 3 dispatches the binary `COPY` path on the resulting `InsertOnly` batches).

## Tests

- `DaemonTests.Bug_4685_rebuild_insert_only_multipage` (2 tests, both green)
- Regression sweep green: `DaemonTests.Aggregations` (36), `DaemonTests.MultiTenancy` (8), `rebuilding_an_inline_projection`.

Draft until Phase 4 lands.

### DIFF
--- a/src/DaemonTests/Bug_4685_rebuild_insert_only_multipage.cs
+++ b/src/DaemonTests/Bug_4685_rebuild_insert_only_multipage.cs
@@ -1,0 +1,121 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using DaemonTests.TestingSupport;
+using JasperFx.Events.Projections;
+using Marten;
+using Marten.Events.Aggregation;
+using Shouldly;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace DaemonTests;
+
+public record Bug4685Incremented();
+
+public class Bug4685Counter
+{
+    public Guid Id { get; set; }
+    public int Count { get; set; }
+}
+
+public partial class Bug4685CounterProjection: SingleStreamProjection<Bug4685Counter, Guid>
+{
+    public Bug4685CounterProjection()
+    {
+        Name = "Bug4685Counter";
+        Options.TeardownDataOnRebuild = true;
+
+        // Force a single stream's events to span multiple rebuild pages: 30 events / BatchSize 10
+        // => 3 pages => 3 ProjectionUpdateBatch flushes that each re-store this stream's aggregate.
+        Options.BatchSize = 10;
+    }
+
+    public void Apply(Bug4685Incremented _, Bug4685Counter counter) => counter.Count++;
+}
+
+/// <summary>
+/// #4685 PR 2 — proving the blocker. Marten's rebuild flushes page-by-page (one
+/// <c>ProjectionUpdateBatch</c> per <c>EventPage</c>, see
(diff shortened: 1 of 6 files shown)


jeremydmiller:
Closing with #4685 as part of the July 2026 rebuild-optimization consolidation (https://github.com/JasperFx/CritterWatch/issues/208#issuecomment-5005009049). The load-bearing artifact here — the empirical proof that INSERT-only rebuild routing requires the single-flush-per-aggregate guarantee (Bug_4685_rebuild_insert_only_multipage) — is now the foundation of jasperfx#525 (Phase 4, P0 in the consolidated plan). This branch is preserved and will be re-picked-up as P3 once Phase 4 lands: flip the pinned test, promote RebuildWithInsertOnly, then dispatch binary COPY on InsertOnly batches.

Issue #4705, opened 2026-06-09
Composite projection shards stall at sequence 1 under per-tenant event partitioning (sharded conjoined tenancy)


erdtsieck:
## Summary

With per-tenant event partitioning on a sharded, conjoined-tenancy store, async **composite** projection shards never advance past `last_seq_id = 1`, even though the high-water mark is fully caught up. A plain standalone `ProjectionLifecycle.Async` projection on the *same* store does advance to the high-water. No work-around found.

## Environment

- Marten 9.7.0 / JasperFx 2.9.1 / Weasel.Postgresql 9.0.3
- WolverineFx 6.5.1 (Wolverine-managed projection daemon distribution: `UseWolverineManagedEventSubscriptionDistribution = true`)
- .NET 10, PostgreSQL 16
- Same behavior observed on Marten 9.6.0 and 9.7.0.

## Store configuration

```csharp
storeOptions.MultiTenantedWithShardedDatabases(x =>
{
    x.ConnectionString = masterConnectionString;   // registry DB
    foreach (var (name, conn) in shards) x.AddDatabase(name, conn);
    x.UseSmallestDatabaseAssignment();
});
storeOptions.Events.TenancyStyle = TenancyStyle.Conjoined;
storeOptions.Events.UseArchivedStreamPartitioning = false;
storeOptions.Events.UseTenantPartitionedEvents = true;
// AppendMode left at the 9.x default (QuickWithServerTimestamps)
```

Every projected document is tenant-partitioned via `MultiTenantedWithPartitioning(x => x.ByList())`. Tenants are provisioned with `store.Advanced.AddTenantToShardAsync(tenantId)` (which creates the per-tenant LIST partitions + per-tenant event sequence) before any append.

## Repro

1. Fresh databases (master + shards), schema applied at startup.
2. Seed ~11,266 events for a single tenant; it gets assigned to one shard via `UseSmallestDatabaseAssignment`.
3. Start the async daemon.

## Observed

`mt_event_progression` on the shard that holds the tenant:

```
11266 | HighWaterMark
11266 | InvoiceJournalEntries:V2:All     <-- the only standalone Async projection
    1 | ProvidedCare:V10:All
    1 | provided-cares:V10:All           <-- composite bundle
    1 | Invoice:V7:All
    1 | invoices:V7:All                  <-- composite bundle
    1 | InsurerClaimLine:V7:All
    1 | claim_lines:V7:All               <-- composite bundle
[...]

PR #4714, opened 2026-06-10
#4712 + #4717 — per-tenant partitioning daemon: safe-harbor high-water + per-tenant progression (JasperFx 2.9.4)

### LINKED ISSUE #4712: Composite projection rebuilds hang under per-tenant event partitioning (9.7.1) — SafeHarborTime computed as DateTime.MinValue (follow-up to #4705)
## Summary

Follow-up to #4705 (closed) — **9.7.1 improved the situation but composite projection rebuilds still hang** under per-tenant event partitioning on a sharded, conjoined-tenancy store. With #4705's fix, some composites now catch up to the high-water, but **two of four composite projections start a single-pass rebuild and never finish** — they freeze partway with no error. The daemon repeatedly fails to compute a "safe harbor" sequence and logs a **`SafeHarborTime` of `0001-01-01` (≈ `DateTime.MinValue`)**, which looks like the root cause.

Related: #4706 (separate symptom, same setup).

## Environment

- Marten **9.7.1** / JasperFx **2.9.2** / Weasel.Postgresql 9.1.2
- WolverineFx **6.6.0** (`UseWolverineManagedEventSubscriptionDistribution = true`)
- .NET 10, PostgreSQL 16

## Configuration

```csharp
storeOptions.MultiTenantedWithShardedDatabases(x =>
{
    x.ConnectionString = masterConnectionString;     // references/registry DB
    foreach (var (name, conn) in shards) x.AddDatabase(name, conn);
    x.UseSmallestDatabaseAssignment();
});
storeOptions.Events.TenancyStyle = TenancyStyle.Conjoined;
storeOptions.Events.UseArchivedStreamPartitioning = false;
storeOptions.Events.UseTenantPartitionedEvents = true;
```
Documents are tenant-partitioned via `MultiTenantedWithPartitioning(x => x.ByList())`; tenants provisioned with `store.Advanced.AddTenantToShardAsync(tenantId)`. The application registers four `CompositeProjectionFor(...)` bundles: `provided-cares`, `invoices`, `claim_lines`, `import-reconciliation` (each groups several `SingleStreamProjection`s), plus standalone async projections.

## Repro

1. Fresh databases (drop + recreate the master + shard DBs).
2. Seed ~11,266 events for a single tenant (assigned to one shard).
3. Start the Wolverine-managed async daemon.

## Observed

The daemon starts a **single-pass composite rebuild** (`floor=0, ceiling=11266`) for all four composites, but only two finish:

```
Starting single-pass composite rebuild for provided-cares:All       over events 0 to 11266
Starting single-pass composite rebuild for import-reconciliation:All over events 0 to 11266
Starting single-pass composite rebuild for invoices:All             over events 0 to 11266
Starting single-pass composite rebuild for claim_lines:All          over events 0 to 11266
Finished with optimized rebuild for projection/subscription provided-cares:All
Finished with optimized rebuild for projection/subscription import-reconciliation:All
   <-- invoices:All and claim_lines:All never log "Finished"
```

Resulting `mt_event_progression` on the shard holding the tenant (HighWaterMark = 11266):

```
   11266 | HighWaterMark
   -- finished composites + standalone shards (19): all at 11266
   11266 | provided-cares:All  /  ProvidedCare:All  /  ProvidedCareLifecycle:All  /  ProvidedCareReport:All  /  ProvidedCareTimeline:All
   11266 | FactuurProvidedCareIndex:All  /  MigratedMidwiferyProvidedCare:All  /  PreGoLiveMidwiferyProvidedCare:All  /  ClaimLineWithValidationError:All
   11266 | import-reconciliation:All + all Imported*:All
   11266 | InvoiceJournalEntries:All   (standalone async)
   -- invoices composite (8): FROZEN at 10000
   10000 | invoices:All  /  Invoice:All  /  InvoiceOverviewItem:All  /  InsurerInvoice:All  /  InstitutionInvoice:All  /  ServiceBureauInvoice:All
   10000 | GPFileImportOverviewItem:All  /  ManualClaimImportOverviewItem:All
   -- claim_lines composite (12): FROZEN at 50
      50 | claim_lines:All  /  ClaimLineByReferenceNumber:All  /  ClientClaimLine:All  /  InstitutionClaimLine:All  /  InsurerClaimLine:All  /  ServiceBureauClaimLine:All
      50 | PendingClientClaimLine:All  /  PendingInstitutionClaimLine:All  /  PendingInsurerClaimLine:All  /  PendingServiceBureauClaimLine:All
      50 | RejectedOrPartiallyApprovedCare:All  /  FinalizedRejectedOrPartiallyApprovedCare:All
```

The two frozen composites do **not** advance with more time (measured stable over several minutes). There are **no `Error`/`Fatal` log entries and no `ShardStopped`/exception** — it is a silent hang.

## Likely root cause — bogus safe-harbor time

The daemon logs these two messages **~1340 times each** while the rebuilds are hung:

```
Daemon projection high water detection was not able to determine a safe harbor sequence, will try again soon

High Water agent is stale at 11266 for database postgresql://.../feature2_claims1/public
High Water agent is stale after threshold of 3 seconds, skipping gap to events marked
  after 0001-01-01T01:00:03+01:00 for database postgresql://.../feature2_claims1/public
```

So the high-water mark itself reaches 11266 (the detector sees all events), but the computed **`SafeHarborTime` is `0001-01-01T01:00:03` (≈ `DateTime.MinValue` + the 3s threshold)**. The gap-skip "to events marked after 0001-01-01" is effectively a no-op, the detector can never determine a safe harbor sequence, and the in-flight composite rebuilds for `invoices` and `claim_lines` never get a valid ceiling to complete — they hang at 10000 and 50 respectively.

## Expected

All composite rebuilds complete to the high-water (11266), as `provided-cares` and `import-reconciliation` do.

## Notes

- Same store/config as #4705 and #4706. #4705 (composite shards stuck at `last_seq_id = 1`) is clearly improved — most shards now advance and two composites fully rebuild — but `invoices` and `claim_lines` still hang.
- The `0001-01-01` `SafeHarborTime` suggests the safe-harbor timestamp can't be read for these shards/databases under sharded + per-tenant partitioning and defaults to `DateTime.MinValue`.
- Happy to share the full JSON daemon log or give access to the environment.

jeremydmiller:
Fixes #4712. Follow-up to #4705.

## Problem
Under `UseTenantPartitionedEvents` on a sharded conjoined store, composite projection rebuilds hang. The daemon logs a `SafeHarborTime` of `0001-01-01` (≈ `DateTime.MinValue` + the 3s stale threshold), the gap-skip becomes a no-op, and the store-global high-water agent loops forever — silently freezing some composite rebuilds.

## Root cause (the #4705 bug class, one query that was missed)
The store-global `HighWaterStatisticsDetector` reads `select last_value from mt_events_sequence` for `HighestSequence`. Under per-tenant partitioning the store-global `mt_events_sequence` is never advanced (each tenant draws seq_ids from its own `mt_events_sequence_{suffix}`), so `HighestSequence` reads 1 while the true mark is far higher → the agent treats the store as perpetually `Stale`. And because no store-global `HighWaterMark` progression row is read, `HighWaterStatistics.Timestamp` is left at `default(DateTimeOffset)` = `0001-01-01` — the source of the bogus SafeHarborTime.

## Fix (mirrors the #4705 `FetchHighestEventSequenceNumber` change)
- Read `coalesce(max(seq_id), 0)` from `mt_events` when `UseTenantPartitionedEvents`.
- Stamp `Timestamp` from that first result (which always returns a row), so it can never be left at `0001-01-01` when the progression row is absent.
- Non-partitioned stores keep reading `last_value from mt_events_sequence`.

## Test
`Bug_4712_safe_harbor_high_water` drives `HighWaterDetector.Detect` directly under per-tenant partitioning. Before the fix: `HighestSequence=1`, `Timestamp=0001-01-01` (with `CurrentMark=40`). After: `HighestSequence=40` and a real timestamp. Deterministic single-DB single-tenant repro (the detector-level seam; the sharded multi-composite hang is the downstream symptom). Non-partitioned high-water detection tests unchanged (10/10).

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -43,13 +43,13 @@
          JasperFx 2.9.2: jasperfx#431 — ProjectionErrorHandlingDescriptor on EventStoreUsage
          so monitoring tools (CritterWatch) can read the daemon error-handling policy off
          the wire instead of presuming skip-and-DLQ behaviour. See JasperFx/ProductSupport#3. -->
-    <PackageVersion Include="JasperFx" Version="2.9.2" />
-    <PackageVersion Include="JasperFx.Events" Version="2.9.2" />
-    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.9.2">
+    <PackageVersion Include="JasperFx" Version="2.9.4" />
+    <PackageVersion Include="JasperFx.Events" Version="2.9.4" />
+    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.9.4">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageVersion>
-    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.9.2" />
+    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.9.4" />
     <PackageVersion Include="Jil" Version="3.0.0-alpha2" />
     <PackageVersion Include="Lamar" Version="7.1.1" />
     <PackageVersion Include="Lamar.Microsoft.DependencyInjection" Version="15.0.0" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -43,13 +43,13 @@
          JasperFx 2.9.2: jasperfx#431 — ProjectionErrorHandlingDescriptor on EventStoreUsage
          so monitoring tools (CritterWatch) can read the daemon error-handling policy off
          the wire instead of presuming skip-and-DLQ behaviour. See JasperFx/ProductSupport#3. -->
-    <PackageVersion Include="JasperFx" Version="2.9.4" />
-    <PackageVersion Include="JasperFx.Events" Version="2.9.4" />
-    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.9.4">
+    <PackageVersion Include="JasperFx" Version="2.9.5" />
+    <PackageVersion Include="JasperFx.Events" Version="2.9.5" />
+    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.9.5">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageVersion>
-    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.9.4" />
+    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.9.5" />
     <PackageVersion Include="Jil" Version="3.0.0-alpha2" />
     <PackageVersion Include="Lamar" Version="7.1.1" />
     <PackageVersion Include="Lamar.Microsoft.DependencyInjection" Version="15.0.0" />
--- a/src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
+++ b/src/Marten/Events/Daemon/HighWater/HighWaterDetector.cs
@@ -337,6 +337,42 @@ left join {schema}.mt_event_progression prog
         return _runner.Query(new MarkHighWaterQueryHandler(_graph, currentMark), token);
     }
 
+    // #4717: persist a durable per-tenant high-water row (HighWaterMark:<tenant>) so each tenant's mark
+    // survives a daemon restart. Under per-tenant event partitioning the store-global mt_events_sequence
+    // is never advanced and a single HighWaterMark row cannot represent multiple tenants. Invoked by
+    // JasperFx's TenantedHighWaterCoordinator on each vectorized per-tenant poll.
+    public Task MarkHighWaterForTenantAsync(string tenantId, long sequence, CancellationToken token)
+    {
+        return _runner.Query(new MarkTenantHighWaterQueryHandler(_graph, tenantId, sequence), token);
+    }
+
+    public class MarkTenantHighWaterQueryHandler: ISingleQueryHandler<bool>
+    {
+        private readonly EventGraph _graph;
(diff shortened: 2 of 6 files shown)


jeremydmiller:
## Follow-up: #4717 per-tenant progression (consume JasperFx 2.9.4)

Building on the #4712 fix, this PR now also adopts the per-tenant continuous-progression work shipped in **JasperFx.Events 2.9.4** and resolves **#4717**.

- Bumped `JasperFx*` 2.9.2 → **2.9.4** (per-tenant continuous agents fan out one per `(shard, tenant)`; also carries the projections-rebuild subscription fix, jasperfx#438).
- `HighWaterDetector.MarkHighWaterForTenantAsync` — Marten implements the new `IHighWaterDetector` hook to persist a durable `HighWaterMark:<tenant>` row.
- **Un-skipped `Bug_4717_per_tenant_progression`** — now green: two tenants of different heights, composite + standalone, each get per-tenant `<Projection>:All:<tenant>` rows **and** per-tenant high-water rows at their own height.
- `sharded_daemon_per_shard_progression` (×2) now **poll** for the per-tenant rows/docs. `WaitForNonStaleData`'s caught-up check counts store-global shards, so it can return before a per-tenant agent commits on a partitioned store — making the immediate assert racy. Per-tenant progression itself is correct; only the test wait was racy.

**Verification:** full `TenantPartitionedEventsTests` **186/186** on clean DBs; the sharded tests are stable across repeated runs (including under within-run shard accumulation).

> Investigation note: the earlier "sharded regression" / "1-event-per-tenant commit bug" turned out to be a flaky `WaitForNonStaleData` + dirty-DB artifact — **not** a product bug. Per-tenant progression is production-correct, so no JasperFx 2.9.5 was needed. Hardening `WaitForNonStaleData` to be per-tenant-aware (so the public helper is reliable on partitioned stores) is a worthwhile separate follow-up.

PR #4757, opened 2026-06-17
Fix #4753/#4754: stop destructive rebuild of tenant-partitioned event tables

### LINKED ISSUE #4706: Schema migration destructively rebuilds tenant LIST-partitioned document tables on every startup over existing data
## Summary

On a store using `MultiTenantedWithPartitioning(x => x.ByList())` document tenancy, `ApplyAllConfiguredChangesToDatabaseAsync` at startup performs a full **table rebuild** (`CREATE <table>_temp` → `INSERT ... SELECT FROM <table>_temp` → drop/rename) for the partitioned document tables — **even when nothing changed** (no projection version bump, same `_N` suffix between deploys). The rebuilt partitioned table does **not** have its per-tenant LIST partitions recreated before the data copy, so the copy fails with `23514: no partition of relation "<table>" found for row` (or `57014: canceling statement due to user request` once the copy is large enough to hit the command timeout).

Net effect: every deploy over a database that already holds tenant data fails at startup. A fresh/empty database is fine (the tables are created empty, so there is no copy).

## Environment

- Marten 9.7.0 / JasperFx 2.9.1 / Weasel.Postgresql 9.0.3
- .NET 10, PostgreSQL 16
- Same behavior on 9.6.0 and 9.7.0.

## Configuration

```csharp
storeOptions.MultiTenantedWithShardedDatabases(...);   // sharded conjoined tenancy
storeOptions.Events.TenancyStyle = TenancyStyle.Conjoined;
storeOptions.Events.UseTenantPartitionedEvents = true;
storeOptions.Events.UseArchivedStreamPartitioning = false;

// every projected document:
mapping.StartIndexesByTenantId().MultiTenantedWithPartitioning(x => x.ByList());
```

Tenants/partitions are created with `store.Advanced.AddTenantToShardAsync(tenantId)`.

## Observed (startup against a DB that already has data)

`MartenSchemaException: DDL Execution for 'All Configured Changes' Failed!`, whose script contains rebuild copies for many partitioned doc tables:

```sql
insert into public.mt_doc_providedcare_10(tenant_id, id, data, mt_last_modified, mt_version, mt_dotnet_type, correlation_id, causation_id, headers)
    select tenant_id, id, data, mt_last_modified, mt_version, mt_dotnet_type, correlation_id, causation_id, headers
    from public.mt_doc_providedcare_10_temp;
insert into public.mt_doc_claimlinebyreferencenumber_7(...) select ... from public.mt_doc_claimlinebyreferencenumber_7_temp;
insert into public.mt_doc_claimlinewithvalidationerror_10(...) select ... from public.mt_doc_claimlinewithvalidationerror_10_temp;
insert into public.mt_doc_afgeleidcontractueeltarief(...) select ... from public.mt_doc_afgeleidcontractueeltarief_temp;
-- ... many more partitioned doc tables
```

Inner exception, depending on data size:

```
Npgsql.PostgresException: 23514: no partition of relation "mt_doc_providedcare_10" found for row
    Routine: ExecFindPartition
-- or, on larger copies:
Npgsql.PostgresException: 57014: canceling statement due to user request
```

The `_N` version suffixes are unchanged between the previous and current deploy, so this is not a real schema change — it looks like a **non-idempotent schema diff** for `MultiTenantedWithPartitioning(ByList)` tables that flags them for a destructive rebuild on every startup.

## Expected

Either:
- the partitioned document tables should not be diffed as "needs rebuild" when they are unchanged, or
- the rebuild path should recreate the per-tenant LIST partitions on the new table before copying the data (and preserve them across the swap).

## Notes

- The perpetual diff may be tied to `StartIndexesByTenantId()` prepending `tenant_id` to computed indexes on partitioned tables (the generated index DDL reads back differently than configured).
- Adjacent to #4665 / #4679 (per-tenant event partitioning + sharded tenancy). Filed alongside a separate report about composite projection shards stalling under the same setup.

### LINKED ISSUE #4753: Event-store tables (mt_streams/mt_events) not covered by #4706 fix — still destructively rebuilt over existing data under UseTenantPartitionedEvents
### Summary

The fix for #4706 (managed ByList tenant-partitioned tables destructively rebuilt on every startup) was applied to **document** tables but **not** to the **event-store** tables. With `Events.UseTenantPartitionedEvents = true`, `public.mt_streams` and `public.mt_events` are still diffed as "needs rebuild" on every startup and destructively rebuilt over existing data, failing with `23514: no partition of relation "mt_streams" found for row` on any database that already holds events.

This is the exact scenario the closed #4706 fixed for documents (`reapply_over_existing_tenant_data_is_idempotent`), but the event tables regressed/were never covered.

### Environment
- Marten 9.8.2 (the #4706 fix `58634c1` + #4713 / Weasel 9.1.5 are present in 9.8.x), Weasel 9.1.x, Npgsql 10, PostgreSQL 17, .NET 10
- `MultiTenantedWithShardedDatabases` + `Events.TenancyStyle = Conjoined` + `Events.UseTenantPartitionedEvents = true` + `Policies.PartitionMultiTenantedDocumentsUsingMartenManagement(...)`; tenants provisioned via `AddTenantToShardAsync` / `AddMartenManagedTenantsAsync`.

### Repro
1. Start the store, `ApplyAllConfiguredChangesToDatabaseAsync`, `AddMartenManagedTenantsAsync(ct, "acme")`, append events for `acme` (so `mt_streams`/`mt_events` get the per-tenant LIST partition + rows).
2. Restart the app (nothing changed).
3. Startup aborts: `CreateMigrationAsync().Difference` is **not** `None` — Marten emits a destructive rebuild of `mt_streams` and `mt_events`:
```
create table public.mt_streams_temp as select * from public.mt_streams;
drop table public.mt_streams cascade;          -- drops the per-tenant child partitions too
CREATE TABLE public.mt_streams (...) PARTITION BY LIST (tenant_id);   -- no child partitions
insert into public.mt_streams(...) select ... from public.mt_streams_temp;   -- 23514
```
→ `Npgsql.PostgresException 23514: no partition of relation "mt_streams" found for row` (same for `mt_events`). Verified the live table already matched the desired definition exactly (columns, `PARTITION BY LIST (tenant_id)`, child partition `mt_streams_08004464`, tenant registered in `tenants.mt_tenant_partitions`).

### Root cause
The #4706 fix sets `IgnorePartitionsInMigration` on the per-mapping **DocumentTable**:
- `src/Marten/Storage/DocumentTable.cs:108` `IgnorePartitionsInMigration = mapping.IgnorePartitions;`
- `src/Marten/Storage/DocumentTable.cs:119` `IgnorePartitionsInMigration = true;` (managed partitions)

But the **event-store tables never set it**:
- `src/Marten/Events/Schema/StreamsTable.cs:79` — `Partitioning = new ListPartitioning { Columns = [TenantIdColumn.Name] }.UsePartitionManager(manager);` (no `IgnorePartitionsInMigration`)
- `src/Marten/Events/Schema/EventsTable.cs` — same: `Partitioning` set via the manager, `IgnorePartitionsInMigration` never set.

So for the event tables, `ListPartitioning.CreateDelta` (Weasel) still compares the manager's partition list against the actual DB partitions; when they don't line up (e.g. the managed partition manager isn't seen as authoritative the way it is for documents), it returns `PartitionDelta.Rebuild` and the destructive migration above is generated.

### Expected
Re-applying an unchanged `UseTenantPartitionedEvents` store over existing data is idempotent — `migration.Difference == None`, no destructive rebuild of `mt_streams`/`mt_events`, no 23514 — exactly as #4706's `Bug_4706_partitioned_table_idempotent_migration` now asserts for documents.

### Suggested fix
Apply the same `IgnorePartitionsInMigration = true` (managed-partition) treatment to `StreamsTable` and `EventsTable` when `UseTenantPartitionedEvents` is on, mirroring `DocumentTable`. A regression test analogous to `Bug_4706_...` but exercising the event store (append events + restart) would lock it in.

### LINKED ISSUE #4754: Managed ByList tenant-partition Rebuild migration drops child partitions and copies data back before recreating them (23514 / data loss)
### Summary

When the schema diff decides a Marten-managed **ByList** tenant-partitioned table needs a rebuild (`PartitionDelta.Rebuild`), the generated migration is **partition-unaware and data-destructive**: it `DROP ... CASCADE`s the parent (removing every per-tenant child partition), recreates only the parent, and then copies the data back **before** the managed LIST child partitions are recreated — so the copy fails with `23514: no partition of relation "..." found for row` (and on an empty-but-once-populated table it would silently lose the partitions).

This is the failure mode behind #4706 / this report's sibling issue, but filed separately because it is the **rebuild path itself**: even when a rebuild is legitimately required, it must not destroy/omit the managed partitions before re-inserting data.

### Observed migration SQL (Marten 9.8.2 / Weasel 9.1.x, production)
```
create table public.mt_streams_temp as select * from public.mt_streams;
drop table public.mt_streams cascade;
CREATE TABLE public.mt_streams ( ... ) PARTITION BY LIST (tenant_id);
insert into public.mt_streams(...) select ... from public.mt_streams_temp;   -- 23514: no partition found
drop table public.mt_streams_temp cascade;
```
The recreated parent has `PARTITION BY LIST (tenant_id)` but no child partitions; the `tenants.mt_tenant_partitions` registrations and the per-tenant child partitions created by `AddMartenManagedTenantsAsync` are gone after the `CASCADE`, so the `INSERT ... SELECT` has nowhere to land.

### Where (indicative — please confirm against current Weasel)
In the local Weasel 8.15.0 source this rebuild SQL is generated in `src/Weasel.Postgresql/Tables/TableDelta.cs` `WriteUpdate` under `PartitionDelta.Rebuild` (~lines 148-164): `create <t>_temp as select *` → `drop table <t> cascade` → `Expected.WriteCreateStatement(...)` → `insert into <t> select ... from <t>_temp`. `WriteCreateStatement` writes only the parent `PARTITION BY LIST`; the managed child partitions from the `IListPartitionManager` are not (re)created before the insert.

### Expected
For a managed-list-partitioned table, a rebuild must either (a) recreate the managed child partitions (from the partition manager / `mt_tenant_partitions`) **before** copying data back, or (b) not be generated at all for managed partitions (treat the managed partition set as not-a-diff). Otherwise the rebuild is a data-loss / startup-failure hazard.

### Notes
- Closely related to #4706 (which fixed *triggering* the rebuild for document tables via `IgnorePartitionsInMigration`). This issue is about the rebuild SQL being unsafe whenever it *is* generated for a managed-partition table.
- Reproduces today on the event-store tables (`mt_streams`/`mt_events`) with `UseTenantPartitionedEvents` — see the sibling issue about those tables not getting the #4706 `IgnorePartitionsInMigration` treatment.

jeremydmiller:
Follow-up to #4756, which was squash-merged with only its first commit (the CI workflow). The #4753/#4754 fix that had been pushed to the same branch landed after the merge and was therefore missed. This PR carries that commit forward unchanged.

Closes #4753. Closes #4754.

## Problem

With `Events.UseTenantPartitionedEvents = true` on a sharded store, re-applying an unchanged schema (every app restart) destructively rebuilt the managed ByList partitioned `mt_streams` / `mt_events` (`CREATE _temp` → `DROP … CASCADE` → recreate parent → `INSERT … SELECT`) **without recreating the per-tenant LIST partitions first**, so the copy of existing events failed with `23514: no partition of relation "mt_streams" found for row`.

## Fix

#4706 fixed exactly this for document tables by setting `IgnorePartitionsInMigration = true` on the managed-partition `DocumentTable`; the flag was never applied to the event tables. This mirrors it on `StreamsTable` and `EventsTable` in the `UseTenantPartitionedEvents` block. The per-tenant partitions are created out-of-band by `AddMartenManagedTenantsAsync` / `AddTenantToShardAsync` (the additive path), so the generic schema diff must not reconcile them. With the flag set, Weasel's `ListPartitioning.CreateDelta` short-circuits to `PartitionDelta.None`, making the destructive Rebuild branch (#4754) **unreachable for Marten's managed tables** — no Weasel change is required to resolve the reported symptom.

## Test

`Bug_4753_sharded_partitioned_event_rebuild` (sharded, mirrors `Bug_4706_sharded_partitioned_doc_rebuild` but appends **events** so the rebuild's `INSERT … SELECT` actually has rows to fail on): **RED before the fix** with the exact 23514, **GREEN after**. Full `TenantPartitionedEventsTests` suite green (189/189). Now that #4756's CI workflow is on master, this test is exercised by the `TenantPartitionedEvents` CI job.

### DIFF
--- a/src/Marten/Events/Schema/EventsTable.cs
+++ b/src/Marten/Events/Schema/EventsTable.cs
@@ -198,6 +198,16 @@ internal class EventsTable: Table
             var manager = events.Options.TenantPartitions!.Partitions;
             Partitioning = new ListPartitioning { Columns = [TenantIdColumn.Name] }
                 .UsePartitionManager(manager);
+
+            // #4753 (mirrors #4706 for DocumentTable): a Marten-managed LIST partitioning keeps its
+            // PARTITION BY clause but is exempt from child-partition reconciliation in the generic
+            // schema diff. The per-tenant partitions are created out-of-band by
+            // AddMartenManagedTenantsAsync / AddTenantToShardAsync (the additive path). Without this,
+            // re-applying an unchanged schema over existing data sees the live per-tenant partitions
+            // as "unexpected" and destructively rebuilds mt_events (CREATE _temp / DROP CASCADE /
+            // recreate parent / INSERT…SELECT), failing with 23514 because the rebuilt parent has no
+            // partitions yet. Setting this makes Weasel's destructive Rebuild path unreachable.
+            IgnorePartitionsInMigration = true;
         }
     }
 
--- a/src/Marten/Events/Schema/PerTenantEventSequences.cs
+++ b/src/Marten/Events/Schema/PerTenantEventSequences.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Npgsql;
 using Weasel.Core;
 using Weasel.Postgresql;
 using DbCommandBuilder = Weasel.Core.DbCommandBuilder;
@@ -158,9 +159,24 @@ internal class PerTenantEventSequences: ISchemaObject
         foreach (var suffix in partitionSuffixes)
         {
             var sequenceName = $"\"{eventSchema}\".\"mt_events_sequence_{suffix}\"";
-            await conn
-                .CreateCommand($"create sequence if not exists {sequenceName} as bigint;")
-                .ExecuteNonQueryAsync(token).ConfigureAwait(false);
+            try
+            {
+                await conn
+                    .CreateCommand($"create sequence if not exists {sequenceName} as bigint;")
+                    .ExecuteNonQueryAsync(token).ConfigureAwait(false);
+            }
+            catch (PostgresException e) when (e.SqlState is PostgresErrorCodes.UniqueViolation
+                                                  or PostgresErrorCodes.DuplicateTable
+                                                  or PostgresErrorCodes.DuplicateObject)
+            {
+                // #4596/#4757: CREATE SEQUENCE IF NOT EXISTS is NOT atomic against a concurrent create.
+                // Two tasks registering the same tenant (e.g. concurrent AddMartenManagedTenantsAsync)
+                // can both pass the IF NOT EXISTS check and race the catalog insert; the loser gets
+                // 23505 on pg_class_relname_nsp_index (or 42P07 / 42710). The sequence now exists — which
+                // is precisely the idempotent outcome we want — so swallow the concurrent-create race.
+                // Each statement runs in its own implicit transaction (no outer transaction here), so the
+                // failure does not poison the connection for the remaining suffixes.
+            }
         }
         await conn.CloseAsync().ConfigureAwait(false);
     }
--- a/src/Marten/Events/Schema/StreamsTable.cs
+++ b/src/Marten/Events/Schema/StreamsTable.cs
@@ -78,6 +78,14 @@ internal class StreamsTable: Table
             var manager = events.Options.TenantPartitions!.Partitions;
             Partitioning = new ListPartitioning { Columns = [TenantIdColumn.Name] }
                 .UsePartitionManager(manager);
+
+            // #4753 (mirrors #4706 for DocumentTable): exempt this Marten-managed LIST partitioning
+            // from child-partition reconciliation in the generic schema diff. The per-tenant
+            // partitions are created out-of-band by AddMartenManagedTenantsAsync /
+            // AddTenantToShardAsync. Without this, re-applying an unchanged schema over existing data
+            // sees the live partitions as "unexpected" and destructively rebuilds mt_streams,
+            // failing with 23514 because the rebuilt parent has no partitions yet.
+            IgnorePartitionsInMigration = true;
         }
     }
 }
--- a/src/TenantPartitionedEventsTests/Regressions/Bug_4596_partition_race_and_MT002_rebuild.cs
+++ b/src/TenantPartitionedEventsTests/Regressions/Bug_4596_partition_race_and_MT002_rebuild.cs
@@ -48,8 +48,13 @@ public class Bug_4596_partition_race_and_MT002_rebuild : IAsyncLifetime
     private string _schema = null!;
     private DocumentStore _store = null!;
 
-    public async Task InitializeAsync()
+    public Task InitializeAsync() => freshStoreAsync();
+
+    // Build a brand-new isolated store against a brand-new schema. Used by InitializeAsync and re-used
+    // per attempt by the concurrent-race test so each retry starts from a clean slate.
+    private async Task freshStoreAsync()
     {
+        _store?.Dispose();
         _schema = $"tp_4596_{Environment.ProcessId}_{Guid.NewGuid():N}".Substring(0, 32);
 
         await using (var conn = new NpgsqlConnection(ConnectionSource.ConnectionString))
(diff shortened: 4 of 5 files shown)


jeremydmiller:
Pushed an extra commit addressing the CI flake in `concurrent_AddMartenManagedTenantsAsync_for_same_tenant_is_idempotent` (the `test-tenant-partitioned-events` job). Diagnosing it found **two** independent races:

1. **Real Marten bug, fixed here** — `PerTenantEventSequences.EnsureSequencesAsync` used `create sequence if not exists`, which is not atomic against a concurrent create. Two tasks registering the same tenant both pass the existence check and race the catalog insert; the loser threw `23505` on `pg_class_relname_nsp_index` — exactly the leak this test guards against. Now swallows the concurrent-create races (23505/42P07/42710); the sequence exists afterward, which is the intended idempotent outcome.

2. **Known Weasel transient, hand-off** — under concurrent DDL, `Table.readExistingAsync` NREs on a torn catalog read (`ColumnFor(pk)!`). Tracked as JasperFx/weasel#307. The test now retries the race scenario only when every fault is that specific NRE transient; a real 23505/42P07 is never swallowed.

Verified 25/25 clean runs of the previously ~1-in-10 flaky test; full `TenantPartitionedEventsTests` suite green (189/189).

Issue #4761, opened 2026-06-18
Per-tenant progression (#4717) not materialized for multiple tenants sharing a shard DB under UseTenantPartitionedEvents (9.9.0) — only store-global rows


erdtsieck:
### Summary

The per-tenant progression mechanism added for #4717 ("a single store-global `:All` shard can't track multiple tenants") does **not** materialize on a shard database that hosts **multiple tenants** under `UseTenantPartitionedEvents` in Marten 9.9.0. Only the store-global `HighWaterMark` and `<projection>:All` rows are created — no `HighWaterMark:<tenant>` / per-tenant progression rows — re-exposing exactly the skip risk #4717 was meant to fix.

### Environment
- Marten 9.9.0 (the #4717 fix commit `41e6d9d0b` is present — tagged V9.7.2…V9.8.1, so included in 9.9.0), Npgsql 10, PostgreSQL 17, .NET 10
- `MultiTenantedWithShardedDatabases` (`UseSmallestDatabaseAssignment`) + `Events.TenancyStyle = Conjoined` + `Events.UseTenantPartitionedEvents = true`; tenants provisioned with `AddTenantToShardAsync`.

### Observed (live shard DB `…_claims1`)
Two tenants landed on the same shard, each with its own independent event sequence:
```
mt_events_sequence_08004464 = 11266      events for 08004464 = 11266
mt_events_sequence_01059910 =  8495      events for 01059910 =  8495
```
The two sequences are independent and their `seq_id` values overlap (both cover 1…8495). Yet `public.mt_event_progression` contains **44 rows, 0 of them per-tenant**:
```
HighWaterMark            = 11266
<projection>:All         = 11266   (for every projection)
```
No `HighWaterMark:08004464`, no `HighWaterMark:01059910`, no `<projection>:All:<tenant>`.

### Why this is a problem (the #4717 rationale)
With one store-global high-water (11266) over two independent, overlapping per-tenant sequences, a tenant whose sequence lags (here `01059910` at 8495) and later appends events at `seq_id` 8496…11266 produces rows **below** the already-advanced store-global high-water. A daemon paging `seq_id > last_processed` will not pick them up → **skipped events** for the lagging tenant. (In a one-shot bulk seed + single catch-up — as tested here — all data is present before the high-water advances, so read models did materialize correctly for both tenants; the hazard is incremental appends after the shard's high-water has moved past a lagging tenant's range.)

### Expected
[...]

Issue #4751, opened 2026-06-17
ForceAllMartenDaemonActivityToCatchUpAsync does not catch up an async CompositeProjection under MultiTenantedWithShardedDatabases + UseTenantPartitionedEvents


erdtsieck:
### Summary

On a store configured with `MultiTenantedWithShardedDatabases` + `Events.UseTenantPartitionedEvents = true` (conjoined tenancy, Marten-managed LIST partitioning), an **async `CompositeProjection`** is **not driven to catch up** by `IHost.ForceAllMartenDaemonActivityToCatchUpAsync()`. After the call returns, the composite's stage‑2 read models are still missing for the tenant whose events were just appended (`LoadAsync<T>(id)` on that tenant returns `null`).

The exact same composite projection catches up correctly when the store is **not** sharded (single database, `BuildProjectionDaemonAsync()` + `Daemon.WaitForNonStaleData()`), and the same code path works on a non-sharded store driven by `ForceAllMartenDaemonActivityToCatchUpAsync()` as well. Only the sharded + tenant-partitioned configuration reproduces the problem.

### Environment

- Marten 9.8.2, JasperFx 2.12.0
- WolverineFx 6.11.0 (Wolverine-managed event subscription distribution enabled; daemon runs in `DurabilityMode.Solo` in the integration-test host)
- Npgsql 10.0.3, PostgreSQL 17
- .NET 10

### Store configuration (abridged)

```csharp
storeOptions.Events.TenancyStyle = TenancyStyle.Conjoined;
storeOptions.Advanced.DefaultTenantUsageEnabled = false;

storeOptions.MultiTenantedWithShardedDatabases(x =>
{
    x.ConnectionString = referencesConnectionString; // tenant registry / master
    x.AddDatabase("claims1", "...");
    x.AddDatabase("claims2", "...");
    x.AddDatabase("claims3", "...");
    x.AddDatabase("claims4", "...");
    x.UseSmallestDatabaseAssignment();
});

storeOptions.Events.UseArchivedStreamPartitioning = false;
storeOptions.Events.UseTenantPartitionedEvents = true;

// "invoices" is registered as an ASYNC composite projection:
storeOptions.Projections.CompositeProjectionFor("invoices", projection =>
{
    projection.Add(new InvoiceProjection(...));            // stage 1: snapshot
    projection.Add<ReturnDecisionsForInvoiceProjection>(); // stage 1
[...]

PR #4762, opened 2026-06-18
Failing repros: sharded tenant-partitioned daemon catch-up + tenant-count divergence (#4751, #4761, #4763)

erdtsieck:
Adds **failing** integration tests under `src/TenantPartitionedEventsTests/Sharded/` that reproduce three problems under `MultiTenantedWithShardedDatabases` + `Events.UseTenantPartitionedEvents = true`. The daemon repros use Marten's own daemon (`BuildProjectionDaemonAsync` + `StartAllAsync` + `WaitForNonStaleData`) — no Wolverine involved — and reuse the existing `ShardedPartitionedFixture`.

### `Bug_4751_composite_catchup_under_sharded` — reproduces #4751
An async `CompositeProjectionFor` is not driven to a non-stale state by the normal async catch-up path. After `StartAllAsync` + `WaitForNonStaleData`, **both** composite stages are empty (`stage1 = 0, stage2 = 0`, expected 10). The same composite shape *does* materialize via `RebuildProjectionAsync` (see the existing `composite_side_effects_rebuild_under_sharded_partitioning` test), so the projection logic is fine — the gap is the continuous async catch-up of a composite on a sharded, tenant-partitioned store.

### `Bug_4761_per_tenant_progression_same_shard` — reproduces #4761
Two tenants assigned to the **same** shard DB, with divergent event counts (3 and 2). The per-tenant projection docs **do** materialize and the **per-tenant** progression rows **do** advance (`HighWaterMark:<tenant>`, `<projection>:All:<tenant>` reach 3/2) — so this corrects the original #4761 premise — but the **store-global** `<projection>:All` mark stays at `0`. `WaitForNonStaleData` reads the store-global mark, so it never sees the shard as non-stale and times out, even though every tenant is fully caught up.

### `Bug_4763_reassign_count_divergence` — reproduces #4763
Assigns a tenant to shard A, then re-assigns it to shard B via `AddTenantToShardAsync`, and reads the pool table back: shard A still reports `tenant_count = 1` after the tenant has left (expected 0). `AssignTenantAsync` recomputes only the target shard count and never decrements the source, so `UseSmallestDatabaseAssignment` mis-ranks shards. (The non-atomic provisioning sub-bug of #4763 — assignment committed before `createPartitionsForTenant` — needs fault injection and is **not** covered here.)

### Shared root of the two daemon repros
Both #4751 and #4761 look like the same underlying issue: on a shard that hosts more than one tenant, the **store-global** `<projection>:All` / `HighWaterMark` progression mark is never advanced (only the per-tenant marks are), and the catch-up helpers (`WaitForNonStaleData`, and by extension `ForceAllMartenDaemonActivityToCatchUpAsync`) key off the store-global mark.

### Notes
- Tests are written to **fail** on `master` (9.9.0) to demonstrate the bugs; they are not skipped.
- Verified failing locally against PostgreSQL 17.

Relates to: #4751, #4761, #4763

### DIFF
--- a/src/TenantPartitionedEventsTests/Sharded/Bug_4751_composite_catchup_under_sharded.cs
+++ b/src/TenantPartitionedEventsTests/Sharded/Bug_4751_composite_catchup_under_sharded.cs
@@ -1,0 +1,157 @@
+#nullable enable
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using JasperFx;
+using JasperFx.Core;
+using JasperFx.Events;
+using JasperFx.Events.Projections;
+using Marten;
+using Marten.Events;
+using Marten.Events.Aggregation;
+using Marten.Events.Projections;
+using Marten.Storage;
+using Marten.Testing.Harness;
+using Npgsql;
+using Shouldly;
+using Weasel.Core;
+using Weasel.Postgresql;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace TenantPartitionedEventsTests.Sharded;
+
+public record CmpLeg(double Distance);
+
+public class CmpTrip
+{
+    public Guid Id { get; set; }
+    public double Distance { get; set; }
+}
+
+public class CmpTripCount
+{
+    public Guid Id { get; set; }
+    public int Legs { get; set; }
+}
+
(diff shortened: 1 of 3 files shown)


jeremydmiller:
This was superseded by #4764

Issue #4765, opened 2026-06-18
Close mt_events_sequence gap on Quick + ExpectedVersion OCC failure (route non-partitioned Quick through bulk function)


jeremydmiller:
## Background

Follow-up planning issue for #4749. When a `FetchForWriting` / `AppendOptimistic` / `AppendExclusive` append loses an optimistic-concurrency race under `EventAppendMode.Quick` (or `QuickWithServerTimestamps`) with the **non-partitioned** event store, the failure leaves a permanent gap in `mt_events_sequence`: the sequence advances past the last persisted `seq_id` with no tombstone. The async daemon's high-water detector then stalls on the unreachable sequence value and `QueryForNonStaleData` hangs indefinitely (the symptom KMDjkb reported).

The **partitioned** path (`UseTenantPartitionedEvents`) is already safe — #4614 routes every append through `mt_quick_append_events`, which checks the expected version **before** any `nextval()`. This issue extends that hardening to the non-partitioned path.

## Root cause

`QuickEventAppender.registerOperationsForStreams` (non-partitioned, existing stream with `ExpectedVersionOnServer.HasValue`) queues two operations into a single Npgsql pipeline batch:

1. `UpdateStreamVersion` — `UPDATE mt_streams SET version = N+M WHERE version = N` (the OCC guard)
2. M × `QuickAppendEventWithVersionOperation` — `INSERT INTO mt_events (..., nextval('mt_events_sequence'))`

The OCC check is C#-side: `UpdateStreamVersion.PostprocessAsync` inspects `reader.RecordsAffected` **after** the entire batch returns. On a concurrent loser:

- `UPDATE` matches 0 rows (silent at the SQL layer — no error)
- The first `INSERT` raises `23505` on `pk_mt_events_stream_and_version`
- Txn rolls back — but `nextval()` is **non-transactional**, so the sequence stays advanced
- Daemon high-water reaches `MAX(mt_events.seq_id)` and stalls forever waiting for the unreachable id

## Fix — route every non-partitioned Quick append through `mt_quick_append_events`

The bulk function already reads `mt_streams.version` into a local variable and raises `MT003` (translated to `EventStreamUnexpectedMaxEventIdException` by `QuickAppendEventsOperationBase.TryTransform`) **before** the foreach loop that calls `nextval()`. We just need to make that path the only Quick path for non-partitioned too. Three changes:

### 1. `QuickAppendEventFunction.WriteCreateStatement`

[...]

Issue #4749, opened 2026-06-17
QueryForNonStaleData blocks indefinitely after duplicate key violation (event sequence gap) with AppendMode.quick


KMDjkb:
**Version**
Marten 9.7.2
postgres:18.1
.NET 10
**Configuration**
<img width="532" height="259" alt="Image" src="https://github.com/user-attachments/assets/58525b10-e7ce-43be-89c4-5f21fbe61a9a" />

After a duplicate key violation during event append, QueryForNonStaleData consistently times out and does not recover. This appears to be caused by a gap between mt_events_sequence and persisted events.

Exception:

```
Npgsql.PostgresException (0x80004005)
SQLSTATE: 23505
Message:
  duplicate key value violates unique constraint
  "pk_mt_events_stream_and_version"
Detail:
  (Redacted — enable 'Include Error Detail' in connection string to view)
Stack Trace:
  at Npgsql.Internal.NpgsqlConnector.ReadMessageLong(...)
  at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1...
  at Npgsql.NpgsqlDataReader.NextResult(...)
  at Marten.Internal.Sessions.OperationPage.ApplyCallbacksAsync(...)
     in OperationPage.cs:line 81
  at Marten.Internal.Sessions.AutoClosingLifetime.ExecuteBatchPagesAsync(...)
     in AutoClosingLifetime.cs:line 240
  at Marten.Internal.Sessions.AutoClosingLifetime.ExecuteBatchPagesAsync(...)
     in AutoClosingLifetime.cs:line 241
Severity:       ERROR
SqlState:       23505
Schema:         public
Table:          mt_events
Constraint:     pk_mt_events_stream_and_version
File:           nbtinsert.c
Line:           666
Routine:        _bt_check_unique
```

**Steps to Reproduce**
	1. Trigger two identical PUT requests concurrently (can reproduce on both multi-node and single-node setups).
	2. One request succeeds.
	3. The second request fails with: 
duplicate key value violates unique constraint
	4. Execute QueryForNonStaleData for any projection.

**Expected Behavior**
	• The failed transaction should not leave the system in a permanently stale state.
	• mt_events_sequence should not advance on failure

**Actual Behavior**
	• All QueryForNonStaleData calls time out: 
The query for [PROJECTION] timed out while waiting for non-stale data.
	• The mt_evemts_sequence have been bumped by one for the failed transaction without any tombstone mt_tombstone events.

[...]

PR #4771, opened 2026-06-19
fix: close mt_events_sequence gap on concurrent Quick OCC failures (#4749)

KMDjkb:
Description:

 ## Problem
 
 After #4765 fixed the *sequential* OCC gap (two saves on the same session, one
 after the other), a second race remained under truly concurrent writes.
 
 When two transactions both use `FetchForWriting` + Quick append mode against the
 same stream, READ COMMITTED allows both to pass the `SELECT version` check in
 `mt_quick_append_events` before either commits. Both then call `nextval()`, one
 inserts successfully, and the other fails with a `23505` on
 `pk_mt_events_stream_and_version`. Because PostgreSQL sequences are
 non-transactional, the loser's `nextval()` is never rolled back — leaving a
 permanent gap between `mt_events_sequence` and the highest committed `seq_id`.
 The async daemon's high-water mark can never reach the sequence value, so every
 `QueryForNonStaleData` call waits until timeout.
 
 ## Fix
 
 Add `FOR UPDATE` to the `SELECT version FROM mt_streams` — but **only in the OCC
 path** (`expected_version IS NOT NULL`):
 
 ```sql
 if expected_version IS NOT NULL then
     -- Loser blocks here until winner commits, then reads the updated
     -- version and raises MT003 *before* nextval() fires.
     select version, is_archived into event_version, stream_is_archived
     from mt_streams where ... for update;
     if COALESCE(event_version, 0) != expected_version then
         RAISE EXCEPTION ... USING ERRCODE = 'MT003';
     end if;
 else
     select version, is_archived into event_version, stream_is_archived
     from mt_streams where ...;  -- unchanged, no extra locking
 end if;

The loser now blocks at the SELECT until the winner commits, reads version N+1,
and raises MT003 before any nextval() call — no gap.

Non-OCC appends (StartStream, plain AppendMany) use the original SELECT
and are completely unaffected.

Test

Added concurrent_occ_appends_to_same_stream_leave_no_sequence_gap to
Bug_4765_quick_occ_no_sequence_gap. Uses a System.Threading.Barrier with
dedicated OS threads to release three SaveChangesAsync calls simultaneously,
reliably triggering the DB-level race. Asserts last_value == max committed seq_id (no gap), exactly 1 winner and 2 ConcurrencyException losers.

### DIFF
--- a/src/EventSourcingTests/Bugs/Bug_4765_quick_occ_no_sequence_gap.cs
+++ b/src/EventSourcingTests/Bugs/Bug_4765_quick_occ_no_sequence_gap.cs
@@ -43,7 +43,8 @@ public class Bug_4765_quick_occ_no_sequence_gap: OneOffConfigurationsContext
     }
 
     private async Task<DocumentStore> BuildStoreAsync(
-        string label, EventAppendMode mode, bool asyncSnapshot, TimeProvider timeProvider = null)
+        string label, EventAppendMode mode, bool asyncSnapshot, TimeProvider timeProvider = null,
+        bool exclusiveLock = false)
     {
         // Keep schema names well under PostgreSQL's 63-char identifier limit —
         // a too-long name gets silently truncated by PG, which then trips schema
@@ -63,6 +64,7 @@ public class Bug_4765_quick_occ_no_sequence_gap: OneOffConfigurationsContext
             opts.DatabaseSchemaName = schema;
             opts.AutoCreateSchemaObjects = AutoCreate.All;
             opts.Events.AppendMode = mode;
+            opts.Events.UseExclusiveLockOnConcurrentAppends = exclusiveLock;
             if (timeProvider != null)
             {
                 opts.Events.TimeProvider = timeProvider;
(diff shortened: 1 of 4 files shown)


jeremydmiller:
@KMDjkb I'm deeply unenthusiastic about this one. I think you need to find other ways to address the concurrency issues in your system. This would impact other people, and I'm not sure how the FOR UPDATE helps you here when Marten wouldn't be holding the transaction open. Anything like this would need to be an opt in behavior anyway because I'm not ramming DDL changes down everybody's throats because of one guy's special needs. We can't work that way. 

I'll take this, but only after you make it an opt in change.

KMDjkb:
Thanks for the quick reply! Just to clarify the transaction concern — the FOR UPDATE is entirely inside the PG function itself, so the lock is acquired and released within a single SaveChangesAsync call. Marten isn't holding anything open at the application layer; it's just two PG connections racing inside the function.
 
 That said, opt-in makes total sense. I've updated the PR with a UseExclusiveLockOnConcurrentAppends flag on StoreOptions.Events (defaults to false). When enabled it generates the FOR UPDATE branch in mt_quick_append_events; when disabled the function is identical to today's behaviour.

jeremydmiller:
And @KMDjkb, is KMD your company, or just part of your GH name? If it's the company, maybe y'all go talk to us again about a support agreement and we can help you out with whatever design is leading you to this much concurrency.

Issue #3281, opened 2024-06-24
Tombstone processing needs to track tenant id 


jeremydmiller:
Here's the culprit:

```csharp
    private async Task executeTombstoneBlock(UpdateBatch batch, CancellationToken cancellationToken)
    {
        await using var session = (DocumentSessionBase)_store.LightweightSession();
        await session.ExecuteBatchAsync(batch, cancellationToken).ConfigureAwait(false);
    }
```

That method needs to be tenant aware

jeremydmiller:
This was fixed, lost sight of when exactly

Issue #482, opened 2016-08-12
Tests failing on master


tim-cools:
```
==============================
==============================
All ActiveProject views match!
==============================
==============================
All ActiveProject views match!
==============================
==============================
All ActiveProject views match!
==============================
==============================
All ActiveProject views match!
==============================
    Marten.Testing.Events.query_against_event_documents_Tests.can_fetch_all_events_before_now [FAIL]
      Shouldly.ShouldAssertException : results.Count
    should be
0
    but was
      4
      Stack Trace:
           at Shouldly.ShouldlyCoreExtensions.AssertAwesomely[T](T actual, Func`2 specifiedConstraint, Object originalActual, Object originalExpected, Func`1 customMessage, String shouldlyMethod)
           at Shouldly.ShouldBeTestExtensions.ShouldBe[T](T actual, T expected, Func`1 customMessage)
           at Shouldly.ShouldBeTestExtensions.ShouldBe[T](T actual, T expected)
        Events\query_against_event_documents_Tests.cs(115,0): at Marten.Testing.Events.query_against_event_documents_Tests.can_fetch_all_events_before_now()
    Marten.Testing.Events.query_against_event_documents_Tests.can_fetch_all_events_after_now [FAIL]
      Shouldly.ShouldAssertException : results.Count
    should be
4
    but was
      0
      Stack Trace:
           at Shouldly.ShouldlyCoreExtensions.AssertAwesomely[T](T actual, Func`2 specifiedConstraint, Object originalActual, Object originalExpected, Func`1 customMessage, String shouldlyMethod)
           at Shouldly.ShouldBeTestExtensions.ShouldBe[T](T actual, T expected, Func`1 customMessage)
           at Shouldly.ShouldBeTestExtensions.ShouldBe[T](T actual, T expected)
        Events\query_against_event_documents_Tests.cs(101,0): at Marten.Testing.Events.query_against_event_documents_Tests.can_fetch_all_events_after_now()
    Marten.Testing.Schema.DocumentSchemaOnOtherSchemaTests.fixing_bug_343_double_export_of_events [FAIL]
      System.UnauthorizedAccessException : Access to the path 'C:\_\Marten\git-tim-cools\src\Marten.Testing\bin\Debug\net46\win7-x64\bin\allsql\all.sql' is denied.
      Stack Trace:
           at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
[...]

Issue #4682, opened 2026-06-08
Conjoined → UseTenantPartitionedEvents migration tool


jeremydmiller:
### Background

Marten 9 supports `Events.UseTenantPartitionedEvents = true` (introduced #4596), which physically partitions `mt_events` and `mt_streams` by `tenant_id` and gives each tenant its own event sequence. The feature requires `TenancyStyle.Conjoined`.

**No tool exists to migrate an existing conjoined event store to use tenant partitioning.** Customers maturing into this feature today have to write custom SQL and accept the operational risk of bespoke data movement at scale. This issue tracks the first-class migration tool.

Reverse migration (partitioned → conjoined) is **deferred** — only build if a customer needs it. Per-tenant sequences may diverge, complicating the reverse path; cost outweighs current benefit.

### Constraints from prior analysis

- PG partitioning requires the partition key to be part of the PK. `mt_events.pk` shape changes from `(stream_id, seq_id)` to `(stream_id, seq_id, tenant_id)`.
- Each tenant gets its own sequence: `mt_events_sequence_{suffix}`.
- A 20M-event single-transaction migration locks the table for hours. The tool uses the `DETACH/ATTACH PARTITION` pattern so per-tenant work happens in bounded windows.
- `mt_event_progression` row identities must be constructed via `ShardName.Compose` (per [jasperfx/jasperfx#419](https://github.com/JasperFx/jasperfx/issues/419) design principle — never hand-roll).

### Data policy

**Do not renumber historical events.** Renumbering invalidates every downstream consumer that captured a sequence position (progression rows, downstream warehouses, audit logs, external integrations). The blast radius vastly outweighs the cosmetic value of contiguous per-tenant sequences.

**Per-tenant sequence start value:** `MAX(seq_id) FROM mt_events WHERE tenant_id = $1` + 1, computed per tenant during migration. Avoids the giant sequence hole that would otherwise appear at each tenant's first post-migration event. The `(stream, seq, tenant)` PK makes seq-id reuse safe across tenants (different rows).

### Migration shape

Three phases. The whole tool is **offline-first in v1** — operators take a downtime window for the migration. Online migration (writes continue during migration via per-tenant brief-lock windows) is a follow-up if needed.

#### Phase 1 — Prepare

- Validate prerequisites:
  - `TenancyStyle = Conjoined` (the source state required for migration)
  - `Events.UseTenantPartitionedEvents = true` in target config
  - Every event row has a non-null `tenant_id`
[...]

PR #4799, opened 2026-06-30
Per-tenant agent distribution, database-affine assignment + order-preserving streaming bulk import (wolverine#3280, #4806)

erdtsieck:
Marten piece of the per-tenant agent distribution work for sharded + tenant-partitioned stores. Addresses JasperFx/wolverine#3280 and #4806; **depends on JasperFx/jasperfx#482**.

The branch has grown beyond the original surface-only scope — it now carries everything Marten needs to run (and migrate an existing event log onto) a sharded, per-tenant-partitioned store under managed distribution:

### Distribution surface
- `IEventStore.DistributesAgentsPerTenant` override: true only for `MultiTenantedWithShardedDatabases` + `UseTenantPartitionedEvents` — the one tenancy where tenants are co-located in a database with independent sequences. Everything else keeps one agent per database (unchanged).
- **Opt-in database-affine assignment** (#4806): new `StoreOptions.Events.UseDatabaseAffineAgentAssignment` (default `false`), surfaced as `IEventStore.GroupAgentAssignmentsByDatabase` and gated on sharded tenancy + per-tenant partitioning. Lets a distribution-aware host keep each shard database's agents on one node, so connection pools scale with the number of databases rather than nodes × databases.

### Per-tenant high-water correctness
- Per-tenant high water is read from `max(seq_id)` instead of the unused per-tenant sequence.
- `BulkInsertEventsAsync` advances each affected tenant's own `HighWaterMark:<tenant>` row on partitioned stores (the store-global row is never read there, and would hold a misleading cross-tenant max); non-partitioned stores are unchanged (regression-tested).

### Order-preserving streaming bulk import (migration support)
- New `IDocumentStore.BulkInsertEventStreamAsync(tenantId, streamHeaders, orderedEvents, …)`: streams an existing event log in bounded memory (no whole-tenant materialization / OOM) and assigns target `seq_id`s in arrival order, **preserving cross-stream ordering** — required by multi-stream projections/read models that compare `Sequence` across streams. `mt_streams` is seeded from the supplied headers first (FK), one transaction per tenant.
- The batch `BulkInsertAsync` path now assigns seq_ids globally in the events' existing `Sequence` order (stable sort, so fresh seeding with all-zero sequences is byte-for-byte unchanged); previously each stream got a contiguous seq_id block, silently flattening the original interleaving.
- COPY-state fix: sequence blocks are only fetched while no `mt_events` COPY is open on the connection (Npgsql rejects commands mid-COPY).

### Tests & docs
Config gating, high-water regressions (partitioned and non-partitioned), streaming + batch ordering incl. multi-block sequence refills; documentation for sharded multi-tenancy, per-tenant event partitioning, projection distribution, and bulk append.

**Merge order:** jasperfx#482 → this → Wolverine #3281. Builds against a locally-packed JasperFx.Events carrying #482.

### DIFF
--- a/docs/configuration/multitenancy.md
+++ b/docs/configuration/multitenancy.md
@@ -467,14 +467,9 @@ opts.Events.UseTenantPartitionedEvents = true;
 
 // Group a shard database's per-tenant agents onto a single node (opt-in, default off).
 opts.Events.UseDatabaseAffineAgentAssignment = true;
-
-// Bounded fan-out ("the mix"): let one shard database's agents spread across up to N nodes for
-// more parallelism (default 1 = strict affinity). Choose N so databases × N × pool stays under
-// the server's max_connections. Surfaced as IEventStore.MaxNodesPerDatabaseForAgents (clamped to >= 1).
-opts.Events.DatabaseAffineAgentFanout = 2;
 ```
 
-These options only take effect for a sharded, per-tenant-partitioned store driven by a distribution-aware host; on any
+This option only takes effect for a sharded, per-tenant-partitioned store driven by a distribution-aware host; on any
 other store `IEventStore.GroupAgentAssignmentsByDatabase` stays `false` and distribution is unchanged.
 
 ## Dynamically applying changes to tenants databases
--- a/docs/configuration/multitenancy.md
+++ b/docs/configuration/multitenancy.md
@@ -449,6 +449,34 @@ runs asynchronous projections across all of them. When new databases or tenants 
 the daemon's periodic health check picks them up and starts projection processing without any
 downtime or reconfiguration.
 
+#### Node-Distributed Daemons and Connection Fan-out
+
+When the sharded model is combined with [per-tenant event partitioning](/events/multitenancy#per-tenant-event-partitioning)
+and a node-distributed async daemon (e.g. [Wolverine-managed event subscription distribution](https://wolverinefx.net/guide/durability/marten/distribution.html)),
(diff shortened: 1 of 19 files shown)


jeremydmiller:
Thank you @erdtsieck — this work is moving forward, split by concern since the PR bundled several goals with different dependencies and readiness (per review under JasperFx/jasperfx#486):

1. **Per-tenant high water from `max(seq_id)`** → **#4847** (already open): your commit cherry-picked with authorship intact. Independent of the JasperFx contract change, so it can land immediately.
2. **`DistributesAgentsPerTenant` surface + the per-tenant agent-start regression test** → follow-up PR once JasperFx/jasperfx#487 (the revised #482) merges and publishes. Note `UseDatabaseAffineAgentAssignment` will not return: Wolverine now derives database-affine grouping from `IEventStore.DatabaseCardinality`, so the Marten option, the `EventGraph` property, and the gating surface disappear entirely.
3. **`BulkInsertEventStreamAsync`** → parked for rework before it returns: the streaming import allocates seq_ids from the store-global `mt_events_sequence` but never advances the per-tenant `mt_events_sequence_{suffix}`, so a migrated tenant's first live append draws seq_id 1 (PK collision, or an event silently below the tenant's high water); the streaming path also needs to force-stamp tenant ids like the batch overload does, plus partitioned-store test coverage. It also needs reconciling with the #4682 in-place migration path so there's one canonical migration story.

Authorship carries into every successor (cherry-picks where clean, `Co-authored-by` where reworked). Epic: JasperFx/jasperfx#486.

PR #4802, opened 2026-07-01
Scope AssertStreamVersion to tenant under conjoined tenancy

### LINKED ISSUE #4803: AssertStreamVersion scans mt_streams instead of point-seeking the PK under conjoined tenancy
## Summary

Under `TenancyStyle.Conjoined`, the optimistic-concurrency "assert expected version" path queries `mt_streams` by stream id **without a `tenant_id` predicate**. This has two consequences:

- **Performance:** without the leading PK column constrained, the query cannot point-seek the `(tenant_id, id)` primary key, so it falls back to a scan of the whole shared `mt_streams` table — O(total streams across all tenants), which grows unbounded as the system scales.
- **Correctness:** because the same stream id/key can exist for multiple tenants, the concurrency check can be evaluated against a different tenant's stream row.

## Cause

`AssertStreamVersionById` and `AssertStreamVersionByKey` build:

```sql
select version from {schema}.mt_streams where id = <param>
```

with no tenant filter (`AssertStreamVersionById.cs:25-31`, `AssertStreamVersionByKey.cs:25-31`). Under conjoined tenancy `mt_streams` has a composite PK `(tenant_id, id)` with `tenant_id` first (`StreamsTable.cs:29-38`), and there is no standalone index on `id`. A predicate on `id` alone leaves the leading PK column unbounded, so no point seek is possible:

- Supported PostgreSQL 13–17: sequential scan or full index scan — O(number of streams).
- PostgreSQL 18+: btree skip scan does one sub-seek per distinct `tenant_id` — still degrades with tenant count.

Because the query is not tenant-scoped, it can also match another tenant's row with the same id. `PostprocessAsync` reads only the first row returned (`AssertStreamVersionById.cs:37-45`), so the version assertion is evaluated against whichever tenant's row Postgres returns first (ordering undefined) — a spurious `ConcurrencyException`, or a missed conflict if the foreign row's version happens to match.

Every sibling path that touches `mt_streams` by id already scopes by tenant under conjoined tenancy — e.g. `UpdateStreamVersion` (`PostgresEventStoreDialect.cs:265-269`) and the `AppendOptimistic`/`AppendExclusive` pre-reads (`EventStore.ConcurrentAppends.cs:24-28,45-49`) — and therefore point-seek the PK cleanly. These two Assert operations are the outlier. `stream.TenantId` is already populated before the op is queued (`RichEventAppender.cs:80`), so the value is on hand.

## When it runs

Both appenders queue these operations for a stream that enforces consistency, has an expected version, but appends no new events (`RichEventAppender.cs:78-89`, `QuickEventAppender.cs:30-41`):

```csharp
Where(x => !x.Events.Any() && x.AlwaysEnforceConsistency && x.ExpectedVersionOnServer.HasValue)
```

Reachable through the public API: `FetchForWriting<T>` sets `ExpectedVersionOnServer`, the caller sets `stream.AlwaysEnforceConsistency = true`, and `SaveChangesAsync` runs even with no appended events (`UnitOfWork.cs:226` keeps such streams as outstanding work). It's a shipped, tested scenario (`src/EventSourcingTests/FetchForWriting/always_enforce_consistency.cs`). When events *are* present, the tenant-scoped `UpdateStreamVersion` runs instead, so these two Assert ops are the only path with this problem — the "assert version, append nothing" no-op save.

This is a documented, recommended feature — see the "Enforcing Consistency Without Appending Events" section of the [command handler workflow docs](https://martendb.io/scenarios/command_handler_workflow), which describes the two-branch behavior (`UpdateStreamVersion` when events are appended, `AssertStreamVersion` when not) and steers command handlers, read-then-validate, and saga/process-manager patterns toward it. Those are frequently-executed paths, so the scan is not on the main append path but is far from theoretical.

The scan is latent: invisible on small tables and degrading as `mt_streams` grows, so an application that uses `AlwaysEnforceConsistency` guards frequently can see this become a significant cost at scale.

## Scope

- Only under `TenancyStyle.Conjoined` (shared `mt_streams` table).
- The cross-tenant correctness effect additionally requires the same stream id/key to exist across tenants — realistic for the `ByKey` (string-identity) variant and for deterministic per-tenant Guids, effectively impossible for random Guids. The scan cost applies regardless.

## Suggested fix

Append `and tenant_id = <tenant>` in both `ConfigureCommand` methods when `TenancyStyle.Conjoined` (mirroring `PostgresEventStoreDialect.cs:265-269` / `EventStore.ConcurrentAppends.cs:24-28`). This restores the PK point seek and scopes the check to the correct tenant in one change. Optionally guard `PostprocessAsync` against multi-row results.

svenclaesson:
Fixes #4803

## Problem

Under `TenancyStyle.Conjoined`, the optimistic-concurrency "assert expected version" path queries `mt_streams` by stream id **without a `tenant_id` predicate**.

`AssertStreamVersionById` and `AssertStreamVersionByKey` build:

```sql
select version from {schema}.mt_streams where id = <param>
```

Under conjoined tenancy `mt_streams` has a composite PK `(tenant_id, id)` with `tenant_id` first, and the same stream id/key can exist for multiple tenants. Consequences:

- **Correctness:** `PostprocessAsync` reads only the first row returned (undefined ordering), so the version assertion can be evaluated against a *different tenant's* same-keyed stream — a spurious `ConcurrencyException`, or a missed conflict if the foreign row's version happens to match `ExpectedVersionOnServer`.
- **Performance:** with the leading PK column unconstrained the query cannot point-seek the PK; it degrades to a scan of the whole shared `mt_streams` table (O(total streams)), and defeats per-tenant partition pruning when `UseTenantPartitionedEvents` is enabled.

## When it runs

Only on the `AlwaysEnforceConsistency` + zero-events-appended path (`FetchForWriting<T>` → set `AlwaysEnforceConsistency = true` → `SaveChangesAsync` with no events). When events are appended, the correctly-scoped `UpdateStreamVersion` runs instead. Every sibling stream-by-id query (`UpdateStreamVersion`, the `AppendOptimistic`/`AppendExclusive` pre-reads) already adds `and tenant_id` under conjoined tenancy — these two operations were the outlier.

## Fix

Append `and tenant_id = <stream tenant>` in both `ConfigureCommand` methods when `TenancyStyle.Conjoined`, mirroring the sibling paths. `Stream.TenantId` is already populated before the op is queued, so this restores the PK point seek and scopes the check to the correct tenant in one change.

## Tests

Added `always_enforce_consistency_conjoined_tenancy` with two cases under conjoined + string identity, where another tenant holds a same-keyed stream that sorts/inserts first (so an unscoped query reads its row first):

- **happy path:** our tenant's version is unchanged → the check must succeed (previously threw a spurious `ConcurrencyException` from the other tenant's higher version).
- **sad path:** our tenant's stream advances out of band → the check must throw (previously missed the conflict by reading the other tenant's still-matching version).

Both fail on `master` and pass with the fix. Existing `always_enforce_consistency`, multi-tenancy, and FetchForWriting suites remain green (242 tests).

### DIFF
--- a/src/EventSourcingTests/FetchForWriting/always_enforce_consistency_conjoined_tenancy.cs
+++ b/src/EventSourcingTests/FetchForWriting/always_enforce_consistency_conjoined_tenancy.cs
@@ -1,0 +1,89 @@
+using System.Threading.Tasks;
+using EventSourcingTests.Aggregation;
+using JasperFx;
+using JasperFx.Events;
+using Marten;
+using Marten.Events;
+using Marten.Storage;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests.FetchForWriting;
+
+public class always_enforce_consistency_conjoined_tenancy: OneOffConfigurationsContext
+{
+    // Under conjoined tenancy the same stream key can exist for multiple tenants.
+    // The AlwaysEnforceConsistency "assert version, append nothing" path must scope
+    // its version check to the fetching tenant, not read another tenant's same-keyed
+    // stream row. The "wrong" tenant here sorts first by (tenant_id, id) and is
+    // inserted first, so an unscoped query would read its version first.
+
+    [Fact]
+    public async Task enforce_consistency_no_events_scopes_to_tenant_happy_path()
+    {
+        StoreOptions(opts =>
+        {
+            opts.Events.TenancyStyle = TenancyStyle.Conjoined;
+            opts.Events.StreamIdentity = StreamIdentity.AsString;
+        }, true);
+
+        const string streamKey = "shared-key";
+
+        // Other tenant's same-keyed stream sits at version 5.
+        theSession.ForTenant("aaa-other").Events.StartStream<SimpleAggregateAsString>(streamKey,
+            new AEvent(), new BEvent(), new CEvent(), new DEvent(), new EEvent());
+
+        // Our tenant's stream sits at version 3.
+        theSession.ForTenant("zzz-target").Events.StartStream<SimpleAggregateAsString>(streamKey,
+            new AEvent(), new BEvent(), new CEvent());
+
+        await theSession.SaveChangesAsync();
+
+        var stream = await theSession.ForTenant("zzz-target").Events
+            .FetchForWriting<SimpleAggregateAsString>(streamKey);
+        stream.AlwaysEnforceConsistency = true;
+
+        // No events appended. Our tenant's version (3) has not changed, so the
+        // consistency check must succeed. If the check reads the other tenant's
+        // version (5), it throws a spurious ConcurrencyException.
+        await Should.NotThrowAsync(async () => await theSession.SaveChangesAsync());
+    }
+
+    [Fact]
+    public async Task enforce_consistency_no_events_scopes_to_tenant_sad_path()
+    {
+        StoreOptions(opts =>
+        {
+            opts.Events.TenancyStyle = TenancyStyle.Conjoined;
+            opts.Events.StreamIdentity = StreamIdentity.AsString;
+        }, true);
+
+        const string streamKey = "shared-key";
+
+        // Other tenant's same-keyed stream sits at version 3 — coincidentally the
+        // same version our tenant will be fetched at.
+        theSession.ForTenant("aaa-other").Events.StartStream<SimpleAggregateAsString>(streamKey,
+            new AEvent(), new BEvent(), new CEvent());
+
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Thanks @svenclaesson — closing this in favor of #4808, which fixes the same #4803 by lifting the tenancy branch out of `ConfigureCommand` and into the closed-shape event storage (separate single-tenant vs conjoined operations, selected once from the descriptor). Your two `always_enforce_consistency_conjoined_tenancy` test cases were carried over intact, and you're credited as a co-author on the commit.

PR #4804, opened 2026-07-02
test(events): failing repro for seq_id not following append order

Blackclaws:
Within a single SaveChangesAsync, global event seq_ids are assigned by iterating session.WorkTracker.Streams in first-touch order and dealing consecutive sequences per stream, rather than following the order events were appended in code.

Repro: append B1, then A1, then B2 -> seq_ids B1=1, A1=3, B2=2, so A1 (appended before B2) gets a HIGHER seq_id than B2. Reproduces under both the default Rich mode and QuickWithServerTimestamps.

### DIFF
--- a/src/EventSourcingTests/seq_id_reflects_append_order_within_unit_of_work.cs
+++ b/src/EventSourcingTests/seq_id_reflects_append_order_within_unit_of_work.cs
@@ -1,0 +1,75 @@
+using System.Linq;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using Marten.Events;
+using Marten.Testing.Harness;
+using Shouldly;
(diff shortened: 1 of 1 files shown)


mysticmind:
As such it is not a bug. "A1 appended before B2 → A1 should have a lower seq_id" is an assumption Marten never promised and, IMO, shouldn't. Within a stream, order is preserved (B1 < B2), which is the guarantee that actually matters.

jeremydmiller:
@Blackclaws This isn't practical, and like @mysticmind said, we've **never** stated that was a guarantee. You'd have to use advisory locks to make appends single file to pull that off I think, and we're sure as hell not doing that.

PR #4805, opened 2026-07-03
perf(rls): wrap current_setting() in the generated tenant policy for per-statement evaluation

dmitrymaranik:
Marten's conjoined-tenancy RLS generator (`RlsPolicySchemaObject`) emits `current_setting('<setting>')` **unwrapped** in the policy's `USING` / `WITH CHECK`. Postgres re-evaluates a bare `current_setting(...)` **once per row** the policy scans; wrapping it in a scalar subquery — `(select current_setting('<setting>'))` — makes it an InitPlan the planner evaluates **once per statement** and caches. It's predicate-equivalent (tenant isolation is unchanged) and the documented Postgres RLS performance pattern; the benefit grows with table size.

- Wraps the generated `USING` / `WITH CHECK` clauses in `RlsPolicySchemaObject`.
- Updates the two exact-string assertions in `row_level_security_unit_tests`. The integration tests use `ShouldContain("current_setting('<setting>'")` substring checks and the existing-policy detection pattern (`%current_setting('<setting>'%`) still match, so both are unaffected.

Spotted with [pgrls](https://github.com/pgrls/pgrls) (an open-source Postgres RLS analyzer). Happy to adjust.

### DIFF
--- a/src/CoreTests/row_level_security_unit_tests.cs
+++ b/src/CoreTests/row_level_security_unit_tests.cs
@@ -142,8 +142,8 @@ public class row_level_security_unit_tests
         createSql.ShouldContain("ALTER TABLE public.mt_doc_target FORCE ROW LEVEL SECURITY;");
         createSql.ShouldContain("DROP POLICY IF EXISTS marten_tenant_isolation ON public.mt_doc_target;");
         createSql.ShouldContain("CREATE POLICY marten_tenant_isolation ON public.mt_doc_target");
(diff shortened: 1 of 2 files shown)


dmitrymaranik:
Thanks for merging, @jeremydmiller! Every Marten conjoined-tenancy RLS policy now evaluates `current_setting()` once per statement (an InitPlan) instead of once per scanned row.

This came out of running [pgrls](https://github.com/pgrls/pgrls) (an open-source Postgres RLS analyzer) over the generated policies — if it was handy, a star would help others find it. Either way, glad it landed!

PR #4857, opened 2026-07-06
BulkInsertEventStreamAsync: streaming, order-preserving bulk event import (parked #4799 item)

erdtsieck:
Reopens the `BulkInsertEventStreamAsync` work parked when #4799 was closed, reworked per the three requirements in JasperFx/jasperfx#486 WS0 and field-hardened on a production-copy migration (500 tenants, sharded + tenant-partitioned).

**Stacked on #4854** (per-tenant sequence allocation for the batch path) — this branch is based on it and rebases trivially once it merges.

## What it adds

A streaming counterpart to `BulkInsertEventsAsync` for migrating an existing event store: consumes `IAsyncEnumerable<IEvent>` lazily in `batchSize` blocks — a tenant with millions of events imports in bounded memory — with small per-stream `BulkEventStreamHeader`s seeding `mt_streams` up front (the FK target), one transaction per tenant for trivial resume logic.

## The three WS0 requirements

- ✅ **Per-tenant sequence advancement** — seq_ids are drawn from the tenant's own `mt_events_sequence_{suffix}` via the resolver from #4854, so the sequence ends at the tenant's height and the first live append after a migration continues seamlessly (no `setval` needed — allocation and position are one mechanism).
- ✅ **Tenant-id force-stamping** — every streamed event is stamped with the import tenant, mirroring the batch overload; stale `TenantId`s from the source store cannot leak into the target partition.
- ✅ **Partitioned-store test coverage** — `streaming_bulk_import_on_a_partitioned_store` proves cross-stream order preservation, the sequence end-position, and a live append directly after the import, on a `UseTenantPartitionedEvents` store. Plus the ported `BulkEventStreamAppendTests` (multi-block sequence refill regression: a command cannot run mid-COPY, so the writer closes before every refill).

## Design notes

- **Cross-stream ordering**: each event gets the next ascending seq_id in arrival order, so supplying the source log's global order reproduces the interleaving multi-stream projections compare on. The batch path now assigns seq_ids the same way (stable sort on the carried `Sequence` — fresh seeding where all are 0 is byte-identical to previous behavior).
- **High water**: under per-tenant partitioning the import writes NO progression row — per-tenant detection derives from `max(seq_id)` since #4847, and advancing the store-global row with a tenant-local seq would misrepresent it. Non-partitioned stores keep the store-global upsert like the batch path.
- **No schema work, by contract**: the caller applies schema at startup and registers the tenant first (`AddMartenManagedTenantsAsync` / `AddTenantToShardAsync`). Field data behind this: schema application from inside the import loop costs O(registered tenants) DDL per fresh database under the shared tenant-partition registry, which ground the 500-tenant migration to a near-halt as the count grew.
- **#4682 reconciliation**: this is the bulk-seed *primitive*; the conjoined→partitioned migration tool can consume it so there is one canonical migration story. Docs section added to `docs/events/bulk-appending.md` covering the contract and use-case.

Authorship note: continues the closed #4799 with the rework requested there; part of the JasperFx/jasperfx#486 hardening cycle.

### DIFF
--- a/docs/events/bulk-appending.md
+++ b/docs/events/bulk-appending.md
@@ -190,6 +190,46 @@ The bulk append API supports all combinations of:
 | Archived stream partitioning | On, Off |
 | Metadata columns | Correlation ID, Causation ID, Headers, User Name (any combination) |
 
+## Streaming Import of an Existing Event Log
+
+`BulkInsertEventsAsync` materializes every event of the import in memory, which is fine for seeding but
+not for migrating a large existing event store. For that scenario use the streaming overload,
+`BulkInsertEventStreamAsync`, which consumes an `IAsyncEnumerable<IEvent>` lazily in `batchSize` blocks —
+a tenant with millions of events imports in bounded memory:
+
+```cs
+// One small header per stream seeds mt_streams up front (the foreign-key target),
+// while the events themselves stream through in COPY blocks.
+var headers = new[]
+{
+    new BulkEventStreamHeader { Id = streamId, Version = 3, AggregateType = typeof(Order) }
+};
+
+await store.BulkInsertEventStreamAsync(
+    tenantId,
+    headers,
+    ReadSourceEventsInSequenceOrderAsync(), // IAsyncEnumerable<IEvent>, in source seq_id order
+    batchSize: 1000);
+```
+
+Three properties make this suitable as the primitive for event-store migrations:
+
+- **Cross-stream ordering is preserved.** Each event receives the next ascending `seq_id` in arrival
+  order, so supplying the source log's global order (read by the source's `seq_id`) reproduces the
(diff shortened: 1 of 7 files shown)


jeremydmiller:
Rebased this branch onto current master (maintainer-edit push; authorship unchanged):

- The first commit ("Bulk event import draws seq_ids from the tenant's own sequence") was dropped by the rebase because it already landed on master via #4854 — this PR is now just the `BulkInsertEventStreamAsync` streaming import commit.
- Re-validated locally on top of the Weasel.Storage extraction (#4861) and the per-tenant sequence changes (#4852/#4853/#4854): Debug + Release build clean; all 4 `BulkEventStreamAppendTests` + the 2 tenant-partitioned regression tests green (net10); full EventSourcingTests suite green on the earlier rebase.

🤖 Rebased with [Claude Code](https://claude.com/claude-code)

Issue #4880, opened 2026-07-07
Tenant REMOVE lifecycle e2e: retire per-tenant agents + progression cleanup mid-run (epic jasperfx#486 WS1 residue; blocked on #4868)


jeremydmiller:
Carved out of the closed epic [jasperfx#486](https://github.com/JasperFx/jasperfx/issues/486) (WS1 'verify dynamic tenant add/remove mid-run').

**ADD is fully covered** — marten dynamic-lifecycle tests + wolverine#3334's churn e2e prove a new tenant's shards/agents appear on some node mid-run with progression rows. **REMOVE is not**: the sharded-tenancy usage descriptor never shrinks (#4868), so `RemoveTenantAsync`/`DisableTenantAsync` can't retire the tenant's daemon agents until restart.

Deliverables once #4868 lands:
- Native HotCold e2e: removing a tenant mid-run retires its per-tenant agents on the owning node (coordinator reconciliation/reaping path from jasperfx#491)
- Wolverine-managed e2e: flip/extend the wolverine churn test to cover removal (agents stop, no orphaned progression writes)
- Pin the progression-row semantics on removal (rows retained for re-add? cleaned? document the contract)

Blocked by #4868.

jeremydmiller:
Status now that #4868 (the blocker) is closed — most of this is already done:

**Done (native + registry, merged in #4893):**
- `sharded_tenancy_remove_lifecycle_tests` — 12 tests: usage descriptor shrinks on remove/disable **without restart** (local + cross-store-instance), reassignment follows the tenant, remove drops partitions + per-tenant sequence + per-tenant progression rows and recomputes tenant count, `remove_is_idempotent_and_a_removed_tenant_can_be_readded_fresh`, disable/enable soft-delete, hyphenated-id sanitization.
- `dynamic_tenant_lifecycle_on_shard_during_daemon` Phase 2 — a **running HotCold coordinator** reaps the removed tenant's per-tenant agent via jasperfx#491 reconciliation; progression + high-water rows cleaned; survivor keeps processing.

**Done now (PR #4908):**
- **Daemon-level re-add** (Phase 3 of the dynamic-lifecycle test): re-adding the removed tenant while the coordinator keeps running gives it a **fresh** per-tenant sequence + a **new** agent that catches up; the pre-removal projection doc is gone; only the newly-appended events project. This pins the "re-add starts fresh" half of the contract at the daemon level (registry level was already covered).
[...]

Issue #4763, opened 2026-06-18
ShardedTenancy: tenant-assignment count drift + non-atomic provisioning (orphaned assignment / broken tenant on partition-create failure)


erdtsieck:
### Summary

Three correctness issues in `ShardedTenancy`'s tenant-assignment bookkeeping + provisioning (verified against 9.9.0 source, `src/Marten/Storage/ShardedTenancy.cs`). They make `UseSmallestDatabaseAssignment` mis-rank shards over time and can leave a tenant permanently broken after a partition-create failure.

### 1. `tenant_count` is blindly incremented on the auto-assign path even when the assignment already existed
`findOrAssignTenantDatabaseAsync` upserts the assignment row idempotently, then **unconditionally** `+1`s the pool count:
```csharp
insert into {schema}.{TenantAssignmentTable} (tenant_id, database_id) values (:tid,:did)
  on conflict (tenant_id) do update set database_id = :did;          // idempotent
update {schema}.{DatabasePoolTable} set tenant_count = tenant_count + 1 where database_id = :did;  // blind +1
```
If the `on conflict` branch fires (tenant already assigned — e.g. two nodes race a first-touch, or a re-touch), `tenant_count` still increments without a new tenant ⇒ the count drifts above reality. `UseSmallestDatabaseAssignment` then mis-ranks shards (favours DBs that only *look* fuller). Contrast the explicit `AssignTenantAsync` which **recomputes** the count via `select count(*)` — the auto-assign path should do the same.

### 2. Re-assigning a tenant to another shard never decrements the source shard's count
`AssignTenantAsync(tenantId, databaseId)` recomputes the count for the **target** only:
```csharp
update {pool} set tenant_count = (select count(*) from {assignments} where database_id = :did and disabled = false) where database_id = :did;
```
On a move A→B, B is recomputed but **A is never updated**, so A's count stays inflated forever ⇒ permanent `UseSmallestDatabaseAssignment` skew. (Also: the old shard's per-tenant partitions/data are left orphaned on A — no validation/cleanup.)

### 3. Assignment is committed before partitions are created — non-atomic ⇒ orphaned assignment / broken tenant
In `findOrAssignTenantDatabaseAsync` the assignment row is committed (and cached) and **then** `createPartitionsForTenant` runs:
```csharp
// ... assignment INSERT committed above, cache updated ...
await createPartitionsForTenant(database, tenantId, CancellationToken.None);  // AddPartitionToAllTables + per-tenant sequence
```
[...]

Issue #4683, opened 2026-06-08
Drop-tenant cleanup gaps under UseTenantPartitionedEvents (orphaned sequences + progression rows)


jeremydmiller:
### Background

`AdvancedOperations.DeleteAllTenantDataAsync(tenantId)` (`src/Marten/Storage/AdvancedOperations.cs:618`) delegates to `TenantDataCleaner` (`src/Marten/Internal/TenantDataCleaner.cs`) and works correctly for the partition-drop path — it cheaply removes the tenant's partition from all conjoined tables in O(metadata), no row-level deletes (`TenantDataCleaner.cs:34-38`).

But under `Events.UseTenantPartitionedEvents = true`, two side-tables are **not** cleaned up by the partition drop:

1. **Per-tenant event sequences** (`mt_events_sequence_{suffix}`) — created by `PerTenantEventSequences.EnsureSequencesAsync` (`src/Marten/Events/Schema/PerTenantEventSequences.cs:150-166`) as freestanding sequences (no `OWNED BY` linkage). Dropping the partition leaves the sequence orphaned. Pin test exists at `src/TenantPartitionedEventsTests/Admin/delete_all_tenant_data_orphan_sequence_pin.cs` — no fix.
2. **Per-tenant progression rows** in `mt_event_progression` — every projection's per-tenant catch-up row (`{Projection}:All:{tenantId}`) and the per-tenant high-water row (`HighWaterMark:{tenantId}`) survive the partition drop. Silent leak.

Both leak slowly but unboundedly across drop-tenant cycles (e.g., trial-account churn). Neither is currently caught by integration tests at scale.

### Scope

Three fixes, one issue. All Marten-side.

#### Fix 1: Drop the per-tenant event sequence

In `TenantDataCleaner.DeleteAllAsync` (or whatever the dispatching method is — `TenantDataCleaner.cs:34-38`), after the partition drop and **only when `Events.UseTenantPartitionedEvents = true`**, also execute:

```sql
DROP SEQUENCE IF EXISTS {schema}.mt_events_sequence_{suffix};
```

The `{suffix}` is the same value the partition uses — look it up via `TenantPartitions` (same registry that drives `PerTenantEventSequences`). Use `IF EXISTS` so re-running the drop is idempotent.

[...]

PR #4893, opened 2026-07-07
Sharded tenancy usage descriptor shrinks on tenant removal (closes #4868, #4880)

### LINKED ISSUE #4868: Sharded tenancy: a running store's usage descriptor never shrinks — RemoveTenantAsync/DisableTenantAsync can't retire the tenant's daemon agents until restart
Found by the WS1 runtime-churn e2e for epic JasperFx/jasperfx#486. The tenant-ADD side converges end-to-end under Wolverine-managed distribution (new tenant → new per-tenant agent + progression rows, no restart). The REMOVE side cannot: `MartenDatabase.TenantIds` is add-only (`Fill` at ShardedTenancy.cs:211/:364) and `DescribeDatabasesAsync` (:735) copies that in-memory list, so after `ShardedTenancy.RemoveTenantAsync`/`DisableTenantAsync` a running store keeps reporting the tenant in `TryCreateUsage` — Wolverine's supported-set diff (and the native coordinator's re-enumeration from jasperfx#491) never see the tenant leave, and its agents run until process restart.

Fix direction: tenant removal should update the in-memory registry (or `DescribeDatabasesAsync`/`FetchManagedTenantIdsAsync`-style fresh reads should be the source for sharded tenancy too, mirroring marten#4864's fresh-read rationale for the single-DB path — one source, invalidation-free).

The Wolverine-side retirement itself is proven by unit test (`running_per_tenant_agent_is_stopped_when_its_tenant_is_removed`); this is purely the store-side signal. Part of JasperFx/jasperfx#486 (WS1).

jeremydmiller:
## What & why

Closes #4868, delivers the #4880 REMOVE-lifecycle e2e.

Found by the WS1 runtime-churn e2e for epic jasperfx#486. The tenant-**ADD** side already converges end-to-end under Wolverine-managed distribution (new tenant → new per-tenant agent + progression rows, no restart). The **REMOVE** side could not: `MartenDatabase.TenantIds` was add-only (`Fill` in `BuildDatabases` / `AssignTenantAsync`) and `DescribeDatabasesAsync` copied that in-memory list, so after `ShardedTenancy.RemoveTenantAsync` / `DisableTenantAsync` a running store kept reporting the tenant in `IEventStore.TryCreateUsage`. Wolverine's supported-set diff (and the native coordinator's re-enumeration from jasperfx#491) never saw the tenant leave, and its agents ran until process restart.

## The fix (`ShardedTenancy`)

- **`BuildDatabases` reconciles** the in-memory registry against the fresh active-assignment read — it now **prunes** `TenantIds` / `_tenantToDatabase`, not just grows them. Every `BuildDatabases` / `DescribeDatabasesAsync` becomes a fresh-read re-enumeration, so a removal made from **another node** converges too. This is the "fresh reads should be the source for sharded tenancy" direction from #4868, mirroring #4864's single-database rationale (one source, invalidation-free).
- **`RemoveTenantAsync` is now the full symmetric inverse of `AddTenantAsync`**: drops the tenant's LIST partitions **and** its `mt_tenant_partitions` registry rows on the shard (so jasperfx#491's `PerTenantShardExpansion` over `ICrossTenantRebuildSource` stops producing the tenant's shard name and its agents get reaped), runs the #4683 per-tenant event-sequence + progression/high-water cleanup, deletes the master assignment row, recomputes `tenant_count` (#4763-style), and shrinks the in-memory registry immediately. Idempotent; cleans a soft-deleted tenant too.
- **`Disable` / `EnableTenantAsync`** now shrink / re-grow `TenantIds` + `tenant_count` so the descriptor tracks soft-delete state. Non-destructive — data, partitions, and progression rows are retained; `Enable` restores in place.
- **`DescribeDatabasesAsync`** uses `Fill` (was `AddRange`) so each tenant id appears exactly once.

### `AdvancedOperations`
- New **`RemoveTenantFromShardAsync`** — the documented inverse of `AddTenantToShardAsync`.
- **`RemoveMartenManagedTenantsAsync`** now routes the sharded case to the removal path (mirrors `AddMartenManagedTenantsAsync`'s existing sharded routing; suffix == tenant id under the sharded 1:1 shape).

## Progression-row contract on removal (#4880)

- **REMOVE is destructive** on the shard — partitions, per-tenant sequence, and per-tenant progression/high-water rows are dropped/cleaned (mirrors the single-DB `RemoveMartenManagedTenantsAsync` / #4683 semantics). A re-added tenant starts **fresh**.
- **DISABLE is the non-destructive soft-delete** — descriptor shrinks, all shard-side data + registry rows retained, `Enable` restores in place.

## Weasel note (deliberate local reimplementation)

The shard partition drop is a small **quoted** reimplementation rather than `Weasel.Postgresql`'s `ManagedListPartitions.DropPartitionFromAllTables`. That path interpolates the partition table name **unquoted** into the `DETACH` / `DROP` DDL and fails with `42601 syntax error` for any tenant id that isn't a bare PostgreSQL identifier (hyphens, uppercase, GUID-shaped) — and the sharded assignment path never restricted tenant ids the way `AddMartenManagedTenantsAsync`'s `AssertValidPostgresqlIdentifiers` does, so quoted partitions exist in the wild and the drop must quote too.

**JasperFx / Weasel ask:** teach `ManagedListPartitions.DropPartitionFromAllTables` (and `...ForValue`) to quote the partition identifier in the `DETACH PARTITION` / `DROP TABLE` DDL (it already quotes on the CREATE side). Once that ships, this local drop loop can fold back to the Weasel call.

## Tests

- **`sharded_tenancy_remove_lifecycle_tests`** (new, `MultiTenancyTests`, 10 tests): descriptor shrink after `RemoveTenantAsync` on the same running store; shrink-on-disable / regrow-on-enable; **cross-store-instance** removal (the multi-node shape); reassignment follows between shard descriptors; each tenant reported exactly once; destructive-removal contract (partitions + registry rows dropped, `tenant_count` recomputed); `tenant_count` shrink/regrow on disable/enable; idempotent removal + fresh re-add; removing a disabled tenant still cleans its shard; `RemoveMartenManagedTenantsAsync` routing.
- **`dynamic_tenant_lifecycle_on_shard_during_daemon` Phase 2** (extended e2e): while a **native HotCold coordinator** keeps running, `RemoveTenantFromShardAsync` mid-run shrinks the client store's descriptor immediately, a fresh describe on the running coordinator node's store also drops the tenant, the coordinator's jasperfx#491 re-expansion + `reapOrphanedAgentsAsync` **reaps the removed tenant's agent**, its progression + high-water rows are cleaned and stay gone, and the surviving tenant keeps processing new events on the same shard daemon.

Results (Postgres on 5432, one project / net9.0 at a time):
- `MultiTenancyTests` sharded surface: **33/33 passed** (10 new + 23 existing).
- `TenantPartitionedEventsTests` `Sharded` collection: **26/26 passed**. In a full-project run 3 sharded tests intermittently fail on the shared physical shard databases (`marten_shard_a/b/c`) under cross-collection load; all pass in isolation and the whole `Sharded` collection passes as a group — the known shared-resource flake for these suites, not a regression.

## Follow-up (NOT in this PR)

- **Wolverine-managed removal e2e** belongs in the `wolverine` repo — flip/extend wolverine#3334's churn test to cover removal (agents stop, no orphaned progression writes) against this descriptor-shrink signal. Tracked under jasperfx#486 WS1.
- **Weasel quoted-drop** fold-back described above.

Part of jasperfx#486 (WS1).

### DIFF
--- a/src/Marten/AdvancedOperations.cs
+++ b/src/Marten/AdvancedOperations.cs
@@ -596,8 +596,26 @@ public class AdvancedOperations
 
         if (_store.Tenancy is not DefaultTenancy)
         {
+            // #4880: route the sharded case to the full per-tenant removal path — the
+            // symmetric inverse of AddMartenManagedTenantsAsync's sharded routing. Under the
+            // sharded 1:1 tenant↔suffix shape, ShardedTenancy.RemoveTenantAsync drops the
+            // tenant's partitions + per-tenant event sequence + progression rows on its shard
+            // AND deletes the master-registry assignment row (see #4868 for the descriptor
+            // shrink that lets running hosts retire the tenant's daemon agents).
+            if (_store.Tenancy is ShardedTenancy shardedTenancy)
+            {
+                foreach (var suffix in suffixes)
+                {
+                    await shardedTenancy.RemoveTenantAsync(suffix, token).ConfigureAwait(false);
+                }
+
+                return;
+            }
+
             throw new InvalidOperationException(
-                "This option is not (yet) supported in combination with database per tenant multi-tenancy");
+                "RemoveMartenManagedTenantsAsync supports DefaultTenancy and ShardedTenancy. " +
+                "MasterTableTenancy tenants own their whole database — call " +
+                "IServiceProvider.RemoveTenantAsync(tenantId) instead.");
         }
 
         var database = (PostgresqlDatabase)_store.Tenancy.Default.Database;
@@ -653,6 +671,25 @@ public class AdvancedOperations
         // entry point and the store-agnostic JasperFx admin extension drive ONE code path
         // (auto-assign → createPartitionsForTenant → per-tenant event sequence).
         return sharded.AddTenantAsync(tenantId, ct);
+    }
+
+    /// <summary>
+    ///     Fully remove a tenant from the sharded pool — the symmetric inverse of
+    ///     <see cref="AddTenantToShardAsync(string, CancellationToken)"/>. Drops the tenant's list
+    ///     partitions (and, under <c>UseTenantPartitionedEvents</c>, its per-tenant event sequence
+    ///     and event-progression rows) from its shard database, deletes the master-registry
+    ///     assignment, and shrinks the store's usage descriptor so running hosts can retire the
+    ///     tenant's daemon agents without a restart (#4868/#4880). DESTRUCTIVE on the shard —
+    ///     use the <c>IDynamicTenantSource</c> disable lifecycle for a non-destructive soft-delete.
+    ///     Only available with sharded tenancy.
+    /// </summary>
+    public async Task RemoveTenantFromShardAsync(string tenantId, CancellationToken ct)
+    {
+        var sharded = _store.Options.Tenancy as ShardedTenancy
+            ?? throw new InvalidOperationException(
+                "RemoveTenantFromShardAsync is only available when using MultiTenantedWithShardedDatabases()");
+
+        await sharded.RemoveTenantAsync(tenantId, ct).ConfigureAwait(false);
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


jeremydmiller:
Filed the Weasel fold-back as JasperFx/weasel#338 — the canonical fix is to apply `ListPartition.SanitizeSuffix` in `ManagedListPartitions.DropPartitionFromAllTables` so the drop path resolves the same table name the create path sanitizes to. Once that lands, the local quoted-drop workaround here can be removed.

Issue #4874, opened 2026-07-07
Intermittent connection abort mid-operation, introduced in 9.11.0-alpha.1


steve-ziegler:
## Summary

Upgrading Marten **9.10.0 → 9.11.0** introduced an **intermittent** failure under a running async
projection daemon: a pooled connection is closed/disposed underneath an in-flight operation, aborting
the socket. It strikes **arbitrary DB operations** — a batched `SaveChangesAsync` in one run, a plain
`SELECT max(seq)` (`FetchHighestEventSequenceNumber`) in another — always the same signature, a
different (usually trivial) victim test each run, at low latency.

```
System.ObjectDisposedException : NpgsqlTransaction         (when the abort lands in SaveChanges' batch)
 ---> Npgsql.NpgsqlException : Exception while reading from stream
   ---> System.IO.IOException : Unable to read data from the transport connection:
        The I/O operation has been aborted because of either a thread exit or an application request.
     ---> System.Net.Sockets.SocketException : The I/O operation has been aborted ...
```

Two representative top-of-stacks (same root, different callers):

```
# run A — batched SaveChanges
at Marten.Internal.Sessions.AutoClosingLifetime.ExecuteBatchPagesAsync (…:239 ExecuteReaderAsync / :256 RollbackAsync)
at Marten.Internal.Sessions.DocumentSessionBase.SaveChangesAsync

# run B — a plain read
at Marten.Storage.MartenDatabase.FetchHighestEventSequenceNumber (ExecuteScalar, SELECT max(seq))
at Marten.Events.TestingExtensions.WaitForNonStaleProjectionDataAsync
```

## Environment

.NET 10, PostgreSQL 17, Npgsql **9.0.5** (identical across the good/bad pins — not a driver change).
`StreamIdentity.AsString`, `EventAppendMode = Quick`, several **Inline** single-stream snapshot
projections + several **Async** projections, async daemon in **Solo**. ~500-test xUnit suite,
collections run **serially** (`parallelizeTestCollections: false`), shared host per collection.

## Bisect (via the published 9.11 alpha packages)

| Marten | JasperFx (transitive) | Result |
|---|---|---|
| 9.10.0 | 2.13.2 | **clean** (3 full runs) |
| **9.11.0-alpha.1** | **2.14.2** | **flakes** (abort signature) |
| 9.11.0 | 2.16.0 | flakes |

**First bad version = `9.11.0-alpha.1`.** Its delta over the clean 9.10.0 is exactly two changes:

[...]

PR #4903, opened 2026-07-08
Don't dispose a caller-owned NpgsqlDataSource (#4874)

jeremydmiller:
Fixes one concrete, deterministic instance of #4874.

## Root cause
`MartenDatabase.Dispose()` disposed its `NpgsqlDataSource` **unconditionally**, contradicting the documented `Connection(NpgsqlDataSource)` / `UseNpgsqlDataSource` contract (*"you need to handle data source disposal"*). Disposing an `NpgsqlDataSource` aborts **every** physical connection rented from it — so when a caller shares one data source across stores (or a running daemon still holds connections), a Marten teardown pulls the socket out from under an in-flight operation. That is the `ObjectDisposedException` / "I/O operation aborted … application request" socket abort reported in #4874.

## Fix
Track ownership on `StoreOptions.OwnsPrimaryDataSource` — `false` when the data source is caller-supplied via `Connection(NpgsqlDataSource)` (which both `UseNpgsqlDataSource` overloads route through), `true` (default) when Marten builds it from a connection string and owns its lifetime — and guard the disposal.

## Test
`Bug_4874_shared_datasource_not_disposed`: two stores sharing one external `NpgsqlDataSource`; disposing one must not abort the other. **Failing-first** — throws `System.ObjectDisposedException` on `master`, green with the fix. Marten-owned (connection-string) data sources still dispose normally (default `OwnsPrimaryDataSource = true`); connection/session/store-options suites 31/32 green (1 pre-existing skip).

## Follow-up
The Weasel base `PostgresqlDatabase.DisposeAsync()` disposes the data source unconditionally too — the async path needs the same ownership guard. Filed as JasperFx/weasel (linked below).

Diagnostic branch available for the reporter to confirm their exact disposer in-suite (opt-in `MARTEN_TRACE_DATASOURCE_DISPOSAL=1` stack-trace capture) if their case is a Marten-owned data source disposed by a different path rather than a shared external one.

### DIFF
--- a/src/CoreTests/Bug_4874_shared_datasource_not_disposed.cs
+++ b/src/CoreTests/Bug_4874_shared_datasource_not_disposed.cs
@@ -1,0 +1,71 @@
+using System.Threading.Tasks;
+using JasperFx;
+using Marten;
+using Marten.Testing.Harness;
+using Npgsql;
+using Shouldly;
+using Xunit;
+
+namespace CoreTests;
+
+// #4874: Marten must not dispose a caller-owned NpgsqlDataSource. Two stores sharing one external
+// data source is the deterministic shape of the reported "pooled connection disposed underneath an
+// in-flight operation" abort — before the fix, disposing store A tore down the shared data source and
+// aborted store B's next operation.
+public class Bug_4874_shared_datasource_not_disposed
+{
+    [Fact]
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Weasel async-path follow-up filed: JasperFx/weasel#345.

PR #4912, opened 2026-07-08
#4874: regression test for case-B coordinator-drain ordering storm (closes case B)

jeremydmiller:
## What

Adds `Bug_4874_coordinator_drain_ordering`, the regression test for the **case-B** disposal-ordering storm in #4874 — the `ObjectDisposedException: 'Npgsql.PoolingDataSource'` aborts the reporter measured, which #4907's async-tenancy foundation intentionally did not resolve.

Reporter @steve-ziegler [confirmed the shape](https://github.com/JasperFx/marten/issues/4874#issuecomment-4914607213): `await host.StopAsync()` is awaited before disposal, and the aborting `OpenAsync` is purely the native HotCold projection coordinator's leadership poll:

```
ProjectionCoordinatorBase.executeAsync
 └ SingleTenantProjectionDistributor.TryAttainLockAsync
    └ Weasel.Postgresql.AdvisoryLock.TryAttainLockAsync
       └ Medallion PostgresDistributedLock.TryAcquireAsync
          └ Npgsql OpenAsync  ⇒ ObjectDisposedException 'Npgsql.PoolingDataSource'
```

On a HotCold **cold** node, every leadership poll re-opens a connection from the owned `NpgsqlDataSource` to retry the advisory lock; tearing that node down disposes the pool out from under an in-flight `OpenAsync` the coordinator loop's `StopAsync` did not drain.

## The test

Boots a hot lock-holder, then repeatedly boots + tears down a cold node under an active poll (`LeadershipPollingTime=50`, connection-string / Marten-owned source, reporter's `await StopAsync(); Dispose();` shape) and asserts **zero** `PoolingDataSource` aborts via a first-chance counter. Failed on the pre-fix stack; **passes now** that master references the fixes.

## Fix (already on master)

- **Weasel 9.16.2** (weasel#349/#350): `AdvisoryLock` disposing-guard + swallow disposed-pool `ObjectDisposedException` in `TryAttainLockAsync`.
- **JasperFx 2.24.1** (jasperfx#499/#500): `ProjectionCoordinatorBase.executeAsync` returns on cancellation / treats `ObjectDisposedException` as terminal instead of re-polling.

Both landed on master via the dependency bump (commit c2b117257). This PR rebases on that and un-skips the test as the verification.

Closes the **case-B** half of #4874.

### DIFF
--- a/src/CoreTests/Bug_4874_coordinator_drain_ordering.cs
+++ b/src/CoreTests/Bug_4874_coordinator_drain_ordering.cs
@@ -1,0 +1,165 @@
+using System;
+using System.Collections.Concurrent;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using JasperFx.Events.Daemon;
+using JasperFx.Events.Projections;
+using Marten;
+using Marten.Events.Aggregation;
+using Marten.Testing.Harness;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Shouldly;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace CoreTests;
+
+// #4874 (case B — the disposal-ORDERING storm, distinct from the tenancy async-disposal foundation in
+// Bug_4874_async_tenancy_disposal). Reporter steve-ziegler confirmed
+// (https://github.com/JasperFx/marten/issues/4874#issuecomment-4914607213):
+//
+//   * `await host.StopAsync()` IS awaited before disposal, and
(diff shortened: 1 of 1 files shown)


jeremydmiller:
Upstream tracking issues for the gated fixes (un-skip this test once both land + are consumed):

- **P1 — Weasel `AdvisoryLock` shutdown guard:** https://github.com/JasperFx/weasel/issues/349
- **P2 — JasperFx `ProjectionCoordinatorBase` drain / terminal-ObjectDisposed:** https://github.com/JasperFx/jasperfx/issues/499

Release chain: Weasel → publish → JasperFx.Events → publish → Marten bump both + un-skip.

PR #4920, opened 2026-07-09
Support non-string IComparable.CompareTo in LINQ Where (e.g. Guid)

### LINKED ISSUE #4919: Guid.CompareTo not translated in LINQ Where (only string.CompareTo is)
`SimpleExpression.VisitMethodCall` only recognizes `CompareTo` when the declaring type is `string`, so `x.SomeGuid.CompareTo(g) < 0` throws `BadLinqExpressionException`. `Guid` has no `<` operator, so `CompareTo` is the only LINQ-expressible form — this blocks keyset/cursor pagination on a `Guid` tiebreaker (#3735).

`CompareToComparable` is already type-agnostic; widening the guard to any `int`-returning instance `CompareTo` translates `Guid` (and `DateTimeOffset`/`DateTime`/`int`) without breaking `string`. Repro + fix in PR.

zidad:
Fixes #4919. Widens the LINQ `CompareTo` guard beyond `string` so `Guid.CompareTo` (and other `int`-returning `IComparable.CompareTo`) translate.

### DIFF
--- a/src/LinqTests/Operators/guid_compare_operator.cs
+++ b/src/LinqTests/Operators/guid_compare_operator.cs
@@ -1,0 +1,41 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Marten.Testing.Documents;
+using Marten.Testing.Harness;
+using Shouldly;
+using Marten;
+
+namespace LinqTests.Operators;
+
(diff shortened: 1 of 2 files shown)


mysticmind (approved):
Looks good to me.

PR #4922, opened 2026-07-09
GetProjectionStatusesAsync resolves the named database instead of the default tenant

### LINKED ISSUE #502: GetProjectionStatusesAsync throws NotSupportedException on a database-per-tenant Marten store, including the per-tenant overload
Found while fixing [JasperFx/CritterWatch#678](https://github.com/JasperFx/CritterWatch/issues/678).

## Repro

A Marten store configured with `MultiTenantedDatabasesWithMasterDatabaseTable` (database-per-tenant), two registered tenant databases, one async projection.

```csharp
var store = (IEventStore)documentStore;

// Both of these throw:
await store.GetProjectionStatusesAsync(ct);
// NotSupportedException: Default tenant does not supported

foreach (var db in await store.AllDatabases())
{
    await store.GetProjectionStatusesAsync(db.Identifier, ct);
    // NotSupportedException: Default tenant does not supported
}
```

The store-wide overload throwing is understandable — there is no default tenant to resolve a session against. But the **per-tenant overload throws identically**, even when handed an identifier that `AllDatabases()` just returned and that `BuildProjectionDaemonAsync(identifier)` accepts happily. It looks like the tenant/database argument isn't reaching the session resolution.

## Why it matters

`IEventStore.GetProjectionStatusesAsync` is the only store-agnostic way to enumerate the *registered* projections and their lifecycles. Consumers use it to answer "is this progression row's projection still registered, or is it an orphan left behind by a rename/removal?"

On a database-per-tenant store that question currently has no answer through the abstraction. CritterWatch has had to suppress orphan detection entirely for such stores — without a registry, every live projection's progression row looks orphaned, and reporting that would be worse than reporting nothing.

The same gap costs the "registered but idle" baseline: a registered async projection with no running agent can't be surfaced at all on these stores.

## Expected

`GetProjectionStatusesAsync(tenantIdOrDatabaseIdentifier, ct)` should resolve against the named database and return its projection statuses, mirroring `BuildProjectionDaemonAsync(string)`, which already works on exactly these stores with exactly these identifiers.

A store-wide overload that fans out across `AllDatabases()` would also be reasonable, though per-database is the more useful primitive for monitoring.

jeremydmiller:
Fixes [jasperfx#502](https://github.com/JasperFx/jasperfx/issues/502).

## Problem

`IEventStore.GetProjectionStatusesAsync(tenantIdOrDatabaseIdentifier, ct)` always opened its explorer session against the **default tenant**, no matter what was passed. On a database-per-tenant store there is no default tenant, so it threw `DefaultTenantUsageDisabledException` — even when handed an identifier that `AllDatabases()` had just returned, and that `BuildProjectionDaemonAsync(identifier)` accepts happily.

The argument only ever reached the shard-name composition; it never reached session resolution.

```csharp
var store = (IEventStore)documentStore;   // MultiTenantedDatabasesWithMasterDatabaseTable

await store.GetProjectionStatusesAsync(ct);                        // throws
foreach (var db in await store.AllDatabases())
    await store.GetProjectionStatusesAsync(db.Identifier, ct);     // also throws
```

## Change

The argument now means what it means everywhere else in the API, resolved exactly the way `BuildProjectionDaemonAsync(string)` resolves it:

- **Single database** (conjoined tenancy with per-tenant event partitioning) — the argument names a tenant co-located in the one database. Shard identities keep the trailing `:tenantId` suffix and the progression lookup matches the tenant-bearing row. **Unchanged.**
- **Database-per-tenant / sharded tenancy** — the argument names a physical database. The session is opened against it, and its shards are **not** tenant-suffixed: each database runs its own daemon over the same shard identities, so suffixing would name a shard that exists nowhere and its progression lookup would miss.

The single/multi discrimination uses `Options.Tenancy.Cardinality`, matching the existing precedent in `DocumentStore.DocumentDiagnostics.cs`.

## Why it matters

This is the only store-agnostic way to enumerate the **registered** projections and their lifecycles. Consumers use it to answer *"is this progression row's projection still registered, or is it an orphan left behind by a rename or removal?"*

On a database-per-tenant store that question currently has no answer through the abstraction, so CritterWatch has had to suppress orphan detection entirely for those stores: without a registry, every live projection's progression row looks orphaned, and reporting that is worse than reporting nothing. The same gap costs the "registered but idle" baseline for a registered async projection with no running agent.

## Tests

New `MultiTenancyTests/projection_statuses_per_database`:

- Every database `AllDatabases()` reports can be asked for its statuses, and each returns **its own** progression and head sequence (tenant-a = 5, tenant-b = 2), read from its own database.
- A database-per-tenant store's shards are **not** tenant-suffixed (`WidgetTally:All`, not `WidgetTally:All:tenant-a`).

Both are red without the fix (`DefaultTenantUsageDisabledException`). The databases and schema are suffixed per target framework, because the test assembly runs once per TFM concurrently against the same Postgres.

Regression: `TenantPartitionedEventsTests/admin_api_tenant_overloads` 6/6 (the single-database per-tenant path), `EventSourcingTests` Explorer 21/21, full `MultiTenancyTests` 159/159.

### DIFF
--- a/src/Marten/DocumentStore.EventStoreExplorer.cs
+++ b/src/Marten/DocumentStore.EventStoreExplorer.cs
@@ -333,18 +333,39 @@ public partial class DocumentStore
     }
 
     /// <summary>
-    /// #4596 Phase 1 Session 4 — override the jasperfx#407 default-throwing
-    /// per-tenant overload. Non-null <paramref name="tenantId"/> returns the
-    /// per-tenant slice of each registered projection: the shard identities
-    /// reported carry the trailing tenant suffix
-    /// (<c>{ProjName}:{ShardKey}:{tenantId}</c>) and their ProcessedSequence
-    /// comes from the matching tenant-bearing row in <c>mt_event_progression</c>.
-    /// Null preserves the today's-behavior "every registered shard, no tenant
-    /// suffix" semantics.
+    /// #4596 Phase 1 Session 4 — override the jasperfx#407 default-throwing per-tenant overload.
+    ///
+    /// <para>
+    /// The argument means one of two things, depending on how the store is tenanted, exactly as it does for
+    /// <see cref="BuildProjectionDaemonAsync(string?, ILogger?)"/>:
+    /// </para>
+    /// <list type="bullet">
+    /// <item><b>Single database</b> (conjoined tenancy with per-tenant event partitioning): the argument names a
+    /// tenant co-located in the one database. Each registered projection reports its per-tenant slice — the shard
+    /// identities carry the trailing <c>{ProjName}:{ShardKey}:{tenantId}</c> suffix that Phase 1 Session 3 writes,
+    /// and their ProcessedSequence comes from the matching tenant-bearing progression row.</item>
+    /// <item><b>Database-per-tenant / sharded tenancy</b>: the argument names a physical database. Its shards are
+    /// NOT tenant-suffixed — each database runs its own daemon over the same shard identities — so the statuses
+    /// are read from that database with the plain shard names. This is the path that previously threw
+    /// <c>DefaultTenantUsageDisabledException</c>, because the session was always opened against the default
+    /// tenant no matter what was passed. See jasperfx#502.</item>
+    /// </list>
+    ///
+    /// <para>
+    /// A null argument preserves the "every registered shard, no tenant suffix, default database" semantics.
+    /// </para>
     /// </summary>
     async Task<IReadOnlyList<ProjectionStatus>> IEventStore.GetProjectionStatusesAsync(string? tenantId, CancellationToken ct)
     {
-        await using var session = openExplorerSession();
+        // On a store spanning several databases the argument selects the physical database whose progression
+        // rows we read. On a single-database store it selects a tenant living inside the one database, and only
+        // then does the shard identity carry the tenant suffix.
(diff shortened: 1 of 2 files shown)


jeremydmiller:
CI's `test-core` job failed on `CoreTests.Bug_4874_coordinator_drain_ordering.cold_node_teardown_does_not_abort_open_against_disposed_datasource` — a coordinator drain/timing test added two days ago in #4912.

This PR touches only `DocumentStore.EventStoreExplorer.cs` (which database an explorer session opens against), so it cannot reach coordinator teardown ordering. The test passes locally on this branch. Re-running the failed job.

jeremydmiller:
Confirmed pre-existing, not this PR.

`CoreTests.Bug_4874_coordinator_drain_ordering.cold_node_teardown_does_not_abort_open_against_disposed_datasource` fails the same way on the `Build & Test - NET9 - PG15 - Newtonsoft` job for two unrelated branches currently open:

- [`repro/linq-guid-keyset`, run 29013767734](https://github.com/JasperFx/marten/actions/runs/29013767734)
- [`aotgraph`, run 29007199040](https://github.com/JasperFx/marten/actions/runs/29007199040)

It passes locally on this branch. The test is a coordinator drain/teardown timing test added a few days ago in #4912; this PR touches only which database an explorer session opens against, so it has no path to coordinator teardown ordering. Every other CI job on this PR is green.

PR #4923, opened 2026-07-10
fix: drain the projection coordinator on disposal (#4915)

### LINKED ISSUE #4915: 9.14.0 regression: weasel#349's ODE swallow in AdvisoryLock.TryAttainLockAsync makes jasperfx#500's coordinator termination unreachable (follow-on to #4874 case B)
## Summary

Thanks for shipping the #4874 case-B chain in 9.14.0 — the two halves are individually sound, but they
don't compose. Weasel 9.16.2 (weasel#349) now catches the disposed-pool `ObjectDisposedException` inside
`AdvisoryLock.TryAttainLockAsync` and returns `false`. That means `ProjectionCoordinatorBase.executeAsync`
(jasperfx#499/#500) never sees the ODE its new terminate-on-ODE catch was added for — `false` reads as
"another node holds the lock", and the loop keeps re-polling the disposed data source on the
`LeadershipPollingTime` cadence indefinitely. In our 512-test integration suite (the same suite that
produced the #4874 case-B stacks), Marten 9.14.0 fails intermittently where 9.13.0 is green, and we can
reproduce the failure by pinning **only** Weasel.Postgresql 9.16.2 onto an otherwise-green Marten 9.13.0
stack.

## Bisection (full 512-test suite, Testcontainers Postgres, ~300 host lifecycles, Wolverine 6.17.0-alpha.1 held constant)

| Marten | JasperFx.Events | Weasel.Postgresql | Runs | Failed tests per run | first-chance disposed-pool ODEs per run |
|---|---|---|---|---|---|
| 9.10.0 (Wolverine 6.14.0) | 2.13.2 | 9.2.3 | 2 | 0, 0 | ~7.7k / ~8.3k |
| 9.13.0-alpha.4 | 2.24.0 | 9.16.1 | 3 | 0, 0, 0 | ~7k–12k |
| 9.13.0 | 2.24.0 | 9.16.1 | 1 | 0 | ~3.9k |
| 9.13.0 + **Weasel 9.16.2 pinned** | 2.24.0 | **9.16.2** | 2 | **1, 1** | **~154k / ~149k** |
| 9.13.0 + **JasperFx.Events 2.24.1 pinned** | **2.24.1** | 9.16.1 | 3 | 0, 0, (1)† | ~220 (all three) |
| 9.14.0 | 2.24.1 | 9.16.2 | 2 | **3, 7** | ~76k / ~143k |

- Marten `9.13.0..9.14.0` is only the dependency bump (`c2b117257`) plus the #4912 regression test, so the
  Marten code itself is exonerated; and the split rows show the failing half is the Weasel bump.
- Every config containing the weasel#349 swallow failed every run (4/4); every config without it was green
  (9/10 runs — † the one exception ran immediately after a killed/orphaned test-host incident on our side
  and did not reproduce in two clean follow-ups; we don't attribute it).
- Failures are intermittent with different victims each run, all in the connection/timing family:
  `IOException`/`SocketException` "I/O operation aborted" mid-read inside
  `MartenDatabase.FetchHighestEventSequenceNumber`, `NpgsqlReadBuffer.EnsureLong` aborts, one
  `42P01 relation does not exist` at host boot, SSE frame-read timeouts.
- Note the abort *count* is not the signal — jasperfx#500 alone drops it to ~220 (working as designed when
  the ODE actually reaches it), and green baselines run fine with ~8k. The signal is the test failures,
  which appear exactly in the swallow configs. We are **not** claiming the ~20× churn is causally what
  breaks the collateral tests — only that the swallow configs fail and the others don't.

## Why the two fixes don't compose (from the 9.16.2 / 2.24.1 sources)

`AdvisoryLock`'s `_disposed` short-circuit only helps once **its own** `DisposeAsync` has run. In the
#4874 case-B window — the data source is disposed while the AdvisoryLock is still live and the coordinator
is still polling — every poll takes the *catch* path instead:

```csharp
// Weasel.Postgresql.AdvisoryLock.TryAttainLockAsync (9.16.2)
catch (ObjectDisposedException)
{
    return false;   // coordinator reads this as "lock held elsewhere" → re-poll forever
}
```

while the code added to handle exactly this state one layer up can never fire:

```csharp
// JasperFx.Events.Daemon.ProjectionCoordinatorBase.executeAsync (2.24.1)
catch (ObjectDisposedException e)   // unreachable: Weasel swallows below
{
    ...
    return;   // terminate the loop — the data source is gone for good
}
```

Neither regression test covers the composition: the jasperfx#500 test uses a fake distributor that
*throws* the ODE directly at the coordinator (so the catch fires), and marten#4912 tears down the whole
cold host, which disposes the AdvisoryLock first — `_disposed` short-circuits the polls, so zero opens are
attempted. The uncovered state is *source disposed, AdvisoryLock live, coordinator polling*, which is what
a test suite with hundreds of host lifecycles hits constantly.

## Suggested direction (happy to test a patch build)

Since a disposed data source can never come back, "not attained during shutdown" seems like the wrong
translation — the lock can never be attained again through this instance. Options that would preserve
both fixes' intent:

1. In the ODE catch, set `_disposed = true` before returning `false` — self-latch so the *next* poll
   short-circuits without touching the pool (smallest change; the loop still lives but stops opening).
2. Or rethrow / return a distinguishable terminal signal so `ProjectionCoordinatorBase`'s existing
   terminate-on-ODE path can do its job.

We can run any candidate against the same 512-test suite and report back within a day.

/cc @jeremydmiller — follow-on from #4874 case B (reporter: steve-ziegler).

jeremydmiller:
Closes #4915 (with JasperFx/weasel#354).

## The gap

`IHostedService` only promises a `StopAsync` on graceful shutdown, and `IHost.Dispose()` does not stop hosted services. `ProjectionCoordinator` implemented **no disposal at all**, so `using var host = ...` — and any teardown that disposes without stopping — tears the container down underneath a running coordinator. The container disposes the `DocumentStore` and its owned `NpgsqlDataSource`, while the HotCold leadership loop keeps polling `AdvisoryLock.TryAttainLockAsync` on its cadence against a pool that is already gone.

Measured on a cold node contending with a lock-holding hot node, counting disposed-pool `ObjectDisposedException`s in the 1.5s after teardown:

| teardown | aborts |
|---|---|
| `await host.StopAsync()` then `host.Dispose()` | **0** |
| `host.Dispose()` alone | **210** (~140/s, sustained for the life of the process) |

That asymmetry is the bug, and it explains the reporter's observation in #4915 that "once a poller is in the disposed-source state it keeps churning for the rest of the process". `Bug_4874_coordinator_drain_ordering` could never reach it, because it stops gracefully first.

## The change

`ProjectionCoordinator` is now `IAsyncDisposable`/`IDisposable` and drains the loop and releases the advisory locks.

Ordering is safe by construction: the coordinator is built from an `IDocumentStore`, so the container necessarily creates the store *before* it and therefore disposes it *after* — reverse-order disposal runs the drain while the data source is still alive and the locks can still be released cleanly. The tests confirm this (a wrong order would surface as aborts from `ReleaseAllLocks`).

The drain is **unconditional**, not skipped after a graceful `StopAsync`. Neither `StartAsync` nor `ResumeAsync` is `virtual` on `ProjectionCoordinatorBase`, so there is no hook to distinguish "already stopped" from "stopped, then `ResumeAsync` started a fresh loop" — and skipping on a stale flag leaks exactly the loop this exists to drain. `StopAsync` is idempotent, so the second call is free. I wrote that optimization first, and `disposal_drains_a_loop_that_resumeasync_restarted_after_a_stop` catches it: 210 aborts with the flag, 0 without.

## Tests

`Bug_4915_coordinator_disposal_drain`, two tests, both red before / green after (399 and 189 aborts pre-fix).

They attribute aborts to the `TryAttainLockAsync` path rather than counting every disposed-pool `ObjectDisposedException` in the process. `CoreTests` runs serially, but other tests leak background work (daemon `StopAllAsync`, schema migrations, session reads) that touches their own disposed data sources later, and `AdvisoryLock.DisposeAsync` benignly touches one as well. A bare global count picks all of that up — which is precisely why `Bug_4874_coordinator_drain_ordering` passes in isolation but fails in a full-suite run today. Worth fixing there the same way in a follow-up.

## Validation

Full `CoreTests` on a clean database, net9.0: **475 passed, 1 failed**. The one failure, `describing_database_usage_from_DefaultTenancy.create_usage`, fails identically on stock `master` in this environment and is unrelated. `DaemonTests.Internals`: 55/55, twice.

## Relationship to weasel#354

Independent fixes for the same failure, from opposite sides:

| | Weasel 9.16.2 | Weasel + weasel#354 |
|---|---|---|
| Marten stock | 210 | 9 |
| Marten + this PR | 0 | 0 |

The 9 is one poll's worth of async-frame rethrows before jasperfx#500's terminate-on-`ObjectDisposedException` catch ends the loop. This PR keeps the loop from outliving its data source; weasel#354 keeps a loop that does from churning. Neither depends on the other, and this PR does not require a Weasel bump.

Reported, diagnosed, and bisected by @steve-ziegler in #4915.

### DIFF
--- a/src/CoreTests/Bug_4874_coordinator_drain_ordering.cs
+++ b/src/CoreTests/Bug_4874_coordinator_drain_ordering.cs
@@ -40,14 +40,19 @@ namespace CoreTests;
 // out from under an OpenAsync the coordinator loop's StopAsync did not drain.
 //
 // The fix is upstream and is now consumed on master:
-//   Priority 1 - Weasel 9.16.2 (weasel#349/#350): AdvisoryLock disposing-guard + swallow disposed-pool
-//                ObjectDisposedException in TryAttainLockAsync (return false = "not attained during shutdown").
+//   Priority 1 - Weasel (weasel#349/#350, then weasel#354): AdvisoryLock disposing-guard, plus a disposed
+//                data source treated as terminal in TryAttainLockAsync -- latch and rethrow. 9.16.2 swallowed
+//                the ObjectDisposedException and returned false instead, which read as "lock held elsewhere"
+//                and made the coordinator re-poll a dead pool forever; 9.16.3 restores the escape (marten#4915).
 //   Priority 2 - JasperFx 2.24.1 (jasperfx#499/#500): ProjectionCoordinatorBase.executeAsync returns on
 //                cancellation / treats ObjectDisposedException as terminal instead of re-polling.
+//   Priority 3 - marten#4923: ProjectionCoordinator drains its loop on disposal, so a host disposed without a
+//                graceful StopAsync cannot leave the loop polling a data source the container already killed.
 //
 // This test reproduces the abort in the reporter's shape (HotCold, connection-string / Marten-owned
-// source, boot+teardown under an active cold-node poll) and asserts zero PoolingDataSource aborts. It
-// failed on the pre-fix stack and passes now that Weasel 9.16.2 + JasperFx 2.24.1 are referenced.
+// source, boot+teardown under an active cold-node poll) and asserts zero leadership-poll aborts. It failed
+// on the pre-fix stack. Note that it exercises the GRACEFUL teardown (StopAsync then Dispose); the
+// dispose-without-stop shape is Bug_4915_coordinator_disposal_drain.
 [Collection("integration")]
 public class Bug_4874_coordinator_drain_ordering
 {
@@ -63,15 +68,20 @@ public class Bug_4874_coordinator_drain_ordering
     {
         const string schema = "bug4874_coordinator_drain";
 
-        // Count only the specific abort this bug produces: an OpenAsync against a disposed Npgsql pool.
+        // Count only the specific abort this bug produces: the coordinator's leadership poll opening a
+        // connection against a disposed Npgsql pool.
+        //
+        // #4915: this used to match ANY disposed-pool ObjectDisposedException in the process, which made the
+        // test fail in a full CoreTests run while passing in isolation -- other tests leak background work
+        // that touches their own disposed data sources, and AdvisoryLock.DisposeAsync benignly touches one
+        // too. FirstChanceException is AppDomain-wide, and CoreTests already runs serially, so the only fix is
+        // to attribute the abort to the poll. See DisposedPoolAborts.
         var aborts = new ConcurrentQueue<string>();
         void handler(object? sender, FirstChanceExceptionEventArgs e)
         {
-            if (e.Exception is ObjectDisposedException ode &&
-                (ode.ObjectName?.Contains("PoolingDataSource") == true ||
-                 ode.Message.Contains("PoolingDataSource")))
+            if (DisposedPoolAborts.IsAdvisoryLockPollAbort(e.Exception))
             {
-                aborts.Enqueue(ode.ToString());
+                aborts.Enqueue(e.Exception.ToString());
             }
         }
 
@@ -115,7 +125,7 @@ public class Bug_4874_coordinator_drain_ordering
         }
 
         aborts.Count.ShouldBe(0,
-            $"Expected no ObjectDisposedException('PoolingDataSource') aborts, but saw {aborts.Count}. " +
+            $"Expected no disposed-pool aborts on the leadership poll path, but saw {aborts.Count}. " +
             "The cold-node leadership poll opened a connection against an already-disposed data source " +
             "during teardown (see #4874).");
     }
--- a/src/CoreTests/Bug_4915_coordinator_disposal_drain.cs
+++ b/src/CoreTests/Bug_4915_coordinator_disposal_drain.cs
@@ -1,0 +1,239 @@
+using System;
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using JasperFx.Events.Daemon;
+using JasperFx.Events.Projections;
+using Marten;
+using Marten.Events.Aggregation;
+using Marten.Testing.Harness;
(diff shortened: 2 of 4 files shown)


jeremydmiller:
Reopened — the red CI on the first push (`0f10d4e9`) was **not** this fix. Both failing `build` jobs failed on `Bug_4874_coordinator_drain_ordering`, with `475 passed, 1 failed`; my two new tests passed.

`Bug_4874` counts every disposed-pool `ObjectDisposedException` in the process through the AppDomain-wide `FirstChanceException` hook. I confirmed it **fails on pristine `origin/master`** in a full `CoreTests` run and passes in isolation, so the flake predates this PR. The cause is that other tests leak background work that later touches their own disposed data sources (daemon `StopAllAsync`, schema migrations, session reads), and `AdvisoryLock.DisposeAsync` benignly touches one too. `CoreTests` already runs serially, so no collection attribute can fix it.

`11e354681` attributes the abort to the leadership poll instead — `TryAttainLockAsync` re-opening a dead pool is the bug both tests are about; anything else touching one is not. Shared between the two tests in `DisposedPoolAborts`. Since `Exception.StackTrace` isn't populated at first-chance time, it walks the throwing thread's live stack.

Local validation on a clean database after that commit:

- full `CoreTests` net9.0: **475 passed, 1 failed**, three consecutive runs
- full `CoreTests` net10.0: **475 passed, 1 failed**
- the one failure, `describing_database_usage_from_DefaultTenancy.create_usage`, reproduces on pristine `master` in my environment and passes on CI

Also worth noting: **Weasel 9.16.3 is now published**, carrying JasperFx/weasel#354. This PR does not require it — the two fixes are independent, and this one brings the abort count to zero on Weasel 9.16.2 as well. The `Directory.Packages.props` bump to 9.16.3 is deliberately not in this PR.

PR #4933, opened 2026-07-10
fix(#4926): drop the duplicate Attribute suffix on IsSoftDeletedMetadata

### LINKED ISSUE #4926: Duplicate "Attribute" on IsSoftDeletedMetadataAttribute
Something that confused me today was the `IsSoftDeletedMetadataAttributeAttribute` ([source](https://github.com/JasperFx/marten/blame/692656d5f9241faa7611da4eceb474fd02aaf654/src/Marten/Schema/MetadataAttributes.cs#L47)).  The .NET Convention is to leave off the "Attribute" when it is referenced in code.  Because of the naming, this does not compile.
```
    //[IsSoftDeletedMetadata] (does not compile)
    [IsSoftDeletedMetadataAttribute] //works but differs from .NET conventions
    public bool Deleted { get; set; }

    [SoftDeletedAtMetadata]
    public DateTimeOffset? DeletedAt { get; set; }

```

This is a minor annoyance and would be a breaking change.  Cleaning it up would avoid confusion but feel free to ignore this.  Also, if the attribute approach is not recommended for some reason (I didn't find it mentioned in the docs), that might be good to document.

Thanks for all you do!

jeremydmiller:
Closes #4926.

The metadata attribute class was named `IsSoftDeletedMetadataAttributeAttribute`, so the .NET-conventional usage `[IsSoftDeletedMetadata]` did **not** compile — callers had to write the redundant `[IsSoftDeletedMetadataAttribute]`, out of step with every sibling (`LastModifiedMetadata`, `CreatedAtMetadata`, `SoftDeletedAtMetadata`, `TenantIdMetadata`, `DocumentTypeMetadata`).

Renamed the class to `IsSoftDeletedMetadataAttribute`.

**Source-compatible.** `[IsSoftDeletedMetadataAttribute]` still binds (now directly to the class name), and `[IsSoftDeletedMetadata]` now works too. The only thing that would break is a reference to the type's *full* name in code (e.g. `typeof(IsSoftDeletedMetadataAttributeAttribute)`) — there were none in the repo.

### DIFF
--- a/src/Marten/Schema/MetadataAttributes.cs
+++ b/src/Marten/Schema/MetadataAttributes.cs
@@ -44,7 +44,7 @@ public class TenantIdMetadataAttribute: MartenAttribute
 ///     Direct Marten to copy the is soft deleted metadata to this member
 /// </summary>
 [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
-public class IsSoftDeletedMetadataAttributeAttribute: MartenAttribute
+public class IsSoftDeletedMetadataAttribute: MartenAttribute
 {
     public override void Modify(DocumentMapping mapping, MemberInfo member)
     {


jeremydmiller:
No, sorry, this shouldn't be done

Issue #4941, opened 2026-07-12
Sharded + tenant-partitioned daemon: per-tenant subscription agents silently never start / composite agent silently stalls at a fixed sequence (no log output)


erdtsieck:
Marten 9.14.0 / JasperFx 2.24.1 / Wolverine 6.17.0, PostgreSQL 18.4 (RDS). Sharded store: `MultiTenantedWithShardedDatabases` + `UseTenantPartitionedEvents`, 512 shard databases, `UseWolverineManagedEventSubscriptionDistribution = true`, ~526 tenants (mostly 1 per database, some co-located).

## Symptom — one tenant's async projections silently never complete; zero log output

One shard database hosts two tenants:

- **Tenant A** (provisioned first, ordinary domain events): fully healthy — every per-tenant agent has a progression mark at its high-water mark, read models populated.
- **Tenant B** (added later; 1,080,517 events written through the bulk-import path; event log is 551k of type X, 528k of type Y, then a contiguous tail of 1,891 events of type Z + a handful of others):

1. **Three of its six per-tenant subscription agents never start.** They are present in the Wolverine node-assignment registry (assigned to a node), but never write a single `mt_event_progression` row and never process anything. No error, no log line. The other three agents for the *same tenant* run fine.
2. **The composite projection agent that does run stops deterministically at seq 1,078,600** — exactly the sequence before the first type-Z event — and never advances past it. Also completely silent: no exception, no "paused" message, nothing in the logs at any level we capture.
3. This reproduces exactly: after deleting the tenant's progression rows and restarting, the composite happily reprocesses 0 → 1,078,600 in ~2.5 minutes and then stops at the same sequence again. Multiple pod restarts over two days never move it.
4. `pg_stat_activity` shows no blocked or long-running query and no open transaction. The daemon's pooled connection is idle; its *last* statement is an Npgsql `pg_type` OID lookup whose `state_change` timestamp coincides exactly with the moment progression stopped — i.e. the hang is client-side inside the event loader/processing path, not in the database.
5. The per-tenant `HighWaterMark:<tenant>` rows keep ticking normally the whole time, so high-water detection is alive; it is specifically the subscription agents that are (a) not starting or (b) silently stalling.
6. Switching the affected node to `DurabilityMode.Solo` (single node, no Wolverine-managed distribution) did **not** change anything: high-water kept ticking, but the missing agents still did not start and the composite stayed at the same sequence for 20+ minutes on an otherwise idle system.

## Questions / asks

[...]

Issue #4607, opened 2026-06-03
Soft-delete (Disable/Enable) tenant lifecycle on ShardedTenancy / IDynamicTenantSource<string>


jeremydmiller:
## Summary

Follow-up explicitly invited by #4605. `ShardedTenancy` now implements `IDynamicTenantSource<string>` (jasperfx#413), but three of the interface's lifecycle methods are intentionally left unimplemented and throw `NotSupportedException`:

- `DisableTenantAsync(tenantId)` — soft-delete
- `EnableTenantAsync(tenantId)` — re-enable
- `AllDisabledAsync()` — list disabled tenants

`AddTenantAsync` (both overloads) and `RemoveTenantAsync` (hard drop of the assignment row) **are** implemented. From `ShardedTenancy.cs`:

```csharp
Task IDynamicTenantSource<string>.DisableTenantAsync(string tenantId)
    => throw new NotSupportedException(
        "Disable/enable tenant lifecycle is not yet supported on sharded tenancy. " +
        "Use RemoveTenantAsync to drop a tenant assignment, or open an issue to request the soft-delete surface.");
```

## Why it's deferred

The sharded tenant-assignment table (`TenantAssignmentTable`, columns `tenant_id`, `database_id`) has **no `disabled` column**, so there's nowhere to record soft-delete state. The code comment notes this is deferred until a consumer needs it.

## Proposed surface

- Add a `disabled boolean not null default false` column to `TenantAssignmentTable` (with an additive Weasel migration so existing pools upgrade in place).
- `DisableTenantAsync` / `EnableTenantAsync` flip the flag; `AllActive*` filters out disabled rows; `AllDisabledAsync` returns the disabled set.
- Disabled tenants should fail tenant resolution (mirroring `MasterTableTenancy`'s soft-delete semantics) so the two dynamic sources behave consistently behind the store-agnostic `IDynamicTenantSource<string>` admin extensions.

## Consumer driving the request

CritterWatch#269 migrates its tenant-management handlers onto the store-agnostic `IDynamicTenantSource<string>`. With sharded soft-delete unimplemented, the CritterWatch tenant-management UI has to gate Disable/Enable by tenancy capability (full CRUD only for `MasterTableTenancy`; sharded is add/remove only). Implementing this surface would make the dynamic-tenancy management experience uniform across both database-per-tenant models.

## References

- PR #4605 (the "open an issue to request the soft-delete surface" note)
[...]

PR #4945, opened 2026-07-12
GH-4942: run the idempotent provisioning repair on the sharded auto-assign path

### LINKED ISSUE #4942: Sharded tenancy: auto-assign AddTenantToShardAsync skips createPartitionsForTenant when the assignment already exists — half-provisioned tenants are never repaired
Marten 9.14.0 / JasperFx 2.24.1, sharded tenancy (`MultiTenantedWithShardedDatabases` + `UseTenantPartitionedEvents`).

## Problem

The two `AddTenantToShardAsync` overloads behave differently for an already-assigned tenant:

- **Explicit** `AddTenantToShardAsync(tenantId, databaseId)` → `AssignTenantAsync` always runs `createPartitionsForTenant` after the registry upsert. The comment above that call (ShardedTenancy.cs, the "outside the advisory lock" block) explicitly documents re-running it as the idempotent provisioning repair: *"with the assignment row committed but partitions missing, a re-run of AssignTenantAsync (or any idempotent provisioning repair) completes the tenant."*
- **Auto-assign** `AddTenantToShardAsync(tenantId, ct)` → `AddTenantAsync` → `findOrAssignTenantDatabaseAsync` returns at **Step 1** ("Check assignment table") as soon as an assignment row exists — `createPartitionsForTenant` and `PerTenantEventSequences.EnsureSequencesAsync` never run.

So the documented repair semantics hold for one overload but not the other, and the auto-assign overload is the one a caller without a database id naturally uses.

## Real-world impact (production)

Our conversion tool calls the auto-assign overload before importing a tenant. The shard databases were recreated while the master registry survived, so the assignment row existed but the shard had no per-tenant partitions. The early return skipped all partition DDL; the event/stream partitions got created implicitly by the append path, but **all 71 partitioned document tables were missing the tenant's list partition**. Every projection write then failed with `23514 no partition of relation …` — and because of the silent-failure issues described in #4941, this ran undetected for two days until users noticed empty read models. Recovery required hand-writing 71 `CREATE TABLE … PARTITION OF` statements and resetting the tenant's progression rows.

## Ask

Make `findOrAssignTenantDatabaseAsync` run the idempotent `createPartitionsForTenant` (+ `EnsureSequencesAsync`) for the existing-assignment path too — or at minimum align the documented "re-run repairs a half-provisioned tenant" semantics across both overloads. The DDL is `IF NOT EXISTS` all the way down, so for a fully-provisioned tenant it's a handful of cheap catalog checks; a `TenantIds`-cache guard could keep the hot path free of them (only run the repair when the tenant wasn't already in the local cache, i.e. once per process per tenant).

Happy to PR this. Related: #4941 (the silent-failure presentation of exactly this state).

jeremydmiller:
Closes GH-4942. Also resolves the root-cause leg of GH-4941 (the half-provisioned-tenant state that then failed silently).

## Root cause

`ShardedTenancy.findOrAssignTenantDatabaseAsync` returned early as soon as an assignment row existed — on both the Step 1 registry check and the under-advisory-lock re-check — skipping `createPartitionsForTenant` and `PerTenantEventSequences.EnsureSequencesAsync` entirely.

That early return is unsafe by design of the provisioning sequence itself: the assignment row is deliberately committed **before** the (idempotent, potentially slow) partition DDL, which runs outside the advisory lock. The comment above that block documents the crash-safety contract as *"with the assignment row committed but partitions missing, a re-run of AssignTenantAsync (or any idempotent provisioning repair) completes the tenant."* The explicit `AddTenantToShardAsync(tenantId, databaseId)` overload honored that contract; the auto-assign `AddTenantToShardAsync(tenantId, ct)` overload did not — a half-provisioned tenant (assignment row present, list partitions and/or per-tenant event sequence missing) was **never** repaired by auto-assign, so every write to a missing document partition failed with `23514 no partition of relation ...` forever.

Production impact from the report: 512-tenant `MultiTenantedWithShardedDatabases` + `UseTenantPartitionedEvents` deployment, one tenant with 71 partitioned document tables missing its list partitions, silently broken for two days (#4941 covers the silent presentation).

## Fix

Both existing-assignment paths in `findOrAssignTenantDatabaseAsync` now run the same idempotent `createPartitionsForTenant` (which includes `EnsureSequencesAsync` under `UseTenantPartitionedEvents`) that the fresh-assignment and explicit-assignment paths run. For the under-lock path the DDL still executes **outside** the advisory lock, per the existing rationale (the lock guards registry rows, not shard-local idempotent DDL).

## Hot-path guard

The repair only runs when the tenant was **not** already in the local `_tenantToDatabase` cache — i.e. at most once per process per tenant. The hot-path callers (`GetTenant` / `GetTenantAsync` / `FindOrCreateDatabase`) consult that cache before ever reaching `findOrAssignTenantDatabaseAsync`, so steady-state tenant resolution is unchanged; first sight of a tenant costs one batch of `IF NOT EXISTS` catalog checks. The explicit `AssignTenantAsync` overload keeps repairing unconditionally, matching its documented re-run semantics.

## Tests

New `TenantPartitionedEventsTests/Sharded/Bug_4942_auto_assign_provisioning_repair.cs` (4 tests, all verified to fail on master before the fix except the last, which pins pre-existing behavior):

- `auto_assign_overload_repairs_half_provisioned_tenant_in_new_process` — assignment row intact, document partition + per-tenant event sequence dropped; a fresh process calling the auto-assign overload recreates both and the tenant is writable again (pre-fix: 23514).
- `lazy_tenant_resolution_repairs_on_first_sight` — plain `GetTenantAsync` heals the same damage, asserted purely from the catalog (no writes that could mask it via lazy schema apply).
- `repair_runs_at_most_once_per_process_per_tenant` — with the tenant already cached, a second auto-assign call does no repair work (the dropped sequence stays dropped); a new process then heals it.
- `explicit_overload_repairs_even_when_tenant_is_cached` — `AddTenantToShardAsync(tenantId, databaseId)` keeps repairing unconditionally, warm cache or not.

Full `TenantPartitionedEventsTests` (232/232) and `MultiTenancyTests` (159/159) pass; full solution builds clean in Release.

### DIFF
--- a/src/Marten/Storage/ShardedTenancy.cs
+++ b/src/Marten/Storage/ShardedTenancy.cs
@@ -425,7 +425,9 @@ public class ShardedTenancy : ITenancy, ITenancyWithMasterDatabase, ITenantDatab
         // bulk migration measured ~26s between tenant starts on exactly this, while the median tenant's
         // actual work took 4s) and pushes concurrent waiters past their command timeout. Crash-safety is
         // unchanged: with the assignment row committed but partitions missing, a re-run of
-        // AssignTenantAsync (or any idempotent provisioning repair) completes the tenant.
+        // AssignTenantAsync (or any idempotent provisioning repair) completes the tenant — and since
+        // #4942 the auto-assign path (findOrAssignTenantDatabaseAsync) runs the same repair on first
+        // sight of an already-assigned tenant, so both AddTenantToShardAsync overloads honor it.
         if (_databasesById.TryFind(databaseId, out var database))
         {
             database.TenantIds.Fill(tenantId);
@@ -790,6 +792,13 @@ public class ShardedTenancy : ITenancy, ITenancyWithMasterDatabase, ITenantDatab
         await maybeApplyChanges().ConfigureAwait(false);
         await maybeSeedDatabases().ConfigureAwait(false);
 
+        // #4942: remember whether THIS process had already resolved the tenant before this call.
+        // The existing-assignment repair below runs only on first sight per process — the callers
+        // on the hot path (GetTenant/GetTenantAsync/FindOrCreateDatabase) consult _tenantToDatabase
+        // before ever reaching this method, so the repair costs at most one batch of IF NOT EXISTS
+        // catalog checks per tenant per process.
+        var previouslyKnown = _tenantToDatabase.TryFind(tenantId, out _);
+
         // Step 1: Check assignment table
         var databaseId = await FindDatabaseForTenantAsync(tenantId, CancellationToken.None)
             .ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


jeremydmiller:
Review notes (self-review since this is our own PR): both early-return holes covered — the Step-1 registry check and the under-lock race path the issue didn't catch; the first-sight-per-process guard anchors on `_tenantToDatabase`, which is the actual hot-path resolution cache; DDL stays outside the advisory lock per the #4607 rationale; red-then-green verified (3/4 new tests fail pre-fix) with 232/232 TenantPartitionedEventsTests + 159/159 MultiTenancyTests. Merging on green — root-cause leg of #4941.

PR #4960, opened 2026-07-16
Fix AdvancedSql/raw-SQL scalar queries for reference-typed columns (byte[], IPAddress, etc.)

### LINKED ISSUE #4959: AdvancedSql.QueryAsync<...> throws ArgumentException/TypeLoadException when a requested type is byte[] (bytea column)
**What happens**

Any `AdvancedSql.QueryAsync<...>` call where one of the requested types is `byte[]` throws at query construction time, e.g.:

```csharp
await session.AdvancedSql.QueryAsync<Guid, byte[], string>(
    "select row(id), row(data), row(media_type) from blob where id = ?", ct, id);
```

```
System.ArgumentException: GenericArguments[0], 'System.Byte[]', on
'Marten.Linq.SqlGeneration.ScalarSelectClause`1[T]' violates the constraint of type 'T'.
 ---> System.TypeLoadException: GenericArguments[0], 'System.Byte[]', on
'System.Nullable`1[T]' violates the constraint of type parameter 'T'.
```

Same failure with a single-type `QueryAsync<byte[]>(...)` — not specific to the tuple/`row()` form.

**Root cause**

`ScalarSelectClause<T>` (`src/Marten/Linq/SqlGeneration/ScalarSelectClause.cs`) is declared `where T : struct` (needed for its `ISelector<T?>`/`Nullable<T>` null-handling). `AdvancedSqlQueryHandlerBase<TResult>.GetSelectClause<T>` (`src/Marten/Linq/QueryHandlers/AdvancedSqlQueryHandler.cs`) special-cases `string` with a dedicated `ScalarStringSelectClause`, but has no equivalent for `byte[]` — so a `byte[]` scalar falls into the generic `ScalarSelectClause<T>` branch and blows up on the `struct` constraint before Npgsql's own automatic `bytea` ↔ `byte[]` mapping ever comes into play.

**Reproduction**

Table:

```csharp
var blobTable = new Table("blob");
blobTable.AddColumn<Guid>("id").AsPrimaryKey();
blobTable.AddColumn("data", "bytea").NotNull();
blobTable.AddColumn<string>("media_type").NotNull();
blobTable.AddColumn<int>("size").NotNull();
opts.Storage.ExtendedSchemaObjects.Add(blobTable);
```

Failing test:

```csharp
public class advanced_sql_query_bytea_column: BugIntegrationContext
{
    [Fact]
    public async Task can_query_single_bytea_scalar()
    {
        // ... seeds the blob table above via ExtendedSchemaObjects, inserts a row ...
        var data = (await session.AdvancedSql.QueryAsync<byte[]>(
            "select data from public.blob where id = ?", CancellationToken.None, theId)).Single();
        data.ShouldBe(theData);
    }

    [Fact]
    public async Task can_query_bytea_column_in_row_tuple()
    {
        var (id, data, mediaType) = (await session.AdvancedSql.QueryAsync<Guid, byte[], string>(
            "select row(id), row(data), row(media_type) from public.blob where id = ?",
            CancellationToken.None, theId)).Single();
        // ...
    }
}
```

Both fail with the `ArgumentException`/`TypeLoadException` above.

**Workaround**

Drop to a raw `NpgsqlCommand` and read the `bytea` column with `reader.GetFieldValueAsync<byte[]>(...)` directly — bypasses `AdvancedSql` entirely for that one column.

**Note:** A PR with a failing-test reproduction and a fix will follow.

mdissel:
## Summary

Fixes #4959.

- `ScalarSelectClause<T>` is `where T : struct` because it also implements `ISelector<T?>` via `Nullable<T>`. Any Npgsql-mapped **reference** type (`byte[]`/`bytea`, `IPAddress`/`inet`, `BitArray`, `JsonDocument`, ...) blew up at query-construction time with `ArgumentException`/`TypeLoadException` before Npgsql's own type mapping ever got a chance to run. Only `string` had a dedicated workaround (`ScalarStringSelectClause`) — `byte[]` and other reference types did not.
- Added `ScalarClassSelectClause<T> where T : class` as the reference-type counterpart — relies on natural reference-type nullability instead of `Nullable<T>`, mirroring the existing `ScalarStringSelectClause` pattern.
- Branch on `typeof(T).IsValueType` in both `AdvancedSqlQueryHandlerBase.GetSelectClause<T>` (`session.AdvancedSql.QueryAsync`/`StreamAsync`) and `UserSuppliedQueryHandler.GetSelectClause` (`session.QueryAsync<T>("select ...")`), which had the identical latent bug.
- Confirmed this isn't `byte[]`-specific but *is* specific to types with a genuine Npgsql **scalar** type mapping — CLR arrays like `int[]`/`Guid[]` are composed from their element type by Npgsql rather than registered as scalar mappings, so they don't hit this code path at all (a separate, unrelated limitation, out of scope here).

## Test plan

- [x] New `src/DocumentDbTests/Bugs/advanced_sql_query_scalar_reference_types.cs`: single-scalar, row-tuple (`row(...)`), and `StreamAsync` cases for a `byte[]`/`bytea` column, plus a single-scalar case for an `IPAddress`/`inet` column, against a plain (non-document) table added via `ExtendedSchemaObjects`.
- [x] Full `CoreTests` suite: 479 passed, 0 failed.
- [x] Full `DocumentDbTests` suite: 1055 passed, 0 failed.
- [x] `LinqTests` scalar-select suites (`Scalar`/`select_scalar` filters): 36 passed, 0 failed — confirms the existing LINQ scalar-projection paths (which never route arbitrary reference types through `ScalarSelectClause<T>` in the first place) are unaffected.
- [x] Verified pre-fix: both new tests reproduce the exact reported `ArgumentException`/`TypeLoadException` before the fix, and pass after.

### DIFF
--- a/src/DocumentDbTests/Bugs/advanced_sql_query_scalar_reference_types.cs
+++ b/src/DocumentDbTests/Bugs/advanced_sql_query_scalar_reference_types.cs
@@ -1,0 +1,121 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Marten.Testing.Harness;
+using Npgsql;
+using Shouldly;
+using Weasel.Postgresql;
+using Weasel.Postgresql.Tables;
+using Xunit;
+
+namespace DocumentDbTests.Bugs;
+
+public class advanced_sql_query_scalar_reference_types: BugIntegrationContext
+{
+    private readonly byte[] theData = Encoding.UTF8.GetBytes("hello blob");
+    private readonly IPAddress theUploadedFrom = IPAddress.Parse("203.0.113.42");
+    private readonly Guid theId = Guid.NewGuid();
+
+    private async Task seedBlobTableAsync()
+    {
+        StoreOptions(opts =>
+        {
+            // Plain table, deliberately not a Marten document, so its bytes live in a
+            // real `bytea` column instead of base64-in-JSONB. Schema-qualified with the
+            // test's own schema so BugIntegrationContext's per-test schema drop cleans it up.
+            var blobTable = new Table(new PostgresqlObjectName(SchemaName, "blob", SchemaUtils.IdentifierUsage.General));
+            blobTable.AddColumn<Guid>("id").AsPrimaryKey();
+
+            // Explicit "bytea" pg type: Weasel's generic AddColumn<byte[]>() maps byte[] to
+            // a Postgres smallint[] array (element-wise), not bytea.
+            blobTable.AddColumn("data", "bytea").NotNull();
+            blobTable.AddColumn<string>("media_type").NotNull();
+            blobTable.AddColumn<int>("size").NotNull();
+
+            // Second reference-type scalar (a real Postgres "inet" column, not bytea) to prove the
+            // fix isn't byte[]-specific: any reference type with a direct Npgsql scalar mapping
+            // (byte[], BitArray, JsonDocument, IPAddress, PhysicalAddress, ...) hit the same
+            // ScalarSelectClause<T> where T : struct crash. Note this does *not* extend to CLR
+            // arrays like int[]/Guid[] — Npgsql composes those from their element type rather than
+            // registering a scalar type mapping, so HasTypeMapping(typeof(int[])) is false and they
+            // never reach this code path at all.
+            blobTable.AddColumn<IPAddress>("uploaded_from").NotNull();
+
+            opts.Storage.ExtendedSchemaObjects.Add(blobTable);
+        });
+
+        await theStore.Storage.ApplyAllConfiguredChangesToDatabaseAsync();
+
+        await using var conn = new NpgsqlConnection(ConnectionSource.ConnectionString);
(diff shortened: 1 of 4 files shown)


mysticmind (changes requested):
The `AdvancedSql` coverage is solid. Could you please also add a unit test for the `UserSuppliedQueryHandler` path, i.e. a raw-SQL scalar case like `QueryAsync<byte[]>("select …")`? Right now all the new tests go through `AdvancedSql`, so that second fixed path has no coverage.

mdissel:
Good catch — added \`can_query_single_bytea_scalar_via_raw_sql\` which exercises \`session.QueryAsync<byte[]>("select ...")\`, i.e. the \`UserSuppliedQueryHandler.GetSelectClause\` path directly (pushed in 586b34f).

Note: the tuple/\`row(...)\` form (\`QueryAsync<T1, T2, T3>\`) only exists on \`AdvancedSql\`, not on the base \`QueryAsync<T>\` API, so that variant is already fully covered by the existing \`AdvancedSql\` row-tuple test — there's no equivalent to add on the raw-SQL side.

mysticmind (approved):
Thanks @mdissel, looks good to me!

Issue #4966, opened 2026-07-17
Natural key is not updated when rebuilding a projection


ytqsl:
Scenario:

A snapshot with natural key. During projection rebuild, the natural key is set at first, but not updated when the key changes.

Repro test - https://github.com/JasperFx/marten/pull/4967

PR #4967, opened 2026-07-17
#4966: Natural key is not updated when rebuilding a projection

ytqsl:


jeremydmiller:
Thanks for the clear reproduction, @ytqsl! Superseded by #4970, which retains your test (with you as co-author), bumps to JasperFx 2.28.1 carrying the actual fix (JasperFx/jasperfx#528), and closes #4966.

Root cause: natural-key discovery silently skipped a two-arg static evolve method (`static TDoc Apply(TEvent e, TDoc current)`), so the update event produced no natural-key mapping and the changed key was never written to `mt_natural_key_X`. Your repro fails on 2.28.0 and passes on 2.28.1.

PR #4995, opened 2026-07-21
Marten.TimescaleDB extension: projection + document hypertables (#4980)

### LINKED ISSUE #4980: Marten.TimescaleDB extension: time-series hypertables for projections and event storage
> Migrated from [JasperFx/CritterWatch#75](https://github.com/JasperFx/CritterWatch/issues/75) — this is a Marten extension proposal, not a CritterWatch feature, so it belongs here. Original issue closed in favor of this one.

## Context

[TimescaleDB](https://github.com/timescale/timescaledb) is a PostgreSQL extension that adds automatic time-based partitioning (**hypertables**), columnar compression, continuous aggregates, and retention policies — all transparent to normal SQL queries. This proposes a `Marten.TimescaleDB` extension that wires those capabilities into Marten's event store and projection system, following the same shape as the existing `Marten.PgVector` / `Marten.PostGIS` extensions under `tools/`.

Two things make it a natural fit for Marten:

1. **`mt_events` is an append-only, monotonically time-ordered table** — hypertable partitioning by `timestamp` gives large speedups on time-window queries plus automatic retention via `drop_chunks(...)`, as a more powerful alternative to `UseArchivedStreamPartitioning`.
2. **Aggregate/flat-table projections that summarise per-minute/per-hour metrics** can be hypertables with continuous aggregates (TimescaleDB-managed materialized views that auto-refresh on insert) — real-time dashboards with pre-computed rollups and no custom downsampling code.

### 1. Flat-table projections as hypertables (recommended starting point)

**Lowest risk, highest value** — doesn't touch `mt_events`. Marten's [FlatTableProjection](https://martendb.io/events/projections/flat.html) writes event data to custom tables; if that table is a hypertable, you get time-series analytics on projected data for free.

```csharp
opts.UseTimescaleDB(ts =>
{
    ts.ProjectionAsHypertable<MetricsProjection>("recorded_at", hyper =>
    {
        hyper.ChunkInterval = TimeSpan.FromHours(1);
        hyper.CompressAfter = TimeSpan.FromDays(30);
        hyper.ContinuousAggregate("hourly_metrics", "1 hour",
            "avg(value) as avg_val, max(value) as max_val");
    });
});
```

```csharp
public class MetricsProjection : FlatTableProjection
{
    public MetricsProjection() : base("sensor_metrics", SchemaNameSource.EventSchema)
    {
        Table.AddColumn<DateTimeOffset>("recorded_at").NotNull();
        Table.AddColumn<string>("sensor_id").NotNull();
        Table.AddColumn<double>("value").NotNull();

        Project<SensorReading>(map =>
        {
            map.Map(x => x.RecordedAt, "recorded_at");
            map.Map(x => x.SensorId, "sensor_id");
            map.Map(x => x.Value);
        });
    }
}
```

The async daemon populates the hypertable from events; TimescaleDB then partitions by time chunks, compresses old chunks (up to ~95% storage reduction), maintains continuous aggregates for fast dashboard queries, and drops chunks past a retention window — all declaratively.

### 2. Event store hypertable — `mt_events` partitioned by `timestamp`

```csharp
ts.EventStoreAsHypertable(hyper =>
{
    hyper.ChunkInterval = TimeSpan.FromDays(7);
    hyper.CompressAfter = TimeSpan.FromDays(30);
    hyper.RetainFor    = TimeSpan.FromDays(365);
});
```

**Advantages:** replaces manual `UseArchivedStreamPartitioning` with automatic time-based chunking; old events compressed to columnar format (ideal for sequential rebuild scans); retention policies as a more powerful alternative to stream archiving; chunk exclusion speeds up the daemon's `WHERE seq_id > X` scans (seq_id correlates with time).

**Challenges (this is the advanced/high-risk phase — see Risks):** TimescaleDB requires the partitioning column in the PK → composite PK `(seq_id, timestamp)`; FKs *to* hypertables are restricted → `mt_event_tag_*` and stream FKs need rework; compressed chunks are read-only → affects header/metadata updates on old events; the async daemon orders by `seq_id`, not `timestamp`.

### 3. Document tables as hypertables — historical/append-heavy documents

```csharp
ts.DocumentAsHypertable<AuditEntry>(x => x.CreatedAt, hyper =>
{
    hyper.ChunkInterval = TimeSpan.FromDays(30);
    hyper.CompressAfter = TimeSpan.FromDays(90);
});
```

Best for append-heavy document types (audit logs, metrics, activity records) where old data is rarely queried but must be retained.

## Proposed API

```csharp
opts.UseTimescaleDB(ts =>
{
    // 1. Projection tables as hypertables
    ts.ProjectionAsHypertable<MetricsProjection>("recorded_at", hyper =>
    {
        hyper.ChunkInterval = TimeSpan.FromHours(1);
        hyper.CompressAfter = TimeSpan.FromDays(30);
        hyper.RetainFor     = TimeSpan.FromDays(365);
        hyper.ContinuousAggregate("hourly_metrics", "1 hour", "avg(value), max(value), min(value)");
    });

    // 2. Event store as hypertable (advanced)
    ts.EventStoreAsHypertable(hyper =>
    {
        hyper.ChunkInterval = TimeSpan.FromDays(7);
        hyper.CompressAfter = TimeSpan.FromDays(30);
    });

    // 3. Document tables as hypertables
    ts.DocumentAsHypertable<AuditEntry>(x => x.CreatedAt, hyper =>
    {
        hyper.ChunkInterval = TimeSpan.FromDays(30);
        hyper.CompressAfter = TimeSpan.FromDays(90);
    });
});
```

Attribute-driven equivalents for projection/document authors are also on the table: `[Hypertable("timestamp")]` on a document/projection table to mark it time-partitioned, and `[ContinuousAggregate]` on a projection to have Marten generate the `CREATE MATERIALIZED VIEW ... WITH (timescaledb.continuous)` DDL.

## Extension architecture — `Marten.TimescaleDB`

A separate package under `tools/` (alongside `Marten.PgVector` and `Marten.PostGIS`). What it does:

1. Registers the `timescaledb` extension via Weasel's `ExtendedSchemaObjects` (`CREATE EXTENSION IF NOT EXISTS timescaledb`) — same pattern as `tools/Marten.PgVector/PgVectorExtensions.cs`.
2. Post-schema-creation hooks to run `create_hypertable()` after table creation.
3. Adds `add_compression_policy()` for configured tables.
4. Adds `add_retention_policy()` where configured.
5. Generates `CREATE MATERIALIZED VIEW ... WITH (timescaledb.continuous)` DDL for continuous aggregates.
6. Handles FK-constraint adjustments (hypertables can't be FK targets).

No special Npgsql plugin is needed — TimescaleDB uses standard PostgreSQL types and SQL; all configuration is DDL.

**Why a separate library (Option A) over folding into Marten core gated on the extension (Option B):** Option A keeps Marten core PG-vanilla and matches the established `Marten.PgVector` / `Marten.PostGIS` pattern, rather than pulling a TimescaleDB-specific concept into Marten core's public API. Recommendation: **Option A.**

## Implementation priority

1. **Phase 1 — FlatTableProjection hypertables:** lowest risk, doesn't touch `mt_events`, clear value for IoT/metrics/monitoring.
2. **Phase 2 — document hypertables:** moderate risk, useful for audit logs and append-heavy types.
3. **Phase 3 — event store hypertable:** highest risk/reward, requires FK rework and PK changes.

The first useful PR can be the Docker test environment + extension verification — a runnable target before any user-facing API changes.

## Key risks

- **FK restriction** is the biggest blocker for `mt_events` as a hypertable — the tag tables have FKs to `mt_events`.
- **PK must include the partitioning column** — a schema change for existing stores.
- **Compressed chunks are read-only** — affects header/metadata updates on old events.
- **Ordering semantics** — the daemon orders by `seq_id`, not `timestamp`.

## Open questions

- **Continuous-aggregate refresh policies** (`add_continuous_aggregate_policy`) — does Marten own the schedule, or does it just provide the DDL hooks and leave the policy to the operator? Cleanest: Marten provides the hooks, operators set the policy.
- **Multi-tenancy matrix** — hypertables work in conjoined tenancy; database-per-tenant means each tenant DB needs its own `timescaledb` extension. Document the matrix.
- **Docker test env** — the `pgvector/pgvector:pg17` image doesn't include TimescaleDB; need a parallel `docker/postgres-timescaledb/Dockerfile` (mirror of the existing postgres Dockerfile with `timescaledb-2-postgresql-17` layered in) + CI wiring.

## Rough effort

- Core extension wrapper + hypertable conversion: ~1 week (mostly DDL plumbing through Weasel).
- Continuous-aggregate attribute + DDL: ~3 days.
- Test infra (Docker + CI): ~2 days.

Greenfield extension, similar in shape and scope to `Marten.PgVector`; nothing structural blocks it.

## References

- [TimescaleDB GitHub](https://github.com/timescale/timescaledb)
- [create_hypertable API](https://docs.timescale.com/api/latest/hypertable/create_hypertable/)
- [Continuous aggregates](https://docs.timescale.com/use-timescale/latest/continuous-aggregates/)
- [Compression & retention policies](https://docs.timescale.com/use-timescale/latest/compression/)
- [EF Core TimescaleDB extension](https://github.com/cmdscale/CmdScale.EntityFrameworkCore.TimescaleDB)
- Prior art in-repo: `Marten.PgVector` and `Marten.PostGIS` under `tools/`

jeremydmiller:
Closes #4980.

Adds a `Marten.TimescaleDB` companion package (mirroring the `Marten.PgVector` / `Marten.PostGIS` pattern) that turns Marten-managed tables into [TimescaleDB](https://github.com/timescale/timescaledb) hypertables — automatic time-partitioning with columnar compression, retention policies, and continuous aggregates.

## What's included

**Extension (`src/Marten.TimescaleDB`)**
- `UseTimescaleDB()` — registers the `timescaledb` extension on every managed database
- `ProjectionAsHypertable<T>("captured_at", …)` — turns a time-bucketed `FlatTableProjection` table into a hypertable (chunk interval, compression, retention, continuous aggregates). The timestamp-in-PK requirement is **explicit**: the projection declares the time column as its primary key, and the extension fails fast with a clear error otherwise.
- `DocumentAsHypertable<T>(x => x.CreatedAt, …)` — turns an append-heavy document table (audit logs, metrics, activity records) into a hypertable partitioned by one of its own timestamp members. The extension drives the composite `(id, created_at)` primary key under the hood, so the document author writes a single line.
- Hypertable / continuous-aggregate DDL is emitted through a custom `IFeatureSchema` that Marten yields **last**, so it runs after the target tables exist. Delta detection makes re-apply idempotent, and `create_hypertable(..., create_default_indexes => FALSE)` keeps Marten's own schema model authoritative — **no schema drift** (`AssertDatabaseMatchesConfiguration` passes).

**Marten core — minimal, opt-in, default-off**
- `DuplicatedField.PartOfPrimaryKey`: a duplicated field can join the document table's primary key. Honored in `DocumentTable`; threaded through `DocumentMapping.Duplicate` / `MartenRegistry.Duplicate` (`partOfPrimaryKey`, which forces `NOT NULL` and skips the redundant index). The existing v9 partition-aware upsert/update/delete/bulk codegen already reads `IsPrimaryKey` generically (built for list/range-partitioned documents), so the composite key flows through automatically — no storage-codegen changes.
- `InternalsVisibleTo` for `Marten.TimescaleDB` and its test project.

**Intentionally out of scope**
- `mt_events` / `mt_streams` are left untouched. Making the event table a hypertable would require weakening the `(stream_id, version)` optimistic-concurrency guarantee, so it is deliberately not supported; the docs point users to `UseArchivedStreamPartitioning` / tenant-partitioned events instead.

## Semantics / caveats
- `DocumentAsHypertable` requires the partition member to be **immutable** for a given document id (the update path matches the full primary key). Intended for append-heavy types. Load/delete by id still work (one row per id) but a load by id alone scans all chunks.

## Docs & packaging
- New docs page `docs/documents/timescaledb.md` + sidebar entry (markdownlint + cspell clean).
- `docker-compose.timescaledb.yml` (port 5433, official image) for local dev, and a dedicated CI workflow using the `timescale/timescaledb-ha` image.
- Nuke `TestTimescaleDB` target; `Marten.TimescaleDB` added to the pack list so the manual NuGet publish workflow ships it.

## Testing
- 13 extension tests green on **net9.0 and net10.0** against TimescaleDB 2.28.3: extension creation, hypertable conversion, compression/retention policies, continuous aggregate, projection writes, full document lifecycle (store/load/update/delete), composite-PK verification, LINQ queries, idempotent re-apply, and no-drift assertions.
- Existing duplicated-field, composite-PK (`Bug_4223`), and document-CRUD regression tests remain green.


jeremydmiller:
## Proposal: fold `Marten.TimescaleDB` into core Marten

### Why (the dependency case)

Unlike the other extension packages, this one isolates **nothing**. Its csproj has zero `PackageReference`s — only `ProjectReference ..\Marten\Marten.csproj` — and every namespace it touches (Marten, Weasel, JasperFx.Core, JasperFx.Events, BCL) is already a transitive dependency of core Marten.

The sibling extensions exist to quarantine a third-party CLR type **and its Npgsql type-handler registration on the data source**, which genuinely can't live in core without forcing it on every user:

| Extension | What it forces on the data source |
|-----------|-----------------------------------|
| Marten.PgVector | `Pgvector` pkg + `ConfigureNpgsqlDataSourceBuilder(b => b.UseVector())` — a `Pgvector.Vector ↔ "vector"` OID mapping |
| Marten.PostGIS | `Npgsql.NetTopologySuite` type mapping |
| Marten.NodaTime | `Npgsql.NodaTime` type mapping |
| **Marten.TimescaleDB** | **nothing — no CLR type, no Npgsql mapping** (grep for `NpgsqlDataSource`/`DataSourceBuilder` in the extension: 0 hits) |

TimescaleDB is a pure schema/DDL feature (`create_hypertable(...)`, `CREATE MATERIALIZED VIEW ... WITH (timescaledb.continuous)`) plus one `bool` on `DuplicatedField`. The only thing keeping it a separate *package* is namespace hygiene — and it's paying for that with an `InternalsVisibleTo` into Marten's guts (`FlatTableProjection.Table`, `EventGraph.DatabaseSchemaName`, `DuplicatedField.PartOfPrimaryKey internal set`). Folding in makes that IVT unnecessary and lets those seams stay `internal`.

`UseTimescaleDB()` stays opt-in at runtime, so non-users pay nothing.

### Keep the `Marten.TimescaleDB` namespace

This fold-in is **assembly-only**: the code moves into the `Marten` project/assembly, but every type **keeps its existing `Marten.TimescaleDB` / `Marten.TimescaleDB.Internal` namespace**. Do not re-namespace anything to `Marten` or `Marten.Storage`.

Consequences:
- The public API is unchanged for consumers — they still write `using Marten.TimescaleDB;` to reach `UseTimescaleDB()` / `ProjectionAsHypertable` / `DocumentAsHypertable`. This deliberately keeps the time-series surface *scoped* behind its own namespace rather than lighting up with a bare `using Marten;`, so core's default surface stays lean even though the code ships in the core assembly.
- Only the **physical location** (project/assembly) and **packaging** change; the namespace, type names, and call sites are identical to today. Migrating an existing user is a no-op beyond dropping the separate `PackageReference`.

### Concrete fold-in steps

**1. Move the ~8 source files** from `src/Marten.TimescaleDB/` into a dedicated folder in core, e.g. `src/Marten/TimescaleDB/` (mirror the current `Internal/` subfolder). **Namespaces stay exactly as they are** — `Marten.TimescaleDB` and `Marten.TimescaleDB.Internal`; folder layout need not match namespace.

| From | To (folder only — namespace unchanged) |
|------|-----|
| `TimescaleDBExtensions.cs` (`UseTimescaleDB`) | `src/Marten/TimescaleDB/` — stays `namespace Marten.TimescaleDB` |
| `TimescaleDBOptions.cs` (`ProjectionAsHypertable`/`DocumentAsHypertable`) | `src/Marten/TimescaleDB/` — stays `namespace Marten.TimescaleDB` |
| `HypertableOptions.cs`, `ContinuousAggregateDefinition` | `src/Marten/TimescaleDB/` — stays `namespace Marten.TimescaleDB` |
| `Internal/TimescaleDBFeatureSchema.cs` | `src/Marten/TimescaleDB/Internal/` — stays `namespace Marten.TimescaleDB.Internal` |
| `Internal/HypertableSchemaObject.cs` | `src/Marten/TimescaleDB/Internal/` — stays `namespace Marten.TimescaleDB.Internal` |
| `Internal/ContinuousAggregateSchemaObject.cs` | `src/Marten/TimescaleDB/Internal/` — stays `namespace Marten.TimescaleDB.Internal` |
| `Internal/IHypertableTarget.cs` (+ `ProjectionHypertableTarget`, `DocumentHypertableTarget`) | `src/Marten/TimescaleDB/Internal/` — stays `namespace Marten.TimescaleDB.Internal` |

Visibility: types under `Marten.TimescaleDB.Internal` can now be plain `internal` (same assembly as the internals they consume); the public entry points (`TimescaleDBExtensions`, `TimescaleDBOptions`, `HypertableOptions`, `ContinuousAggregateDefinition`) stay `public` under `Marten.TimescaleDB`.

**2. Delete the IVT lines** in `src/Marten/Properties/AssemblyInfo.cs`:
```
[assembly: InternalsVisibleTo("Marten.TimescaleDB")]
[assembly: InternalsVisibleTo("Marten.TimescaleDB.Tests")]
```
With the code in-assembly, `IHypertableTarget`/`HypertableSchemaObject`/etc. become plain `internal`. Types that must stay reachable from the (now in-core) public API stay `public`; the internals they consume no longer need IVT.

**3. Tests:** keep a dedicated `Marten.TimescaleDB.Tests` project (it still needs the TimescaleDB image and its own CI lane) that now `ProjectReference`s core Marten — no IVT needed since the public API stays public under `Marten.TimescaleDB`. Because the namespace is unchanged, the 13 tests (extension creation, hypertable conversion, compression/retention, continuous aggregate, projection writes, full document lifecycle, composite-PK, LINQ, idempotent re-apply, no-drift) carry over **with no `using` changes at all** — only the removed dependency on the deleted extension project.

**4. Packaging / build / CI:**
- Remove `Marten.TimescaleDB` from the pack list in `build/build.cs` (no separate NuGet package).
- Keep the `TestTimescaleDB` Nuke target and the `on-push-do-ci-build-timescaledb.yml` workflow — but **fix the workflow first** (see below); it now runs the in-core tests against the TimescaleDB image.
- Remove `src/Marten.TimescaleDB/` and its `Marten.slnx` entry; keep/repoint the test project entry.

**5. Docs:** `docs/documents/timescaledb.md` and the sidebar entry stay; just drop any "install the Marten.TimescaleDB package" language — the feature ships in core, enabled via `UseTimescaleDB()`.

### Pre-existing blocker to fix in the same pass (independent of fold-in)

CI is red for a reason unrelated to the code: the workflow's "Optimize database for running tests faster" step writes to `/var/lib/postgresql/data/postgresql.conf`, but the `timescale/timescaledb-ha` image keeps its data dir at `/home/postgres/pgdata/data`. Under `bash -e -o pipefail` the failed redirect kills the job **before any test runs**. Fix by tuning via the service container's `command:` args (image-agnostic) rather than exec-editing the conf file, e.g.:
```
command: postgres -c fsync=off -c full_page_writes=off -c synchronous_commit=off
```
Only after this lands will we get a real green test signal for the extension.

### Also worth confirming during the fold-in

- **Transaction-block behavior:** `CREATE MATERIALIZED VIEW ... WITH (timescaledb.continuous)` and `create_hypertable(..., migrate_data => TRUE)` have historically had "cannot run inside a transaction block" restrictions. The 13 tests pass on a clean bootstrap; verify it still holds when hypertable DDL is batched into one migration transaction alongside other pending changes.
- **Policy drift:** compression/retention settings are only emitted when the hypertable delta is `Create`; later changes to `CompressAfter`/`RetainFor`/segmentby aren't diffed/re-applied. Fine as a documented limitation — add a one-line note to the docs.

### Direction (decided)

Fold into the core Marten assembly, **keeping the `Marten.TimescaleDB` namespace**. This gets the best of both: no separate package / csproj / IVT / packaging entry to maintain, zero runtime cost to non-users (`UseTimescaleDB()` is opt-in), **and** the time-series API stays scoped behind its own `Marten.TimescaleDB` namespace so core's default surface (`using Marten;`) stays lean. Consumers see an unchanged API — only the extra `PackageReference` goes away.

_Drafted during PR review as a hand-off for a dedicated agent._

Issue #554, opened 2016-10-12
LINQ Queries with .Any() or .Count() throw or return wrong results


lngr:
The following code on an `List<Guid>` returns the wrong number of results if the list is empty:

`theSession.Query<Target>().Where(x => x.GuidList.Any()).ToList();
`

The following code throws an exception "Sequence contains no elements".

`theSession.Query<Target>().Where(item => item.NumberArray.Length == 3)`
`theSession.Query<Target>().Where(item => item.GuidList.Count() == 1)`

The attached pull-request has testcases.

jeremydmiller:
This one's probably going to be easy to fix by just sticking an AND [locator] IS NOT NULL in the where clause.
```
theSession.Query<Target>().Where(x => x.GuidList.Any()).ToList(); 
```

These two?

```
theSession.Query<Target>().Where(item => item.NumberArray.Length == 3)
theSession.Query<Target>().Where(item => item.GuidList.Count() == 1)
```

You can work around that by just having a calculated field for the child count and query against that. I bet it'd be more efficient to do it that way than trying to do the JSON operators to get in there anyway. I'd honestly vote against trying to "fix" that in the Linq support.

lngr:
OK, then we'll use the workaround here.

jeremydmiller:
The `Any()` looks like it'll be fixed by #564. I'll try to address the `Count()` usage this week.

jeremydmiller:
Got it this morning.

lngr:
@jeremydmiller nice, thanks!

PR #5008, opened 2026-07-22
Batched extended-progression telemetry write in one round trip (jasperfx#553)

erdtsieck:
The Marten half of jasperfx#553 (pairs with JasperFx/jasperfx PR "Batch extended-progression heartbeat writes per database flush interval").

The JasperFx.Events `ExtendedProgressionWriter` now coalesces every shard's heartbeat on a database into one batch per flush interval and drives a new batched `IEventDatabase.WriteExtendedProgressionAsync` overload, because the per-shard single-row write does not scale under per-tenant agent fan-out (agents = projections × tenants): on a sharded multi-tenant deployment the one-connection-rent-per-shard-per-interval write path produced ~780 `mt_mark_event_progression_extended` calls/sec per shard database and drove the server to its connection ceiling.

Marten's implementation lands the whole batch in **one round trip on one rented connection**, as a plain `UPDATE ... FROM unnest` join rather than a new batched database function — **no schema object is added**, so deployments running `AutoCreate.None` pick the batching up without a migration. Semantics match `mt_mark_event_progression_extended` exactly: update-only telemetry decoration of existing progression rows, never INSERT, never touch `last_seq_id` / `last_updated`, shards without a progression row are skipped silently. Single-state batches delegate to the existing function path.

The method compiles against JasperFx.Events 2.33.1 as a plain overload and becomes the interface implementation automatically once the dependency bumps to the JasperFx.Events release carrying the batched default interface member. Verified locally against a 2.33.2 pre-release pack of the JasperFx PR: the daemon → `ExtendedProgressionWriter` → `MartenDatabase` path dispatches into the batch overload and the existing `Bug_5001_running_on_node_persisted` tests still pass.

New `DaemonTests.extended_progression_batch_write` covers: multi-shard batch updates in one call, never-insert + progression untouched (missing shard skipped silently), and empty/single-state batches.

### DIFF
--- a/src/DaemonTests/extended_progression_batch_write.cs
+++ b/src/DaemonTests/extended_progression_batch_write.cs
@@ -1,0 +1,177 @@
+using System;
+using System.Threading.Tasks;
+using DaemonTests.TestingSupport;
+using JasperFx.Core;
+using JasperFx.Events.Daemon;
+using JasperFx.Events.Projections;
+using Marten.Events.Aggregation;
+using Marten.Storage;
+using Shouldly;
+using Weasel.Core;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace DaemonTests;
+
+public record BatchTelemetryEvent();
+
+public class BatchTelemetryStream { public Guid Id { get; set; } }
+
+public partial class BatchTelemetryProjection: SingleStreamProjection<BatchTelemetryStream, Guid>
+{
+    public void Apply(BatchTelemetryEvent @event, BatchTelemetryStream projection) { }
+}
+
+public partial class OtherBatchTelemetryProjection: SingleStreamProjection<BatchTelemetryStream, Guid>
+{
+    public OtherBatchTelemetryProjection()
+    {
+        Name = "OtherBatchTelemetry";
+    }
+
+    public void Apply(BatchTelemetryEvent @event, BatchTelemetryStream projection) { }
+}
+
+// jasperfx#553 — the batched extended-progression write. The JasperFx.Events ExtendedProgressionWriter
+// coalesces every shard's heartbeat on a database into one batch per flush interval; Marten's overload
+// must land the whole batch in ONE round trip with the exact semantics of
+// mt_mark_event_progression_extended: update-only telemetry decoration of existing progression rows,
+// never INSERT, never touch last_seq_id / last_updated.
+public class extended_progression_batch_write: DaemonContext
+{
+    public extended_progression_batch_write(ITestOutputHelper output): base(output)
+    {
+    }
+
+    private async Task<(object? heartbeat, object? status, object? reason, object? node, object? seq)> readRowAsync(
+        string shard)
+    {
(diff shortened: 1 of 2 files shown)


jeremydmiller:
## Resolution: this work has already landed on `master` — recommend closing (no rebase / superseding PR needed)

Thanks @erdtsieck — this is already in `master`, with your authorship intact. Writing up the state so the close is unambiguous.

### What's on `master` now

- **Production code:** commit [`334283ee7`](https://github.com/JasperFx/marten/commit/334283ee7) *"Add the batched extended-progression telemetry write (jasperfx#553) (#5008)"* — **authored by Anne Erdtsieck** (`anne.erdtsieck@topicus.nl`), committed by GitHub as a squash with `Co-authored-by: Jeremy D. Miller`. The `MartenDatabase.WriteExtendedProgressionAsync(IReadOnlyList<ShardState>)` method on `master` is **byte-identical** to this PR's. Attribution is already correct — no superseding PR is required to preserve it.
- **The `#5022` flake is already fixed** on `master` via the merged #5023 (`fix/5022-drain-shutdown-telemetry-race`). The test no longer stands up a live daemon and `StopAllAsync()`s it before asserting; it seeds the committed progression rows directly through `mt_mark_event_progression` (`seedProgressionRowsAsync()`), so the shutdown-telemetry `"Stopped"` heartbeat can't race in and clobber the asserted rows.
- **The dependency bump is in:** `master` pins **JasperFx.Events 2.34.0**, which carries the batched default interface member `IEventDatabase.WriteExtendedProgressionAsync(IReadOnlyList<ShardState>, …)` (added in jasperfx `6f30c83`, #554/#553). That means the method is no longer dead code — it now overrides the interface DIM, and the daemon's `ExtendedProgressionWriter` dispatches into Marten's one-round-trip batch path.

Net: both merge blockers I flagged in review (the flaky test and "the overload is dead until the JasperFx bump") are resolved on `master`, and the production code there is identical to this branch's.

### Why not rebase this branch

A rebase onto `master` would contribute **nothing new** and would actively regress the test:

- The production method already matches `master` byte-for-byte → zero production delta.
- This branch still carries the **original daemon-based (flaky) version** of `extended_progression_batch_write.cs`. Rebasing would either no-op against master's fixed test or, worse, reintroduce the `#5022` race. We want master's `seedProgressionRowsAsync()` version, not this branch's.

### Recommended action

1. **Close #5008** as already-landed, linking commit `334283ee7` and PR #5023 so the trail is clear.
2. No superseding PR needed — Anne is already the author of the commit on `master`.
3. When the other in-flight PRs rebase onto `master`, they inherit all three of the above (batched write, flake fix, JasperFx 2.34.0) for free.

If for record-keeping you'd rather this PR show as *merged* than *closed*, note that the squash landed on `master` outside this PR's merge button, so GitHub can't retroactively mark it merged — but the authorship on `334283ee7` already credits Anne, so closing with this comment is the accurate resolution.

_Reviewed and reconciled against `origin/master` @ `334283ee7`._

PR #5018, opened 2026-07-22
feat: Cache LINQ query plans per filter shape (fixes #5013)

### LINKED ISSUE #5013: Cache LINQ query plans per filter shape — compiled-query performance for conditional Where chains
## Motivation

Compiled queries (`ICompiledQuery` / `ICompiledListQuery`) remove the LINQ expression parsing cost per execution, and with `UseMartenCompiledQueryResultPolicy` they compose beautifully with raw JSON streaming in Wolverine.HTTP. But they freeze the query *shape* at plan time, which makes them unusable for the most common real-world endpoint: a list with optional filters.

Every overview endpoint in our claims platform looks like this:

```csharp
IQueryable<PendingClientClaimLine> query = session.Query<PendingClientClaimLine>();
if (vanaf.HasValue)                       query = query.Where(x => x.Begindatum >= vanafAsDateOnly);
if (totEnMet.HasValue)                    query = query.Where(x => x.Begindatum <= totEnMetAsDateOnly);
if (bsn.HasValue)                         query = query.Where(x => x.ClientBsn == bsn);
else if (!string.IsNullOrWhiteSpace(client)) query = query.Where(x => x.ClientNaam.VolledigeNaam.NgramSearch(client));
if (prestatieId.HasValue)                 query = query.Where(x => x.PrestatieId == prestatieId);
```

Five optional filters → up to 2^5 distinct shapes, but each *individual* shape is completely static: same expression trees, same operators, only parameter values differ. Today every request re-parses and re-plans the whole expression. Encoding this as a compiled query is impossible (`myParam == null || ...` guards would be baked in at plan compilation), so these endpoints are locked out of the compiled-query tier entirely.

## Proposal

An opt-in caching layer that recognizes recurring query shapes and reuses their compiled plans:

```csharp
var opts = new StoreOptions();
opts.Linq.QueryPlanCache = QueryPlanCache.PerShape(maxEntries: 1024);
```

Sketch:
1. Before full LINQ compilation, compute a **shape key**: a structural hash of the expression tree that treats captured closure values/constants as parameter *slots* (type + position) rather than values. The conditional-`Where` pattern above yields one key per filter combination.
2. Cache hit → reuse the compiled SQL + parameter-extraction delegates; only re-bind the slot values from the new expression's closures. Miss → compile as today, store.
3. Bounded LRU; anything that defeats safe parameterization (dynamically built member access, non-parameterizable constructs like the explicit OR-chain we build for `id IN (...)` workarounds) simply never caches — correctness identical to today, caching is purely a fast path.

Explicit opt-in per query is a fine first cut if global shape-hashing is too aggressive:

```csharp
var lines = await query.ToListAsync(cancellation, QueryPlan.Cached);
```

## Why in Marten and not user land

The shape key and the value-slot extraction need the internals of Marten's LINQ-to-SQL compilation (which subtrees become parameters, which become SQL structure). Outside Marten you can only cache whole result sets, not plans.

## Expected impact

For hot list endpoints the LINQ parse/plan step is the dominant CPU cost after raw streaming removes serialization (the SQL itself is index-backed and fast). Shape caching gives conditional-filter endpoints the same per-request profile as hand-written compiled queries — without contorting them into 2^N compiled-query classes or forfeiting streaming.

## Prior art

EF Core does exactly this transparently (query cache keyed on expression shape with parameter extraction); the concept is proven, the ask is the Marten/Weasel equivalent.

erdtsieck:
## Summary

Implements #5013: an **opt-in** query plan cache that reuses compiled LINQ plans across ad-hoc queries sharing the same structural "shape" (e.g. conditional `Where()` chains behind optional filters) but differing only in the captured filter values. This lets endpoints with dynamic filter combinations get compiled-query-like performance without giving up `IQueryable` ergonomics.

Correctness is unaffected for every existing code path -- the cache is entirely additive and disabled by default.

## Design

1. **`ExpressionShapeVisitor`** (`src/Marten/Linq/Caching/`) walks the expression tree and computes a stable SHA-256 shape key. It only understands a narrow, allow-listed subset -- `Where`, `OrderBy`/`OrderByDescending`, `ThenBy`/`ThenByDescending`, `Skip`, `Take`, `Select`, plus simple comparisons/member access. Anything else (`Include()`, `Stats()`, custom SQL, `StartsWith()`, `GroupBy`, `SelectMany`, etc.) flags the shape as unsupported, and it's never cached -- falls straight back to today's behavior. While walking, each closure-captured/constant sub-expression is recorded as a positional "slot" whose value may vary between calls.
2. **`LinqPlanRecorder`** builds the cached plan on a miss: it replaces every slot with a distinct sentinel value (reusing the same internal `UniqueValueSource`/`QueryCompiler.Finders` machinery that already backs `ICompiledQuery`), compiles that template exactly once through the normal `LinqQueryParser` pipeline, and matches the resulting `NpgsqlBatch`'s parameters back to slots purely by sentinel-value identity. **If any parameter can't be attributed to a slot** (e.g. a multi-tenant `tenant_id`), the shape is rejected outright and never cached -- this is the key correctness guard against replaying a stale/foreign value.
3. On a **cache hit**, the template `NpgsqlBatch` is cloned with fresh parameter values pulled from the new expression's slots (`NpgsqlParameter.Value` is mutable, unlike the underlying Weasel `CommandParameter` that produced the original SQL). This skips LINQ parsing and SQL generation entirely and jumps straight to execution with the cached `IQueryHandler`.
4. **`QueryPlanCache`** is a bounded, thread-safe, FIFO-approximate-LRU cache (`ConcurrentDictionary` + insertion-order `ConcurrentQueue`).

## Configuration

```csharp
var opts = new StoreOptions();
opts.Linq.QueryPlanCache = QueryPlanCache.PerShape(maxEntries: 1024);
// or:
opts.Linq.EnableQueryPlanCaching(maxEntries: 1024);
```

Per-query opt-in:

```csharp
var lines = await query.ToListAsync(cancellation, QueryPlanCaching.Cached);
```

### Naming deviation from the issue

The issue's example used `QueryPlan.Cached`, but `Marten.Linq.QueryPlan` already exists as the Postgres `EXPLAIN` plan model -- overloading it with a `.Cached` static member would be confusing. This PR introduces a distinct `QueryPlanCaching` enum (`Default` / `Cached`) instead.

## Scope (MVP)

Only the `ToListAsync<TResult>(CancellationToken, QueryPlanCaching)` terminal operator is wired up to the cache in this pass, matching the issue's own example. `CountAsync`, `AnyAsync`, streaming (`ToAsyncEnumerable`), etc. are left uncached for now -- the underlying `ExpressionShapeVisitor` / `QueryPlanCache` / `LinqPlanRecorder` infrastructure is generic enough to extend to those later.

Sentinel-value generation currently supports `string`, `Guid`, `int`, `long`, `float`, `decimal`, `DateTime`, `DateTimeOffset`, enums, and arrays of those (reusing `QueryCompiler.Finders`). Types without a registered finder (e.g. `bool`) simply never get cached -- a safe MVP limitation, not a correctness risk.

## Documentation

Added a "Query Plan Cache for Ad Hoc Linq Queries" section to `docs/documents/querying/compiled-queries.md` covering configuration, shape-key mechanics, and current scope. Verified locally with `markdownlint-cli` and `cspell` per repo conventions -- both clean.

## Tests

New `src/LinqTests/query_plan_cache_Tests.cs` covers:
- Same shape / different values -> same cache key
- Different shapes -> different keys
- Different filter combinations (e.g. `Where` alone vs. `Where` + `OrderBy`) -> different keys
- Unsupported shapes (e.g. `StartsWith`) are correctly flagged as not cacheable
- Cached plan produces identical results to the uncached path
- Cache respects `maxEntries`
- Default (disabled) behavior requires explicit opt-in; cache stays empty until enabled

## Verification

- `dotnet build` on `src/Marten`, `src/LinqTests`, and `src/CoreTests` -- no errors.
- Full `LinqTests` suite (single TFM, net10.0) run against Postgres: **1405/1405 passed**, including the 7 new tests, with zero regressions vs. baseline.

### DIFF
--- a/docs/documents/querying/compiled-queries.md
+++ b/docs/documents/querying/compiled-queries.md
@@ -779,7 +779,7 @@ public async Task use_as_batch()
 <sup><a href='https://github.com/JasperFx/marten/blob/master/src/DocumentDbTests/Reading/query_plans.cs#L34-L71' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_using_query_plan_in_batch_query' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
-## Query Plan Cache for Ad Hoc Linq Queries
+## Query Plan Cache for Ad Hoc Linq Queries <Badge type="tip" text="9.18" />
 
 ::: info
 Not to be confused with the `IQueryPlan<T>` "Specification" pattern described above, or with the Postgres `EXPLAIN`
--- a/docs/documents/querying/compiled-queries.md
+++ b/docs/documents/querying/compiled-queries.md
@@ -778,3 +778,89 @@ public async Task use_as_batch()
 ```
 <sup><a href='https://github.com/JasperFx/marten/blob/master/src/DocumentDbTests/Reading/query_plans.cs#L34-L71' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_using_query_plan_in_batch_query' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
+
+## Query Plan Cache for Ad Hoc Linq Queries
+
+::: info
+Not to be confused with the `IQueryPlan<T>` "Specification" pattern described above, or with the Postgres `EXPLAIN`
+output returned by [`ExplainAsync()`](/diagnostics#previewing-the-postgresql-query-plan). This is a distinct, opt-in
+performance feature for ordinary `IQueryable<T>` Linq queries.
+:::
+
+`ICompiledQuery` gives you the best raw performance by completely bypassing Linq expression parsing, but it comes at
+the cost of a fixed query shape -- every compiled query class represents exactly one filter/sort/paging combination.
+That is a poor fit for endpoints with _optional_ or _conditional_ filters, where the same handler builds up a
+different `Where()` chain depending on which query string parameters were supplied. Historically, every distinct
+combination of filters on that kind of endpoint paid the full cost of Linq expression parsing and SQL generation on
+every single call, because each combination is technically a different Linq expression tree.
+
+The query plan cache closes that gap. It is an opt-in, bounded cache that recognizes when two Linq queries share the
+same _structural shape_ -- the same sequence of `Where()`, `OrderBy()`/`OrderByDescending()`, `ThenBy()`/
+`ThenByDescending()`, `Skip()`, `Take()`, and `Select()` calls -- even though the actual filter values differ. On a
+cache hit, Marten skips Linq parsing and SQL generation entirely and just substitutes the new filter values into the
+previously-built database command.
+
+### Enabling the cache
+
+The cache is disabled by default. Turn it on for the whole `DocumentStore` with either of these equivalent options:
+
+```cs
+var opts = new StoreOptions();
+
+// Full control over the maximum number of cached shapes
+opts.Linq.QueryPlanCache = QueryPlanCache.PerShape(maxEntries: 1024);
+
+// Or the shorthand version, also with a default of 1024 entries
+opts.Linq.EnableQueryPlanCaching();
+```
+
+Once enabled, opt individual Linq queries into the cache by passing `QueryPlanCaching.Cached` to `ToListAsync()`:
+
+```cs
+IReadOnlyList<Invoice> lines = await session
+    .Query<Invoice>()
+    .Where(x => x.CustomerId == customerId)
+    .Where(x => status.HasValue ? x.Status == status.Value : true)
+    .OrderBy(x => x.CreatedAt)
+    .ToListAsync(cancellation, QueryPlanCaching.Cached);
+```
+
+Every call to that code with a different `customerId` and/or `status` reuses the same cached plan the second and
+subsequent times it's called, as long as which `Where()` clauses are present (not their values) stays the same.
+
+### How shape keys work
+
+Behind the scenes, Marten walks the Linq expression tree for the query and builds a structural "shape key":
+
+* Method calls, operators, and member names contribute to the key by their _name_ and _position_ in the tree, never
+  by value.
+* Anything that looks like a captured local variable, method parameter, or literal constant -- i.e., the actual
+  filter values -- is treated as a positional "slot" rather than part of the shape. Two queries with the identical
+  Linq structure but different slot values always produce the same shape key.
+* The key is scoped by both the source document type and the requested result type, so two structurally identical
+  queries against different document types never collide.
+
+Only a deliberately narrow, safe subset of Linq is recognized as cacheable: `Where`, `OrderBy`, `OrderByDescending`,
+`ThenBy`, `ThenByDescending`, `Skip`, `Take`, and `Select`, built from simple comparisons and member access. Anything
+outside that -- `Include()`, `Stats()`, raw/custom SQL, `StartsWith()`/`Contains()`, `GroupBy()`, `SelectMany()`, and
+so on -- is automatically recognized as unsupported for caching. Unsupported queries are not an error; they simply
+execute through Marten's normal, always-correct Linq pipeline exactly like they do today, with no caching applied.
+
+::: tip
+Correctness always wins over caching. If Marten can't fully account for every parameter that will end up in the
+generated SQL command for a shape -- for example, a conjoined multi-tenancy `tenant_id` parameter, which comes from
+the session rather than from the query expression -- that shape is simply never cached rather than risk ever
+replaying a stale or unrelated value.
+:::
+
+### Bounded cache size
+
+`QueryPlanCache.PerShape(maxEntries)` bounds the number of distinct shapes that will be cached at once. Once that
+limit is reached, the oldest cached shapes are evicted first to make room for new ones, so the cache can't grow
+without bound even for endpoints that produce a very large number of distinct filter combinations.
+
+### Current scope
+
+As of this writing, the query plan cache only applies to the `ToListAsync<TResult>(CancellationToken,
+QueryPlanCaching)` overload. Other terminal operators such as `CountAsync()`, `AnyAsync()`, and streaming queries do
+not yet participate in the cache.
--- a/src/LinqTests/query_plan_cache_Tests.cs
+++ b/src/LinqTests/query_plan_cache_Tests.cs
@@ -1,0 +1,155 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Marten;
+using Marten.Linq;
+using Marten.Linq.Caching;
+using Marten.Testing.Documents;
+using Marten.Testing.Harness;
+using Shouldly;
+
+namespace LinqTests;
+
+public class query_plan_cache_Tests: OneOffConfigurationsContext
+{
+    [Fact]
+    public void same_shape_different_values_produces_same_key()
+    {
+        int number1 = 1;
+        int number2 = 2;
+
+        var (expr1, _) = whereExpression(number1);
+        var (expr2, _) = whereExpression(number2);
+
+        var shape1 = ExpressionShapeVisitor.Analyze(expr1);
+        var shape2 = ExpressionShapeVisitor.Analyze(expr2);
+
+        shape1.IsSupported.ShouldBeTrue();
+        shape2.IsSupported.ShouldBeTrue();
+
+        shape1.BuildKey(typeof(Target), typeof(Target))
+            .ShouldBe(shape2.BuildKey(typeof(Target), typeof(Target)));
+    }
+
+    [Fact]
+    public void different_shapes_produce_different_keys()
+    {
+        var (whereExpr, _) = whereExpression(1);
+        var (skipTakeExpr, _) = skipTakeExpression(1, 2);
+
+        var shape1 = ExpressionShapeVisitor.Analyze(whereExpr);
+        var shape2 = ExpressionShapeVisitor.Analyze(skipTakeExpr);
+
+        shape1.IsSupported.ShouldBeTrue();
+        shape2.IsSupported.ShouldBeTrue();
+
+        shape1.BuildKey(typeof(Target), typeof(Target))
+            .ShouldNotBe(shape2.BuildKey(typeof(Target), typeof(Target)));
+    }
+
+    [Fact]
+    public void different_filter_combinations_produce_different_keys()
+    {
+        var (whereOnlyExpr, _) = whereExpression(1);
+        var (whereAndOrderExpr, _) = whereAndOrderExpression(1);
+
+        var shape1 = ExpressionShapeVisitor.Analyze(whereOnlyExpr);
+        var shape2 = ExpressionShapeVisitor.Analyze(whereAndOrderExpr);
+
+        shape1.BuildKey(typeof(Target), typeof(Target))
+            .ShouldNotBe(shape2.BuildKey(typeof(Target), typeof(Target)));
+    }
+
+    [Fact]
+    public void unsupported_shapes_are_flagged()
+    {
+        IQueryable<Target> queryable = new Target[0].AsQueryable();
+        var expr = queryable.Where(x => x.String.StartsWith("A")).Expression;
+
+        var shape = ExpressionShapeVisitor.Analyze(expr);
+
+        shape.IsSupported.ShouldBeFalse();
+    }
+
+    [Fact]
+    public async Task cached_plan_produces_same_results_as_uncached()
+    {
+        StoreOptions(_ => _.Linq.EnableQueryPlanCaching());
+
+        using var session = theStore.LightweightSession();
+        session.Store(new Target { Number = 1, String = "one" });
+        session.Store(new Target { Number = 2, String = "two" });
+        session.Store(new Target { Number = 3, String = "three" });
+        await session.SaveChangesAsync();
+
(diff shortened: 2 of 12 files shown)


jeremydmiller:
## Deferred to the [9.19 milestone](https://github.com/JasperFx/marten/milestone/122) — not taking this into 9.18

Thanks @erdtsieck — the design here is genuinely thoughtful and I want to take it, just not rushed into the next release. A batch-cloning / parameter-swapping cache is exactly the kind of feature that needs to bake, so parking it for 9.19 to harden correctness and coverage first.

### What's solid (confirmed by reading)

- **The tenant-bleed guard is real**: `LinqPlanRecorder` rejects (never caches) any shape where a batch parameter can't be attributed to a captured slot by sentinel-value identity, so a `tenant_id` (or any non-slot param) forces the shape uncacheable.
- **Thread-safety looks right**: `RebindValues` builds a *fresh* `NpgsqlBatch` per hit and only reads the shared template — it doesn't mutate cached state.
- Clean opt-in surface (default-off, distinct `QueryPlanCaching` enum, MVP-scoped to the one `ToListAsync` overload).

### What 9.19 needs before merge

**Test coverage is thin relative to the risk.** The current 7 tests cover the happy path but not the dangerous ones. Before this ships it needs:
- **Multi-tenant tests** — the single most important guard (reject-if-`tenant_id`-unattributable) is completely untested. Add: a multi-tenant shape is not miscached, and two tenants never bleed.
- **Concurrency test** — validate the fresh-batch-per-hit design under parallel load on a shared cached plan.
- **Literal-constant test** — `x.Status == "Active"` vs `"Inactive"`: confirm a changed *literal* yields a different key/result and can't stale.
- **Reused captured value** — `x.A == v || x.B == v`.
- **Type variety end-to-end** — only `int` (`Number`) is exercised today; add string / Guid / DateTime / enum / array filters (these are also where `RebindValues` dropping `NpgsqlDbType`/type metadata could mis-bind).

**Correctness items under review** (I'm completing an adversarial pass; will follow up with specifics): sentinel-vs-real-value collision escaping the rejection guard, shape-key completeness (every SQL-affecting dimension hashed), literal-vs-slot handling, `RebindValues` type-metadata fidelity for arrays/enums/jsonb, and eviction races in the `ConcurrentDictionary` + `ConcurrentQueue` bound.

Milestone: **9.19**. Not a rejection — a "let's get the correctness envelope nailed down first."

jeremydmiller (changes requested):
Thanks for this — the architecture is solid and I checked the two things I worried about most, both of which are **safe**:

- **No hot-path regression when disabled.** The existing `ToListAsync<T>(token)` path is untouched; the cache is reachable only via the new `ToListAsync<T>(token, QueryPlanCaching.Cached)` overload, and `ExecuteListWithPlanCacheAsync` early-outs on `!cache.Enabled` (the default `QueryPlanCache.Disabled`) before any shape analysis. 👍
- **No tenant bleed.** `CurrentTenantFilter` emits `tenant_id` as a real parameter carrying the actual tenant value (never a sentinel), so the recorder's "every batch parameter must map to a captured sentinel slot, otherwise refuse to cache" guard (`LinqPlanRecorder`) correctly rejects conjoined-tenant shapes. Concurrency is fine too — `RebindValues` clones into a fresh `NpgsqlBatch` and only reads the template.

## Blocker before this can merge: null filter values return wrong (empty) results on a cache hit

Marten generates **structurally different SQL** depending on whether a comparison value is null. See `QueryableMember.CreateComparison` (`src/Marten/Linq/Members/QueryableMember.cs:143-153`):

```csharp
if (unwrappedValue == null)
{
    return op == "=" ? new IsNullFilter(this) : new IsNotNullFilter(this);   // "col is null" — NO parameter
}
var def = new CommandParameter(Expression.Constant(unwrappedValue));
return new MemberComparisonFilter(this, def, op);                            // "col = $1" — a parameter
```

`SimpleExpression.CompareTo` does the same for the captured-member path. The plan template is always recorded with **non-null sentinel values** (`UniqueValueSource` yields GUIDs / negative ints), so a cached plan for `.Where(x => x.Name == name)` always bakes the `col = $1` form. On a later cache hit where `name` is null, `RebindValues` binds `DBNull.Value`, producing `col = NULL`, which matches nothing in Postgres — the correct result is the `IS NULL` rows. The `!=`/null case is symmetric (`col != NULL` instead of `IS NOT NULL`).

Repro shape:

```csharp
// first call caches the plan with a non-null value
await q.Where(x => x.Name == "foo").ToListAsync(t, QueryPlanCaching.Cached);
// cache hit — returns [] instead of the rows where Name is null
await q.Where(x => x.Name == null).ToListAsync(t, QueryPlanCaching.Cached);
```

The shape key can't distinguish these — the null-ness is a runtime value of a captured slot, so both calls hash to the same shape yet need different SQL. Optional/nullable filters are exactly the "conditional filter endpoint" scenario this feature targets, so this is a common, silent, wrong-results path.

**Suggested fix (simplest safe option):** on a cache hit, if any current slot value is null, bypass the cache and fall through to the uncached path. A regression test with a nullable filter (one non-null call priming the plan, then a null call) would lock it in.

## Minor (non-blocking)
- The flagship doc example in `docs/documents/querying/compiled-queries.md` uses `status.HasValue ? ... : true`, a `Conditional`, which `ExpressionShapeVisitor` flags unsupported — so that example never actually caches. Worth switching the doc to an example that does.

I'm holding this out of the 9.19.0 release for now because of the null-value blocker; the feature is opt-in and not yet merged, so it doesn't gate the release. Happy to re-review as soon as the null path is handled.

jeremydmiller:
Status update: **holding this until the next release cycle** rather than closing it. The mechanism is sound in outline and the code is careful, but we are not willing to take on a permanent performance feature whose central claim — that it is faster — has not been measured. Below is the benchmark plan we would want to see, so this can be picked back up with something concrete.

## Why this is the blocker

The seven tests here are all correctness and shape-key behaviour: same/different shape keys, unsupported-shape flagging, cached-equals-uncached results, max-entries, opt-in gating. All useful, none of them measure time or allocations.

That matters more than usual because of what the cache costs on the way in. Every consulted query pays a full expression-tree walk, `StringBuilder` accumulation, a UTF-8 encode, and a `SHA256.HashData(...)`. So one tree walk (Linq parsing) is traded for another tree walk plus a cryptographic hash plus a dictionary lookup; the saving is SQL generation and handler construction, not traversal. On a miss you pay all of that *plus* the sentinel-template recording pass. With a bounded FIFO and diverse shapes, a store could pay the overhead on nearly every call and never gain.

Incidentally, SHA-256 looks like the wrong tool for a process-local dictionary key — there is no security boundary here, and a non-cryptographic hash (or the shape string itself) would be materially cheaper. Worth revisiting once the numbers exist.

## The benchmark plan

`src/MartenBenchmarks/LinqActions.cs` already has a `CreateLinqCommand` benchmark with `[MemoryDiagnoser]` attached, measuring exactly the cost this PR targets. Extending it is the natural home.

**Measure both levels.** This is the part most likely to decide the feature:

1. **Command construction only** — the existing `CreateLinqCommand` shape. Shows the raw translation saving.
2. **Full round trip against Postgres.** If a query costs, say, 500µs at the database, saving 20µs of translation is ~4% and the feature is hard to justify. The construction-only number will flatter the cache; the end-to-end number is the one that decides it.

**Arms to compare:**

| Arm | What it tells us |
| --- | --- |
| Ad-hoc Linq, cache disabled | today's baseline |
| Cache hit — same shape, different values | the actual win |
| Cache miss — shape never seen | worst case: hash + record pass, no benefit |
| **Unsupported shape** (e.g. the `status.HasValue ? ... : true` conditional) | **pure overhead, zero benefit** |
| `ICompiledQuery` | the floor — how much is even available to win |
| Ad-hoc + Npgsql `Max Auto Prepare` | how much is recoverable for free |

The unsupported-shape arm is not a footnote. The flagship example in this PR's own documentation is `.Where(x => status.HasValue ? x.Status == status.Value : true)`, and `ExpressionShapeVisitor`'s default case reads `// Anything else (conditional, indexers, ...) is out of scope`. So the documented headline scenario is one the cache silently refuses, and anyone who follows the docs gets the overhead and none of the benefit. That arm measures what that costs them.

The last arm matters too. Marten currently has no reference to `Max Auto Prepare` or `Auto Prepare Min Usages` anywhere in `src/` or `docs/`. Since Marten already parameterises everything, SQL text is stable while filter values vary — exactly what Npgsql's automatic preparation keys on. It is a connection-string setting with no Marten code and no correctness surface, and it helps every query path rather than an allow-listed subset. If it recovers most of the same win, that reframes this PR considerably. (It is not free of caveats: prepared statements are per-connection and cost server memory, they break under PgBouncer transaction pooling unless configured, and Postgres may switch to a generic plan after five executions, which can be *worse* on skewed data. Worth measuring, not assuming.)

**Also worth capturing:** allocations per query on the hit path (the `NpgsqlBatch` clone is not free), and behaviour under shape churn — N distinct shapes against a cache of size N/2, to show what eviction thrashing costs.

## What would justify merging

- The ad-hoc → `ICompiledQuery` gap is large enough **end to end** to be worth closing at all.
- The cache hit recovers a worthwhile share of that gap end to end, not just in command construction.
- Overhead on the miss and unsupported paths is small enough that enabling it store-wide cannot regress a workload whose shapes mostly do not cache.

If the first point does not hold, that is a useful result in itself and we would close this without prejudice — it would also tell us something about how strongly the docs should keep recommending compiled queries.

## Independent of the benchmarks

Two things need fixing regardless, both from the earlier review:

- **The null-filter correctness bug.** Marten emits `col is null` (no parameter) versus `col = $1` structurally, so a template recorded with non-null sentinels always bakes `col = $1`; a later hit with a null value rebinds to `col = NULL` and matches nothing. Optional filters going null is precisely the target scenario. Simplest fix is to bypass the cache on a hit when any slot value is null, plus a nullable-filter regression test.
- **The doc example** must use a shape the visitor actually supports.

This branch is also now `CONFLICTING` against master and will need a rebase — note that #5163, #5164 and #5165 have all touched adjacent daemon/progression code, though nothing in the Linq path.

Thanks for your patience on this one. The deferral is about evidence, not about the quality of the work.

jeremydmiller:
Closing this, and I want to give you a real answer rather than a third deferral — you've put careful work into this and you're owed the reasoning.

We benchmarked it. BenchmarkDotNet, local Postgres, two worktrees (master vs this branch), same queries on both sides.

## What a plan cache can possibly save

All this removes is expression parsing plus SQL generation, so that is the ceiling:

| scenario | parse + SQL gen | full query | parse share |
|---|---|---|---|
| unindexed JSONB scan, 10 rows | 3.54 us | 40,102 us | 0.009% |
| indexed column, 1 row — the fastest realistic Marten query | 1.19 us | 430 us | 0.28% |

## Before and after, on this branch

Everything below runs on the indexed single-row setup, deliberately, because that is where per-call overhead is the largest share it can be:

| | mean | allocated |
|---|---|---|
| plain ad-hoc, cache off | 316.3 us ±34.9 | 25.18 KB |
| **cache HIT**, cacheable shape, warm | 335.3 us ±41.8 | **26.32 KB** |
| opted in, conjoined multi-tenant | 353.6 us ±55.8 | **39.81 KB** |
| multi-tenant control, no opt-in | 295.2 us ±28.6 | 25.92 KB |

**Every timing pair overlaps within noise.** That is the heart of it: the quantity being optimized is 1.19 us, and the run-to-run standard deviation on the operation is 28–56 us — 25 to 45 times larger than the entire thing being removed. The improvement is not small, it is below the measurement floor, on localhost, on the shape most favorable to it. Over a real network it only shrinks.

Allocation is the clean signal, since it carries no timing jitter, and it goes the wrong way: **a cache hit allocates 4.5% more than no cache at all**, for nothing measurable back.

## The multi-tenant result

This is the one we did not anticipate. `LinqPlanRecorder.TryBuild` refuses any shape carrying a batch parameter it cannot attribute to a captured slot — correctly, it is the guard against replaying a foreign value — and `tenant_id` is never a slot. Instrumenting the cache directly: after 30 warmup iterations the plain store holds **1** entry and the conjoined multi-tenant store holds **0**.

But `ExecuteListWithPlanCacheAsync` still walks the expression, SHA-256s a key, builds a full template through `LinqQueryParser`, discards it, and *then* falls through to `ExecuteListAsync`, which parses again. Two parses per query, forever, zero hits — that is the +54% allocation. There is no negative cache, so it never learns that this shape will never be cacheable.

## Where that leaves it

Combined with the null-filter correctness issue from the earlier review (the template bakes `col = $1`; a hit with a null value rebinds to `col = NULL` and silently returns nothing — exactly the optional-filter case this targets), and the documented flagship example using a `Conditional` that `ExpressionShapeVisitor` refuses outright, we do not think there is something here to unlock by iterating.

For context on the alternative: the same query through an `ICompiledQuery` measured 364 us against 430 us ad-hoc — also inside the noise band. Both routes say the same thing from opposite directions, which is that LINQ parsing is simply not where an ad-hoc Marten query spends its time.

None of this is a knock on the implementation. The tenant-attribution guard was right, the batch-cloning was concurrency-safe, and the default-off gating was clean. The mechanism worked; the thing it optimizes turned out not to be a cost worth removing.

Closing #5013 alongside this. Thank you for the work and for your patience through two deferrals — and please do keep bringing performance ideas.

Issue #5041, opened 2026-07-24
Natural key handling issues


ytqsl:
2 issues discovered while testing rebuilding snapshots with natural keys:
1. conventional static handlers with IEvent signature are not discovered by the rebuild
2. the old natural key is not cleaned up from the table
3. natural key source handlers do not receive the snapshot current state when invoked

repro PR: https://github.com/JasperFx/marten/pull/5042

PR #5042, opened 2026-07-24
repro for https://github.com/JasperFx/marten/issues/5041

ytqsl:


jeremydmiller:
This was superseded by 5041 with the fix and co-authorship by @ytqsl.

Thanks for the reproduction here!

Issue #569, opened 2016-10-20
GroupBy support in Linq queries


jeremydmiller:
No real plans about how far to go with this yet or timing.

AKlaus:
`Queryable.GroupBy()` throws an error:
```
Could not determine an 'id/Id' field or property for requested document type System.Linq.IGrouping`2
   at Marten.Schema.DocumentMapping.Validate()
   at Marten.Storage.DocumentTable..ctor(DocumentMapping mapping)
   at Marten.Schema.DocumentMapping.<toSchemaObjects>d__129.MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Marten.Storage.Tenant.generateOrUpdateFeature(Type featureType, IFeatureSchema feature)
   at Marten.Storage.Tenant.ensureStorageExists(IList`1 types, Type featureType)
   at Marten.Linq.MartenQueryable`1.executeAsync[TResult](Func`2 source, CancellationToken token)
   at Marten.Linq.MartenQueryable`1.ToListAsync[TResult](CancellationToken token)
   at Marten.QueryableExtensions.ToListAsync[T](IQueryable`1 queryable, CancellationToken token)
```

Therefore:
- Could you please confirm that there is no option of using `GROUP BY` except manually writing an SQL query?
- Are there plans to implement the functionality? (looks unlikely, as it's marked for 1.3 milestone and the current version is 2.4).
- RavenDb provides this functionality via [Map-Reduce indexes](https://ravendb.net/docs/article-page/4.0/http/indexes/map-reduce-indexes). Is there a different (or recommended) approach to reduce data by grouping on a field set?

P.S. I understand that Marten is an open source project and public contribution is more than welcome, but I was very surprised to see lack of grouping functionality.

jeremydmiller:
@AKlaus You're commenting on the issue in the backlog to add this feature, it's not in the docs, and a very quick search would have shown you that it doesn't appear in any unit test in the codebase.

So,

1. Marten does not yet support GroupBy in Linq
2. Yes, obviously by the existence of this issue
3. Good for them, and no, not yet

"I was very surprised to see lack of grouping functionality" -- please don't ever use a phrase like that in an OSS issue. 

[...]

Issue #565, opened 2016-10-18
Multi-document Select() operations


jeremydmiller:
See @mdissel 's comments on https://github.com/JasperFx/marten/issues/396.

What if we had some mechanism to do a Linq `Select()` using data from related documents?

jeremydmiller:
Calling this a duplicate with #1599

PR #5054, opened 2026-07-26
Adopt JasperFx.Events 2.36.0: shard failure classification, drain timeout docs, natural key extraction

### LINKED ISSUE #5041: Natural key handling issues

### LINKED ISSUE #5047: Docs: DaemonSettings.StopAndDrainTimeout (jasperfx#564)
### Summary

JasperFx/jasperfx#564 adds a new async daemon knob that needs documentation in the projections/daemon docs:

```csharp
opts.Projections.StopAndDrainTimeout = 30.Seconds(); // default is 5 seconds
```

`DaemonSettings.StopAndDrainTimeout` (also on `IReadOnlyDaemonSettings`) bounds how long the daemon waits for a **single** subscription or projection shard to gracefully finish its in-flight page and flush its progression when that agent is stopped. It applies to all three stop paths: `StopAgentAsync`, `StopAllAsync` (the SIGTERM/shutdown path), and the internal stop-if-running agent replacement. The default is 5 seconds, which is exactly what was hardcoded before, so nothing changes unless it's configured.

### What the docs should say

- **What it bounds.** `StopAndDrainAsync` is the *graceful* path — it lets the in-flight page/batch complete and flushes the progression row. Exceeding the timeout cancels that drain mid-flight, which abandons the progression flush.
- **Why you'd raise it.** If the drain is cut off before progression is flushed, the shard restarts against a stale progression row and throws `ProgressionProgressOutOfOrderException` on the next start. Raise it when in-flight batches legitimately take longer than 5 seconds: large `BatchSize`, expensive projections, heavy rebuild load, or a slow/contended database. This is most visible on shutdown of a host with a large agent universe — e.g. database-per-tenant with thousands of (projection × tenant) shards draining inside a Kubernetes termination grace window (the motivating incident behind jasperfx#564).
- **Pair it with the host's shutdown grace period.** The per-shard bound is only useful if the process actually lives long enough to use it, so a raised value should be matched by `HostOptions.ShutdownTimeout` and the Kubernetes `terminationGracePeriodSeconds`.
- **Why you'd lower it.** A host that would rather cut a wedged shard loose quickly and take the progression replay hit — for instance to keep node-failover/reassignment latency low — can set it below the default.
- **Opting out.** `Timeout.InfiniteTimeSpan`, or any non-positive value, removes the separate bound; the drain is then limited only by the daemon's own cancellation. Worth calling out that this means a genuinely wedged shard can hold up shutdown indefinitely.

### Notes

- Docs only — no code change needed here; the setting arrives with the JasperFx.Events version carrying jasperfx#564.

### LINKED ISSUE #5048: Implement IEventFailureContext + extended progression failure columns (jasperfx#565)
### Summary

JasperFx/jasperfx#565 (PR: JasperFx/jasperfx#567) adds a classified pause/stop reason for projection and subscription shards — `ShardFailure` — so an external supervisor (Wolverine's assignment plane, CritterWatch) can finally say *why* a shard paused instead of only that it did. Two halves of that land in **Marten**, and they need to release together with the JasperFx.Events version that carries the new types.

There is deliberately **no fallback** in the daemon (no type-name sniffing): a store's exception declares its own category, so until these changes ship, a serialization failure in Marten classifies as `ShardFailureCategory.Other` with no event details.

### Part 1 — implement `IEventFailureContext` on the read-side exceptions

```csharp
public interface IEventFailureContext
{
    ShardFailureCategory Category { get; }   // the exception declares its own kind
    long Sequence { get; }                   // the only guaranteed member
    string? EventTypeName { get; }
    Guid? EventId { get; }
    Guid? StreamId { get; }
    string? StreamKey { get; }
    string? TenantId { get; }
    long? Version { get; }
}
```

`ShardFailure.For(exception, occurredAt)` walks the whole exception graph (inner exceptions and `AggregateException.InnerExceptions`) looking for one of these, so it works even when the exception arrives wrapped.

- **Deserialization failure** (`EventDeserializationFailureException` in Marten; the equivalent in Polecat) → `Category => ShardFailureCategory.EventSerialization`. It already carries `Sequence`. It is handed an `IEventType` and currently **drops it** — retain it so `EventTypeName` can be reported instead of living only inside the message string. Everything else can stay null: this is raised while reading a row, before there is an `IEvent`.
- **Unknown event type** (`UnknownEventTypeException`) → `Category => ShardFailureCategory.UnknownEventType`, and report the sequence if it is available at the throw site. Kept separate from serialization on purpose: a missing registration or a rollback is a different operator action than bad data.
- Anything else the store throws from the load path that can name a single event is fair game for the same treatment.

Marten note: `EventDeserializationFailureException.ToDeadLetterEvent` builds its `DeadLetterEvent` with an object initializer, so its `Id` stays `Guid.Empty` and is assigned by document identity generation. jasperfx now assigns `Id` in the `DeadLetterEvent` constructor (version 7, so ids stay time-ordered) precisely so the id is known before the background write lands — worth doing the same on this path.

### Part 2 — extended progression columns

`ShardState.Failure` now rides along on the paused/stopped states the daemon publishes, and `IEventDatabase.WriteExtendedProgressionAsync` documents the contract. The extended progression row already persists agent status, heartbeat, pause reason and running node (jasperfx#537 / #553); add the classified fields so a consumer polling the database — CritterWatch when the publishing node is *down*, which is exactly when this matters — sees the same reason an in-process observer does:

| column | source |
|---|---|
| `failure_category` | `ShardFailure.Category` (store the enum name, not the ordinal) |
| `failure_event_sequence` | `ShardFailure.Event?.Sequence` |
| `failure_event_type` | `ShardFailure.Event?.EventTypeName` |
| `failure_event_tenant_id` | `ShardFailure.Event?.TenantId` |

`ShardFailure.Detail` is exactly what `ShardState.PauseReason` has always carried, so the existing pause-reason column covers the text — no new column for it.

Contract points worth getting right:

- **Clear the failure columns** when a state arrives with `Failure == null` on a `ShardAction.Started` — a shard that recovered must stop reporting the reason it paused an hour ago, or every supervisor built on this alerts forever.
- Same best-effort rules as the rest of extended progression: never throw for a missing row, and **never** touch `last_seq_id`-equivalent progression from this path.
- Hydrate `ShardState.Failure` on the read side (`AllProjectionProgress` / `ProjectionProgressFor`) from those columns, so a poller gets the same shape as a live observer. `ShardFailure` is a record with `required` members — build it with `Category`, `ExceptionType`, `RootExceptionType`, `Message`, `Detail` and `OccurredAt` filled from what the row has (`Detail`/`Message` from the pause-reason column is fine).

### Acceptance

- A projection paused by a poison event reports `Category = ApplyEvent` with the failing sequence (this already works via `ApplyEventException`).
- A projection paused by a corrupted event body reports `Category = EventSerialization` with the sequence and event type — the case that motivated the issue.
- A recovered shard's persisted failure columns go back to null.

### LINKED ISSUE #5052: Update NaturalKeyProjection for the widened JasperFx.Events natural key extraction contract (jasperfx#569)
Follow-up to [JasperFx/jasperfx#569](https://github.com/JasperFx/jasperfx/issues/569) / PR [JasperFx/jasperfx#571](https://github.com/JasperFx/jasperfx/pull/571), which fixes the upstream half of #5041.

`NaturalKeyEventMapping.Extractor` changes from `Func<object /* event data */, object?>` to `Func<IEvent, object?>`. Widening it to the whole event is what makes an `IEvent<T>` `[NaturalKeySource]` handler bindable at all, and what lets a key be derived from event metadata.

Both Marten call sites already hold the real `IEvent`, so this is a two-line change:

- `NaturalKeyProjection.ApplyAsync` — `mapping.Extractor(@event.Data)` → `mapping.Extractor(@event)`
- `NaturalKeyProjection.QueueUpsertsForEvents` — same

Worth doing at the same time, since jasperfx#571 makes them possible/observable:

- the failing test from #5042, `bug_5041_natural_key_should_be_updated_during_rebuild_when_handler_is_ievent`, should pass once the pin is bumped
- an unbindable `[NaturalKeySource]` method now throws `InvalidProjectionException` from `AssembleAndAssertValidity()` instead of silently registering nothing — worth a pass over the Marten test suite for projections that were quietly relying on a natural key mapping that never existed
- `NaturalKeyFor(x => x.SetBy<TEvent>(...))` is now the supported explicit registration path and should be documented alongside `[NaturalKeySource]`

Blocked on a JasperFx.Events release carrying jasperfx#571.

Note: jasperfx#569 was written believing Marten was the only consumer. Polecat maintains its own `pc_natural_key_X` lookup table off the same `NaturalKeyDefinition` and needs the equivalent one-line change — tracked in [JasperFx/polecat#369](https://github.com/JasperFx/polecat/issues/369).

jeremydmiller:
Adopts JasperFx/JasperFx.Events 2.36.0 and lands the Marten-side issues that ride with it. One commit per issue, plus the bump.

Closes #5047. Closes #5048. Closes #5052. Closes #5041.

## Commits

| | |
|---|---|
| `Bump JasperFx/JasperFx.Events 2.35.0 -> 2.36.0` | jasperfx#564 + #565/#567 + #569/#571 |
| `docs(#5047)` | `DaemonSettings.StopAndDrainTimeout` |
| `feat(#5048)` | `IEventFailureContext` + extended progression failure columns |
| `fix(#5052)` | natural key extraction on the widened contract |
| `test(#5041)` | ties off the umbrella issue across all three reported items |

### Why the bump is not purely a version change

`NaturalKeyEventMapping.Extractor` widens from `Func<object, object?>` to `Func<IEvent, object?>`, which is source breaking. Marten does not compile on 2.36.0 without adapting `NaturalKeyProjection`'s two call sites, so those two lines ride in the bump commit to keep every commit buildable. The rest of #5052 is its own commit.

## #5047 — drain timeout docs

New *Graceful Shutdown and the Drain Timeout* section: what the timeout bounds (the in-flight page plus the progression flush, across all three stop paths), the `ProgressionProgressOutOfOrderException` that follows a drain cut short, pairing a raised value with `HostOptions.ShutdownTimeout` and `terminationGracePeriodSeconds`, why a deployment might instead lower it, and what opting out with `Timeout.InfiniteTimeSpan` costs. Docs only.

## #5048 — why a shard is down

**Part 1.** `EventDeserializationFailureException` declares `EventSerialization` and now retains the `IEventType` it was always handed but only used for the message string. `UnknownEventTypeException` declares `UnknownEventType` — kept separate on purpose, since a missing registration is a deployment fix rather than a data fix — and carries the sequence, threaded down from the row being read. `ToDeadLetterEvent` assigns its `Id` up front instead of leaving `Guid.Empty` for identity generation, so the creating process can correlate its `ShardFailure` with the row before that background write lands.

**Part 2.** Four columns behind the existing `EnableExtendedProgressionTracking` gate, read back onto `ShardState.Failure`. The category persists as the enum *name*, so reordering `ShardFailureCategory` can never silently re-label old rows. No column for the reason text — `ShardFailure.Detail` is exactly what `pause_reason` already carried.

The failure columns follow a different write rule from the rest of extended progression: written when the state carries a `Failure`, cleared on a `Started` that has none, and otherwise **left alone**. That last case is load-bearing — `SubscriptionAgent` publishes a plain `Stopped` right behind a `Paused`, and an unconditional write would erase the reason microseconds after recording it.

Both write overloads now share one `UPDATE ... FROM unnest` so that conditional lives in one place. This retires Marten's use of `mt_mark_event_progression_extended` rather than growing its signature, following the reasoning already recorded on the batch overload: no schema object to migrate. The function stays installed for anything calling it directly.

> [!NOTE]
> **This adds a schema migration.** Stores that already have `EnableExtendedProgressionTracking` on need `ApplyAllConfiguredChangesToDatabaseAsync`. The columns are nullable so no backfill is required, and until the migration runs the write degrades to no telemetry rather than failing anything — extended progression is best-effort by contract.

## #5052 — natural key extraction

The two skipped tests from #5042 could not simply be un-skipped. Widening to the whole event fixes one half; the other half is now an error **by design**, because the reporter's aggregate declares `required` members and jasperfx#571 stopped fabricating blank aggregates that C# itself would never have let you construct — that fabrication was the original `ArgumentNullException` out of `SaveChangesAsync`. So the file pins three things:

- an unbindable source **fails loudly**, naming the method, the reason, and both ways out. The original bug was that *nothing* happened — no mapping, no log, no error — so silence is the regression actually worth guarding;
- the rename works when the key comes from the event alone (a static `[NaturalKeySource]` returning the key type, taking `IEvent<T>` — #5042's failing case, and the strategy that could not bind before);
- the rename works for the reporter's own aggregate through the explicit `NaturalKeyFor(x => x.SetBy/SetByEvent(...))` registration.

Docs: the handler-signature section described one strategy and warned that `IEvent<T>` "is not currently supported here, and silently ignored". Replaced with the three ranked strategies, when the blank-aggregate fallback is skipped and why, the loud failure, and a new section for `NaturalKeyFor`.

## #5041 — the umbrella issue

Three defects were reported. Two are fixed in code; the third is a *cannot be supported here* that had never been written down.

| | Item | Where |
|---|---|---|
| 1 | static `IEvent<T>` handlers not discovered | the `#5052` commit here |
| 2 | old natural key not cleaned up | shipped in #5049 |
| 3 | handlers do not receive the current snapshot state | unsupported by design — now documented |

Item 2's existing tests only exercise a source shape discovery could *always* bind (a static handler taking the raw event). On the two paths #5052 newly enables, the question could not even be asked before, because item 1 meant nothing was written for those event types at all — so this PR asserts retirement there too, on both the `IEvent<T>` source and the explicit registration.

Item 3 stays unsupported because the lookup table is maintained *inline at append time*. That is exactly what makes a natural key lookup work under an `Async` snapshot lifecycle, where no aggregate exists yet, and it is why the key has to be a function of the event alone. jasperfx#571 stopped papering over it — the blank-aggregate strategy is ranked last, skipped when the aggregate cannot be safely constructed, and reported loudly rather than silently producing a wrong key. The docs now state that the method never sees the current aggregate, why that follows from when the write happens, and what to do instead.

Swept the rest of the suite for projections quietly relying on a mapping that never existed — 34 natural key tests in `EventSourcingTests` plus 7 in `DaemonTests`, all green, so nothing else depended on the silence.

## Verification

| Suite | Result |
|---|---|
| CoreTests | 493 passed, 1 skipped |
| EventSourcingTests | 1470 passed, 7 skipped |
| DaemonTests | 249 passed, 1 pre-existing failure |
| markdownlint + cspell | clean over all of `docs/` |

The DaemonTests failure is `blue_green_side_effect_gate`, which flaps pass/fail run-to-run on a clean worktree at the merge-base before any change here. Unrelated to this work — it touches neither natural keys, extended progression, nor event deserialization.


jeremydmiller:
### CI: `HighWaterAgentTests.skips_multiple_gaps_and_keeps_advancing`

The one failing job (`test-daemon`, net10.0) is **not** caused by anything in this PR's own commits — it comes from the JasperFx.Events 2.36.0 bump, and I've filed it upstream as [JasperFx/jasperfx#572](https://github.com/JasperFx/jasperfx/issues/572).

Bisected to the bump alone; a commit changing nothing but the pin reproduces it.

| | Result |
|---|---|
| This branch (2.36.0), net10.0 Release, unmodified test | 8 of 10 fail |
| Clean master (2.35.0), same test | 0 of 8 fail |
| Interleaved, to rule out DB/ordering effects | 2.36.0 fails 2/3, 2.35.0 passes 3/3 |

**It is not a daemon slowdown.** With generous budgets both versions reach the head at an identical ~5.05s on the same per-gap cadence, and when the test passes `Tracker.HighWaterMark` equals the expected head — the mark is reached, the waiter just never observes it.

**It is not a tight test deadline either.** I tried raising the per-step budget 2s → 10s; still 9 of 10 failures. The wakeup is lost, not late, so widening the deadline cannot fix it — which is why this PR carries no test-timeout change.

Root cause is a race in `ShardStateTracker`/`ShardStatusWatcher` that jasperfx#568 narrowed but did not close: `publish` walks a captured listener snapshot and refreshes `_states` from inside that same walk, so a watcher that subscribes mid-walk gets neither the `OnNext` nor a current-state snapshot containing it. Details and a suggested direction are in the upstream issue.

Holding this PR until an upstream fix ships.

jeremydmiller:
### Resolved by JasperFx.Events 2.36.1

[jasperfx#572](https://github.com/JasperFx/jasperfx/issues/572) is fixed and shipped ([jasperfx#573](https://github.com/JasperFx/jasperfx/pull/573), *"close the lost-wakeup race rather than narrowing it"*). Pins bumped 2.36.0 → 2.36.1.

The fix makes subscribe-and-read-snapshot atomic against the publication walk: `ShardStateTracker` now takes one lock across recording a published state and capturing its listener list, and `ShardStatusWatcher` goes through it via a new `SubscribeAndCaptureCurrentStates`. The two sides are totally ordered, so a watcher can no longer fall between them and see the state on neither path.

**Verification on net10.0 Release** — the same protocol that produced the failures:

| | 2.36.0 | 2.36.1 |
|---|---|---|
| `skips_multiple_gaps_and_keeps_advancing`, 10 runs | 2 passed / **8 failed** | **10 passed / 0 failed** |
| Full DaemonTests suite | 249 / 1 failed | **250 / 0 failed** |

Ten consecutive runs rather than one, deliberately: this bug passed roughly 1 in 10 by chance, so a single green run would have proved nothing.

**No Marten code changed for this.** The test is untouched — widening its timeouts was tried and failed (9 of 10 still failed), and would have masked a real defect on the public `WaitForHighWaterMark` path rather than fixing it.

Issue #5063, opened 2026-07-29
OrderBy on a dictionary indexer ignores the key and sorts by the whole JSON object


jeremydmiller:
`OrderBy(x => x.SomeDictionary["key"])` silently drops the key and orders by the whole JSON object instead.

## Repro

```csharp
public class Car
{
    public Guid Id { get; set; }
    public Dictionary<string, string> Attributes { get; set; } = new();
}

var car1 = new Car { Attributes = { { "one", "5" }, { "anumber", "5" }, { "color", "red" } } };
var car2 = new Car { Attributes = { { "anumber", "3" }, { "color", "green" } } };
var car3 = new Car { Attributes = { { "anumber", "7" }, { "color", "blue" } } };
var car4 = new Car { Attributes = { { "anumber", "4" }, { "color", "purple" } } };

session.Store(car1, car2, car3, car4);
await session.SaveChangesAsync();

var ordered = await query.Query<Car>().OrderBy(x => x.Attributes["anumber"]).ToListAsync();
// expected: car2 (3), car4 (4), car1 (5), car3 (7)
// actual:   car3 (7), car2 (3), car4 (4), car1 (5)
```

## Generated SQL

```sql
select d.data from public.mt_doc_car as d order by d.data -> 'Attributes';
```

The indexer never makes it into the ORDER BY. It should be something like:

```sql
select d.data from public.mt_doc_car as d order by d.data -> 'Attributes' ->> 'anumber';
```

Because the whole `Attributes` object is what gets sorted, the result is the same
regardless of which key you ask for — `OrderBy(x => x.Attributes["color"])` and
`OrderBy(x => x.Attributes["anumber"])` return byte-identical orderings.

Reproduced on master @ 9.20.2, PostgreSQL 15 and latest, both Newtonsoft and System.Text.Json.
Behaves the same whether `Attributes` is a field or an auto-property.

## Why this has gone unnoticed

The wrong SQL still returns *plausible* results. Ordering by the whole JSONB object
coincidentally matches ordering by a value whenever:

- the dictionary has a single key, or
- the values happen to sort the same way the serialized objects do.

In the case above, `OrderBy(x => x.Attributes["color"])` asserts successfully purely by
coincidence — blue/green/purple/red happens to be the same order the JSONB objects sort in.
Only the `anumber` assertion in the same test diverges.

`Where` clauses on the same indexer (`x => x.Attributes["color"] == "red"`) are unaffected;
this appears specific to the ordering path.

[...]

PR #5064, opened 2026-07-29
Migrate the test suite to xunit v3

jeremydmiller:
Migrates the test suite from xunit 2.9.3 to xunit.v3 3.2.2.

Execution stays on **VSTest** (`Microsoft.NET.Test.Sdk` + `xunit.runner.visualstudio` 3.1.5, which supports both), so the Nuke targets and the GitHub workflows keep working unchanged. Microsoft Testing Platform is deliberately deferred — adopting it would mean reworking every `dotnet test` invocation and replacing coverlet.

`src/samples/Helpdesk/Helpdesk.Api.Tests` stays on v2 (Ogooreck is v2-bound, separate solution), which is why both package ids remain in `Directory.Packages.props`.

## Why this is one commit series and not per-project

22 of 27 xunit projects form a single atomic cohort. Seven `Compile`-link xunit-typed harness files out of `Marten.Testing`, so changing `IntegrationContext`'s signatures breaks every linker at once, and test-project → test-project references pull in the rest. Only 5 could have moved independently.

## What v3 forced

- **`SerializerTypeTargetedFact/Theory` lose their v2 discoverers.** v3 deleted that extensibility model; it turns out none of it was needed — assigning `Skip` from the `RunFor` setter is enough.
- **The `XunitTestFramework` subclasses become `[ModuleInitializer]`s.** v3 reworked that extensibility point and its runner docs are still marked forthcoming. `src/TestSetup.cs` was an orphan naming a nonexistent assembly and is deleted.
- **`IAsyncLifetime` is now `ValueTask`-based and inherits `IAsyncDisposable`.** This is the dangerous one — see below.
- **`Xunit.Abstractions` is gone**; `ITestOutputHelper` moved to `Xunit` and gained an `Output` member.

## The correctness landmine

42 classes inherit an `IDisposable`-only harness base *and* implement `IAsyncLifetime`. Under v2 the runner called both. Under v3 only the async path runs, so their base `Dispose()` would have silently stopped disposing sessions — a suite-wide resource leak that compiles cleanly.

`OneOffConfigurationsContext` and `StoreContext<T>` now own a virtual async teardown that chains to `Dispose()`, and 92 derived members became `override` + `base.DisposeAsync()`. All 48 test-side `override DisposeAsync` implementations were audited to confirm they reach base.

Worth knowing: the hiding warnings that reveal this (CS0114/CS0108) **only appear on a full rebuild**. An incremental build reports zero and looks clean.

## Tests that had never run

The v2 attribute named its discoverer by assembly (`"Marten.Testing"`) while the attribute type lived in each `Compile`-linking project's own assembly. v2 could not resolve the discoverer and **silently discovered no test at all** — not skipped, absent.

| | baseline | this branch |
|---|---|---|
| DocumentDbTests | 1082 | **1086** |
| LinqTests | 1409 | **1418** |

Those 13 are exactly the `SerializerTypeTargetedFact` methods in the two linking projects. Confirmed by diffing `--list-tests`; nothing was lost in the other direction.

Restoring them surfaced real problems that had no way of being noticed:

- `Bug_1189` called synchronous `FirstOrDefault()`. Marten 9.0 made data access async-only, and this test was invisible when that landed, so it simply rotted.
- `Bug_1219` fails on a genuine LINQ defect — filed as #5063 and skipped with that link rather than adjusting the assertion to match the broken behaviour.

Separately, in projects that only `ProjectReference` the harness the discoverer *did* load, but read `GetNamedArgument<SerializerType?>` against a non-nullable enum, so it never skipped. The 3 affected PatchingTests turned out to pin their own store via `UseNewtonsoftForSerialization()`, making the annotation redundant; they are now plain `[Fact]` so the SystemTextJson leg keeps running them.

**Net CI effect: the Newtonsoft leg gains 13 tests, the SystemTextJson leg is unchanged.**

## Order-dependent tests

v3 orders test methods within a class differently than v2, which exposed tests that only ever passed because a sibling ran first. Each is the same shape — a sibling resets shared state and this one doesn't:

- `delete_many_documents_by_query` and `compiled_query_by_string_fragments` — assert on global document counts
- `get_raw_json_Tests` — both store an "Issue 1" then call `ToJsonSingle()`, which requires exactly one
- `identity_map_mechanics` — both seed the same three "James" users
- `advanced_sql_query` — several store `DocWithMeta #1`, whose revision trips optimistic concurrency on re-store
- `marten_managed_tenant_id_partitioning` — the additive test's `b1`/`b2` survive in Marten's in-memory managed-tenant cache across a schema drop
- `full_text_index.migration_from_v3_to_v4` — simulated a v3 index in this context's schema but asserted against the unrelated `fulltext` schema, so it only passed once a sibling had populated that schema. Now points at the schema it actually sets up, which makes it deterministic *and* meaningful: it passes in isolation on a fresh database, which it never did on master.

None are migration defects; all are latent isolation bugs the new ordering revealed.

## Also in here

- **`Marten.CommandLine.Tests` removed.** It was not in `Marten.slnx`, so the solution build never touched it, and it fails to compile identically on master. The Nuke `TestCli` target had `ProceedAfterFailure`, which is why CI never surfaced it. Target, schema entry, both workflow steps and the CLAUDE.md references are gone.
- **`FSharpProjections` FSharp.Core pin.** It compiled against the SDK's implicit FSharp.Core 10 while CPM deployed 9.0.100 — latent until something reflected over the assembly, which v3 does because tests now run from the test exe rather than `testhost`. Verified identical on master.
- **`DaemonTests/TestSetup.cs` deleted rather than ported.** Its attribute named a type and assembly that stopped existing at some rename, so v2 could never load it and the suite has always run under SystemTextJson regardless of `DEFAULT_SERIALIZER`. Deleting preserves that exactly; a working module initializer would have silently flipped the suite's serializer.
- **`src/Tests.props`** carries the shared `OutputType=Exe`, the entry-point opt-outs, and a `NoWarn` for `xUnit1051` (~5,650 candidate sites; threading real cancellation tokens through the DB tests is a separate opt-in change).
- **`EventSourceWorker`** needed `XunitAutoGeneratedEntryPoint=false` — xunit.v3 ships entry-point generation in `buildTransitive/`, so it leaks into non-test projects that reference a test project.

## Verification

Every suite was run against a v2 baseline worktree built from the same commit, on both TFMs, and both serializer legs for the suites where that matters. The DB-backed runs used a freshly dropped and recreated database so residue could not be mistaken for a regression.

### Known pre-existing flake, not addressed here

`DaemonTests.blue_green_side_effect_gate` fails intermittently on **the v2 baseline** too — 3 consecutive baseline runs on net9.0 produced 1, 1 and 2 failures. It is unrelated to this change.

`PLAN-xunit3-migration.md` carries the full findings with reproduction detail.


jeremydmiller:
Follow-ups filed from this work:

- #5063 — `OrderBy` on a dictionary indexer drops the key and sorts by the whole JSON object (the defect behind the one skipped test in this PR)
- #5065 — `blue_green_side_effect_gate` is flaky on master (measured: 3 baseline runs → 1, 1, 2 failures)
- #5066 — DaemonTests has never honoured `DEFAULT_SERIALIZER`; fixing it reveals 2 Newtonsoft round-trip failures
- #5067 — thread `TestContext.Current.CancellationToken` (the xUnit1051 suppression added here)
- #5068 — re-evaluate the 13 remaining `SerializerTypeTargetedFact` annotations; they may be unnecessary like the PatchingTests trio turned out to be
- #5069 — finish the migration: Helpdesk.Api.Tests is blocked on Ogooreck, which is why both xunit package ids remain
- #5070 — `IntegrationContext` shares one store and never clears between tests, which is the root cause of the six order-dependent tests fixed here

PR #5074, opened 2026-07-29
Drop 4 unnecessary SerializerTypeTargetedFact annotations

### LINKED ISSUE #5068: Re-evaluate the 13 remaining SerializerTypeTargetedFact annotations — they may be unnecessary
13 tests carry `[SerializerTypeTargetedFact(RunFor = SerializerType.Newtonsoft)]` and, because of the v2 discoverer defect described in #5064, **have never executed under either serializer**. There is therefore no evidence that any of them actually needs the annotation.

| project | count | tests |
|---|---|---|
| DocumentDbTests | 4 | `persist_and_query_via_dynamic.CanPersistAndQueryDynamic`, `end_to_end_versioned_docs.overwrite_behavior{,_with_props,_async}` |
| LinqTests | 9 | `distinct_operator.get_distinct_numbers`, `Bug_1189.can_select`, `Bug_1219.can_order_by_string_string_dictionaries`, `select_many.select_many_with_select_transformation`, 5 in `select_clause_usage` |

## Precedent

The same exercise on the 3 annotated `PatchExpressionTests` found the annotation was pure redundancy — each test already calls `opts.UseNewtonsoftForSerialization()` on its own store, so it pins Newtonsoft irrespective of the ambient setting. Those 3 are plain `[Fact]` as of #5064, which is why the SystemTextJson leg keeps running them.

None of the 13 above configure a serializer of their own, so unlike the PatchingTests trio they genuinely read the ambient one. That makes them *plausibly* Newtonsoft-only — `persist_and_query_via_dynamic` uses `dynamic`, for instance — but it is untested either way.

## What to do

Temporarily drop the annotations, run DocumentDbTests and LinqTests with `DEFAULT_SERIALIZER=SystemTextJson`, and keep the annotation only on the ones that actually fail. Every one that passes is coverage the SystemTextJson leg should be getting.

Note `Bug_1219.can_order_by_string_string_dictionaries` is currently skipped against #5063 and should be handled there instead.

jeremydmiller:
Closes #5068.

The 13 annotated tests had never executed under either serializer (see #5064), so there was no evidence any of them needed `RunFor = Newtonsoft`. I stripped all of them, ran DocumentDbTests and LinqTests under `DEFAULT_SERIALIZER=SystemTextJson`, and kept the annotation only where the test genuinely fails.

It's a real split — **8 of 12 are load-bearing, 4 are not.** Not the blanket redundancy the PatchingTests trio turned out to be.

## Removed — these pass under System.Text.Json and now run on that leg

- `distinct_operator.get_distinct_numbers`
- `Bug_1189_can_select_transform_without_an_id.can_select`
- `select_many.select_many_with_select_transformation`
- `select_clause_usage.transform_with_deep_properties_to_anonymous_type`

## Kept, each now carrying the observed reason

| test | reason |
|---|---|
| `persist_and_query_via_dynamic` | STJ deserializes to `JsonElement`, which has no dynamic member access — `'JsonElement' does not contain a definition for 'temperature'`. Inherent to the serializer. |
| `select_clause_usage` ×4 | the multi-field and constructor projections |
| `end_to_end_versioned_docs` ×3 | see below |

### The versioned-docs three are not actually a serializer problem

They fail when the suite runs under the System.Text.Json configuration, but the cause is **not** serialization:

```csharp
doc1.Version.ShouldBe(originalVerion);              // passes — round-trips fine
session1.VersionFor(doc1).ShouldBe(originalVerion); // fails — returns null
```

The document's own `Version` property deserializes correctly; it's `session.VersionFor(doc)` that comes back null after `LoadAsync`. Sibling tests in the same class call `VersionFor` under either serializer and pass.

So the annotation here is keeping the suite green rather than describing a genuine serializer limitation, and the comment says exactly that. The real cause is unknown and probably deserves its own issue.

## Verification

net9.0, both legs, from a freshly recycled database:

| leg | suite | result |
|---|---|---|
| SystemTextJson | DocumentDbTests | 1086 total, 1081 passed, 5 skipped |
| SystemTextJson | LinqTests | 1418 total, **1413 passed**, 5 skipped |
| Newtonsoft | DocumentDbTests | 1086 total, 1085 passed, 1 skipped |
| Newtonsoft | LinqTests | 1418 total, 1417 passed, 1 skipped |

LinqTests on the System.Text.Json leg goes from 1409 passed / 9 skipped to 1413 passed / 5 skipped.

`Bug_1219` is excluded — it's skipped on master pending #5063.

### DIFF
--- a/src/DocumentDbTests/Metadata/end_to_end_versioned_docs.cs
+++ b/src/DocumentDbTests/Metadata/end_to_end_versioned_docs.cs
@@ -49,7 +49,7 @@ public class end_to_end_versioned_docs: IntegrationContext
     // NOTE the cause is NOT serialization: doc1.Version round-trips correctly, it is
     // session.VersionFor(doc) that comes back null after LoadAsync. The sibling tests in
     // this class call VersionFor happily under either serializer. Left targeted so the
-    // suite stays green; the underlying reason is not understood.
+    // suite stays green; tracked as #5075.
     [SerializerTypeTargetedFact(RunFor = SerializerType.Newtonsoft)]
     public async Task overwrite_behavior()
     {
@@ -105,7 +105,7 @@ public class end_to_end_versioned_docs: IntegrationContext
     // NOTE the cause is NOT serialization: doc1.Version round-trips correctly, it is
     // session.VersionFor(doc) that comes back null after LoadAsync. The sibling tests in
     // this class call VersionFor happily under either serializer. Left targeted so the
-    // suite stays green; the underlying reason is not understood.
+    // suite stays green; tracked as #5075.
     [SerializerTypeTargetedFact(RunFor = SerializerType.Newtonsoft)]
     public async Task overwrite_behavior_with_props()
     {
@@ -161,7 +161,7 @@ public class end_to_end_versioned_docs: IntegrationContext
     // NOTE the cause is NOT serialization: doc1.Version round-trips correctly, it is
     // session.VersionFor(doc) that comes back null after LoadAsync. The sibling tests in
     // this class call VersionFor happily under either serializer. Left targeted so the
-    // suite stays green; the underlying reason is not understood.
+    // suite stays green; tracked as #5075.
     [SerializerTypeTargetedFact(RunFor = SerializerType.Newtonsoft)]
     public async Task overwrite_behavior_async()
     {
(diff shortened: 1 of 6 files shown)


jeremydmiller:
Filed the versioned-docs anomaly as #5075.

Narrowed it further before filing — the class run in isolation against a fresh database gives **5 passed under Newtonsoft, 3 failed under System.Text.Json**, so it is the serializer configuration rather than ordering or pollution. The split is between tests that call `VersionFor` on a document stored in that same session (fine on both serializers) and tests that call it after `LoadAsync` in a new session (null under System.Text.Json). `doc.Version` itself holds the right value either way, so it is not a round-trip problem.

The three annotations stay here to keep the suite green, now pointing at #5075.

Issue #5127, opened 2026-08-02
ProjectionScenario ergonomics and quality pass


jeremydmiller:
Findings from a read of `Marten/Events/TestSupport/` (796 lines across 7 files) while assessing the cross-store lift. None of these are correctness bugs — the one bug found is marten#5126, filed separately so it can be fixed without waiting on any of this.

Grouped by how much they cost a user.

## Visible to users

**1. `Append` / `StartStream` return a `StreamAction` that never executes.** Each method queues a closure *and* separately constructs a throwaway result:

```csharp
public StreamAction Append(Guid stream, params object[] events)
{
    var step = action(e => e.Append(stream, events));   // this is what runs
    ...
    return StreamAction.Append(_store.Events, stream, events);   // this is not
}
```

The returned value looks like the operation's result and is not connected to it. Either return `void`/the `ScenarioStep`, or document loudly.

**2. Assertions throw bare `Exception`.**

```csharp
throw new Exception($"Document {typeof(T).FullNameInCode()} with id '{id}' does not exist");
```

Nothing can distinguish an assertion failure from an infrastructure failure. A typed `ProjectionAssertionException` gets that without pulling an assertion library into Marten core.

**3. `DoNotDeleteExistingData` is a double negative on an API that wipes the database by default.** `EventProjectionScenario` deletes all event data plus every projection's storage types unless you set this. The XML doc says "Only use this in testing," which is right, but the opt-out reads backwards at the call site.

**4. No documentation.** `git grep EventProjectionScenario docs/` returns nothing. The feature is only discoverable from the `AdvancedOperations` XML comment.

**5. Public constructor, internal `Execute`.** `ProjectionScenario(DocumentStore)` is public but `Execute` is `internal`, so a directly-constructed scenario can be configured and never run. Either make the ctor internal or make `Execute` public.

## Internal quality

**6. ~120 lines of copy-paste in the assertions.** `DocumentShouldExist<T>` and `DocumentShouldNotExist<T>` are each written four times — Guid/int/long/string — with byte-identical bodies apart from the id parameter type. One private helper over `object id` collapses both sets; this is exactly the shape `LoadDocumentAsync(session, object id, ct)` already uses in `JasperFx.Events.ComplianceTests`.

**7. Double enumeration of `IEnumerable<object>` overloads.**

```csharp
var step = action(e => e.Append(stream, events));
if (events.Count() > 3)                                  // walk #1, just for the description
    ...
[...]

Issue #5126, opened 2026-08-02
ProjectionScenario silently drops a trailing append — the events are never committed


jeremydmiller:
`ProjectionScenario` only flushes the session when the *next* queued step is an assertion, so any action that is the **last** step in a scenario is never saved. The session is disposed in the `finally` without a `SaveChangesAsync`, and Marten's dispose does not commit.

`Marten/Events/TestSupport/ScenarioAction.cs`:

```csharp
public override async Task Execute(ProjectionScenario scenario, CancellationToken ct = default)
{
    _action(scenario.Session.Events);

    if (scenario.NextStep is ScenarioAssertion)   // <-- null for the last step
    {
        await scenario.Session.SaveChangesAsync(ct).ConfigureAwait(false);
        await scenario.WaitForNonStaleData().ConfigureAwait(false);
    }
}
```

`NextStep` peeks the queue *after* the current step was dequeued, so it is null on the final step and the branch never runs.

## Consequences

- A scenario ending in an append silently loses those events.
- **An arrange-only scenario is a complete no-op that passes.** Nothing is written, nothing is asserted, the test is green.

## Reproduction

Verified against master (`c89ef018e`). Fails with `count should be 2 but was 1`:

```csharp
public class scenario_trailing_action_probe : OneOffConfigurationsContext
{
    [Fact]
    public async Task trailing_action_is_committed()
    {
        StoreOptions(opts => opts.Projections.Add(new UserProjection(), ProjectionLifecycle.Inline));

        var id = Guid.NewGuid();

        await theStore.Advanced.EventProjectionScenario(scenario =>
        {
            // assertion in the middle, so the FIRST append flushes
            scenario.Append(Guid.NewGuid(), new CreateUser { UserId = id, UserName = "First" });
            scenario.DocumentShouldExist<User>(id);

            // trailing append with nothing after it
            scenario.Append(Guid.NewGuid(), new CreateUser { UserId = Guid.NewGuid(), UserName = "Trailing" });
        });

        await using var query = theStore.QuerySession();
        var count = await query.Query<User>().CountAsync();
        count.ShouldBe(2);   // actual: 1
    }
}
```

(`CreateUser` / `UserProjection` are the existing types in `DaemonTests/EventProjections/event_projection_scenario_tests.cs`.)

## Why no existing test catches it

All 8 tests in `event_projection_scenario_tests.cs` happen to end with an assertion, so the final-step path is never exercised.

## Suggested fix

[...]

PR #5133, opened 2026-08-03
Adopt the lifted JasperFx.Events ProjectionScenario

jeremydmiller:
Follow-up to #5127 / #5132. Consumes the cross-store lift in JasperFx/jasperfx#616.

> [!IMPORTANT]
> **Draft: this will not go green until JasperFx.Events 2.38.0 is published.** The `Directory.Packages.props` pin here is the *intended* published version — jasperfx#616 is open, not merged. The branch was built and tested against a local pack of that branch; it needs re-verification against the real package before this leaves draft.

## What changes

Marten's `Events/TestSupport` goes from seven files of harness to a single ~70-line subclass. Everything about how a scenario behaves — step queueing, batched commits, the #5126 trailing flush, action fail-fast, the re-execution guard, daemon lifecycle, `Timeout` — now lives on the shared `ProjectionScenario<TOperations, TQuerySession>` base type in `JasperFx.Events.TestSupport`.

What's left in Marten is only the store-specific seam: `DeleteExistingDataAsync`, `HasAnyAsyncProjections`, `BuildDaemonAsync`, `OpenSession`, `SaveChangesAsync`, `EventsFor`, and the `object`-id `LoadDocumentAsync` dispatch — the same switch `MartenComplianceFixture` already carries. Net −692 / +51 lines.

## API impact

Small, and on top of the already-breaking #5132:

- `scenario.Execute(ct)` → `scenario.ExecuteAsync(ct)`.
- `ProjectionScenarioException` and `ProjectionScenarioAssertionException` move from `Marten.Events.TestSupport` to `JasperFx.Events.TestSupport`. Test code that catches either needs the new `using`.
- `DocumentShouldExist<T>` / `DocumentShouldNotExist<T>` now take `object` ids on the base type instead of four typed overloads each. Existing typed call sites keep compiling — everything converts to `object`.

Everything else a user touches is unchanged from #5132.

## Verification (against a local pack of jasperfx#616)

- `Marten.slnx` builds clean.
- DaemonTests scenario suite 15/15 on net9.0 — the 8 original tests, the 2 from #5126, and the 5 from #5132, all unmodified apart from the `ExecuteAsync` rename and the exception `using`.
- CoreTests exception convention test green.

## Merge order

1. jasperfx#616 merges and JasperFx.Events 2.38.0 publishes.
2. Re-verify this branch against the real package, undraft.
3. JasperFx/polecat#404 then subclasses the same base type — no hand-ported overloads.


jeremydmiller:
Rebased onto master now that JasperFx.Events 2.38.0 is published and #5135 has landed. The `Directory.Packages.props` pin edit dropped out as redundant — master already carries 2.38.0.

Re-verified against the **published** package rather than the local pack this was originally proven with: full `Marten.slnx` builds clean and the scenario suite is 15/15 on net9.0. Out of draft.

PR #5134, opened 2026-08-03
Container-scoped projections work with live aggregation and are validated

### LINKED ISSUE #5095: Service-injected scoped/transient projections are not validated or usable by live aggregation / single stream rebuilds
## Description

`AddProjectionWithServices<T>` behaves differently depending on projection lifetime.

Scoped and transient projections are wrapped in `ScopedAggregationWrapper`. Normal inline and async-daemon execution works, but native live aggregation and targeted single-stream rebuild do not select the actual projection through `AggregatorFor<T>()`.

Registering the projection as a singleton exposes the actual projection, allowing `AggregateStreamAsync<T>` and `RebuildSingleStreamAsync<T>` to work.

## Problem

A scoped service-injected projection works during normal projection processing:

```csharp
services.AddMarten(options => { })
    .AddProjectionWithServices<MyProjection>(
        lifecycle: ProjectionLifecycle.Inline,
        lifetime: ServiceLifetime.Scoped);
```

However, native aggregation paths need the actual projection:

```csharp
await session.Events.AggregateStreamAsync<MyAggregate>(streamId);

await documentStore.Advanced
    .RebuildSingleStreamAsync<MyAggregate>(streamId);
```

Changing the projection registration to `ServiceLifetime.Singleton` makes these APIs work.

This is problematic because a singleton is only safe when the full projection dependency graph is stateless and singleton-compatible.

## Hidden validation

The scoped/transient wrapper also appears to hide projection validation.

Marten correctly rejects a projection that mixes explicit `Evolve`/`EvolveAsync` with conventional aggregate handlers, such as a public event constructor or `Create`/`Apply` method:

```csharp
public sealed class MyAggregate
{
    public MyAggregate(MyEvent @event) { }
}

public sealed class MyProjection : SingleStreamProjection<MyAggregate, string>
{
    public override ValueTask<MyAggregate?> EvolveAsync(
        MyAggregate? aggregate,
        string streamId,
        IQuerySession session,
        IEvent @event,
        CancellationToken cancellationToken)
    {
        // ...
    }
}

```

With singleton registration, Marten throws `InvalidProjectionException` while building the document store.

With scoped/transient `AddProjectionWithServices<T>` registration, Marten reports the projection as valid because validation appears to inspect the wrapper instead of the underlying projection.

Changing the registration to singleton can therefore reveal an invalid projection configuration that was previously hidden.

## Expected behavior

- `AggregateStreamAsync<T>` and `RebuildSingleStreamAsync<T>` should support service-injected projections registered as scoped or transient, resolving them through an appropriate scope.
- Projection validation should always inspect the underlying projection, independent of its DI lifetime.
- Mixing explicit `Evolve`/`EvolveAsync` with conventional aggregate handlers should fail consistently during Marten configuration.

## Why this matters

Applications may use injected services while replaying streams. Scoped/transient projections should not silently lose native live aggregation and targeted rebuild support, nor should they hide invalid projection configuration.

jeremydmiller:
Closes #5095. Consumes JasperFx/jasperfx#617.

> [!IMPORTANT]
> **Draft: blocked on JasperFx.Events 2.38.0.** The pin here is the *intended* published version; jasperfx#617 is open, not published. Everything below was verified against a local pack of that branch and needs re-verification against the real package before this leaves draft.

## The bug

A projection registered through `AddProjectionWithServices` with a **Scoped or Transient** lifetime is wrapped in `ScopedAggregationWrapper`, and two things went wrong as a result.

**Native aggregation silently ran the wrong thing.** The wrapper doesn't implement `IAggregatorSource`, so `AggregatorFor<T>()` never found it and fell through to conventional aggregation built off the aggregate type. `AggregateStreamAsync<T>` and `RebuildSingleStreamAsync<T>` therefore ignored the registered projection's logic entirely — no error, just a different answer. Registering the same projection as a Singleton worked, which is the asymmetry the reporter hit.

**Validation was hidden.** The wrapper's base constructor copied name/version/options off the resolved projection but never called `AssembleAndAssertValidity()` on it — only the wrapper was validated. A projection that a Singleton registration rejects at startup was silently accepted when registered Scoped, so switching lifetime later would surface a configuration error that had been there all along.

## What changed in Marten

Small — the real work is in jasperfx#617. The four `Register` overloads in `SingleStreamProjection` / `MultiStreamProjection`, plus `CompositeProjection`, now call `ScopedAggregationWrapper.Build(...)` instead of closing `typeof(ScopedAggregationWrapper<,,,,>)` themselves. That helper picks the single-stream wrapper (which can serve live aggregation) or the general one, and unwraps the reflection `TargetInvocationException` so a configuration error surfaces as the real `InvalidProjectionException` instead of a reflection wrapper.

Multi-stream deliberately keeps the general wrapper: only single-stream projections implement `IAggregator`, and exposing multi-stream to `AggregatorFor` would be new behavior the Singleton path doesn't have.

## Test

`Bug_5095_scoped_projection_live_aggregation` is a 3×3 matrix — {live aggregation, single stream rebuild, invalid-projection validation} × {Singleton, Scoped, Transient}.

The aggregate deliberately has **no** conventional `Create`/`Apply` methods, so the conventional fallback can't accidentally produce a correct-looking answer; the projection's `Evolve` (which multiplies by an injected factor) is the only thing that can build it.

| | before | after |
|---|---|---|
| Repro matrix | 3 passed / 6 failed | **9 passed** |

Every pre-fix failure was Scoped or Transient; every Singleton case passed throughout.

## Regression sweep (net9.0)

`ContainerScopedProjectionTests` 80/80 · `EventSourcingTests` 1614/1614 · `DaemonTests` 267/267 · full `Marten.slnx` builds clean. `AggregatorFor` is on the hot path for all live aggregation, so the two large suites are the meaningful signal here.

### DIFF
--- a/src/ContainerScopedProjectionTests/Bug_5095_scoped_projection_live_aggregation.cs
+++ b/src/ContainerScopedProjectionTests/Bug_5095_scoped_projection_live_aggregation.cs
@@ -1,0 +1,228 @@
+using System;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using JasperFx.Events.Projections;
+using Marten;
+using Marten.Events.Aggregation;
+using Marten.Events.Projections;
+using Marten.Exceptions;
+using Marten.Testing.Harness;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Npgsql;
+using Shouldly;
+using Weasel.Postgresql;
+using Xunit;
+
+namespace ContainerScopedProjectionTests;
+
+/// <summary>
+/// Repro for https://github.com/JasperFx/marten/issues/5095.
+///
+/// A projection registered through <c>AddProjectionWithServices</c> with a Scoped or
+/// Transient lifetime is wrapped in <c>ScopedAggregationWrapper</c>. That wrapper does
+/// not implement <c>IAggregatorSource&lt;TQuerySession&gt;</c>, so
+/// <c>ProjectionGraph.AggregatorFor&lt;T&gt;()</c> never finds it: the lookup falls
+/// through <c>tryFindProjectionSourceForAggregateType</c> to the conventional
+/// aggregation built off the aggregate type itself. Every native aggregation path --
+/// <c>AggregateStreamAsync</c>, <c>RebuildSingleStreamAsync</c>, <c>AggregateToAsync</c>
+/// -- therefore silently ignores the projection's own logic. Registering the same
+/// projection as a Singleton exposes the real projection and everything works.
+///
+/// Separately, <c>ProjectionSourceWrapperBase</c>'s constructor copies name/version/
+/// options off the resolved source but never calls <c>AssembleAndAssertValidity()</c>
+/// on it, so an invalid projection that a Singleton registration rejects at startup is
+/// silently accepted when registered Scoped.
+///
+/// The aggregates here deliberately have NO conventional Create/Apply methods, so the
+/// conventional fallback cannot accidentally produce a correct-looking answer -- the
+/// projection's Evolve override is the only thing that can build them.
+/// </summary>
+[Collection("ioc")]
+public class Bug_5095_scoped_projection_live_aggregation
+{
+    [Theory]
+    [InlineData(ServiceLifetime.Singleton)]
+    [InlineData(ServiceLifetime.Scoped)]
+    [InlineData(ServiceLifetime.Transient)]
+    public async Task live_aggregation_uses_the_registered_projection(ServiceLifetime lifetime)
+    {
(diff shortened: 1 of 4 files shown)


jeremydmiller:
Rebased onto master now that JasperFx.Events 2.38.0 is published and #5135 has landed. The `Directory.Packages.props` pin edit dropped out as redundant — master already carries 2.38.0, including the changelog entry for jasperfx#617.

Re-verified against the **published** package rather than the local pack this was originally proven with: full `Marten.slnx` builds clean and `ContainerScopedProjectionTests` is 80/80 on net9.0, including all 9 cases of the #5095 repro matrix. Out of draft.

Issue #5136, opened 2026-08-03
now() at time zone 'utc' written into timestamptz columns skews mt_last_modified and mt_events.timestamp by the server's UTC offset


jeremydmiller:
Three write paths store `now() at time zone 'utc'` into a `timestamp with time zone` column. That expression strips the offset to produce a *naive* timestamp holding UTC wall-clock time, and assigning a naive timestamp to a `timestamptz` re-interprets it in the session's `TimeZone`. On any database whose `TimeZone` is not UTC the stored instant is wrong by the UTC offset — it is a genuinely different point in time, not a display artifact.

The matching INSERT defaults (`(now())`, `(transaction_timestamp())`) are correct, so a row is stamped correctly on insert and then jumps by the offset the first time it goes through one of these paths.

## Reproduction

```sql
create database tz_probe;
alter database tz_probe set timezone='Europe/Berlin';
\c tz_probe

create table probe(seq bigserial primary key, ts timestamp with time zone not null default (now()));

insert into probe(ts) values ((now() at time zone 'utc'));  -- what the write paths below do
insert into probe(ts) values (default);                     -- what the column DEFAULT does

select seq, ts, now() as server_now, now() - ts as reads_this_far_in_the_past from probe order by seq;
```

```
 seq |               ts              |          server_now           | reads_this_far_in_the_past
-----+-------------------------------+-------------------------------+----------------------------
   1 | 2026-08-03 13:41:09.340771+02 | 2026-08-03 15:41:09.340771+02 | 02:00:00
   2 | 2026-08-03 15:41:09.340771+02 | 2026-08-03 15:41:09.340771+02 | 00:00:00
```

Two hours in the past on a UTC+2 database. West of UTC it skews the other way (`America/Chicago` → five hours in the future).

## Affected sites

| Site | Column | Impact |
| --- | --- | --- |
| [`LastModifiedColumn.cs:27`](https://github.com/JasperFx/marten/blob/master/src/Marten/Storage/Metadata/LastModifiedColumn.cs#L27) — `WriteMetadataInUpdateStatement` | `mt_last_modified` (`timestamp with time zone`, default `(transaction_timestamp())`) | **Every document UPDATE.** Broadest blast radius: `IDocumentMetadata.LastModified`, `ModifiedSince`/`ModifiedBefore` queries, and anything auditing document freshness. Inserted rows are correct; the value shifts on first update. |
[...]

PR #5165, opened 2026-08-03
Make the stale-gap cap clock survive detector restarts (#5109)

jeremydmiller:
Carries forward the defect and the tests from #5109 by @uniquelau, with a different mechanism.

## The defect (their diagnosis, confirmed)

`SkipStaleGapsDespiteLiveTransactionsAfter` is documented as the bounded override for the transaction-evidence hold — once a gap has been stuck past the cap, the skip proceeds even though a candidate reserver still looks alive. But it measured from `_stuckGap.Since`, an in-memory observation that dies with the detector instance. Under managed-distribution agent churn, or restart cycles shorter than the cap, the clock reset every cycle and the documented bound never fired.

#5163 did **not** address this — it repaired the allocation fence, which is a different mechanism. Confirmed empirically: applying their test file to current master, **4 of 6 pass** (#5163 handles those scenarios) and the **2 detector-churn cases fail**. This fixes those two.

## The mechanism

Record the first sighting durably: a reserved `mt_event_progression` row holding the pinned mark and the moment it was first seen pinned. The mark only ever advances, so while it has not moved that timestamp stays a sound **lower bound** on how long the gap has been stuck — the direction a cap needs. Keying the row on the mark makes it self-invalidating: a different mark is a different gap, and the clock restarts.

The stale threshold deliberately stays on the in-memory clock, so a fresh detector still gives a just-appeared gap a full settle window. A null cap (the default) is untouched.

## Why not the original approach

#5109 dated the stall from the timestamp of the earliest committed event above the mark. That reads `mt_events.timestamp`, which is client-written in Rich append mode (the default) and skewed by the server's UTC offset in Quick mode — see #5136, which was found while reviewing that PR. A behind-clock client or a non-UTC server could age a gap past the cap and drop a live append's events. Using the mark's own pinning time keeps everything in server time and needs no trust in event timestamps at all.

The original PR's `last_updated`-only rejection still stands and its guard test (`idle_store_first_live_append_is_not_cap_skipped`) is carried over intact.

## Tests

All six tests are @uniquelau's from #5109, unchanged in substance, with co-authorship on the commit:

| Test | master | here |
| --- | --- | --- |
| `cap_skips_a_pre_existing_dead_gap_despite_an_idle_listener_and_no_allocation_history` | pass | pass |
| `idle_store_first_live_append_is_not_cap_skipped` (guard) | pass | pass |
| `live_reserver_is_not_cap_skipped_before_the_cap_genuinely_elapses` (guard) | pass | pass |
| `running_daemon_resumed_after_the_gap_formed_skips_within_the_cap` | pass | pass |
| **`detector_churn_does_not_postpone_the_cap_forever`** | **fail** | pass |
| **`conjoined_tenancy_routes_store_global_and_the_cap_still_bounds_detector_churn`** | **fail** | pass |

DaemonTests 280/280 net10.0. Further suites running locally; will report.

## Housekeeping

The new reserved row is excluded from `ProjectionProgressStatement` and from the blanket UPDATE in `TryCorrectProgressInDatabaseAsync`, same as the allocation fence. #5164's `isProgressRow` allowlist already ignores unrecognised bookkeeping names, so `WaitForNonStaleProjectionDataAsync` needed no change — which is that hardening paying for itself immediately.

### DIFF
--- a/src/DaemonTests/Bugs/Bug_4953_cap_bounds_fenceless_detector_holds.cs
+++ b/src/DaemonTests/Bugs/Bug_4953_cap_bounds_fenceless_detector_holds.cs
@@ -1,0 +1,466 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using DaemonTests.TestingSupport;
+using JasperFx.Core;
+using JasperFx.Events;
+using JasperFx.Events.Projections;
+using Marten;
+using Marten.Events.Daemon.HighWater;
+using Marten.Storage;
+using Marten.Testing;
+using Marten.Testing.Harness;
+using Npgsql;
+using Shouldly;
+using Weasel.Postgresql;
+using Xunit;
+
+namespace DaemonTests.Bugs;
+
+/// <summary>
+/// Follow-up to #4953/#5057: SkipStaleGapsDespiteLiveTransactionsAfter is documented as the bounded
+/// override for the transaction-evidence hold — once a gap has been stuck past the cap, the skip
+/// proceeds even though a candidate reserver still appears alive. A detector that first observes an
+/// ALREADY-DEAD gap (daemon resumed mid-gap) has no allocation history, so the fence cannot rule
+/// out a Wolverine-style idle-in-transaction listener and the evidence hold engages — exactly the
+/// situation the cap exists to bound. The per-gap clock is detector-scoped, though, so a cap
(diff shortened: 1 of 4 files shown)


jeremydmiller:
Local sweep complete on net10.0: DaemonTests 280/280, EventSourcingTests 1627/1627, TenantPartitionedEventsTests 238/240 (2 pre-existing skips).

Issue #5158, opened 2026-08-03
StreamOne/WriteSingle with a Select() projection throws "Field not found in row: data" when emitETag is on (regression in 9.18.0)


jeremydmiller:
## Summary

`WriteSingle` / `StreamOne<T>` throws at runtime when the queryable carries a `Select()` projection, because the ETag support appends an `mt_version` column to a `select` that has no `data` column.

Regression introduced by #5027 / PR #5030 (`44f9b44a2`), first shipped in **9.18.0**. Still present on master at `0e8daed37`.

## Repro

```csharp
await using var query = store.QuerySession();
var context = new DefaultHttpContext { Response = { Body = new MemoryStream() } };

await query.Query<Issue>().Where(x => x.Id == issue.Id)
    .Select(x => new { x.Description })      // <-- any Select() projection
    .WriteSingle(context, emitETag: true);   // emitETag: true is the default
```

```text
System.IndexOutOfRangeException : Field not found in row: data
   at Npgsql.BackendMessages.RowDescriptionMessage.GetFieldIndex(String name)
   at Npgsql.NpgsqlDataReader.GetOrdinal(String name)
   at Marten.Services.JsonStreamingExtensions.StreamOneWithVersion(DbDataReader reader, ...)
   at Marten.Internal.Sessions.QuerySession.StreamOneWithVersion(DbCommand command, ...)
   at Marten.Linq.MartenLinqQueryProvider.StreamOneWithVersion[T](Expression expression, ...)
   at Marten.AspNetCore.QueryableExtensions.WriteSingle[T](IQueryable`1 queryable, ...)
```

Passing `emitETag: false` works, which is the current workaround — it routes to the plain `StreamJsonFirstOrDefault` path that never appends the version column.

## Root cause

In `MartenLinqQueryProvider.StreamOneWithVersion<T>`, `T` is the *projected* type once a `Select()` is in play — an anonymous type, or any DTO — not the mapped document type:

```csharp
var mapping = _session.Options.Storage.FindMapping(typeof(T)) as DocumentMapping;

if (numericRevision || mapping is { Metadata.Version.Enabled: true })
{
    main.SelectClause = new VersionSelectClause<T>(main.SelectClause);
    ...
}
```

[...]

Issue #5120, opened 2026-08-02
StreamOne<T> cannot emit an ETag for a projection-target document. Suggestion: support sourcing it from the numeric revision (which already equals the stream version)


uniquelau:
**Environment:** Marten 9.20.2, 9.22.2, and master `de63fc5ed` (all verified empirically); Marten.AspNetCore same versions; .NET 10; Postgres.

## The gap

The conditional-GET ETag support (#5010 → #5015, refined in #5027/#5030) documents that types with version metadata disabled emit no ETag. Projection-target documents are *forced* into that state: `ProjectionDocumentPolicy` (`src/Marten/Events/Projections/ProjectionDocumentPolicy.cs:15-21`) sets `Metadata.Version.Enabled = false` + `UseNumericRevisions = true` for any aggregate-projection doc — deliberately, per #2978/#3057/#3097 (numeric revisions are required for blue/green; Guid OCC on projection targets is unsupported by design).

Net effect: **the common CQRS read-model shape — a `SingleStreamProjection` document served by `StreamOne<T>` — can never produce an ETag**, because `MartenLinqQueryProvider.StreamOneWithVersion<T>` gates on exactly `mapping is DocumentMapping { Metadata.Version.Enabled: true }` (`src/Marten/Linq/MartenLinqQueryProvider.cs:291-294`) — the Guid flavor only.

However, the projected doc's `mt_version` **is populated and meaningful** — `SingleStreamProjection.ConfigureAggregateMapping` sets `UseVersionFromMatchingStream = true` (`src/Marten/Events/Aggregation/SingleStreamProjection.cs:46`), so the bigint `mt_version` equals the source stream version. Verified empirically: after 3 events, the row's `mt_version = 3`, `MetadataForAsync(...).CurrentRevision = 3` (`CurrentVersion = Guid.Empty`). That is byte-for-byte the same value `StreamAggregate<T>` already serves as its ETag (`ETagHelpers.Format(long)`, `src/Marten.AspNetCore/ETagHelpers.cs:22`). The read model knows its version; only the `StreamOne` ETag path can't see it.

## No workaround exists for Inline projections (all verified empirically)

- `Schema.For<T>().UseOptimisticConcurrency(true)` → duplicate `mt_version` column (uuid + bigint): `ArgumentException` from the schema diff on an existing table, invalid DDL (two `mt_version` columns) on a fresh one. Known and by-design per #3097.
[...]

Issue #5027, opened 2026-07-22
StreamOne<T> ETag: inline mt_version (eliminate 2nd round trip) + reverse where T:notnull + ETag test hardening (pre-9.18, follow-up to #5015)


jeremydmiller:
Follow-up to #5015 (ETag / If-None-Match support). **To be done before the 9.18 release.**

### Problem

`StreamOne<T>` currently computes its ETag by, *after* the document JSON is already fetched and buffered:
1. deserializing the buffered JSON back into a `T` (`session.Serializer.FromJson<T>(stream)`), then
2. issuing a **second DB round trip** — `IQuerySession.MetadataForAsync<T>(entity)` — purely to read `mt_version`.

So with `EmitETag = true` (the default), every `StreamOne` request does **two DB round trips instead of one**, and a `304` saves only bandwidth — the document was already fetched. (This is unlike `StreamAggregate`, where `FetchStreamStateAsync` reads the version cheaply *before* the expensive fold, so a `304` skips real work.)

This second call is also the reason `StreamOne<T>`/`WriteSingle<T>` had to tighten to `where T : notnull` (because `MetadataForAsync<T>` is constrained `notnull`).

### Goal

Fetch `mt_version` **inline with the document in the original single round trip** — same idea as the `count(*) OVER() as total_rows` stats-column trick that #5014 uses — so:
- `StreamOne` stays at **one** DB round trip with `EmitETag = true` on by default (which the maintainer has OK'd once this lands), and
- the second `MetadataForAsync` call is removed entirely.

### Sketch

- Have the `StreamOne` streaming path select `mt_version` alongside the `data` column (extend the Marten core streaming query used by `WriteSingle`, analogous to how `StatsSelectClause` appends `count(*) OVER()`), and read the version off the same `DbDataReader` before/while streaming the body — no re-deserialization, no metadata query.
- Once the version no longer comes through `MetadataForAsync<T>`, the `where T : notnull` constraint on the public `StreamOne<T>`/`WriteSingle<T>` signatures can be dropped cleanly (see the separate constraint-reversal that ships in #5015 itself).
- Guard the `mt_version`-disabled case: if the doc type has version metadata turned off, skip emitting an ETag rather than emitting a constant `"00000000-…"` (which would risk false `304`s). Confirm what the inline read returns in that config.

### Acceptance

- `StreamOne` with `EmitETag = true` executes exactly one DB command (assert via command counting / query logging).
[...]

Issue #5010, opened 2026-07-22
ETag / If-None-Match (304) support on StreamOne, StreamMany and StreamAggregate


erdtsieck:
## Motivation

We converted the read-only detail endpoints of a healthcare claims platform to `StreamOne<T>` / `StreamAggregate<T>`. Frontends poll several of these detail endpoints (invoice details, upload feedback) while a user has the screen open. Streaming already removed the serialize cost, but every poll still ships the full body — invoices with hundreds of claim lines are easily 100KB+ — even when nothing changed.

Marten already stores exactly the change token we need, next to the data it streams:

- documents: the `mt_version` column (and `mt_last_modified`);
- event-sourced aggregates: the stream version `FetchLatest` resolves anyway.

So a conditional-GET fast path is one extra column in the select that is already happening.

## Proposal

Teach the streaming result types conditional requests:

1. On every response, set `ETag: "<version>"` (document `mt_version` for `StreamOne`/`StreamMany` single sources, stream version for `StreamAggregate`; optionally `Last-Modified` from `mt_last_modified`).
2. When the request carries `If-None-Match` matching the current version, respond `304 Not Modified` with an empty body.

For `StreamOne` the version can be read in the same query that fetches the document — a match means the body bytes are simply not written (one round-trip either way, zero bytes on the wire on a hit). For `StreamAggregate`, `FetchLatest` already knows the stream version before touching the snapshot, so a match can skip the snapshot/fold work entirely. For `StreamMany` a weak collection ETag (e.g. `max(mt_version)` + count over the filtered set) is a possible follow-up but has more caveats — fine to scope the first cut to the single-source types.

```csharp
// Endpoint code does not change:
[WolverineGet("/api/v1/institution-invoice/{id}")]
public static StreamOne<InstitutionInvoice> Get(string id, IQuerySession session)
    => new(session.Query<InstitutionInvoice>().Where(x => x.Id == id));

// First request:
//   200 OK, ETag: "17", body: {...}
// Poll while unchanged:
//   GET ... If-None-Match: "17"  ->  304 Not Modified, no body
```

Opt-out (or opt-in) flag on the result type if someone's contract cannot tolerate the extra headers:

```csharp
=> new(query) { EmitETag = false };
```

## Why in Marten and not user land

The version travels in Marten's own metadata columns; only the streaming write path can read it in the same round-trip as the data and decide not to write the body. User code would need a separate metadata query per request, which erases most of the win.

## Impact

[...]

PR #5166, opened 2026-08-04
Fix StreamOne streaming the document id as the body on an identity-tracking session

erdtsieck:
`StreamOne<T>` returns **200 with the document's id as the whole body** when the query runs on an identity-tracking session. Found on 9.22.3 in an application whose HTTP endpoints receive their session from Wolverine's Marten integration rather than a `QueryOnly` `IQuerySession`.

```
GET /api/v1/institution-invoice/{id}    200 OK
Content-Length: 38
ETag: "1"
Content-Type: application/json

"422c81ae-d73c-48ac-be1f-4eb65eefb606"
```

which surfaces at the caller as

```
System.Text.Json.JsonException: The JSON value could not be converted to InstitutionInvoice.
Path: $ | LineNumber: 0 | BytePositionInLine: 38
```

38 bytes is a quoted GUID, and that was the clue: the body is the id column.

## Cause

`VersionSelectClause<T>.innerFields()` aliases the payload to `data` — the name the streaming reader looks it up by — **by position**:

```csharp
var fields = Inner.SelectFields().ToArray();
fields[0] = $"{fields[0]} as {VersionSelectClause.DataAlias}";
```

on the strength of its own comment: *"`DocumentTable.SelectColumns` guarantees the payload is the first selected field ('the order of the selection is data, id, everything else')"*.

`SelectColumns` does not do that. It adds the id first whenever the id is selected:

```csharp
if (id != null) { columns.Remove(id); answer.Add(id); }
columns.Remove(data); answer.Add(data);
```

and `IdColumn.ShouldSelect` is `storageStyle != StorageStyle.QueryOnly`. So the select list starts `d.data` **only** for `QueryOnly`; through a lightweight, identity-map or dirty-tracking session it starts `d.id, d.data`. The alias landed on the id, `GetOrdinal("data")` found the id column, and `WriteJsonValueAsync` streamed the id.

The comment in `SelectColumns` is describing what *older code assumes*, not what the method guarantees — worth reading as a warning rather than a contract.

## Why the tests did not catch it

Every streaming endpoint in `IssueService` takes `IQuerySession`, which resolves to a `QueryOnly` session — the one storage style where the positional assumption happens to hold. So the entire ETag suite (#5010, #5015, #5027, #5120, #5157) exercised only the case that works.

This PR adds the missing shape: one endpoint streaming through an `IDocumentSession`, and a test that asserts the body is the document rather than its id. It fails on `main` with *"the body must be the document, not its id"*.

## Fix

Match the payload column by name instead of by position. The positional fallback stays for an inner clause that *projects* rather than selects the column — `SelectDataSelectClause`'s `jsonb_build_object(...)` under a `Select()` (#5158) — where there is no column name to match and the projection is the only candidate anyway. So both shapes that motivated the alias keep working, and the ETag assertion is in the new test to prove the fix did not cost the header this code path exists for.

```
Marten.AspNetCore.Testing: 114 passed, both TFMs
```

## Scope worth a second opinion

Only `StreamOne` goes through `VersionSelectClause`, so `StreamMany` / `StreamPaged` / `StreamPagedByCursor` are unaffected — they read `data` from an unaliased select list. `StreamAggregate` streams a live aggregation, not a document row. But `EmitETag` defaults to `true`, so on 9.22.3 every `StreamOne` endpoint that receives a tracking session returns the id instead of the document; ours was one endpoint of several and the others simply were not covered by a test that deserializes the body. That may be worth a patch release rather than waiting for the next feature drop.

### DIFF
--- a/src/IssueService/StreamingMinimalEndpoints.cs
+++ b/src/IssueService/StreamingMinimalEndpoints.cs
@@ -45,6 +45,15 @@ public static class StreamingMinimalEndpoints
                     ContentType = "application/vnd.marten.issue+json"
                 });
 
+        // Streamed through an identity-tracking session rather than the QueryOnly IQuerySession every
+        // other endpoint here uses. That is not a nicety: IdColumn.ShouldSelect is
+        // storageStyle != QueryOnly, so this is the only shape whose select list starts d.id, d.data, and
+        // it is the shape a Wolverine.Http endpoint gets when the Marten integration hands it a session.
+        // Aliasing the payload by position streamed the id as the whole body here.
+        app.MapGet("/minimal/issue/{id:guid}/from-document-session",
+            (Guid id, IDocumentSession session)
+                => new StreamOne<Issue>(session.Query<Issue>().Where(x => x.Id == id)));
+
         // EmitETag = false opt-out
         app.MapGet("/minimal/issue/{id:guid}/no-etag",
             (Guid id, IQuerySession session)
--- a/src/Marten.AspNetCore.Testing/streaming_result_types_tests.cs
+++ b/src/Marten.AspNetCore.Testing/streaming_result_types_tests.cs
@@ -121,6 +121,39 @@ public class streaming_result_types_tests: IntegrationContext
         });
     }
 
+    [Fact]
+    public async Task stream_one_streams_the_document_through_an_identity_tracking_session()
+    {
+        // Every other endpoint here takes the QueryOnly IQuerySession, where the payload happens to be the
+        // first selected field. IdColumn.ShouldSelect is storageStyle != QueryOnly, so an identity-tracking
+        // session selects d.id, d.data, ... instead — and aliasing the payload by position then put the
+        // alias on the id, so the reader streamed the document's id as the entire body: a 200 whose payload
+        // does not deserialize into the document type.
+        var issue = new Issue { Description = "through a document session", Open = true };
+        await using (var session = theHost.Services.GetRequiredService<IDocumentStore>().LightweightSession())
+        {
+            session.Store(issue);
+            await session.SaveChangesAsync();
+        }
+
+        var result = await theHost.Scenario(s =>
+        {
+            s.Get.Url($"/minimal/issue/{issue.Id}/from-document-session");
+            s.StatusCodeShouldBe(200);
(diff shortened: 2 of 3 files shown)


erdtsieck:
Sharpening the scope, because "identity-tracking session" undersells it: **every Wolverine.HTTP endpoint gets one, including the ones that ask for `IQuerySession`.**

Our endpoint is declared exactly the way the docs suggest:

```csharp
[WolverineGet("/api/v1/institution-invoice/{id}")]
public static StreamOne<InstitutionInvoice> Get(string id, IQuerySession session)
    => new(session.Query<InstitutionInvoice>().Where(x => x.Id == id));
```

and Wolverine's Marten integration generates this:

```csharp
await using var documentSession = _outboxedSessionFactory.OpenSession(messageContext, tenantId);
...
InstitutionInvoiceEndpoint.Get(id, ((Marten.IQuerySession)documentSession));
```

It opens a **document session** for the outbox and casts it to satisfy the parameter. So the storage style is never `QueryOnly` there, whatever the signature says, and it does not depend on `UseLightweightSessions()` either — the default `IdentityMap` session is equally `!= QueryOnly`.

Which makes the blast radius: on 9.22.3, **every `StreamOne` endpoint in a Wolverine.HTTP + Marten application returns the document's id instead of the document**, with a 200 and a correct `Content-Type`. `EmitETag` defaults to `true`, so nothing has to opt in.

Two things follow that I would not have written into the original description:

1. It fails **silently**. No exception server-side, correct status, correct content type — only a client that deserializes into the promised type notices. We caught it because two integration tests happen to call `ReadAsJsonAsync<T>()`; a test asserting `StatusCodeShouldBeOk()` alone would have passed it straight through to production.
2. `Marten.AspNetCore`'s own `IssueService` declares every streaming endpoint with `IQuerySession` and is *not* hosted by Wolverine, so its sessions really are `QueryOnly`. That is the one configuration where the positional assumption holds, and it is the only one the suite covers. The test in this PR adds the other one.

Workaround for anyone hitting this before a release: `EmitETag = false` on the affected endpoints takes the non-versioned path in `WriteSingle` and is unaffected.

I would suggest this warrants a patch rather than riding the next feature release — the failure mode is a wrong 200 on a default-on code path.

Issue #5153, opened 2026-08-03
Lift CompactStreamAsync<T> onto the shared operations surface, then a stream-compacting compliance suite


jeremydmiller:
Part of #5118 (compliance wave growth), epic #5110. **Cross-repo: the first half targets [jasperfx](https://github.com/jasperfx/jasperfx).**

Stream compacting is implemented in both products with the *same signature over the same shared request type*, but declared on each product's own operations interface rather than on JasperFx's:

| | Declaration |
|---|---|
| shared today | `JasperFx.Events/Protected/StreamCompactingRequest.cs:71` — `StreamCompactingRequest<T>` is **already a JasperFx type** |
| Marten | `Marten.Events.IEventStoreOperations.CompactStreamAsync<T>(Guid\|string, Action<StreamCompactingRequest<T>>?)` (`Marten/Events/EventStore.StreamCompacting.cs:35`) |
| Polecat | `Polecat.Events.IEventOperations.CompactStreamAsync<T>(Guid\|string, Action<StreamCompactingRequest<T>>?)` (`Polecat/Events/IEventOperations.cs:61`) |

`JasperFx.Events.IEventStore.CompactStreamAsync(Guid\|string, CancellationToken)` exists but is the store-level, non-generic, no-configuration form — and on Marten both overloads reflect back onto the generic method rather than being the real entry point. It is not what the tests exercise.

## Part 1 (jasperfx): lift the generic method

Declare `CompactStreamAsync<T>(Guid id, Action<StreamCompactingRequest<T>>? configure = null)` and the string overload on the shared `JasperFx.Events.IEventStoreOperations`. Both products already have the exact signature, so adoption is deleting a declaration, not writing an implementation. Confirm this is additive on both sides before shipping — it is a source-breaking change if either product's method differs in optionality or return type.

## Part 2 (compliance): the suite

Once the method is shared, `StreamCompactingCompliance` is portable with no fixture change.

| Marten tests today | Polecat tests today |
|---|---|
| `Aggregation/stream_compacting.cs` | `Events/stream_compacting_tests.cs` |

1. Compacting collapses events below the compaction point into a snapshot event and leaves the aggregate's folded state unchanged
2. `AggregateStreamAsync` after compaction equals the pre-compaction result
3. `FetchStreamAsync` returns the compacted shape; `FetchStreamStateAsync` still reports the true version
4. Compacting an already-compacted stream
5. Compacting a stream with an inline snapshot registered keeps the document consistent
[...]

PR #5240, opened 2026-08-15
Fix store level CompactStreamAsync throwing NullReferenceException

arnelirobles:
# Fix store level `CompactStreamAsync` throwing `NullReferenceException`

## What breaks

`IEventStore.CompactStreamAsync` on `IDocumentStore` throws `NullReferenceException` on every call,
both the `Guid streamId` and the `string streamKey` overload. There is no partial success and no
diagnostic: the caller gets a bare NRE pointing into `DocumentStore.EventStore.cs`.

Against `main` at f686ad7 (9.24.0), on JasperFx.Events 2.48.0:

```
System.NullReferenceException : Object reference not set to an instance of an object.
   at Marten.DocumentStore.JasperFx.Events.IEventStore.CompactStreamAsync(Guid streamId, CancellationToken token)
      in src/Marten/DocumentStore.EventStore.cs:line 713
```

and the same at line 729 for the string overload.

This is the self-contained maintenance entry point: no session in the signature, a
`CancellationToken`, documented in JasperFx.Events as *"Compact a stream by aggregating events into
a snapshot. Resolves aggregate type from stream state."* It is the overload an operator or a
background job reaches for, and it cannot work at all. Every existing compaction test goes through
the session level `IEventStoreOperations.CompactStreamAsync<T>` overload and calls
`SaveChangesAsync` itself (`src/EventSourcingTests/Aggregation/stream_compacting.cs:214`), which is
why nothing covered this.

## The two causes

`src/Marten/DocumentStore.EventStore.cs:711` and `:727`, identical in both overloads:

```csharp
var method = typeof(Marten.Events.IEventStoreOperations).GetMethod(nameof(Marten.Events.IEventStoreOperations.CompactStreamAsync),
    [typeof(Guid), typeof(Action<>).MakeGenericType(typeof(StreamCompactingRequest<>).MakeGenericType(state.AggregateType))]);
var genericMethod = method!.MakeGenericMethod(state.AggregateType);
```

`method` is always null, and the null-forgiving operator dereferences it. Two independent reasons,
both measured against 2.48.0:

**1. The declaration moved.** #5153 lifted both `CompactStreamAsync<T>` overloads off
`Marten.Events.IEventStoreOperations` and onto `JasperFx.Events.IEventStoreOperations`. The comment
left behind at `src/Marten/Events/IEventOperations.cs:21-25` records the move. `Type.GetMethod` and
`GetMethods` on an *interface* do not walk base interfaces, so the derived interface finds nothing:

```
typeof(Marten.Events.IEventStoreOperations).GetMethods().Count(m => m.Name == "CompactStreamAsync")   => 0
typeof(JasperFx.Events.IEventStoreOperations).GetMethods().Count(m => m.Name == "CompactStreamAsync") => 2
```

**2. The overload used cannot match.** The lookup builds a *closed*
`Action<StreamCompactingRequest<TAggregate>>` and matches it against the *open generic definition*
`CompactStreamAsync<T>(Guid, Action<StreamCompactingRequest<T>>)`. The closed-parameter-signature
overload of `GetMethod` cannot match an open generic method definition. Measured: the same
`GetMethod` call against `JasperFx.Events.IEventStoreOperations` also returns null.

So a fix that corrects only the declaring type still returns null and still throws. Both have to go.

**3. A third defect sits behind the null**, and only surfaces once the lookup works. The session
level `CompactStreamAsync<T>` only *queues* a `ReplaceEventOperation<T>` and a
`DeleteEventsOperation` (`src/Marten/Events/EventStore.StreamCompacting.cs:28-32` and `:130-132`);
it does not commit. This method disposes its own `LightweightSession` and never calls
`SaveChangesAsync`, so a lookup-only fix compacts nothing, silently. The `token` was also dropped
after `FetchStreamStateAsync`, so the fetch, the aggregation and the archiver callback inside
`StreamCompactingExecution.ExecuteAsync` all ran uncancellable.

## The fix

Both overloads dispatch through a private generic helper on `DocumentStore` instead of reflecting
into the interface:

```csharp
private static readonly MethodInfo _compactByStreamId =
    typeof(DocumentStore).GetMethod(nameof(compactStreamAsync), BindingFlags.NonPublic | BindingFlags.Static)!;

private static async Task compactStreamAsync<T>(IDocumentSession session, Guid streamId, CancellationToken token)
    where T : class
{
    await session.Events.CompactStreamAsync<T>(streamId, x => x.CancellationToken = token).ConfigureAwait(false);
    await session.SaveChangesAsync(token).ConfigureAwait(false);
}
```

Both causes are gone because the reflected signature is one Marten owns rather than an interface in
a dependency, so the next declaration move cannot break this call site silently the way #5153 did.
It also puts the `SaveChangesAsync` and the `CancellationToken` back in ordinary compiled code
instead of trying to build an `Action<StreamCompactingRequest<T>>` reflectively for an aggregate
type only known at runtime. The `MethodInfo` lookups are resolved once into static readonly fields.

Nothing else changed. `using JasperFx.Events.Protected;` came out of
`DocumentStore.EventStore.cs` because `StreamCompactingRequest<>` was its only reference.

## What the tests prove

Two end to end tests in `src/EventSourcingTests/Aggregation/stream_compacting.cs`, beside the
existing session level ones:

- `store_level_compaction_by_guid_identity`
- `store_level_compaction_by_string_identity`

Each starts a nine event stream, calls `((IEventStore)theStore).CompactStreamAsync(...)`, then
re-reads the stream **through a fresh session**. That last part is deliberate: it fails on the
missing `SaveChangesAsync` as well as on the lookup, so cause 3 cannot regress back in unnoticed.
Both fail on `main` with the `NullReferenceException` quoted above.

## Verification

`src/EventSourcingTests` on net10.0 against Postgres 17: **1852 passed, 7 skipped, 0 failed**,
including the two new tests. Postgres came from the repo's own `docker-compose.yml`.

## One open question

I kept reflection here because the aggregate type is only known from stream state at runtime, and
narrowed it to a signature Marten controls. If you would rather remove the reflection entirely, say
by caching a `Func<IDocumentSession, Guid, CancellationToken, Task>` per aggregate type off the
projection registry, I am happy to switch it in this PR.

### DIFF
--- a/src/EventSourcingTests/Aggregation/stream_compacting.cs
+++ b/src/EventSourcingTests/Aggregation/stream_compacting.cs
@@ -312,6 +312,54 @@ public class stream_compacting : OneOffConfigurationsContext
     }
 
     [Fact]
+    public async Task store_level_compaction_by_guid_identity()
+    {
+        StoreOptions(opts => opts.Projections.Snapshot<Letters>(SnapshotLifecycle.Inline));
+
+        var streamId = Guid.NewGuid();
+        theSession.Events.StartStream<Letters>(streamId, A(), B(), A(), C(), C(), D(), D(), A(), A());
+        await theSession.SaveChangesAsync();
+
+        // The store level overload resolves the aggregate type off the stream state
+        // and owns its own session, so it has to commit for itself
+        await ((IEventStore)theStore).CompactStreamAsync(streamId, CancellationToken.None);
+
+        await using var session = theStore.LightweightSession();
+        var events = await session.Events.FetchStreamAsync(streamId);
+        var compacted = events.Single().ShouldBeOfType<Event<Compacted<Letters>>>();
+        compacted.Version.ShouldBe(9);
+        compacted.Data.Snapshot.ACount.ShouldBe(4);
+        compacted.Data.Snapshot.BCount.ShouldBe(1);
+        compacted.Data.Snapshot.CCount.ShouldBe(2);
+        compacted.Data.Snapshot.DCount.ShouldBe(2);
+    }
+
+    [Fact]
+    public async Task store_level_compaction_by_string_identity()
+    {
+        StoreOptions(opts =>
+        {
+            opts.Events.StreamIdentity = StreamIdentity.AsString;
+            opts.Projections.Add<LetterCountsByStringProjection>(ProjectionLifecycle.Inline);
+        });
+
+        var streamKey = Guid.NewGuid().ToString();
+        theSession.Events.StartStream<LetterCountsByString>(streamKey, A(), B(), A(), C(), C(), D(), D(), A(), A());
+        await theSession.SaveChangesAsync();
+
+        await ((IEventStore)theStore).CompactStreamAsync(streamKey, CancellationToken.None);
+
+        await using var session = theStore.LightweightSession();
+        var events = await session.Events.FetchStreamAsync(streamKey);
+        var compacted = events.Single().ShouldBeOfType<Event<Compacted<LetterCountsByString>>>();
+        compacted.Version.ShouldBe(9);
+        compacted.Data.Snapshot.ACount.ShouldBe(4);
+        compacted.Data.Snapshot.BCount.ShouldBe(1);
+        compacted.Data.Snapshot.CCount.ShouldBe(2);
+        compacted.Data.Snapshot.DCount.ShouldBe(2);
+    }
+
+    [Fact]
     public async Task end_to_end_string_identification_at_specified_version()
     {
         StoreOptions(opts =>
--- a/src/Marten/DocumentStore.EventStore.cs
+++ b/src/Marten/DocumentStore.EventStore.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
 using System.Linq;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using JasperFx;
@@ -14,7 +15,6 @@ using JasperFx.Events.Aggregation;
 using JasperFx.Events.Daemon;
 using JasperFx.Events.Descriptors;
 using JasperFx.Events.Projections;
-using JasperFx.Events.Protected;
 using Marten.Events;
 using Marten.Events.Archiving;
 using Marten.Events.Daemon;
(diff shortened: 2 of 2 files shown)


jeremydmiller:
Reviewed this against the source rather than the decompilation — the diagnosis is right, and it fixes more than the title claims. 👍

## The reflection break

Confirmed each link:

- `Marten.Events.IEventStoreOperations` no longer declares `CompactStreamAsync` — #5153/#5194 lifted both overloads onto `JasperFx.Events.IEventStoreOperations`, leaving only the derivation. `Type.GetMethod` does not walk base interfaces, so the old lookup returned `null` and `method!.MakeGenericMethod` threw.
- The secondary claim in the comment is also correct, and worth having written down: retargeting the lookup at the *moved* interface would still have returned `null`, because `GetMethod(name, Type[])` cannot match an open generic method definition with closed parameter types. Reflecting at a private helper this class owns is the right way out.

## The part the title undersells

`StreamCompactingExecution.ExecuteAsync` only *queues* — `CompletelyReplaceEvent` plus `QueueOperation(new DeleteEventsOperation(sequences))` — and never saves. So the store-level overload opened a `LightweightSession`, would have queued into it, and disposed without committing. **Even with the reflection repaired, that overload was a silent no-op.** The added `SaveChangesAsync` is the second half of the bug, not tidying, and the asymmetry with the session-level overload (which correctly leaves the save to its caller) is the right call.

The `x => x.CancellationToken = token` is a real fix too — the old call passed `null` for `configure`, so the token the store-level overload accepts was being discarded.

## Why it shipped

`DocumentStore.EventStore.cs` had not been touched since #5182, and there is no store-level compaction test anywhere in the tree — `grep` for `(IEventStore)theStore` turns up only the explorer and multi-tenancy suites. #5194 moved an interface and this reflection broke with no compile error and no failing test. The two tests here close exactly that gap, which is the durable part of the change.

## Minor, none blocking

- The two helpers are near-duplicates, but the `Guid`/`string` split is forced by the overload pair — fine as is.
- Neither test asserts the token propagation, so that half of the fix is unverified. Not worth holding the PR for.
- Pre-existing and untouched, noting it only so it is not rediscovered as a regression from this change: calling the `string` overload against a Guid-identified store still NREs inside `ExecuteAsync` on `request.StreamId!.Value`, because the identity branch keys off `StreamIdentity` rather than which overload was called. Same on the session-level path. Deserves its own issue.

jeremydmiller:
Correction to the last bullet of my review above, now that I've run it rather than only read it.

I said the `string` overload against a Guid-identified store "still NREs inside `ExecuteAsync` on `request.StreamId!.Value`". The mechanism is right, the exception type is not — `Nullable<T>.Value` throws `InvalidOperationException` ("Nullable object must have a value"), not `NullReferenceException`. And the store-level path does not reach it at all: `FetchStreamStateAsync` returns null first, so the `state?.AggregateType == null` guard fires with "stream not found or no aggregate type associated" — wrong diagnosis, but not a crash.

The more interesting half turned out to be the direction I did not check: `CompactStreamAsync<T>(Guid)` against a **string**-identified store throws nothing at all. It takes the `AsString` branch, reads a null `StreamKey`, matches no stream, and returns at `if (!events.Any()) return;` — a silent no-op, with no way for a caller to distinguish it from a completed compaction.

Filed as #5244 with all four paths, fix in #5245. None of it is a regression from this PR.

Issue #5062, opened 2026-07-29
mt_quick_append_events returns {NULL} for an empty event list, masking the original exception


erdtsieck:
### Summary

With `QuickAppend` + tenant-partitioned events, `mt_quick_append_events` returns `{NULL}` when it is called with an **empty** event array. Npgsql then fails to read the result into `long[]`:

```
System.InvalidCastException: Cannot read a non-nullable collection of elements because the
returned array contains nulls. Call GetFieldValue with a nullable collection type instead.
```

Because the empty append happens on the tombstone path (`mt_tombstone`), this exception **replaces the original exception** that caused the append to fail in the first place. The real error never reaches the caller, and what surfaces is a non-retryable `InvalidCastException` with no relation to the actual problem. In our case this dead-lettered messages in Wolverine that would otherwise have been retried, and it cost us a long time to trace back.

### Cause

The first element of the returned array is computed as:

```sql
return_value := ARRAY[event_version + array_length(event_ids, 1)];
```

`array_length('{}', 1)` is `NULL` in PostgreSQL, so `event_version + NULL` is `NULL`, and the function returns a `bigint[]` whose only element is `NULL`.

### Reproduction

Against a database with a Marten schema created with `Events.UseTenantPartitionedEvents = true` and a registered tenant:

```sql
BEGIN;
select mt_quick_append_events(
  'mt_tombstone', 'Tombstone', '<a registered tenant id>',
  '{}'::uuid[], '{}'::varchar[], '{}'::varchar[], '{}'::jsonb[], '{}'::bytea[],
  '{}'::varchar[], '{}'::varchar[], '{}'::jsonb[], '{}'::timestamptz[]) as returned;
ROLLBACK;
```

```
 returned
----------
 {NULL}
```

Reading that value with `reader.GetFieldValue<long[]>(0)` throws the `InvalidCastException` above.

### Suggested fix

Treat an empty event array as zero appended events, e.g.:

```sql
return_value := ARRAY[event_version + COALESCE(array_length(event_ids, 1), 0)];
```

Guarding the caller against issuing an append with no events would help too, but the `COALESCE` alone makes the function correct for the empty case and stops it from masking the original exception.

### Environment

- Marten 9.20.0, JasperFx 2.36.1
- PostgreSQL 15/18
- `Events.TenancyStyle = Conjoined`, `Events.UseTenantPartitionedEvents = true`, `AppendMode` = Quick (default)
- Consumed from Wolverine 6.23.0 handlers

PR #5263, opened 2026-08-19
Retried batch loses its events: pooled parameter buffers were released after the first execution (#5262)

### LINKED ISSUE #5262: A retried batch sends empty event arrays: SaveChangesAsync commits the documents, silently drops the events, and reports success
## Summary

On a production incident we found a `SaveChangesAsync()` that

1. failed with an Npgsql read timeout while executing its operation batch,
2. **persisted part of that batch anyway** — all document inserts/deletes committed, but only 8 of 69 appended events, and
3. **never surfaced an exception to the caller** — Wolverine logged the message as successfully processed, one attempt, no retry, no dead letter.

The application only noticed five days later, because the missing events were the ones that remove a work item from a read model, so the item kept coming back on every projection rebuild.

I can't reproduce it on demand (it needs a batch that actually times out mid-flight), so this is a forensic report rather than a repro. Everything below is measured from the event store and the container logs, not inferred.

## Environment

| | |
|---|---|
| Marten | 9.22.6 |
| JasperFx / JasperFx.Events | 2.46.0 |
| Npgsql | 10.0.3 |
| .NET | 10 |
| PostgreSQL | 16 (AWS RDS) |
| `EventAppendMode` | `QuickWithServerTimestamps` (the Marten 9 default) |
| Event tenancy | `TenancyStyle.Conjoined` + `UseTenantPartitionedEvents = true` |
| Tenancy | multi-database, ~512 shard databases, tenants assigned via a database pool table |
| Session | opened by Wolverine's `OutboxedSessionFactory` → `store.OpenSession(new SessionOptions { Tracking = DocumentTracking.None })` |
| Connection lifetime | `AutoClosingLifetime` (per `SessionOptions.buildConnectionLifetime`: `OwnsConnection && OwnsTransactionLifecycle`, `ReadCommitted`, `UseStickyConnectionLifetimes = false`) |

## What the code does

A message handler loads N pending read-model documents, appends one event per document to the document's own stream, deletes the documents, and inserts one guard document per line — all on one session, committed once. Simplified:

```csharp
public static void After(IReadOnlyList<PendingLine> lines, IDocumentSession session)
{
    foreach (var group in lines.GroupBy(x => x.StreamKey))
    {
        session.Events.Append(group.Key, events: group.Select(x => new PreparedForInvoice { LineId = x.Id, Amount = x.Amount }));
    }

    session.DeleteObjects(lines);

    foreach (var line in lines)
    {
        session.Insert(new ConcurrencyGuard { Id = line.Id });
    }
}
```

Wolverine's generated handler shows there is exactly one session and exactly one `SaveChangesAsync` — no second commit anywhere:

```csharp
public override async Task HandleAsync(MessageContext context, CancellationToken cancellation)
{
    await using var serviceScope = _serviceScopeFactory.CreateAsyncScope();
    // Building the Marten session
    await using var documentSession = _outboxedSessionFactory.OpenSession(context);
    var message = (InvoiceLines)context.Envelope.Message;

    (var lines, var config, var routing) = await Handler.LoadAsync(message, documentSession, ...);

    var outgoing1 = Handler.Handle(message, config, routing, lines);

    // Outgoing, cascaded message
    await context.EnqueueCascadingAsync(outgoing1).ConfigureAwait(false);

    RecordCauseAndEffect(context, context.Runtime.Observer);
    Handler.After(lines, documentSession);

    // Save all pending changes to this Marten session
    await documentSession.SaveChangesAsync(cancellation).ConfigureAwait(false);

    await context.FlushOutgoingMessagesAsync().ConfigureAwait(false);
}
```

For this particular run, N = 69 lines spread over 65 streams.

## The failure

`After` created the 69 event objects at `13:19:50.114443Z` (client-assigned timestamps — `QuickWithServerTimestamps` sends `timestamps[]` from the client, per the note in `QuickAppendEventsWithServerTimestampsOperation`). `SaveChangesAsync` then failed ~5.3 s later.

Marten logged the failure through `LogFailure(NpgsqlBatch, Exception)`, which emits one `LogError` per `batch.BatchCommands` entry — so the 215 error records below are **one** batch failure, and they give an exact inventory of the batch:

| operation | count | log window (UTC) |
|---|---|---|
| `select public.mt_quick_append_events($1,…)` | **65** | 13:19:55.390–.403 |
| `insert into public.wolverine_incoming_envelopes` | 10 | .404–.405 |
| `delete from public.mt_doc_<pending>` | **69** | .406–.445 |
| `insert into public.mt_doc_<guard>` | **69** | .445–.459 |
| `update public.wolverine_incoming_envelopes` | 1 | .460 |
| (1 more) | 1 | .461 |

Every one of them carried the same exception:

```
Npgsql.NpgsqlException (0x80004005): Exception while reading from stream
 ---> System.TimeoutException: Timeout during reading attempt
   at Npgsql.Internal.NpgsqlConnector.ReadMessageLong(Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
   at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
   at Npgsql.NpgsqlConnector…
```

### Why it timed out at 5 s (probably worth its own look)

`AutoClosingLifetime.ExecuteBatchPagesAsync` sets `batch.Timeout = CommandTimeout`, which comes from `StoreOptions.CommandTimeout`. That property defaults to `StoreOptions.DefaultTimeout = 5` and is only raised from a connection string in `StoreOptions.Connection(...)` (`StoreOptions.cs:610`).

In a multi-database tenancy setup the shard connection strings never pass through that code path, so `Command Timeout=300` on every shard connection string had no effect and the effective batch timeout stayed at **5 seconds** — for a batch of 215 operations against a loaded database. If that is intended, it is a sharp edge: the connection string of the database the batch actually runs against is ignored, and there is no warning.

## What actually persisted

Verified in the database five days later:

| | expected | present |
|---|---|---|
| appended events | 69 | **8** |
| document deletes | 69 | 69 (all gone) |
| guard document inserts | 69 | 69 |
| `wolverine_incoming_envelopes` writes | 11 | 11 (all 10 cascaded messages were delivered and handled) |

Two details that pin this down:

- The 8 surviving events carry the client timestamps from *this* attempt (`13:19:50.114443Z`/`.114444Z`), so they are that batch's own event objects, not a later retry's.
- The guard documents carry `mt_last_modified = 13:19:55.502928Z` (their insert SQL uses `transaction_timestamp()`), i.e. ~40 ms **after** the last of the 215 error records was written. So something committed after the failure was logged.
- Sequence numbers are contiguous around the survivors (the 8 events sit on consecutive `seq_id`s, immediately followed by the next unrelated events) — the 57 missing appends never drew a sequence number, and there are no tombstones in that range.

## What did not happen

No exception reached the caller. Wolverine logged `Finished processing` for that envelope 4 s later, exactly one `Started processing`, no retry and no dead letter — and the application has explicit retry policies for `MartenCommandException`, `ConcurrencyException`, `EventStreamUnexpectedMaxEventIdException`, `DocumentAlreadyExistsException` and `AggregateException`, so a thrown exception would have been visible.

Reading the two relevant lifetimes, both look like they should have rolled back and rethrown:

- `AutoClosingLifetime.ExecuteBatchPagesAsync` — `try { tx = BeginTransactionAsync(); foreach page { … } } catch { tx.RollbackAsync(); Logger.LogFailure(new NpgsqlCommand(), e); TransformAndThrow(e); }`
- `TransactionalConnection.ExecuteBatchPagesAsync` — same shape, plus `if (exceptions.Count == 1) → rollback + rethrow` and `if (exceptions.Any()) → rollback + AggregateException`.

So the two things I cannot explain from the code are exactly the two that hurt:

1. **How a single transaction ended up partially committed** — 8 of 65 `mt_quick_append_events` calls, but all 138 document operations and all 11 envelope operations. A `ROLLBACK` on a broken connector should have left nothing, and a partial commit should not be reachable at all.
2. **How `SaveChangesAsync` returned normally** after `LogFailure` had been called for all 215 commands.

Is it possible for the batch pages to be executed outside the transaction that `ExecuteBatchPagesAsync` begins — or for a broken-connector rollback path to swallow the original exception and return? A `Timeout during reading attempt` leaves the connector in a state where the rollback itself fails; `AutoClosingLifetime` has an inner `catch (Exception rollbackEx)` around the rollback that logs and rethrows, but `TransactionalConnection` does not, so there at least the rollback failure would replace the original exception.

## Why this is nasty in practice

The write was split across two kinds of storage that the application assumed were atomic together. The document side (delete the work item, claim the line) committed; the event side (record that the work item was handled) did not. Because read models are rebuilt from events, every later projection rebuild resurrected work items that had already been processed — including some already sent to an external party. Nothing anywhere reported a failure.

Happy to run more queries against the affected database or to try building a repro with a deliberately short `CommandTimeout` and a slow trigger, if that would help.

erdtsieck:
Fixes #5262. Two commits, and the second one is the bug that loses data — take that one even if you disagree with the first.

## Commit 2 — a retried batch sends empty event arrays

`OperationPage` calls `ConfigureCommand` **once**, in its constructor; `Compile()` only hands back the already-built `NpgsqlBatch`. So when `Options.ResiliencePipeline` re-executes a page, the same parameter objects go over the wire again.

`QuickAppendEventsOperationBase` rents its parallel column buffers from `ArrayPool` through `PooledList<T>` and released them in `PostprocessAsync`'s `finally`, on the stated assumption that "parameter bytes hit the wire before `ExecuteReaderAsync` returned, so the rentals are safe to release here". True for one execution. On a retry, the released `PooledList` reports `Count = 0` — and `Count` is what Npgsql uses for the array length — so the append sends empty `event_ids` / `event_types` / bodies arrays. `mt_quick_append_events` with an empty array writes nothing, bumps no version and raises no error.

Net effect: the retry's transaction commits the documents, silently skips the events, and `SaveChangesAsync` returns **success**.

`Bug_5262_retry_loses_events` reproduces it on master — first attempt times out, second attempt succeeds:

```
exception from SaveChangesAsync : <none>
batches executed                : 2
attempt counter (nextval)       : 2
persisted                       : 0 events, 30 documents, 0 tombstones
expected                        : 30 events, 30 documents
```

The fix moves ownership of the rentals from a single execution to the unit of work: the operation implements `IDisposable`, `PostprocessAsync` no longer releases, and `SaveChanges` disposes the pages' operations in a `finally` — exactly once, whatever the outcome.

This mechanism also predicted the production data. In the incident, 8 of 69 events survived a retried batch. Since `ApplyCallbacksAsync` walks the operations in order, only the appends it had *not* reached before the timeout still held their payload, so the survivors had to be the last ones in batch order. They were, exactly: positions 58–65 of 65.

## Commit 1 — do not retry a write whose outcome is unknown

Separate concern, and optional as far as I'm concerned. The default policy handles every `NpgsqlException` with 3 retries, and a command timeout is one of those. A client-side read timeout cannot tell a rolled-back transaction from a committed one, so replaying a non-idempotent batch there risks appending the same events twice.

Committing goes through a `WriteResiliencePipeline` built from `AddMartenWriteDefaults()`: same shape, same attempt count, one predicate difference — no retry when the outcome is in doubt.

```csharp
internal static bool IsOutcomeInDoubt(Exception exception)
{
    return exception switch
    {
        PostgresException => false,   // the server answered, so the transaction is definitively aborted
        NpgsqlException => true,      // I/O level: timeout, dropped connection - unknown
        TimeoutException => true,
        AggregateException aggregate => aggregate.InnerExceptions.Any(IsOutcomeInDoubt),
        { InnerException: { } inner } => IsOutcomeInDoubt(inner),
        _ => false
    };
}
```

Everything that retried before still retries: errors PostgreSQL itself reported (serialization failures, deadlocks, constraint violations) leave the transaction definitively aborted, and so does a failure raised in reader post-processing, which `ExecuteBatchPagesAsync` rolls back itself. `CoreTests.retry_mechanism` covers that second case and stays green — its `MartenCommandException` wraps a plain `Exception`, not an I/O failure.

The read path (`MartenDatabase.Execution.cs:58`) keeps the unchanged `ResiliencePipeline`; retrying an idempotent read after a connection blip is useful. `ConfigurePolly` sets both pipelines so a caller who takes over keeps full control; `ExtendPolly` layers the user's policies over the respective defaults.

## Commit order is red, green, red, green

Each test lands in its own commit and is failing at that point, so you can check out either test commit and watch it fail before the fix that follows it:

| commit | |
|---|---|
| `2d006d7c9` | failing test — a successful retry commits its documents but loses its events |
| `10f9511c2` | fix — the pooled parameter buffers belong to the unit of work |
| `b9bdc166e` | failing test — a timed-out batch is retried in full, appends included |
| `d3cabc6ea` | fix — do not retry a unit of work whose outcome is unknown |

Verified at every step, and the first test is red against pristine `origin/master`:

```
exception from SaveChangesAsync : <none>
batches executed                : 2
persisted                       : 0 events, 30 documents, 0 tombstones
expected                        : 30 events, 30 documents
```

## Verification

PostgreSQL 18, `-f net10.0`:

| suite | total | failed | note |
|---|---|---|---|
| `CoreTests` baseline | 548 | 7 | includes the two new tests |
| `CoreTests` with both commits | 549 | 5 | all 5 also fail on the baseline |
| `EventSourcingTests` baseline | 1933 | 5 | `Projections.testing_projections.*` |
| `EventSourcingTests` with both commits | 1933 | 5 | same 5 |

## Corrections to the original issue

The issue as I first filed it claimed a partially committed transaction. That was wrong — `SaveChangesAsync` is one transaction per attempt and it is all-or-nothing. I have rectified that on the issue; the real mechanism is the recycled buffers above.

### DIFF
--- a/docs/configuration/retries.md
+++ b/docs/configuration/retries.md
@@ -28,6 +28,19 @@ return builder
 The general idea is to have _some_ level of retry with an exponential backoff on typical transient errors encountered
 in database usage (network hiccups, a database being too busy, etc.).
 
+Committing a unit of work is held to a stricter policy. A batch of document and event operations is
+**not idempotent** — replaying it appends the same events a second time — so `SaveChangesAsync()` only
+retries when the previous transaction is known to be gone. A command timeout or a dropped connection
+kills the connector, which means the `ROLLBACK` never reaches PostgreSQL and the server may well have
+committed while the response never made it back to the client. Retrying in that situation is what turns
+one lost commit into a duplicated one, so Marten does not do it:
+
+<!-- snippet: sample_default_write_polly_setup -->
+<!-- endSnippet -->
+
+Errors that PostgreSQL itself reports (a serialization failure, a deadlock, a constraint violation) do
+leave the transaction definitively aborted, and are still retried.
+
 You can **replace** Marten's Polly configuration through:
 
 <!-- snippet: sample_configure_polly -->
--- a/src/CoreTests/Bugs/Bug_5262_partial_commit_on_batch_timeout.cs
+++ b/src/CoreTests/Bugs/Bug_5262_partial_commit_on_batch_timeout.cs
@@ -1,0 +1,197 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using Marten;
+using Marten.Services;
+using Marten.Testing.Harness;
+using Npgsql;
+using Shouldly;
+using Xunit;
+
+namespace CoreTests.Bugs;
+
+/// <summary>
+/// Repro for #5262. A SaveChangesAsync that mixes event appends with document writes and blows the
+/// command timeout
+///
+/// 1. is retried in full by <c>Options.ResiliencePipeline</c> (DocumentSessionBase.SaveChanges.cs:170) -
+///    including the event appends, which are not idempotent, and
+/// 2. leaves state behind - mt_streams.version is bumped for streams whose events were never written.
+///
+/// (2) is the exact fingerprint we see in production: streams whose <c>version</c> runs ahead of their
+/// own event count, on the worst one by 216.
+/// </summary>
+public class Bug_5262_partial_commit_on_batch_timeout: OneOffConfigurationsContext
+{
+    private readonly ITestOutputHelper _output;
+
+    public Bug_5262_partial_commit_on_batch_timeout(ITestOutputHelper output)
+    {
+        _output = output;
+    }
+
+    public class Doc
+    {
+        public Guid Id { get; set; }
+        public string Name { get; set; }
+    }
+
+    public record StreamStarted(string Name);
+
+    public record SomethingHappened(string Name);
+
+    private const int Count = 30;
+
+    [Fact]
+    public async Task batch_timeout_rolls_everything_back_and_runs_the_batch_once()
+    {
+        StoreOptions(opts =>
+        {
+            opts.CommandTimeout = 1; // seconds - the Marten default is 5
+            opts.Events.StreamIdentity = StreamIdentity.AsString;
+            opts.Events.AppendMode = EventAppendMode.QuickWithServerTimestamps; // the Marten 9 default
+        });
+
+        var streamKeys = Enumerable.Range(0, Count).Select(i => $"stream-{i}").ToArray();
+
+        // Arrange: the streams already exist, exactly like the production case. Appends to an existing
+        // stream go through mt_quick_append_events.
+        await using (var setup = theStore.LightweightSession())
+        {
+            foreach (var key in streamKeys)
+            {
+                setup.Events.StartStream(key, new StreamStarted(key));
+            }
+
+            await setup.SaveChangesAsync();
+        }
+
+        var logger = new RecordingLogger();
+
+        await using var session = theStore.LightweightSession();
+        session.Logger = logger;
+
+        for (var i = 0; i < Count; i++)
+        {
+            session.Events.Append(streamKeys[i], new SomethingHappened($"event-{i}"));
+            session.Store(new Doc { Id = Guid.NewGuid(), Name = $"doc-{i}" });
+        }
+
+        // Pushes the batch past the 1 second command timeout. Event operations go first in the batch
+        // (UnitOfWork.AllOperations), so the appends have already been sent when this one stalls.
+        session.QueueSqlCommand("select pg_sleep(3)");
+
+        Exception caught = null;
+        try
+        {
+            await session.SaveChangesAsync();
+        }
+        catch (Exception e)
+        {
+            caught = e;
+        }
+
+        _output.WriteLine($"exception  : {caught?.GetType().FullName ?? "<none>"} / inner {caught?.InnerException?.GetType().Name}");
+        _output.WriteLine($"batches executed (OnBeforeExecute) : {logger.BatchesStarted}");
+        _output.WriteLine($"batch commands logged as failed    : {logger.FailedCommands.Count}");
+        foreach (var group in logger.FailedCommands.GroupBy(NormalizeSql).OrderByDescending(g => g.Count()))
+        {
+            _output.WriteLine($"  {group.Count(),4} x {group.Key}");
+        }
+
+        // Measure on a fresh connection, bypassing Marten entirely.
+        var (rows, extraEvents, docs, tombstones) = await MeasureAsync();
+
+        _output.WriteLine($"after the failure: {extraEvents} SomethingHappened events, {docs} documents, {tombstones} tombstones");
+        _output.WriteLine("streams whose version does not match their own event count:");
+        var mismatched = rows.Where(x => x.Id != "mt_tombstone" && x.Version != x.Events).ToList();
+        foreach (var row in mismatched)
+        {
+            _output.WriteLine($"  {row.Id}: mt_streams.version = {row.Version}, actual events = {row.Events}");
+        }
+
+        // Marten's contract: one transaction, so all of it or none of it. And the batch belongs to the
+        // caller to retry - re-appending events behind their back is not safe.
+        caught.ShouldNotBeNull("SaveChangesAsync returned normally after the batch failed");
+        logger.BatchesStarted.ShouldBe(1);
+
+        extraEvents.ShouldBe(0);
+        docs.ShouldBe(0);
+        mismatched.ShouldBeEmpty();
+    }
+
+    private async Task<(List<StreamRow>, int, int, int)> MeasureAsync()
+    {
+        await using var conn = new NpgsqlConnection(ConnectionSource.ConnectionString);
+        await conn.OpenAsync();
+
+        var rows = new List<StreamRow>();
+        await using (var cmd = conn.CreateCommand())
+        {
+            cmd.CommandText = $"""
+                               select s.id, s.version, (select count(*) from {SchemaName}.mt_events e where e.stream_id = s.id)
+                               from {SchemaName}.mt_streams s order by s.id
+                               """;
+            await using var reader = await cmd.ExecuteReaderAsync();
+            while (await reader.ReadAsync())
+            {
+                rows.Add(new StreamRow
+                {
+                    Id = reader.GetString(0), Version = reader.GetInt64(1), Events = reader.GetInt64(2)
+                });
+            }
+        }
+
+        async Task<int> ScalarAsync(string sql)
+        {
+            await using var cmd = conn.CreateCommand();
+            cmd.CommandText = sql;
+            return Convert.ToInt32(await cmd.ExecuteScalarAsync());
+        }
+
+        var extraEvents = await ScalarAsync($"select count(*) from {SchemaName}.mt_events where type = 'something_happened'");
+        var docs = await ScalarAsync($"select count(*) from {SchemaName}.mt_doc_bug_5262_partial_commit_on_batch_timeout_doc");
+        var tombstones = await ScalarAsync($"select count(*) from {SchemaName}.mt_events where type = 'tombstone'");
+
+        return (rows, extraEvents, docs, tombstones);
+    }
+
+    public class StreamRow
+    {
+        public string Id { get; set; }
+        public long Version { get; set; }
+        public long Events { get; set; }
+    }
+
+    private static string NormalizeSql(string sql)
+    {
+        var trimmed = sql.Trim().ReplaceLineEndings(" ");
+        return trimmed.Length <= 70 ? trimmed : trimmed[..70];
+    }
+
+    public class RecordingLogger: IMartenSessionLogger
+    {
+        public List<string> FailedCommands { get; } = new();
+        public int BatchesStarted { get; private set; }
+
+        public void LogSuccess(NpgsqlCommand command) { }
+        public void LogSuccess(NpgsqlBatch batch) { }
+
+        public void LogFailure(NpgsqlCommand command, Exception ex) => FailedCommands.Add(command.CommandText);
+
+        public void LogFailure(NpgsqlBatch batch, Exception ex)
+        {
+            foreach (var command in batch.BatchCommands)
+            {
+                FailedCommands.Add(command.CommandText);
+            }
+        }
+
+        public void LogFailure(Exception ex, string message) { }
+        public void RecordSavedChanges(IDocumentSession session, IChangeSet commit) { }
+        public void OnBeforeExecute(NpgsqlCommand command) { }
+        public void OnBeforeExecute(NpgsqlBatch batch) => BatchesStarted++;
+    }
+}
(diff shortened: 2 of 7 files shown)


jeremydmiller:
Merged the buffer-lifetime half to master — your two commits cherry-picked with authorship intact, at 7939baf84 and 88ba29a5c, plus 40686a8c3 for two cosmetic things (the commit added a UTF-8 BOM to `QuickAppendEventsOperationBase.cs` that master did not have, and the re-indent carried forward a `#5062` comment block that had been duplicated in place).

Verified the red/green here rather than taking it on trust. At 7939baf84 — your test, no fix:

```
Failed CoreTests.Bugs.Bug_5262_retry_loses_events.a_successful_retry_still_writes_its_events
   Shouldly.ShouldAssertException : events
 batches executed                : 2
 persisted                       : 0 events, 30 documents, 0 tombstones
 expected                        : 30 events, 30 documents
```

Your numbers exactly. With the fix, and on PostgreSQL 18 / net10.0: `CoreTests` 541 passed / 0 failed, `EventSourcingTests` 1926 passed / 0 failed. Cleaner than the baseline you measured, so whatever those 5 were, they were not this.

Two things I checked before merging that are worth recording, because they are what makes the fix complete rather than partial:

- `PooledList` has exactly one consumer in the tree — `QuickAppendEventsOperationBase`. No other operation is exposed to this.
- `DocumentSessionBase.ExecuteBatchAsync` is the only site that wraps `ExecuteBatchPagesAsync` in the resilience pipeline, and the daemon's `ProjectionUpdateBatch` routes through it too. So your single `finally` genuinely covers every path, including the daemon — the rentals do not leak there.

The diagnosis is the good part of this. Predicting *which* eight events survived from the operation ordering, then confirming positions 58–65 of 65 against the production data, is about as convincing as a forensic report gets without a repro. Thank you for coming back and correcting the partial-commit claim too — that took the investigation to the right place.

## Holding commit 1

Leaving this PR open for `WriteResiliencePipeline`, which I would like to narrow before taking:

1. **`NpgsqlException => true` is too wide.** Pool exhaustion, connect failures, "connection was reset before anything was sent" are all `NpgsqlException` and all raised *before* anything reaches the server. Those are exactly the blips the write retry exists for, and this turns every one of them into a no-retry. I would want "in doubt" to mean a failure after the command was written — a timeout during read, a connector broken mid-statement — not the whole type.
2. **`ConfigurePolly` sets `WriteResiliencePipeline = ResiliencePipeline`,** so anyone who has ever called it silently opts out of the protection. For a guard against silent double-appends that default is backwards. A separate `ConfigureWritePolly` would be better than a doc note.
3. **`ExtendPolly` invokes the caller's `configure` delegate twice** against two builders. A delegate that registers a shared strategy instance, or has side effects, will misbehave.

Also worth noting the urgency dropped once the buffer bug was fixed: a retry is now a duplicate-append risk rather than a silent-loss one, and duplicates are caught by the server-side version check on every optimistic shape. Plain `Append` with no expected version is the remaining exposure — real, but a much smaller thing than what you found.

Happy to take that as an update to this PR, or as a fresh one if it is easier.

jeremydmiller:
Merged the rest of this — your `b9bdc166e` cherry-picked with authorship intact (e09c00363), plus 75eb1ee26 for the write pipeline. Master is now at 75eb1ee26 and this PR is fully landed.

I did the broader survey of Npgsql/PostgreSQL error states before settling the predicate, and it changed the shape of the answer in two ways worth writing down.

## The predicate is now a three-way classification, not a two-way one

`IsOutcomeInDoubt` had one axis. There are actually three states a failed commit can leave behind:

- **RolledBack** — the server reported it, so the transaction is definitively gone, and the condition may pass. Explicit SQLSTATE allowlist: `40001`/`40P01`/`40000`, the `53xxx` resource class, `55000`/`55006`/`55P03`, `57P03`, `58000`/`58030`.
- **Deterministic** — server-reported, but a replay reproduces it exactly. Constraint violations, syntax errors. No point retrying.
- **Unknown** — the wire failed. Surfaced, never replayed.

## Two things the survey turned up that neither of us had

**Your `NpgsqlException => true` was too wide, but so was my "just narrow it" advice.** I had assumed the safe pre-flight cases (pool exhaustion, connect refused) could be recovered by looking at the exception. They cannot: Npgsql reports pool exhaustion as `NpgsqlException` wrapping `TimeoutException` — structurally identical to a read timeout that stalled mid-batch. One is safe to replay, one is not, and they are the same object graph. So both surface. That is a real behaviour loss versus today and it is deliberate; it is in the docs.

**`MartenCommandException` needed to stay retryable, and for a reason neither predicate expressed.** My first cut classified "anything not from the wire" as safe, on the theory that Marten rolls those back itself. That is true, but it also swept in every domain exception Marten raises from `PostprocessAsync` — and 15 tests went red, `ConcurrencyException` and `DocumentAlreadyExistsException` surfacing as `AggregateException` after three pointless retries. The rule had to be narrowed to the two wrapper types the old policy actually handled. Worth knowing if you ever revisit this: `retry_mechanism` is not the only thing that throws from post-processing, it is just the only thing that *should* be retried when it does.

Also, `NpgsqlException.IsTransient` is not usable here, which is the cleanest way to state the whole problem: Npgsql's definition is built for idempotent work, so it includes the entire `08xxx` class plus `admin_shutdown`, `crash_shutdown`, `idle_session_timeout` and `transaction_resolution_unknown` — every one of which means the connection died, possibly with a `COMMIT` in flight. Right for a `SELECT`, wrong for an append.

## Your three review points

1. **Predicate** — replaced as above.
2. **`ConfigurePolly` silently clearing the protection** — `ConfigurePolly`/`ExtendPolly` now scope to the read pipeline only, with `ConfigureWritePolly`/`ExtendWritePolly` as the explicit opt-in. Your own `Bug_5262_retry_loses_events` proved the point: it used `ConfigurePolly` to force the permissive policy so it could exercise the buffer bug, and after this change that no longer reaches the commit path. One-line fix in the test, and it is now the demonstration that the scoping works.
3. **`ExtendPolly` invoking the delegate twice** — gone, since each method builds one pipeline.

## Verification

PostgreSQL 18 / net10.0. `CoreTests` 577 passed / 4 failed, against a clean-master baseline in a separate worktree of 536 passed / 5 failed — my 4 are a strict subset of the baseline's 5, so this adds none. (Those are the order-dependent shared-schema tests; the same code gave 0 failures earlier in the session against a cleaner database.) `EventSourcingTests` 1935/0. `DaemonTests` 307/0. `retry_mechanism` and both of your tests green.

Docs: `docs/configuration/retries.md` has a new section on why commits are retried differently, the full allowlist, the `IsTransient` warning, and the pool-exhaustion caveat. markdownlint and cspell clean across all 143 doc files.

Thanks for pushing on this one, and for splitting the PR the way you did — being able to take the data-loss fix immediately and think properly about the policy afterwards was exactly the right shape.

jeremydmiller:
Fully landed on master at 75eb1ee26 — both commits merged with your authorship intact. Closing since GitHub will not do it automatically for a cherry-picked merge.

Thanks again for this one. It started as a forensic report with no repro and ended up as two real fixes plus a third issue (#5269, the `CommandTimeout` sharp edge you spotted along the way), which is next up.

PR #5265, opened 2026-08-20
Failing test: Append stores only the first tag of a given tag type

Noblix:
Reproduction only, as discussed on Discord.

An event carrying two tags of the same registered tag type keeps both rows when appended with `StartStream`, but only the first with `Events.Append`. The added test covers both cases. The `Append` (append_stores_every_tag_of_one_type) one fails so CI is expected to be red.

### DIFF
--- a/src/EventSourcingTests/Dcb/dcb_multiple_tags_of_one_type_tests.cs
+++ b/src/EventSourcingTests/Dcb/dcb_multiple_tags_of_one_type_tests.cs
@@ -1,0 +1,76 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using JasperFx.Events.Tags;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests.Dcb;
+
+public record HomeWorkCopied(string AssignmentName, StudentId CopiedFrom, StudentId CopiedBy);
+
+// A homework-copied event names two students, so it carries two tags of the student tag type and
+// has to be findable from either of them. StartStream stores both rows; Events.Append stores only
+// the first, on the store's default append mode.
+[Collection("OneOffs")]
+public class dcb_multiple_tags_of_one_type_tests: OneOffConfigurationsContext
+{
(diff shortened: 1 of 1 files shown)


jeremydmiller:
Fixed on master (790761fc0 / e0362af26). Your repro is merged as-is with authorship intact (c88181c8e) and both of its tests are green.

## What it was

The bulk `mt_quick_append_events` function takes one `varchar[]` per registered tag type, parallel to the events array — so it has exactly one slot per (event, tag type), and `writeAllTagValues` filled it from the first matching tag and `break`ed. The function keeps writing the first of each type; the surplus is now queued as ordinary tag-table rows keyed on the event id, the same way HStore mode already worked. No function signature change, so nothing to migrate.

Worth knowing that the tag tables were never the constraint: the PK is `(value, [tenant_id], seq_id[, is_archived])`, so distinct values on one event were always two legal rows. Only the append path could not get them there.

## One correction to the framing

It is not Append vs StartStream — it is bulk function vs per-event. A non-partitioned `StartStream` takes the per-event `InsertStream` + `QuickAppendEventWithVersion` route, which already queued one insert per tag, which is why that half passed. Everything else routes through the function.

That matters because under `UseTenantPartitionedEvents`, `forceBulkFunction` sends `StartStream` down the bulk route too — so `start_stream_stores_every_tag_of_one_type` fails there as well. I confirmed that against master and added the case in `TenantPartitionedEventsTests`. Rich append mode was unaffected throughout.

## HStore now throws instead of dropping

`DcbStorageMode.HStore` cannot represent this at all: the column maps one key to one value and the key is the tag type. It was silently keeping one and discarding the rest, which is the same silent-absence failure this issue is about — a dropped tag is not an error to a DCB query, it is just missing from the answer. It now throws `MartenNotSupportedException` naming the tag type and pointing at `TagTables`. The DCB docs previously said the second value "overwrites the first"; that is corrected, and `TagTables` is now documented as the mode that supports the case.

Thanks for the clean reproduction — having both halves in one file is what made the bulk-vs-per-event split obvious.

Noblix:
Thank you for the explanation of the cause (and for the fix 😀)

PR #5271, opened 2026-08-20
fix(events): shorten the natural key table's primary key constraint name

jeremydmiller:
`mt_natural_key_X` shortens every name it writes — both foreign keys and its reverse-lookup index go through `PostgresqlIdentifier.Shorten` — but the primary key constraint was left on Weasel's `pkey_{table}_{columns}` default. That default runs past PostgreSQL's 63 character limit for an aggregate type name of any real length:

```
pkey_ + mt_natural_key_ + <aggregate> + _natural_key_value + _tenant_id
  5         15                              18                  10
```

48 characters before the type name, so **sixteen characters is enough to go over** once `tenant_id` joins the key under conjoined tenancy. Without conjoined tenancy the budget is 25.

PostgreSQL truncated the name silently, which is why this went unnoticed. Weasel 9.25 validates primary key constraint names — they were previously validated by no provider at all ([weasel#448](https://github.com/JasperFx/weasel/issues/448)) — and refuses to emit one that would be cut, so the store now fails to build its schema at all:

```
JasperFx.Resources.ResourceSetupException : Failed to setup resource Main of type WeaselDatabase
---- Weasel.Postgresql.PostgresqlIdentifierTooLongException : Database identifier
     pkey_mt_natural_key_nkorderaggregate_natural_key_value_tenant_id
     would be truncated. The PostgresqlMigratorNameDataLength property is currently 64.
```

That identifier is exactly 64 characters, from a 16-character aggregate name.

## The fix

Set `PrimaryKeyName` through `Shorten`, following `EventTagTable`, which already does exactly this. The expression mirrors Weasel's PostgreSQL default (`pkey_{name}_{cols}`) so nothing else changes.

**`Shorten` is a no-op below the limit**, so a name that already fit is spelled exactly as before and no existing database sees a migration. A name that was *over* the limit moves from PostgreSQL's silent truncation to Weasel's truncate-plus-hash — a constraint rename on first apply, for a database whose constraint name was already not the one Marten declared.

## How this was found

Bumping Wolverine to Weasel 9.25.0 turned three `MartenTests` green-to-red — the only failures in a 610-test suite, and the only ones anywhere in that upgrade attributable to Marten. Verified by baselining: they pass on Weasel 9.24.0 and fail on 9.25.0.

Verified end-to-end by pointing Wolverine's `MartenTests` at a locally-built Marten carrying this commit, on Weasel 9.25.0 — `tenant_partitioned_natural_key_aggregate` goes 0/3 → **3/3**.

Note this affects any `[NaturalKey]` aggregate with a type name of 16+ characters under conjoined tenancy, so it is likely to bite broadly once Marten moves to Weasel 9.25.

https://claude.ai/code/session_01HxbBpN6PmRB5CMSSTdGNr3

### DIFF
--- a/src/Marten/Events/Schema/NaturalKeyTable.cs
+++ b/src/Marten/Events/Schema/NaturalKeyTable.cs
@@ -94,6 +94,14 @@ internal class NaturalKeyTable: Table, ISchemaObject
             IsUnique = false,
             Columns = new[] { streamCol }
         });
+
+        // Every other name this table writes is shortened; the primary key constraint was left on
+        // Weasel's pkey_{table}_{columns} default, which runs past PostgreSQL's 63 character limit
+        // for an aggregate type name of any real length -- 16 characters is enough once tenant_id
+        // joins the key under conjoined tenancy. PostgreSQL used to truncate it silently; Weasel
+        // 9.25 validates primary key constraint names and refuses to emit one that would be cut.
+        PrimaryKeyName = PostgresqlIdentifier.Shorten(
+            $"pkey_{Identifier.Name}_{string.Join("_", PrimaryKeyColumns)}");
     }
 
     /// <summary>


jeremydmiller:
Extended this to `DocumentTable` on master (23d9a45df) — same bug one table over, and far more exposed, since that is every document Marten stores.

It was left on Weasel's `pkey_{table}_{columns}` default for the same reason the natural key table was, and it is where both of the Weasel 9.25.0 failures actually came from: `pkey_mt_doc_bug_2025_event_inheritance_in_projection_identity_id` (64) and `pkey_mt_doc_bug_4679_catch_up_per_tenant_23505_tripdistance_tenant_id_id` (72), both from tests in this repo.

## It is not cosmetic

Worth recording, because weasel#487 relaxed the *rejection* and it would be easy to read that as closing the matter. It does not — PostgreSQL still truncates, and a primary key constraint's backing index is a **schema-scoped** object. Two document types whose names agree for long enough collide outright. Confirmed directly against PostgreSQL 15:

```
NOTICE:  identifier "pkey_this_is_a_very_long_..._aaa" will be truncated to "pkey_this_is_a_very_long_..._trunc"
NOTICE:  identifier "pkey_this_is_a_very_long_..._bbb" will be truncated to "pkey_this_is_a_very_long_..._trunc"
ERROR:  relation "pkey_this_is_a_very_long_document_table_name_that_will_be_trunc" already exists
```

So `Shorten`'s deterministic hash suffix is not decoration — it is the thing that keeps the two distinct.

## Blast radius

`Shorten` is a no-op at 63 characters or fewer, so nothing changes for a schema that was not already being truncated. The only databases that see a constraint rename are the ones that already have the bug, and Weasel emits `alter table ... rename constraint` rather than a drop/recreate.

Four tests in `DocumentDbTests`: the length bound, distinct names for two near-identical types, an ordinary type keeping exactly the name it always had, and the one that matters — both tables actually created in one schema, which threw 42P07 before. Verified red on master first (2 of the 4 fail).

Sizing note for anyone touching that test: the type names are tuned so each *table* name is exactly 63 characters, which PostgreSQL accepts, while the derived `pkey_` runs to 71. Making them any longer trips the object-name check instead, which is a different failure.

Suites: `DocumentDbTests` 1095/0, `MultiTenancyTests` 165/0, `EventSourcingTests` 1943/0, `TenantPartitionedEventsTests` 243/0, `CoreTests` at its known order-dependent baseline.

Issue #5281, opened 2026-08-26
DCB: fetching the same tag boundary twice in one session raises a spurious DcbConcurrencyException


jeremydmiller:
Fetching the *same* DCB tag boundary twice in one session and then saving raises a spurious `JasperFx.Events.DcbConcurrencyException`, even though no other session ever touched the tag. The two fetches conflict with each other inside a single unit of work.

## Repro

```csharp
var studentId = new StudentId(Guid.NewGuid());
var query = new EventTagQuery().Or<StudentId>(studentId);
await using var session = theStore.LightweightSession();

await session.Events.FetchForWritingByTags<StudentCourseEnrollment>(query);
var boundary = await session.Events.FetchForWritingByTags<StudentCourseEnrollment>(query);
boundary.AppendOne(new StudentGraded(studentId, new CourseId(Guid.NewGuid()), 95));

await session.SaveChangesAsync();
// JasperFx.Events.DcbConcurrencyException:
//   "DCB consistency violation: new events matching the tag query were appended after sequence 0"
```

There is no second session anywhere in that snippet, and the stream is brand new.

## Cause

`FetchForWritingByTagsHandler.HandleAsync` queues a `DcbTagVersionAssertion` on the session for **every** fetch ([`FetchForWritingByTagsHandler.cs#L255`](https://github.com/JasperFx/marten/blob/master/src/Marten/Events/Dcb/FetchForWritingByTagsHandler.cs#L255)). Two fetches over the same tag capture the *same* version for the same `(tag_table, tag_value)` row — the fetch-time capture is a plain `SELECT ... version`, so nothing has moved between them.

At save time both assertions are in the unit of work, and each emits:

```sql
insert into mt_dcb_tag_version (tag_table, tag_value, tenant_id, version)
values ($1, $2, $3, $captured + 1)
on conflict (tag_table, tag_value, tenant_id)
  do update set version = mt_dcb_tag_version.version + 1
  where mt_dcb_tag_version.version = $captured
returning 1
```

The first one bumps the row `0 -> 1` and returns a row. The second one's `where version = $captured` no longer matches, returns no row, and `PostprocessAsync` reports that as a conflict ([`DcbTagVersionAssertion.cs#L108-L142`](https://github.com/JasperFx/marten/blob/master/src/Marten/Events/Dcb/DcbTagVersionAssertion.cs#L108-L142)).

So the boundary is genuinely being defended against a concurrent writer — the concurrent writer is just the same session's own earlier fetch.

[...]

Issue #5280, opened 2026-08-26
Spurious DcbConcurrencyException when the same DCB boundary is fetched twice in one session


jeremydmiller:
Calling `FetchForWritingByTags` twice with the same `EventTagQuery` on a single session, and then appending through one of the boundaries, throws `DcbConcurrencyException` on `SaveChangesAsync()` — even though nothing else touched the tag.

## Repro

```csharp
[Collection("OneOffs")]
public class dcb_repeated_boundary_fetch_tests: OneOffConfigurationsContext
{
    [Fact]
    public async Task same_boundary_fetched_twice_in_one_session_then_appended()
    {
        StoreOptions(opts =>
        {
            opts.Events.AddEventType<StudentGraded>();
            opts.Events.RegisterTagType<StudentId>("student");
        });

        var studentId = new StudentId(Guid.NewGuid());
        var query = new EventTagQuery().Or<StudentId>(studentId);
        await using var session = theStore.LightweightSession();

        await session.Events.FetchForWritingByTags<StudentCourseEnrollment>(query);
        var boundary = await session.Events.FetchForWritingByTags<StudentCourseEnrollment>(query);
        boundary.AppendOne(new StudentGraded(studentId, new CourseId(Guid.NewGuid()), 95));

        await Should.NotThrowAsync(() => session.SaveChangesAsync());   // throws DcbConcurrencyException
    }
}
```

Reproduced on `master` at 9.29.0 (c921fa5df). Affects both `DcbStorageMode.TagTables` and `HStore` — the consistency check is the same side-table mechanism in either mode.

## Cause

Each `FetchForWritingByTags` builds its own `DcbTagVersionAssertion` (`src/Marten/Events/Dcb/DcbTagVersionAssertion.cs`), and two fetches of the same query capture the **same** version for the same `(tag_table, tag_value)` row. Both assertions end up in one unit of work, and each emits:

```sql
insert into ....mt_dcb_tag_version (tag_table, tag_value, tenant_id, version) values (..., $captured + 1)
on conflict (tag_table, tag_value, tenant_id) do update set version = mt_dcb_tag_version.version + 1
where mt_dcb_tag_version.version = $captured
returning 1
```

The first one bumps the row and returns a row. The second one's `where version = $captured` no longer matches — it returns nothing, `PostprocessAsync` reads no row, and it reports a conflict against a version its own sibling moved.

[...]

PR #5282, opened 2026-08-26
Cover four more repeated-boundary-fetch cases around #5280

jeremydmiller:
> **This PR changed shape.** It opened as a fix for #5281. While it was in review, #5276 landed and
> already fixed that bug — #5281 was a duplicate report of #5280. The implementation half of this PR
> is superseded and has been dropped; what remains is four tests. See the note at the bottom.

## What's left

Four tests around #5280's row-collapsing rule, on ground nothing else covers. No behavior changes —
they pass on master as it stands.

- **hstore mode.** `Bug_5280_repeated_boundary_fetch_tests` and `dcb_noop_boundary_tag_version_tests`
  both run on the `TagTables` default. The capture step and the assertion are shared by both storage
  modes so this should never diverge, but a regression here surfaces as an *absent tag* rather than
  an error — the one failure a consistency boundary must not have.
- **Appending through both boundaries over one row**, where each boundary object carries events of
  its own. #5280's repro appends through the second boundary only, so a dedupe keyed on "last
  boundary wins" rather than on the row would still pass it.
- **Two disjoint boundaries in one session** — the other side of collapsing. Rows that are *not*
  duplicates must all still be asserted; a collapse keyed on anything coarser than
  `(tag_table, tag_value)` would quietly drop one, and the loss shows up only as a boundary that
  stopped being enforced.
- **Re-fetching a boundary after a committed save**, asserting an interloper is still caught.
  `dcb_noop_boundary_tag_version_tests` covers the retry idiom after a *no-op* save and asserts it
  can append; this asserts the opposite half after a save that actually committed.

Verified green alongside the suites they sit next to: 14 tests across
`dcb_repeated_boundary_fetch_edge_cases`, `Bug_5280_repeated_boundary_fetch_tests` and
`dcb_noop_boundary_tag_version_tests`, 0 failures.

## What was dropped, and why

The original PR added a `DcbTagVersionLedger` that claimed rows per unit of work and kept the lowest
captured version per row. #5276 solved the same problem a better way — hold the captured rows on the
session, queue **one** merged assertion, sort it, and collapse duplicate rows keeping the oldest
capture — which also fixed a latent cross-boundary lock-ordering deadlock that my per-fetch
assertions left in place.

The two arrive at the same semantics on the case that matters: a writer landing *between* two fetches
must still conflict, so the oldest/lowest captured version has to win. All ten tests from the original
PR pass against master's implementation **unmodified**, which is what confirmed the fix was complete
rather than merely close. Six of the ten duplicated `Bug_5280_repeated_boundary_fetch_tests` and are
dropped along with the implementation.

#5281 should be closed as a duplicate of #5280. Closing this PR entirely is a perfectly reasonable
call too if the four remaining tests aren't judged worth the file.

### DIFF
--- a/src/EventSourcingTests/Dcb/dcb_repeated_boundary_fetch_edge_cases.cs
+++ b/src/EventSourcingTests/Dcb/dcb_repeated_boundary_fetch_edge_cases.cs
@@ -1,0 +1,126 @@
+#nullable enable
+using System;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using JasperFx.Events.Tags;
+using Marten.Events;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests.Dcb;
+
+/// <summary>
+/// Extra ground around #5280's row-collapsing rule, found while #5281 was open as a duplicate report
+/// of it. <see cref="Bug_5280_repeated_boundary_fetch_tests"/> covers the reported shape — one query
+/// fetched twice, two overlapping queries, and the oldest-capture-wins rule. These are the cases that
+/// probing the bug turned up alongside it and that nothing else pins down.
+/// </summary>
+[Collection("OneOffs")]
+public class dcb_repeated_boundary_fetch_edge_cases: OneOffConfigurationsContext
+{
+    private void ConfigureStore(DcbStorageMode mode = DcbStorageMode.TagTables)
+    {
+        StoreOptions(opts =>
+        {
+            opts.Events.DcbStorageMode = mode;
+            opts.Events.AddEventType<StudentGraded>();
+            opts.Events.RegisterTagType<StudentId>("student");
+            opts.Events.RegisterTagType<CourseId>("course");
+        });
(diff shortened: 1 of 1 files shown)


jeremydmiller:
Closing: superseded by #5276, which landed while this was open and already fixes the bug this PR set out to fix.

#5281 turned out to be a duplicate report of #5280. Master holds the captured rows on the session, queues one merged assertion, sorts it, and collapses duplicate rows keeping the oldest capture — the same semantics this PR's ledger arrived at, plus it closes a cross-boundary lock-ordering deadlock that the per-fetch assertions here left in place.

All ten tests written against #5281 pass on master's implementation unmodified, which is what confirmed the fix was complete rather than merely close. Six of them duplicated `Bug_5280_repeated_boundary_fetch_tests`.

The four non-duplicate tests (hstore mode, appending through both boundaries, two disjoint boundaries surviving the collapse, re-fetch after a committed save still catching an interloper) are not being merged with this. Happy to re-open them as a standalone test-only PR if they're wanted.

Issue #5268, opened 2026-08-20
DCB hstore tag index cannot be built concurrently, and not at all on tenant-partitioned events


erdtsieck:
## Setup

Enabling `DcbStorageMode.HStore` on an existing store adds two things to `mt_events`: a nullable `tags hstore` column, and a GIN index `idx_mt_events_tags` over it (`EventsTable`).

The column is free — a nullable add is metadata-only. The index is not, and there is currently no way to build it without taking the table out of service for the duration:

* I don't see `CONCURRENTLY` emitted anywhere for it; `IsConcurrent` exists on `UniqueIndexAttribute` but nothing sets it for this index, so schema application issues a plain `CREATE INDEX`, which holds an ACCESS EXCLUSIVE lock while it builds. On a large event table that is a write outage, not a migration.
* And with `Events.UseTenantPartitionedEvents = true` a flag would not be enough anyway: Postgres refuses `CREATE INDEX CONCURRENTLY` on a partitioned parent altogether (`cannot create index on partitioned table … concurrently`).

So for the combination Marten itself supports — DCB in hstore mode on a tenant-partitioned event store — there is no non-blocking path to the index today.

## What Postgres does allow

The supported dance for a partitioned table is three steps:

1. `CREATE INDEX ON ONLY <parent> USING gin (tags)` — metadata only, leaves the parent index invalid;
2. `CREATE INDEX CONCURRENTLY` on each partition;
3. `ALTER INDEX <parent_idx> ATTACH PARTITION <child_idx>` for each — the parent index flips to valid once every child is attached.

## Ask

Either of these would be enough, and the second is the cheaper one:

1. build the DCB tag index that way when the event table is partitioned (and plain `CONCURRENTLY` when it is not);
2. or let it be opted out of schema application, so an operator can pre-create it out of band and have the diff accept the existing index rather than trying to recreate it.

For context on why this matters in practice rather than in theory: we run a sharded, tenant-partitioned event store across several hundred shard databases, and turning hstore mode on is a prerequisite for using DCB against history at all — see #5267, which carries the tags through a bulk import. Enabling the mode should not require a maintenance window per shard.

Happy to do the work on this one myself — I'll pick it up. Wanted to raise it first so the approach can be agreed before I send a PR, per CONTRIBUTING.

jeremydmiller:
[...]

PR #5291, opened 2026-08-26
Build the DCB hstore tag index without blocking writes (#5268)

jeremydmiller:
Closes the half of [#5268](https://github.com/JasperFx/marten/issues/5268) that #5275 could not: the reporter's **option 1**, Marten building the index itself rather than handing it to an operator.

## What this adds

`Events.BuildHStoreTagIndexConcurrently` marks `idx_mt_events_tags` concurrent, so the migration emits `CREATE INDEX CONCURRENTLY` as a command of its own outside the transaction instead of a plain `CREATE INDEX` holding `ACCESS EXCLUSIVE` on `mt_events` for the whole GIN build.

That needed Weasel first — this bumps **Weasel 9.25.1 → 9.26.0** for [weasel#502](https://github.com/JasperFx/weasel/issues/502). Also in that bump: weasel#508 (an INVALID index now reports as drift, so a failed concurrent build is repaired by the next apply rather than sitting there unused) and weasel#500 (`DbObjectName.Schema`/`.Name` hold the undelimited name — Marten compares no quoted literals, and emitted DDL is byte-identical).

Opt-in, not the default. A concurrent build cannot run inside a transaction, which changes what `db-patch` and `db-dump` write out: still correct applied against a live database, no longer runnable as one transactional script. Nobody should inherit that by upgrading.

## The tenant-partitioned case is refused, not shipped

PostgreSQL will not build a concurrent index on a partitioned parent at all, and the sequence it accepts instead — `ON ONLY` parent, one concurrent index per partition, `ALTER INDEX ... ATTACH PARTITION` for each — needs the partition list. Weasel takes that from `ListPartitioning.PartitionTableNames`, which reads the table's **declared** partitions and does not consult the `PartitionManager` the way `CreateDelta` and `WriteCreateStatement` both do. Under `UseTenantPartitionedEvents` every partition is manager-owned, so the list is empty.

I ran it rather than reasoning about it. The result against a live store:

```
 tp_diag_66589 | idx_mt_events_tags | f
```

The parent index created, `indisvalid = false`, no children, forever. An index the planner will not use, that every later apply then reports as drift — worse than the blocking build it replaced, and worse still given the flag was set specifically to avoid an outage.

So `StoreOptions.Validate` throws on the combination, with the out-of-band procedure in the message. **Lifting this is a one-line Weasel change**, not a Marten one: `PartitionTableNames` consulting `PartitionManager?.Partitions() ?? _partitions` like its two siblings. Happy to open that separately — it is out of scope for a Marten PR.

Until then, the partitioned route stays what #5275 shipped and documented: `Events.IgnoreIndex(EventGraph.HStoreTagIndexName)` plus the three-step sequence by hand. That is the reporter's actual deployment, and this PR pins it working end to end under partitioning for the first time.

## Tests

Eight, and every assertion reads `pg_index.indisvalid` rather than mere existence — "the index is there" is precisely what a half-finished concurrent build would pass.

Unpartitioned (`EventSourcingTests/Dcb/Bug_5268_hstore_tag_index_can_be_built_out_of_band.cs`): the concurrent build completes; it is not drift on a later apply or `AssertDatabaseMatchesConfigurationAsync` (`CONCURRENTLY` is not part of what `pg_get_indexdef` echoes back, so a canonical form that kept it would rebuild the index on every startup); and it works on a first-time create as well as on a migration.

Partitioned (`TenantPartitionedEventsTests/Dcb/Bug_5268_concurrent_tag_index_under_partitioning.cs`): the refusal fires and its message carries the way out; the flag is harmless when hstore mode is off; the blocking default is unchanged; and the out-of-band three-step index survives a later apply, an assert, and still serves tag queries.

## Verification

CoreTests, EventSourcingTests, DocumentDbTests, TenantPartitionedEventsTests, MultiTenancyTests on net9.0. All green except a four-test cluster in CoreTests (`Bug_4185`, `Bug_4187`, `rolling_range_partitioning`, `divergent_application_assembly_reuse`) that fails identically on a clean `origin/master` worktree here — pre-existing local cross-test interference, not the bump. markdownlint and cspell clean on `docs/events/dcb.md`.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -399,7 +399,7 @@
     <PackageVersion Include="Vogen" Version="7.0.0" />
     <!-- 9.16.2: the hold at 9.2.1 (were newer builds actually being published?) is resolved —
          Weasel.EntityFrameworkCore is on nuget.org, so it tracks the rest of the Weasel line. -->
-    <PackageVersion Include="Weasel.EntityFrameworkCore" Version="9.26.0" />
+    <PackageVersion Include="Weasel.EntityFrameworkCore" Version="9.27.0" />
     <!-- Weasel.Postgresql 9.1.x: managed LIST partitions under multi-database (sharded) apply.
          9.1.2 (marten#4706): managed-partition tables are no longer destructively rebuilt — the
          out-of-band AddPartitionToAllTables path creates partitions while the generic schema diff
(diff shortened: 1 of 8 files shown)


jeremydmiller:
Filed the Weasel side as [JasperFx/weasel#520](https://github.com/JasperFx/weasel/issues/520), and 16dfcdb00 points the guard and its test at it — so the thing that comes out when Weasel ships names the reason it went in.

One detail the issue sharpened that isn't in the description above: `RangePartitioning` does **not** have this bug. Its `PartitionTableNames` goes through `expectedRanges()`, which is the same manager-first resolution `ListPartitioning` uses in its other two call sites. So the two manager-capable strategies disagree and `ListPartitioning` is the outlier — which makes the fix a straight alignment rather than a new behaviour.

jeremydmiller:
Unparked on Weasel **9.27.0**, which carries the [weasel#520](https://github.com/JasperFx/weasel/issues/520) fix ([weasel#522](https://github.com/JasperFx/weasel/pull/522)) — all three `ListPartitioning` call sites now resolve through one `expectedPartitions()` helper, so the partition enumeration is no longer empty for a manager-owned partitioning.

That lifts the `StoreOptions.Validate` refusal, and `Events.BuildHStoreTagIndexConcurrently` now covers the shape #5268 was actually reported against: a sharded, tenant-partitioned event store.

The refusal test is replaced by the three it was standing in for, each reading `indisvalid` rather than mere existence:

- the sequence attaches one index per tenant partition and the parent flips to valid
- the result is not drift on a later apply or `AssertDatabaseMatchesConfigurationAsync`
- a tenant added *afterwards* gets its own partition index without knocking the parent back to invalid

The out-of-band route keeps its test — it is still what a shard running an older Marten has to use.

Rebased onto master (past #5289). TenantPartitionedEventsTests, EventSourcingTests and DocumentDbTests are green on net9.0; CoreTests shows only the same four failures a clean `origin/master` worktree shows on this machine.

PR #5292, opened 2026-08-26
Resolve serializer member aliases when patching (#5290)

### LINKED ISSUE #5290: Patching API builds JSON paths from raw C# member names — `[JsonPropertyName]` / serializer aliases are ignored, so patches on aliased members silently write phantom nodes
**Marten version:** 9.23.0 (the relevant code is unchanged on current master)
**.NET version:** 10.0
**PostgreSQL version:** 18
**Serializer:** SystemTextJsonSerializer, `Casing.Default`

### Repro

```csharp
public class Doc
{
  public Guid Id { get; set; }

  [JsonPropertyName("nm")]
  public string Name { get; set; }
}

// store a Doc, then:
session.Patch<Doc>(id).Set(x => x.Name, "changed");

var reloaded = await session.LoadAsync<Doc>(id);
// reloaded.Name is still the old value
```

The row afterwards holds both nodes:

```json
{ "Id": "…", "nm": "original", "Name": "changed" }
```

The patch reports success, `Load`/`Query` keep returning the old value, and the phantom `Name` node
is erased by the next full document save — so there is no durable evidence that anything went wrong.

### Cause

`PatchExpression.toPath` builds the path from the raw member names with only a casing transform;
member aliases from the serializer (`[JsonPropertyName]`, `JsonProperty`, etc.) are never consulted:

```csharp
private string toPath(Expression expression)
{
  var visitor = new PatchingMemberFinder();
  visitor.Visit(expression);

  // TODO -- don't like this. Smells like duplication in logic
  return visitor.Members.Select(x => x.Name.FormatCase(_session.Serializer.Casing)).Join(".");
}
```

The LINQ side resolves the same expression through the queryable-member machinery and lands on the
correct `nm` path — so `Where(x => x.Name == …)` and `Patch(...).Set(x => x.Name, …)` disagree about
where the same member lives. The `toPathExpression` helper in the same file already routes through
`JsonPathCreator(_session.Serializer)`, which suggests the intended direction.

How we hit it: a 5-second job-progress heartbeat was persisted with
`session.Patch<JobDoc>(id).Set(x => x.Job.Progress, p)` on a document whose `Job` property is stored
as `"jb"`. Every tick wrote `data->'Job'->'Progress'`, every reader read `data->'jb'->'Progress'`,
and each status-transition save (full serialization) deleted the phantom node — the write path
looked healthy for weeks with zero errors anywhere.

### Expected

Either the lambda overload resolves paths through the same member/alias machinery as the LINQ
provider, or patching a member whose serialized name differs from the emitted path throws instead of
succeeding against a location the serializer will never read.

### Related hazard

While diagnosing this we also confirmed that patches leave `.Duplicate()` columns stale (expected,
since patching is server-side SQL and duplicated values are computed during .NET serialization) —
but this is equally silent at the call site. The mapping knows which members are duplicated; a patch
whose path touches one could refuse or at least log. Happy to file that separately if you'd rather
track it on its own.

jeremydmiller:
Fixes [#5290](https://github.com/JasperFx/marten/issues/5290).

`PatchExpression.toPath` built its JSON path from raw C# member names with only a casing transform, so a member carrying a serializer alias — `[JsonPropertyName]`, or `[JsonProperty]` through Marten.Newtonsoft's resolver — was patched at a path the serializer never reads.

The reporter's description of how this fails is the reason it deserves a careful fix rather than a one-liner: the patch reports success, `Load` and `Query` keep returning the old value, and the phantom node sits beside the real one until the next full document save erases it. A five-second job-progress heartbeat wrote into a phantom node for weeks with zero errors anywhere.

## The fix

`MemberInfo.ToJsonKey(Casing)` is the resolution `QueryableMember` already uses — which is exactly why, as the issue notes, the LINQ provider landed on `nm` for the same expression all along. Using it in `toPath` makes `Where(x => x.Name == v)` and `Patch(...).Set(x => x.Name, v)` agree about where a member lives by construction rather than by coincidence.

**The predicate overloads had the same gap in a second path builder.** The issue suggests `toPathExpression` was already right because it routes through `JsonPathCreator` — it isn't: `JsonPathCreator.VisitMember` emitted `@.Name` from the raw member name too, so a JSONPath predicate against an aliased member matches nothing and silently leaves the element in place. Same fix, applied only to its document-member branch; the other branch walks a captured closure value rather than a document path and stays on the raw name.

Dictionary keys reach `toPath` as `IndexerKeyInfo`, a synthetic `MemberInfo` that declares no attributes, so `ToJsonKey` returns exactly what `Name.FormatCase` did. There is a test pinning that — a fix that "resolved" a dictionary key would be a worse bug than the one being fixed.

## Tests

Six, in `src/PatchingTests/Patching/Bug_5290_patch_honors_serializer_member_aliases.cs`:

- `Set` on an aliased member under System.Text.Json, and again under Newtonsoft — genuinely separate routes through `ToJsonKey`, not the same assertion twice
- a nested aliased member (`x.Job.Progress` where both `Job` and `Progress` are aliased) — the reporter's actual shape, where both segments have to resolve
- `Increment` and `Append`, so the fix is not pinned to `Set` alone
- the dictionary-key guard
- `Remove`-by-predicate, for the `JsonPathCreator` half

Every one asserts on the **stored JSON** as well as the round-trip. "The value came back changed" would not catch a second node appearing beside the right one, and that phantom node is the whole shape of this bug. Five of the six fail without the fix; the sixth is the guard and passes both ways by design.

PatchingTests, LinqTests and DocumentDbTests are green on net9.0.

## Not addressed here

The issue's "related hazard" — patches leaving `.Duplicate()` columns stale — is a genuinely separate change (it wants a refusal or a warning at the call site, not a path fix), and the reporter offered to file it separately. Worth its own issue.

### DIFF
--- a/src/Marten/Linq/Parsing/JsonPathCreator.cs
+++ b/src/Marten/Linq/Parsing/JsonPathCreator.cs
@@ -92,7 +92,12 @@ public sealed class JsonPathCreator(ISerializer serializer) : ExpressionVisitor
         }
         else
         {
-            _jsonPathBuilder.Append($"@.{node.Member.Name.FormatCase(serializer.Casing)}");
+            // #5290: same resolution as PatchExpression.toPath and QueryableMember. This visitor is
+            // only reached from the patching API's predicate overloads, where the members named are
+            // document members and so carry the serializer's aliases; casing alone would build a
+            // JSONPath predicate that matches nothing on an aliased member. The branch above is a
+            // captured closure value, not a document path, so it stays on the raw name.
+            _jsonPathBuilder.Append($"@.{node.Member.ToJsonKey(serializer.Casing)}");
 
             if (!_memberIfInBinary.Contains(node) && node.Type == typeof(bool))
                 _jsonPathBuilder.Append($" {LogicalOperators[ExpressionType.Equal]} {(_memberIfInUnary.Contains(node) ? "false" : "true")}");
--- a/src/Marten/Patching/PatchExpression.cs
+++ b/src/Marten/Patching/PatchExpression.cs
@@ -348,8 +348,15 @@ internal class PatchExpression<T>: IPatchExpression<T>
         var visitor = new PatchingMemberFinder();
         visitor.Visit(expression);
 
-        // TODO -- don't like this. Smells like duplication in logic
-        return visitor.Members.Select(x => x.Name.FormatCase(_session.Serializer.Casing)).Join(".");
+        // #5290: ToJsonKey, not Name.FormatCase. Casing alone ignores serializer member aliases
+        // ([JsonPropertyName], and [JsonProperty] through Marten.Newtonsoft's resolver), so a patch
+        // on an aliased member wrote a phantom node beside the real one -- the patch reported
+        // success, every reader kept seeing the old value, and the next full save erased the
+        // evidence. ToJsonKey is the same resolution QueryableMember uses, so the patch path and
+        // the LINQ provider now agree about where a member lives by construction rather than by
+        // coincidence. IndexerKeyInfo (dictionary keys) declares no attributes, so it keeps the
+        // casing-only behavior it had.
+        return visitor.Members.Select(x => x.ToJsonKey(_session.Serializer.Casing)).Join(".");
     }
 
     private string toPathExpression(Expression expression)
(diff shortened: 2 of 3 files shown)


wickedmachinator:
incredible, thank you!

PR #5300, opened 2026-08-27
Add failing test for DCB double append under stress

Noblix:
During a stress test it was discovered that consistency invariants for DCB can be broken under concurrency. In this case a group of 16 concurrent requests resulted in 2 writes, breaking an invariant that only one should be allowed. This PR contains tests replicating that.

A theory for why this is possible is that DCB boundaries are handled with the following steps:
1)	Read events matching event tag query
2)	Read version for that boundary
3)	Do some work and define new events
4)	Write new events based on version being stable since (2)

A small but important detail here is that step (1) and (2) are separate reads. This means that when a write happens between step (1) and (2) then step (3) is working with stale data but when we commit in step (4) we do not get a version conflict.

### DIFF
--- a/src/EventSourcingTests/Dcb/dcb_staggered_concurrent_append_tests.cs
+++ b/src/EventSourcingTests/Dcb/dcb_staggered_concurrent_append_tests.cs
@@ -1,0 +1,117 @@
+#nullable enable
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using JasperFx.Events;
+using JasperFx.Events.Tags;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests.Dcb;
+
+// Concurrent appends guarded by the same DCB tag boundary must serialize: when several
+// writers race the same boundary, only one may commit. This pins that invariant under
+// staggered (non-lockstep) concurrency, exercised with many racers over many rounds.
+// Note: #4591's test only covers the lockstep case (racers synchronized).
+[Collection("OneOffs")]
+public class dcb_staggered_concurrent_append_tests : OneOffConfigurationsContext
+{
+    private const int Racers = 24;
+    private const int Rounds = 200;
+
+    public dcb_staggered_concurrent_append_tests()
+    {
+        StoreOptions(opts =>
+        {
+            opts.Events.AddEventType<Enrolled>();
+            opts.Events.AddEventType<ProgressRecorded>();
+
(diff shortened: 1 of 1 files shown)


jeremydmiller:
Thanks @Noblix — your theory is exactly right, and the repro is a good one. Fix is up in #5301, which carries your test as its first commit with authorship preserved.

To fill in the last detail of the mechanism: the two reads aren't just separate, they're separate *statements*, and at READ COMMITTED Postgres takes a fresh snapshot at the start of each one — even though both ride the same batch and the same implicit transaction. So the race window isn't the gap between them, it's the **entire execution time of the events query**, which is the expensive one. That's why 16 concurrent requests was enough.

The version was captured second, which pairs a fresh version with a stale aggregate, and then the save's `where version = $captured` matches a check the caller never actually made. Capturing the version first can only pair a stale-low version with a fresh aggregate, which fails the assertion — a spurious conflict the caller retries, never a lost one. Same ordering `FetchAsyncPlan.FetchForWriting` already uses for streams; the DCB handler was the odd one out.

Both of your tests fail on master and pass with the fix. Also filed JasperFx/polecat#515 and JasperFx/fisher#133 to check the same conditions against the other two stores — Polecat turns out never to have picked up the #4591 side-table fix at all, so that one is a real find downstream of your report.

Closing this in favour of #5301 once that merges.

jeremydmiller:
Superseded by #5301, merged as fce5ceb7c — your two tests went in unchanged as the first commit of that branch, and you're credited as a co-author on the squashed commit.

Recapping the fix for anyone finding this later: the version capture now runs *ahead* of the events SELECT in `FetchForWritingByTagsHandler`. Your steps (1) and (2) aren't just separate reads, they're separate *statements*, and at READ COMMITTED Postgres takes a fresh snapshot at the start of each one even though both ride the same batch and the same implicit transaction. So the window isn't the gap between them — it's the entire execution time of the events query, which is why 16 concurrent requests was enough to hit it.

Capturing second pairs a fresh version with a stale aggregate, and the save's `where version = $captured` then matches a check the caller never actually made. Capturing first can only pair a stale-low version with a fresh aggregate, which fails the assertion — a spurious conflict the caller retries, never a lost one. Two statements can't be made to share a snapshot, but the pairing only has to fail safe.

Thanks for the report and the repro — chasing it also turned up that Polecat never picked up the earlier #4591 side-table fix at all, so it's likely exposed to the original race too (JasperFx/polecat#515). Fisher gets the same test as regression insurance in JasperFx/fisher#133.

PR #5302, opened 2026-08-27
Invoke transaction participants when the session is in an ambient transaction

arnelirobles:
`AmbientTransactionLifetime.ExecuteBatchPagesAsync` accepts a participants list and never reads it. `BeforeCommitAsync` appears zero times in that file, and it does not delegate to an inner lifetime either.

The asymmetry is three against one:

| Lifetime | invokes `BeforeCommitAsync` |
|---|---|
| `TransactionalConnection` | yes |
| `ExternalTransaction` | yes |
| `AutoClosingLifetime` | yes |
| `AmbientTransactionLifetime` | **no, and does not delegate** |

(`EventTracingConnectionLifetime` looks like a fourth by grep but is a decorator passing `participants` through correctly.)

So a session enlisted in a `TransactionScope` is the one place a registered participant is silently skipped. For the EF Core participant that means the `DbContext` is never handed Marten's connection, and its writes go nowhere the Marten transaction can see.

## Where the call goes

After the pages execute and after the existing failure paths throw. This lifetime never commits: the ambient scope owns that and completes when the caller completes the scope, so this is the last safe point.

## The transaction argument is null there

`ITransactionParticipant.BeforeCommitAsync` now takes `NpgsqlTransaction?`. That is by necessity rather than convenience: the connection is enlisted in the ambient transaction, so no `NpgsqlTransaction` exists to hand over, and EF Core throws if given one while already operating inside a `TransactionScope`. `UseTransactionAsync(null)` is the required call there rather than a tolerated one, which is now written into the EF Core participant.

The signature change is a nullable annotation, so it is metadata rather than a binary break, but it is a public interface and worth naming.

## Tests

`transaction_participant_invocation`, two cases, and the pair is the point:

- `participant_is_invoked_on_an_ordinary_session` **passes on master**. Without it, the other test could be failing because participants are broken generally rather than in this one lifetime.
- `participant_is_invoked_when_enlisted_in_an_ambient_transaction` **fails on master**, passes with the change.

Measured in that order on `b6f6f81`: 1 passed, 1 failed before; 2 passed after. `CoreTests` Sessions suite 26 passed.

## Not addressed

The end-to-end consequence for an inline EF-Core-backed projection under a `TransactionScope` is argued rather than demonstrated here. This proves the participant is not invoked, which is the mechanism; I did not build the projection scenario on top of it.

### DIFF
--- a/src/CoreTests/Sessions/transaction_participant_invocation.cs
+++ b/src/CoreTests/Sessions/transaction_participant_invocation.cs
@@ -1,0 +1,89 @@
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Transactions;
+using Marten;
+using Marten.Services;
+using Marten.Testing.Documents;
+using Marten.Testing.Harness;
+using Npgsql;
+using Shouldly;
+using Xunit;
+
+namespace CoreTests.Sessions;
+
+/// <summary>
+/// Every connection lifetime is handed the session's transaction participants when its batch pages
+/// execute. A participant exists to flush work belonging to another system, EF Core being the only
+/// production implementation, into the same transaction Marten is about to commit. A lifetime that
+/// accepts participants and does not invoke them therefore drops that work with no error.
+/// </summary>
+public class transaction_participant_invocation: BugIntegrationContext
+{
+    [Fact]
+    public async Task participant_is_invoked_on_an_ordinary_session()
+    {
+        StoreOptions(opts => opts.RegisterDocumentType<Target>());
+        await theStore.Storage.ApplyAllConfiguredChangesToDatabaseAsync();
+
+        var participant = new RecordingParticipant();
+
+        await using var session = theStore.LightweightSession();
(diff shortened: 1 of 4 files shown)


jeremydmiller:
@arnelirobles No, just no. You lost me at "ambient transactions". I really, really don't want to have to support anything like this. What's the justification for this?

arnelirobles:
I did it because the participants list is accepted and then silently ignored in this one lifetime. The other three call `BeforeCommitAsync`, `AmbientTransactionLifetime` takes the list and never reads it. So a participant registered on a session inside a `TransactionScope` looks like it works and does nothing. The silent part is what bothered me, not ambient transactions.

I'm not attached to this fix. If you don't want participants working under `TransactionScope` at all I can flip it: make ambient sessions reject participants loudly instead of accepting and skipping them. That's less for you to support, not more. Or just close this. Your call.

jeremydmiller:
Let me be more clear then, I do not want to support ambient transactions inside of projections, full stop. 

What was the use case that got you here?

Issue #4085, opened 2026-01-21
Inconsistent tenant_id values break SingleStream async projection rebuilds


alekbarszczewski:
## Description

SingleStream async projections fail during rebuild/catchup when events in the same stream have inconsistent `tenant_id` values. Some events are tagged with `"DEFAULT"` while others are tagged with `"Marten"`, causing the projection's `Apply` methods to receive documents with default/uninitialized property values.

## Reproduction

When running an async SingleStream projection and events are already in the database (e.g., app restart with existing events), the projection produces incorrect results. The `Apply` method receives a document with all properties set to their default values, as if the `Create` method was never called.

### Example projection
```csharp
public record BillrunLookup
{
    [Identity]
    public required string BillrunId { get; init; }
    public required bool IsCompleted { get; init; }
    public required YearMonth SettlementMonth { get; init; }
    public required int BillrunSequence { get; init; }

    public static BillrunLookup Create(BillrunStarted started)
    {
        return new BillrunLookup
        {
            BillrunId = started.BillrunId,
            IsCompleted = false,
            SettlementMonth = started.ForMonth,
            BillrunSequence = started.BillrunSequence,
        };
    }

    public static BillrunLookup Apply(BillrunCompleted completed, BillrunLookup lookup)
    {
        // lookup has all properties set to defaults here!
        return lookup with { IsCompleted = true };
    }
}

public class BillrunLookupProjection : SingleStreamProjection { }

// Registration
opts.Projections.Add(ProjectionLifecycle.Async);
```

### Expected result
```json
{"BillrunId": "billrun-1", "IsCompleted": true, "BillrunSequence": 1, "SettlementMonth": "2025-12"}
```

### Actual result
```json
{"BillrunId": "billrun-1", "IsCompleted": true, "BillrunSequence": 0, "SettlementMonth": "0001-01"}
```

## Root Cause

Events appended to the same stream have inconsistent `tenant_id` values in the database:

| seq_id | stream_id | tenant_id |
|--------|-----------|-----------|
| 1      | billrun-1 | DEFAULT   |
| 2      | billrun-1 | Marten    |

This appears to be caused by the Wolverine.Marten integration when using `[AggregateHandler]` with `IStartStream` for stream creation and `IEnumerable<object>` for subsequent appends.

### Handler example that produces inconsistent tenant IDs
```csharp
public class StartBillrunHandler
{
    [AggregateHandler]
    public static IStartStream Handle(StartBillrun command)
    {
[...]

PR #5306, opened 2026-08-31
marten#5305: regression guard for #4085, and the JasperFx bump it requires

### LINKED ISSUE #5305: The wolverine#2053 fix never reached the async daemon; add a regression test and confirm IsGlobalWithinConjoinedTenancy
Follow-up to #4085 (transferred here from wolverine#2053), found while auditing the JasperFx side. Filing separately rather than reopening #4085 so the history there stays intact — happy to fold this in if you would rather reopen.

## The fix was structurally incomplete

`Marten.Events.Aggregation.SingleStreamProjection<TDoc,TId>.BuildSlicer` sets `ForceSingleTenancy` from `EventGraph.TenancyStyle`, carrying the comment *"This will address https://github.com/JasperFx/wolverine/issues/2053"*.

But `JasperFx.Events.Grouping.TenantedEventSlicer<TDoc,TId>` honoured that flag in only one of its two overloads. `SliceAsync(IReadOnlyList<IEvent>)` respected it; `SliceAsync(EventRange)` unconditionally did `range.Events.GroupBy(x => x.TenantId)`.

The async daemon reaches **only** the `EventRange` overload — `EventRange.SliceAsync(IEventSlicer)`, called from `GroupedProjectionExecution.groupEventRangeAsync` for `SliceBehavior.Preprocess` and from `AggregationRunner` for `JustInTime`. Inline and live go through the list overload.

Which lines up exactly with what #4085 reported:

> Only async projections during rebuild/catchup are affected
>
> Live aggregation works correctly (not affected)

So the flag was being set on precisely the path that ignored it, and honoured on the path that never needed it. Confirmed with a failing unit test against the slicer, not by inference.

Fixed upstream in JasperFx/jasperfx#721 / [jasperfx#722](https://github.com/JasperFx/jasperfx/pull/722).

## What this means for Marten

**No code change is needed** — Marten already sets the flag correctly. But picking up the JasperFx bump is a real behaviour change on the async daemon, arriving via a dependency rather than a Marten commit:

- Conjoined stores: `isSingleTenanted` is false, nothing moves.
- Single-tenanted stores with clean `*DEFAULT*` rows: already folded to one group, nothing moves.
- Single-tenanted stores with mixed `tenant_id` values: now fold into one group instead of several. **This is the #4085 fix landing.**

Two asks:

1. **A regression test.** There is none today — `2053` appears exactly once in Marten's source, in that `BuildSlicer` comment. Something that seeds a single-tenanted store with mixed `tenant_id` rows on one stream, runs an async single-stream projection, and asserts one correctly-folded document. The shared-suite version of this is JasperFx/jasperfx#724, but Marten is where #4085 was reported and where the coverage is currently zero.
2. **A release note**, since it changes async projection grouping without a corresponding Marten diff.

## Separate question: `IsGlobalWithinConjoinedTenancy`

`Events/Fetching/FetchAsyncPlan.cs:45` treats a stream as single-tenanted when:

```csharp
if (_events.TenancyStyle == TenancyStyle.Single || _events.GlobalAggregates.Contains(typeof(TDoc)))
```

`SingleStreamProjection.BuildSlicer` checks only the first of those:

```csharp
var isSingleTenanted = session.As<QuerySession>().Options.EventGraph.TenancyStyle == TenancyStyle.Single;
```

So a global aggregate on a conjoined store is *fetched* as global but *sliced per tenant* in the daemon. That inconsistency is currently unobservable for the same reason as everything else above — the flag did nothing on that path — and becomes observable the moment the JasperFx fix ships.

Should `isSingleTenanted` be `TenancyStyle == Single || IsGlobalWithinConjoinedTenancy`? I do not know Marten's intent here well enough to answer, but it looks like the same bug one level over, and now is the moment to settle it.

jeremydmiller:
Closes #5305.

## The gap this closes

[#4085](https://github.com/JasperFx/marten/issues/4085) (originally wolverine#2053) was closed as fixed. It had **no regression test** — `2053` appeared exactly once in the whole source tree, in a comment on `SingleStreamProjection.BuildSlicer`.

Worse, the fix was structurally incomplete, and nothing here could have told us. Marten has set `ForceSingleTenancy` since the original fix. But `TenantedEventSlicer` honoured that flag on only **one of its two overloads**: `SliceAsync(IReadOnlyList<IEvent>)` respected it while `SliceAsync(EventRange)` ignored it — and the async daemon reaches only the `EventRange` one.

So the flag was being set on precisely the path that could not read it, and honoured on the path that never needed it. Which lines up exactly with what #4085 reported:

> Only async projections during rebuild/catchup are affected
>
> Live aggregation works correctly (not affected)

Fixed upstream in JasperFx/jasperfx#721, shipped in JasperFx **2.58.0**. That is why this PR bumps the pin rather than adding a test alone — the test cannot pass on 2.56.0.

## The test

`Bug_4085_async_projection_with_mixed_tenant_ids`, and it is a **proven** guard rather than an assumed one:

| JasperFx pin | Result |
|---|---|
| 2.57.2 (pre-fix) | **fails** — `Shouldly.ShouldAssertException : tally.EventCount` |
| 2.59.0 | **passes** |

Verified by pinning back and watching it go red, then forward again — not by reading release notes.

Two deliberate choices worth reviewing:

**It seeds the reported table state with an `UPDATE` on `tenant_id`** rather than reproducing the Wolverine append path. The bug is "given rows in this state, the daemon must fold them as one aggregate"; how they reached that state is not Marten's concern and is fragile to reproduce. The reporter's own workaround was the mirror image of this UPDATE.

**It asserts the UPDATE touched exactly two rows.** If a future change normalises `tenant_id` on write or read, the test would otherwise still pass — for a reason with nothing to do with slicing — and quietly stop guarding #4085.

**The aggregate is additive on purpose.** A stream folded once and a stream folded in two pieces differ in these numbers; a last-write-wins aggregate would hide the split.

## Why this test lives in Marten rather than the shared compliance suite

I first wrote this as a shared suite (JasperFx/jasperfx#724) and it does not work. Polecat and Fisher both **normalise `tenant_id` on write**, so neither can construct the precondition through the shared surface — the fact skips on both, asserting nothing. Marten appears to be the only store that can hold these rows, which is the wrong shape for a compliance suite. Written up in JasperFx/jasperfx#727; this PR is the coverage landing where the condition is actually reachable.

## The `IsGlobalWithinConjoinedTenancy` question — deliberately NOT answered here

#5305 also asks whether `isSingleTenanted` should be `TenancyStyle == Single || IsGlobalWithinConjoinedTenancy`, since `FetchAsyncPlan.cs:45` treats a global aggregate as single-tenanted while `BuildSlicer` does not.

What I established:

- `AddGlobalProjection` sets `IsGlobalWithinConjoinedTenancy = true` **and** `Schema.For<TDoc>().SingleTenanted()`, so the document is single-tenanted while events stay conjoined.
- Stream ids are globally unique under conjoined tenancy — `StartStream` with the same id from a second tenant throws `ExistingStreamIdCollisionException` — so one stream cannot be *created* across tenants.

What I could **not** establish: my probe's events all landed with `*DEFAULT*` tenant_id, so it never actually exercised mixed tenancy. I am not turning a misconfigured probe into a conclusion. The question is still open and wants someone with deeper knowledge of Marten's tenancy intent, or its own issue.

## Test results

`EventSourcingTests` on net9.0: **1991 passed, 4 failed** of 2002. All four investigated against a `master` baseline worktree at its own 2.56.0 pin:

| Failure | Verdict |
|---|---|
| `testing_projections.test_async_aggregation` | **pre-existing** — fails identically on master |
| `testing_projections.test_async_aggregation_with_wait_for` | **pre-existing** |
| `testing_projections.test_async_aggregation_with_wait_for_and_fake_time_provider` | **pre-existing** |
| `Bug_5268...the_concurrent_index_works_under_archived_stream_partitioning` | **environmental** — passes on this branch against a clean database; the failure only appears after repeated full-suite runs against a reused one |

None are caused by this change. A clean-database full run is in flight to confirm the count; I will update here if it says otherwise.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -342,19 +342,35 @@
          verified by diffing the discovered test list against master, not by comparing run totals.
          Inert for Marten: the Event Model provenance ladder (jasperfx#703/#704) only affects hosts
          registering an IEventModelDefinitionSource, which Marten does not, and
-         RetryBlock.ShouldRetry/OnTerminalFailure (jasperfx#701) are additive and default to null. -->
-    <PackageVersion Include="JasperFx" Version="2.56.0" />
-    <PackageVersion Include="JasperFx.Events" Version="2.56.0" />
+         RetryBlock.ShouldRetry/OnTerminalFailure (jasperfx#701) are additive and default to null.
+
+         2.57-2.59 (marten#5305): the bump this branch actually needs. jasperfx#721 fixed
+         TenantedEventSlicer.SliceAsync(EventRange) to honour ForceSingleTenancy. It was respected
+         only by the IReadOnlyList<IEvent> overload, and the async daemon reaches only the EventRange
+         one. Marten has set that flag in SingleStreamProjection.BuildSlicer since the marten#4085
+         fix, so until 2.58.0 the flag was being set on precisely the path that could not read it.
+         Bug_4085_async_projection_with_mixed_tenant_ids fails on anything earlier; that was verified
+         by pinning back to 2.57.2 and watching it go red, not by reading release notes.
+
+         Also arriving, both inert for Marten: jasperfx#723 moves ForceSingleTenancy resolution into
+         JasperFxSingleStreamProjectionBase behind IEventTenancySource on the session, which Marten's
+         own BuildSlicer override already computes identically (Polecat and Fisher, whose subclasses
+         are empty class bodies, are the stores that needed it); and two opt-in compliance suites
+         (jasperfx#724, #725) that add no tests until a store enrolls them. Marten enrolls neither;
+         see jasperfx#727 for why #724 in particular is being reconsidered. -->
+    <PackageVersion Include="JasperFx" Version="2.59.0" />
+    <PackageVersion Include="JasperFx.Events" Version="2.59.0" />
     <!--
       The shared cross-store event sourcing compliance suites (#5116). Source-only package: the
       suites compile inside EventSourcingTests so JasperFx's aggregate source generator can bind
       Marten's own session types. Marten.Testing needs it too because MartenComplianceFixture,
       which closes the seam, lives beside the rest of the harness.
     -->
-    <PackageVersion Include="JasperFx.Events.ComplianceTests" Version="2.56.0" />
+    <PackageVersion Include="JasperFx.Events.ComplianceTests" Version="2.59.0" />
     <!--
-      Deliberately AHEAD of the rest of the JasperFx line, which stays at 2.41.0 until Marten adopts
-      the strong-typed identity compliance suite that landed in 2.42.0. This package is analyzer-only,
+      No longer ahead of the rest of the JasperFx line. As of the 2.59.0 bump (marten#5305) the
+      whole family moves together, and this note is kept for the history below. This package is
+      analyzer-only,
       declares no dependencies, and Marten bundles its dll into Marten.nupkg (see
       _BundleEventsSourceGeneratorAnalyzer in Marten.csproj, marten#4557), so its version is
       independent of the runtime family's.
@@ -369,11 +385,11 @@
       which does not care how the instance was constructed. The generator is otherwise byte-identical
       from 2.38.0 through 2.42.0, so this is an isolated swap of that one fix.
     -->
-    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.56.0">
+    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.59.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageVersion>
-    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.56.0" />
+    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.59.0" />
     <PackageVersion Include="Jil" Version="3.0.0-alpha2" />
     <PackageVersion Include="Lamar" Version="7.1.1" />
     <PackageVersion Include="Lamar.Microsoft.DependencyInjection" Version="15.0.0" />
--- a/src/EventSourcingTests/Bugs/Bug_4085_async_projection_with_mixed_tenant_ids.cs
+++ b/src/EventSourcingTests/Bugs/Bug_4085_async_projection_with_mixed_tenant_ids.cs
@@ -1,0 +1,118 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using JasperFx.Core;
+using JasperFx.Events;
+using Marten;
+using Marten.Testing.Harness;
+using Shouldly;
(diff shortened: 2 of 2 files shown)


jeremydmiller:
Split the `IsGlobalWithinConjoinedTenancy` question out to #5307 so it does not ride along on this PR unanswered.

Recording there what this PR's bump changes about it: until JasperFx 2.58.0 the value of `isSingleTenanted` had no effect on the daemon path at all, because `TenantedEventSlicer` ignored `ForceSingleTenancy` on the overload the daemon reaches. This PR is what makes the flag live — so any consequence of `FetchAsyncPlan` and `BuildSlicer` computing it differently becomes observable for the first time here.

jeremydmiller:
Clean-database full run finished. Same totals — **1991 passed, 4 failed, 7 skipped** of 2002 — so the count is not an artefact of a reused database.

The `Bug_5268` verdict is now sharper than "environmental". Across the two full runs it failed a **different test method each time**:

| Run | Failing method |
|---|---|
| reused DB | `the_concurrent_index_works_under_archived_stream_partitioning` |
| clean DB | `the_tag_index_can_be_built_without_blocking_writes` |

Both at ~30s, which is the timeout rather than an assertion. Run in isolation against a clean database on this branch, the whole class passes **8/8**. A different method failing on each full-suite run, only under load, is flakiness in that class rather than anything this branch changed — `CREATE INDEX CONCURRENTLY` contending with a busy suite.

The three `testing_projections.test_async_aggregation*` failures remain confirmed pre-existing: they fail identically on a `master` worktree at its own 2.56.0 pin.

A full `master` baseline run is in flight to state master's own failure count for comparison; I will post it here.

jeremydmiller:
**CI is green: 30 of 30 jobs pass**, including `EventSourcing` on both legs (net9/PG15, net10/PGlatest).

That resolves the `Bug_5268` question I flagged earlier, and **corrects what I said about it**. I called it "environmental, unrelated to this branch" on two data points. The fuller picture:

| Environment | `Bug_5268` |
|---|---|
| local, all 2002 tests in one process | failed 2 of 2 on this branch |
| local, same class in isolation | passed 8 of 8 |
| local, all 2002 in one process, `master` | passed 2 of 2 |
| **CI, sharded per area** | **passed** |

The relationship to this branch was real, just indirect: `CREATE INDEX CONCURRENTLY` waits behind every transaction open when it started, this branch adds 10 tests (1 new + 9 from the compliance bump) several of which drive the async daemon, and a single-process full-suite run concentrates enough concurrency to push a marginal case past the command timeout. CI shards by area and never gets there.

Nothing here is broken and nothing blocks this PR. Tracked separately as #5308 with a recommendation to make that test's timeout independent of unrelated suite load, so local full-suite runs stay trustworthy for exactly this kind of dependency-bump validation.

The three `testing_projections.test_async_aggregation*` failures remain confirmed pre-existing: `master` at its own 2.56.0 pin fails the same three across two runs (1992 total, 3 failed, both times).

Issue #5315, opened 2026-09-01
Full-text search on a document with two FullTextIndexes silently searches only one of them


jeremydmiller:
Found while analysing #5298 (full-text ranking), where it appears as "Open question 1 — how does rank find its index when a document has several?". It is not only a ranking problem: the **filter** already picks arbitrarily, so filing it on its own.

## The bug

`FullTextWhereFragment.GetDataConfig` resolves the index to search by taking the first one matching the `regConfig` (`src/Marten/Linq/SqlGeneration/Filters/FullTextWhereFragment.cs:69-81`):

```csharp
return mapping
    .Indexes
    .OfType<FullTextIndexDefinition>()
    .Where(i => i.RegConfig == regConfig)
    .Select(i => i.DataConfig)
    .FirstOrDefault() ?? FullTextIndexDefinition.DataDocumentConfig;
```

Every full-text index on a document shares the default `regConfig` of `english` unless someone deliberately varies it, so `FirstOrDefault` over "the ones matching `english`" is `First` over "all of them".

## Repro

```csharp
public class FtsDoc
{
    public Guid Id { get; set; }
    public string Data { get; set; }
    public string Title { get; set; }
}

StoreOptions(opts =>
{
    opts.Schema.For<FtsDoc>()
        .FullTextIndex(x => x.Title)
        .FullTextIndex(x => x.Data);
});

session.Query<FtsDoc>().Where(x => x.PlainTextSearch("hi")).ToCommand();
```

Actual:

```sql
select d.id, d.data from ... as d
where to_tsvector('english'::regconfig, ((d.data ->> 'Title')))
      @@ plainto_tsquery('english'::regconfig, :p0)
```

The second index is not merely unranked — it is **not searched**. A document matching only on `Data` never comes back, and there is no error, no warning, and nothing in the SQL hinting that a second index was considered and dropped.

## Why it is worth its own issue

Two full-text indexes on one document is a reasonable model — separate indexes let you weight, maintain or drop them independently, and Marten accepts the configuration without complaint. The query API then gives no way to say which one you meant: `PlainTextSearch` / `WebStyleSearch` / `PhraseSearch` take a term and a `regConfig`, and `regConfig` is the *only* selector.

The ordering is also whatever `mapping.Indexes` happens to be in, so which index wins is a function of declaration order rather than anything the caller expressed.

## Possible shapes

[...]

Issue #5298, opened 2026-08-26
Ranking with Full text search


mlh758:
PG supports ranking text searches if you add ranks in the index with `setweight`: https://www.postgresql.org/docs/current/textsearch-features.html#TEXTSEARCH-MANIPULATE-TSVECTOR

Marten doesn't currently expose any place to use this, although it allows ranking if you use ngram. I've got a feature in my project where the ranking would be nice to have. I had claude take a swing at a plan and edited it down a bit. At first blush the API it's proposing looks reasonable. If this is something you would take, but don't want to implement, I could take a swing at it as well so I'm not just offloading work on you all.

The two missing features:

- **`setweight`** when configuring a `FullTextIndex`, so a match in a title outranks a match in a
  long description.
- **`ts_rank`** when querying, so results come back in relevance order.

`OrderByNgramRank` (8.29+) already does exactly this for ngram indexes. The proposal is the
equivalent for tsvector search, plus the index-side weighting that makes the rank meaningful.

## What Marten emits today

A multi-member index concatenates the members into one flat vector
(`FullTextIndexDefinitionFactory.GetDataConfig`):

```csharp
options.Schema.For<Achievement>()
    .FullTextIndex(a => a.Title, a => a.Description, a => a.Tagline);
```

```sql
create index mt_doc_achievement_idx_fts on public.mt_doc_achievement using gin (
  to_tsvector('english', ((d.data ->> 'Title') || ' ' || (d.data ->> 'Description') || ' ' || (d.data ->> 'Tagline')))
);
```

And the query side (`FullTextWhereFragment`) reuses that same expression, so the index is hit:

```sql
to_tsvector('english'::regconfig, (...)) @@ websearch_to_tsquery('english'::regconfig, $1)
```

Both halves are boolean. Every matching document is equally relevant.

## Proposal, part 1: `setweight` on the index

Assign each member one of Postgres's four weight labels (`A`–`D`, defaulting to `D`, which is
what an unlabelled vector already means):

```csharp
options.Schema.For<Achievement>()
    .FullTextIndex(idx => idx
        .Weighted(a => a.Title, TextSearchWeight.A)
        .Weighted(a => a.Tagline, TextSearchWeight.B)
        .Weighted(a => a.Description, TextSearchWeight.C));
```

Emitting one `setweight` per member, concatenated as vectors rather than as text:

```sql
create index mt_doc_achievement_idx_fts on public.mt_doc_achievement using gin ((
[...]

PR #5316, opened 2026-09-02
Apply table alias to the data column in `where`

### LINKED ISSUE #5314: Full-text search rewrites the JSON key, not just the column: a member whose serialized name contains "data" searches a key that does not exist
Found while analysing #5298 (full-text ranking). Pre-existing and independent of that feature, but it sits directly on the seam that feature would build on, so filing separately.

## The bug

`FullTextWhereFragment` rebases the index's data config onto the table alias with a naive substring replace (`src/Marten/Linq/SqlGeneration/Filters/FullTextWhereFragment.cs:38`):

```csharp
_dataConfig = GetDataConfig(mapping, regConfig).Replace("data", "d.data");
```

`Replace` has no idea which occurrences of `data` are the **column** and which are part of a **JSON key**, so it rewrites both.

## Repro

```csharp
public class FtsDoc
{
    public Guid Id { get; set; }
    public string Data { get; set; }
    public string Title { get; set; }
}

StoreOptions(opts =>
{
    opts.UseSystemTextJsonForSerialization(casing: Casing.CamelCase);
    opts.Schema.For<FtsDoc>().FullTextIndex(x => x.Data, x => x.Title);
});

session.Query<FtsDoc>().Where(x => x.PlainTextSearch("hi")).ToCommand();
```

Actual, from `ToCommand()`:

```sql
select d.id, d.data from ... as d
where to_tsvector('english'::regconfig, ((d.data ->> 'd.data') || ' ' || (d.data ->> 'title')))
      @@ plainto_tsquery('english'::regconfig, :p0)
```

`->> 'd.data'` is a key that does not exist. Expected `->> 'data'`.

## Two consequences, both quiet

1. **Wrong results.** That member contributes nothing to the search — `->>` on a missing key yields NULL — so documents that should match on `Data` never do. No error, no warning.
2. **The index stops being used.** The `where` expression no longer matches the index's expression, so the GIN index cannot serve the query and it degrades to a scan. The whole point of `FullTextWhereFragment` reading the expression back off the index definition is to keep those two in step, and this silently breaks it.

## Reaching it

Any member whose **serialized** name contains `data` as a substring:

- camelCase serialization plus a property named `Data` (above)
- `[JsonPropertyName("metadata")]`, `"userdata"`, `"data_v2"`, …

It is casing-sensitive rather than exotic: with default casing the key is `Data`, `Replace("data", …)` does not match, and the same model is fine. That is presumably why it has not been reported — it depends on serializer configuration rather than on the query.

## Note on the fix

The replace exists because `FullTextIndexDefinitionFactory.GetDataConfig` builds the expression with `RawLocator.RemoveTableAlias("d")` and the query side needs the alias back. Round-tripping through string surgery is the actual defect; building the aliased and un-aliased forms from the member locators, or making the replace alias-aware (anchored on `data ->>` / `data #>` rather than bare `data`), both avoid it.

Worth fixing before #5298 lands either half, since ranking has to resolve the *same* tsvector expression the `Where` resolved — and today that expression is sometimes wrong.

Related: the sibling defect in the same method is filed separately (multi-index `FirstOrDefault`).

mlh758:
Using a regex apply the alias instead of a simple replace. 

Closes #5314

### DIFF
--- a/src/CoreTests/Util/StringExtensionsTests.cs
+++ b/src/CoreTests/Util/StringExtensionsTests.cs
@@ -1,0 +1,68 @@
+using Marten.Util;
+using Shouldly;
+using Xunit;
+
+namespace CoreTests.Util;
+
+public class StringExtensionsTests
+{
+    [Theory]
+    [InlineData("data", "d.data")]
+    [InlineData("(data ->> 'Title')", "(d.data ->> 'Title')")]
+    [InlineData("((data ->> 'Title') || ' ' || (data ->> 'LastName'))",
+        "((d.data ->> 'Title') || ' ' || (d.data ->> 'LastName'))")]
+    [InlineData("data #>> '{Inner,Name}'", "d.data #>> '{Inner,Name}'")]
+    public void qualifies_the_data_column(string sql, string expected)
+    {
+        sql.ApplyTableAliasToDataColumn("d").ShouldBe(expected);
+    }
+
+    // https://github.com/JasperFx/marten/issues/5314 — a naive Replace("data", "d.data") rewrote the
+    // JSON key as readily as the column, so a member whose serialized name contains "data" searched
+    // a key that does not exist.
+    [Theory]
+    [InlineData("(data ->> 'data')", "(d.data ->> 'data')")]
+    [InlineData("(data ->> 'metadata')", "(d.data ->> 'metadata')")]
+    [InlineData("(data ->> 'data_v2')", "(d.data ->> 'data_v2')")]
+    [InlineData("(coalesce(data ->> 'data', '') || ' ' || coalesce(data ->> 'title', ''))",
+        "(coalesce(d.data ->> 'data', '') || ' ' || coalesce(d.data ->> 'title', ''))")]
+    // a doubled quote escapes a quote inside the literal, so the literal does not end early
+    [InlineData("(data ->> 'it''s data')", "(d.data ->> 'it''s data')")]
+    public void leaves_json_keys_alone(string sql, string expected)
+    {
+        sql.ApplyTableAliasToDataColumn("d").ShouldBe(expected);
+    }
+
+    [Theory]
+    [InlineData("(metadata ->> 'Title')")]
+    [InlineData("(data_v2 ->> 'Title')")]
+    public void leaves_other_identifiers_alone(string sql)
+    {
+        sql.ApplyTableAliasToDataColumn("d").ShouldBe(sql);
+    }
+
+    [Fact]
+    public void is_idempotent()
+    {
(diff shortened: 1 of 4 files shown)


jeremydmiller:
Thanks for picking this up — and apologies for the collision: I had a fix for #5314 in flight on my own branch without checking whether you had already started, which is on me. **Yours is the better fix and it is the one going in.**

Concretely better, not diplomatically:

- It lives in `StringExtensions` as the explicit inverse of `RemoveTableAlias`. That is the right home, because the broken round-trip *is* the defect — index definitions store the un-aliased form and the query side rebases it. Mine was a hand-rolled character walk buried inside `FullTextWhereFragment`, which is the wrong layer and not reusable.
- `round_trips_with_RemoveTableAlias` is the test I did not think to write, and it is the one that actually pins the invariant rather than a symptom of it.
- It takes the alias as a parameter instead of hardcoding `"d"`.
- The literal-first alternation is neater than my state machine, and you cover cases mine handles but never proved: `#>>` json-path, `data_v2`, and the doubled-quote escape in `'it''s data'`.
- Your end-to-end test queries and asserts both documents come back. Mine only asserted on generated SQL, which would not have caught a fix that produced plausible-looking SQL that still matched nothing.

Merged. I am dropping my version and rebasing the rest of my branch on top of yours.

Two things from that branch you may want to know about, since they are adjacent to what you have been looking at:

**#5315 turned out to be two bugs, and the one I filed was the wrong one.** `FullTextIndexDefinition` derives its name from the *table*, never from the members, so two `FullTextIndex()` calls over different members collide by construction and the second was silently discarded — never registered, never created, never searched. The `FirstOrDefault` I reported is only reachable once both indexes carry explicit names. I found it because the fix I wrote for the filed mechanism did not make my test fail. Correcting the issue now.

**Weasel 9.29.0 shipped weasel#541**, so `setweight` weighting is reachable — `TsVectorExpression` / `ForTsVector`, with `IndexedTsVector` as the single property both the DDL and the query-side filter read, which is exactly the "the rank must resolve the same vector the Where resolved" point you made in #5298. The index half is built and verified against a real PostgreSQL, including a no-drift round-trip.

The `ts_rank` **ordering** half is still open and is the interesting one. `OrderByFragment` holds `List<string>` and appends straight to the command, so an ordering cannot carry a parameter today — which is precisely why the ngram precedent inlines its term with quote-doubling, and why your "the new ordering should parameterize" note is a bigger job than it looks. It needs the ordering pipeline to carry fragments rather than strings. Happy to hand that one to you if you want it, or to take it myself — say which.

mlh758:
@jeremydmiller I should have pinged you and said  I was working on it. I just wasn't confident I'd have something so didn't want to speak up yet.

Issue #5225, opened 2026-08-13
Opt-in second-level aggregate cache for FetchForWriting()


erdtsieck:
## The goal

Make `FetchForWriting()` cheap for the very common shape where a *chain* of commands walks over one
aggregate stream, without forcing the aggregate onto an `Inline` or `Async` snapshot.

I'd like to add an **opt-in, second-level aggregate cache** behind `FetchForWriting()`, in the shape of an
additional `IFetchPlanner`. This issue is the design + the safety argument; I'm preparing the proposal
and the PR myself, and would like a sanity check on the direction (and on the two small enabling PRs
below) before I go further.

## The problem

In a CQRS system built on Wolverine's aggregate handler workflow, one business operation is often a chain
of cascading commands over a *single* stream. A real example from our claims system — six handlers, one
`ProvidedCare` stream:

```
DeterminePrestatie → DetermineWlzDebtor → DetermineDebtor
  → ValidateNationality → DetermineTariff → PrepareForInvoicing
```

Every hop is its own message, its own session and its own transaction — by design, because each hop has
to be independently retryable. So each hop calls `FetchForWriting()` again on the same stream.

For a `Live`-registered aggregate that means every hop reads the *entire* stream back and folds it from
scratch, even though the previous hop, milliseconds earlier and usually in the same process, had that
exact aggregate in memory and appended two events to it.

The existing options don't cover this:

* **`Events.UseIdentityMapForAggregates`** is exactly the right idea, but it caches on the *session*, and
  each cascading message gets a fresh session. It never spans hops.
* **`AsyncOptions.CacheLimitPerTenant`** gives the async daemon a bounded MRU cache of aggregates
  (`IAggregateCache<TKey, TItem>` / `RecentlyUsedCache<TKey, TItem>` in `JasperFx.Core`). Command handling
  has no equivalent.
* **Switching the aggregate to `Inline`** does fix the read cost, but it is a much bigger change: a
  snapshot write joins every append transaction, the projection has to be rebuild-safe and serialization-safe,
  and for a large aggregate that is a real trade rather than a free win. It shouldn't be the only answer.

## The proposal

A `FetchCachedPlan<TDoc, TId>` that keeps `(tenantId, streamId) -> (aggregate, version)` in a bounded
`RecentlyUsedCache`, and on a hit issues the same two-statement batch the `Live` plan issues today, with
one predicate added:

```
statement 1:  select version from mt_streams where id = ?            -- unchanged
[...]

PR #5321, opened 2026-09-02
Aggregate write cache for the Live lifecycle (#5225)

erdtsieck:
Closes the remaining scope of #5225: the aggregate write cache on the `Live` lifecycle.

## Why

The cache that landed in #5256 covers `Async` and `Inline`. `Live` is the lifecycle #5225 was written for — *"Make `FetchForWriting()` cheap … **without forcing the aggregate onto an `Inline` or `Async` snapshot**"* — and it is the one with no stored snapshot to skip, so it was the one with the most to gain. Today `CacheAggregatesForWriting<T>()` on a `Live` aggregate silently does nothing: `FetchLivePlan` has no cache path, so the call compiles, enrols the type, and never produces a hit.

That is also the shape the issue was motivated by. In an application where one business operation is a chain of cascading commands over a single stream — each hop its own message, session and transaction, by design, so each is independently retryable — every hop calls `FetchForWriting` on the same stream and replays the entire stream from scratch.

## What a hit does

The same strategy `FetchAsyncPlan` already uses, with the baseline coming from memory instead of the document table: read the stream version, read only the events after the cached version, fold those onto the cached instance.

`IEventIdentityStrategy<TId>.BuildEventQueryHandler` already takes an optional `ISqlFragment` filter, so the delta needs no new SQL machinery — just `new WhereFragment("d.version > ?", cachedVersion)`.

The version handed to `AppendToStream` still comes from the database on every call, so the optimistic concurrency assertion is untouched and a stale entry costs a larger delta query rather than a wrong aggregate — the load-bearing property from `IAggregateWriteCache`.

## Two things specific to Live

**Archived streams.** Under `Live` a zero-event delta does not mean "nothing changed": `IsNotArchivedFilter` removes an archived stream's events, so an uncached fetch answers `null` where a cached entry still holds the aggregate. Left alone the cache would resurrect it. `BuildCommandForReadingVersionForStream` now reads `is_archived` alongside the version (`select version, is_archived from mt_streams …`) and a hit on an archived stream discards the entry and retries uncached. Every other reader of that statement takes the version by ordinal 0 and is unaffected. This is the eviction the issue anticipated under "Other correctness details".

**`FetchLatest` after an append in the same session.** This one bit me while building it, and it is the more interesting half.

`FetchForWriting` → append → `FetchLatest` is the aggregate-handler workflow's shape, and `FetchForReading`'s item-map optimisation folds the appended events onto the very instance the fetch handed out:

```csharp
var starting = stream.Aggregate;
var appendedEvents = stream.Events;
return await _aggregator.BuildAsync(appendedEvents, session, starting, id, _documentStorage, cancellation);
```

For an aggregate that applies events in place — which most class aggregates do, `SimpleAggregate.Apply(AEvent _) { ACount++; }` included — that moves `starting` past the version the cache entry files it under. `RecentlyUsedAggregateWriteCache.Store` keeps the reference, not a copy, so the entry now understates the state it points at, and the next fetch tops it up with events it already has.

Fixed by **evicting on the fold** rather than deferring the write to after the commit. Deferring (the `Inline` strategy) would be wrong whenever `FetchLatest` was *not* called — the instance stays at the fetched version while the stream moves on — and the two cases are not distinguishable at commit time. Eviction fails safe: the next fetch replays the stream. It is guarded on there being events to fold, so a read-only `FetchLatest` keeps its entry.

## Heads-up: the same hazard exists on the merged Async path

While proving the above I checked whether `Async` has it too, and it does. `FetchAsyncPlan.ForUpdate` stores immediately after the fold, justified in a comment as:

> Under the Async lifecycle any events the caller appends are not applied to this instance in session, so it cannot drift ahead of the database.

But `FetchAsyncPlan.ForReading` performs exactly the same item-map fold onto `stream.Aggregate`. Reproduced with a temporary test against `caching_async_aggregates_for_writing` (not included here): after `FetchForWriting` → `AppendOne` → `FetchLatest` → commit, a re-fetch in a new session returns `ACount` 5 where 4 is correct.

I have deliberately **not** fixed that here. It is released behaviour (9.28+), the fix is the same few lines in `FetchAsyncPlan.ForReading`, and it deserves its own review and possibly its own patch — a PR titled "Live support" quietly changing `Async` correctness would be hard to review. Happy to open it as a separate issue + PR, or to fold it in here if you would rather have both at once. Your call.

## Tests

`caching_live_aggregates_for_writing`, mirroring the Async and Inline suites, plus the two Live-specific cases:

- cold fetch matches an independent live replay and seeds the entry
- a hit folds **only** the delta (the baseline is deliberately poisoned to 100, so the correct answer 102 is unreachable by a full replay)
- a hit with an empty delta returns the baseline unchanged — the case a cascading chain actually hits
- a stale entry still yields the correct aggregate
- optimistic concurrency still fires against a warm cache
- a cache ahead of the database falls back to the uncached path and heals in one round
- an archived stream is not resurrected
- `FetchLatest` in the same session does not poison the entry
- entries do not leak across tenants
- caching is off unless the type opts in

Verified red without the change rather than assumed: removing the delta narrowing fails 4 of them with double-counted state, and removing the eviction fails the `FetchLatest` test.

Full `EventSourcingTests` run: **2003 passed, 7 skipped, 5 failed** — all 5 being `Projections.testing_projections`, which hardcodes `Host=localhost;Port=5432;Username=postgres` and cannot reach a database in my setup. They fail in Weasel's `executeMigration` while opening the connection, so they touch none of this.

Docs updated in `docs/events/optimizing.md`: the lifecycle table is now three columns, with a section on what the cache removes under `Live` and the archived-stream note.

### DIFF
--- a/docs/events/optimizing.md
+++ b/docs/events/optimizing.md
@@ -155,8 +155,9 @@ when your projection is constantly updating a relatively small number of differe
 ## Caching Aggregate Snapshots for FetchForWriting <Badge type="tip" text="9.26" />
 
 The cache above is the *daemon's*. This one is the command side's: an opt-in, node-local cache of aggregate
-snapshots that lets `FetchForWriting` skip loading the stored snapshot and read only the events after it.
-Think of it as an identity map for aggregates with a lifetime longer than a session.
+snapshots that gives `FetchForWriting` a baseline to start from, so it reads only the events after it
+instead of the stored snapshot or the whole stream. Think of it as an identity map for aggregates with a
+lifetime longer than a session.
 
 It is off for every aggregate type, and is enabled per type:
 
@@ -165,6 +166,14 @@ opts.Projections.Snapshot<Order>(SnapshotLifecycle.Async);
 
 // Keep up to 1000 recently fetched Order snapshots
 opts.Events.CacheAggregatesForWriting<Order>(sizeLimit: 1000);
+```
+
+It works on all three lifecycles, including `Live` aggregates that have no stored snapshot at all:
+
+```csharp
+opts.Projections.LiveStreamAggregation<Order>();
+
+opts.Events.CacheAggregatesForWriting<Order>();
 ```
 
 Per type rather than store-wide because the win is proportional to how often *one* stream is fetched for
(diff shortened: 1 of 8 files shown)


erdtsieck:
Moved this to **draft**: writing more tests turned up a hole that is not specific to `Live`, and I would rather not have this reviewed as ready while the answer to it is open.

## What the test found

Storing the aggregate at fetch time re-publishes the very instance the caller is still holding. A later fetch then claims that entry, folds its delta onto it, and silently moves the first caller's aggregate to state it never fetched and never asserted a version against.

`a_second_fetch_must_not_fold_onto_an_instance_another_session_still_holds` (deliberately red on this branch):

```csharp
// one writer fetches and holds its aggregate for the length of its session, as a handler does
await using var holder = theStore.LightweightSession();
var held = await holder.Events.FetchForWriting<SimpleAggregate>(streamId);
held.Aggregate.ACount.ShouldBe(3);

// meanwhile the stream moves on somewhere else
await using (var other = theStore.LightweightSession())
{
    other.Events.Append(streamId, new AEvent());
    await other.SaveChangesAsync();
}

// a second writer fetches the same stream and folds its delta
await using (var second = theStore.LightweightSession())
{
    var stream = await second.Events.FetchForWriting<SimpleAggregate>(streamId);
    stream.Aggregate.ACount.ShouldBe(4);
}

held.Aggregate.ACount.ShouldBe(3);   // fails: 4
```

`TryTake`'s take-on-read guarantees a single *claimer* per entry, and its doc comment says why that matters:

> Aggregates are commonly mutable and a store folds delta events onto the instance it is handed, so two callers holding the same instance would corrupt each other's result. Exactly one caller may ever win a given entry.

That holds only until the store hands the same object back. Any `Store` at fetch time does exactly that, while the fetching caller keeps the instance in `stream.Aggregate` and in the session's item map for the rest of its session.

## It is in released behaviour, not just here

`FetchAsyncPlan.ForUpdate` stores at fetch time the same way, so the identical probe against `caching_async_aggregates_for_writing` fails identically — `held.Aggregate.ACount` is 4 where 3 is correct. I ran it as a temporary test and removed it again; it is a two-line edit of the version above if you want it in the repo. So this affects `Async` since 9.28, independently of this PR.

Also worth noting: the same immediate-store choice made `two_writers_on_one_stream_leave_the_cache_in_a_correct_state` report **two** hits rather than one for two sequential writers, because the first writer's store puts the entry back before the second fetches. Aggregate correctness there is fine; the hit count is just the visible symptom of the same thing.

## Where I think this leaves it

#5225 called this out in advance under *"The one genuinely new hazard is instance sharing"*, and answered it:

> **Default: cache the serialized snapshot** and materialize a fresh instance per fetch. Every caller gets its own object, mutation cannot leak, and there is no data race — while still skipping the read and fold of N events in favour of one document deserialization.
> **Opt-in: share the instance** for aggregates that honour the decider contract.

The merged implementation caches the instance. For `Live` the serialized default would be self-contained and still a large win — one deserialize per hit instead of N event deserializes plus N folds — and it would make take-on-read unnecessary rather than load-bearing. But it changes the shape of `IAggregateWriteCache`, which Polecat and Fisher share, so it is not mine to decide inside this PR.

Options as I see them, happy to implement whichever you prefer:

1. Serialized snapshots as the default across all three lifecycles, per the issue, with instance sharing as the documented opt-in. Cleanest, biggest change, needs a jasperfx-side call.
2. Keep instance caching but never publish at fetch time — store only once nothing can still hold the instance. Under `Live` that has to be at the *fetched* version rather than the committed one (the instance only moves if `FetchLatest` folded, which this branch already evicts on). Narrows the window rather than closing it.
3. Document the precondition and leave it. I would argue against: the existing `UseIdentityMapForAggregates` wording exempts mutation "outside of Marten internals", and this fold *is* Marten internals.

The rest of the PR stands on its own and is green: the delta narrowing, the archived-stream eviction, the `FetchLatest` eviction, and the expected-version and batched paths. Tell me which direction you want for the sharing question and I will finish it.

PR #5322, opened 2026-09-02
Cover jasperfx#733: constructor-based Create dropped when ShouldDelete is present

jeremydmiller:
Draft: **pinned at JasperFx 2.61.0, which is not published yet.** The fix lives in [JasperFx/jasperfx#733](https://github.com/JasperFx/jasperfx/issues/733) (branch `gh-733/event-ctor-create-with-should-delete`). CI will not go green here until 2.61.0 is on nuget.org.

## The bug

A self-aggregating snapshot may declare its Create handler as an event-shaped constructor, `public Foo(FooCreated e)`, instead of a named `static Create`. That works on its own. Adding a `ShouldDelete` method to the same aggregate switched the source generator to a different emitter — one that built its dispatch switch from named conventional methods only — so the constructor's event type got **no case arm at all**.

Nothing failed loudly. The generated code compiled, the constructor never ran, and the next Apply-only event built the aggregate through `RuntimeHelpers.GetUninitializedObject`, skipping every field initializer. The reporter saw:

```
JasperFx.Events.Daemon.ApplyEventException: Failure to apply event #0 Id(...)
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at TempGuidAggregate.Apply(TagAdded @event)
```

...out of an `Apply` that appended to a collection property. The quieter outcome, where no `Apply` happens to dereference anything, is a silently blank aggregate.

## What is in this PR

- `Bug_jasperfx_733_event_constructor_create_with_should_delete` — the reporter's shape across inline, async and live aggregation, plus the delete arm itself so the fix is not asserted only on the newly restored Create. All four fail on 2.60.0 and pass on 2.61.0.
- The JasperFx pin bump to 2.61.0, with the usual note in `Directory.Packages.props`.

The test aggregate's parameterless constructor is deliberately private and `[JsonConstructor]`-annotated. The generator only emits `new Foo()` for a **public** parameterless constructor and falls back to `GetUninitializedObject` otherwise, so keeping it private preserves the reported NRE instead of papering over it, while still giving System.Text.Json something to deserialize through.

## Upstream fix

Three changes in `EvolverCodeEmitter`, all mirroring what the Evolve emitters already did:

1. `EmitSelfAggregatingDetermineAction`'s event-type set now concats `info.EventConstructors`, and a type whose only Create is a constructor emits `snapshot = new T(data);` under the same `snapshot == null` guard.
2. `EmitSelfAggregatingCreateApplyStatements` takes the constructor too, so an event covered by *both* a constructor and a `ShouldDelete` predicate returning false still folds in the delete arm's `else` — the #652 shape, which had the same gap.
3. `EmitEventTypesProperty` folds constructors in as well. That omission was wider than the reported bug: it dropped the creating event from the generated `EventTypes` on **every** self-aggregating path, including the ones whose dispatch was already correct.

## Verification

Run against a local 2.61.0 prerelease pack of the fix:

- `EventSourcingTests` — 2002 passed, 0 failed, 7 skipped
- `DaemonTests` — 319 passed, 0 failed
- New suite — 4/4 fail on 2.60.0, 4/4 pass on the fix

Both documented workarounds from the issue (convert the constructor to a `static Create`, or register the delete through `DeleteEvent<T>()` instead of `ShouldDelete`) become unnecessary on this bump.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -376,16 +376,36 @@
          that replays one projection over a stream slice or DCB tag match and prints per-event
          before/after state, writing nothing. It is picked up by JasperFx.CommandLine discovery, so
          every Marten host referencing JasperFx.Events gains the command on this bump with no Marten
-         change, which is worth knowing about when it shows up in the CLI help output. -->
-    <PackageVersion Include="JasperFx" Version="2.60.0" />
-    <PackageVersion Include="JasperFx.Events" Version="2.60.0" />
+         change, which is worth knowing about when it shows up in the CLI help output.
+
+         JasperFx 2.61.0: jasperfx#733, a source generator fix that only bites self-aggregating
+         snapshots whose "Create" handler is an event-shaped constructor, `public Foo(FooCreated e)`,
+         rather than a named `static Create`. That works on its own. Adding a ShouldDelete method
+         to the same aggregate switched the generator to a different emitter, one that built its
+         dispatch switch from named conventional methods only, so the constructor's event type got no
+         case arm at all. Nothing failed loudly: the constructor never ran, and the next Apply-only
+         event built the aggregate through RuntimeHelpers.GetUninitializedObject, skipping every
+         field initializer. The reporter saw an ApplyEventException wrapping a NullReferenceException
+         out of an Apply that appended to a collection property; the general symptom is a silently
+         blank aggregate. Both documented workarounds (converting the constructor to a static Create,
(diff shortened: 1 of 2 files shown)


jeremydmiller:
JasperFx 2.61.0 is [published and indexed](https://www.nuget.org/packages/JasperFx.Events/2.61.0), so this is no longer blocked — marking ready for review.

Re-verified against the real package (not the local prerelease pack): `Bug_jasperfx_733_event_constructor_create_with_should_delete` is 4/4 green on 2.61.0.

Issue #4680, opened 2026-06-08
Upcaster bypassed when the source event type is appended (typed) into the same DocumentStore


erdtsieck:
## Behavior

When an event-type rename is handled with `options.Events.Upcast<TOld, TNew>(...)`, the upcaster is **bypassed on read if a `TOld` instance was appended (typed) into the same `DocumentStore` instance**. The event reads back as `TOld`, not `TNew` — even when read from a brand-new session straight from the database.

Reading the same stored event from a *different* store instance (one that registers the upcaster but has never appended `TOld`) *does* upcast correctly. So it appears that appending the typed source event registers a concrete `EventMapping` for `TOld` that shadows the upcaster when resolving the stored event-type name on read.

## Environment

- Marten 9.6.0
- JasperFx.Events 2.8.2
- .NET 10
- System.Text.Json serializer (`EnumStorage.AsString`)
- PostgreSQL

## Minimal reproduction

```csharp
public record UpcastReproOld(string Value);
public record UpcastReproNew
{
    public required string Value { get; init; }
}

[Fact]
public async Task Same_store_typed_append_is_not_upcast_on_read()
{
    await using var store = DocumentStore.For(opts =>
    {
        opts.Connection(ConnectionString);
        opts.UseSystemTextJsonForSerialization(enumStorage: EnumStorage.AsString);
        opts.Events.Upcast<UpcastReproOld, UpcastReproNew>(x => new() { Value = x.Value });
    });
    await store.Storage.Database.ApplyAllConfiguredChangesToDatabaseAsync();

    var streamId = Guid.NewGuid();
    await using (var write = store.LightweightSession())
    {
        write.Events.StartStream(streamId, new UpcastReproOld("hello"));
        await write.SaveChangesAsync();
    }

    await using var read = store.LightweightSession(); // fresh session, reads from DB
    var events = await read.Events.FetchStreamAsync(streamId);

    // Actual: events[0].Data is UpcastReproOld
    // Hoped:  events[0].Data is UpcastReproNew
    events[0].Data.ShouldBeOfType<UpcastReproNew>();
}
```

## Context / why this matters

We are renaming a large set of event records (with property-name/shape changes, hence upcasters rather than plain aliases) and registering `Upcast<TOld, TNew>` for each. Production works fine, because the running deployment never appends `TOld` anymore — the legacy events were written by a previous deployment, and on read they upcast correctly.

[...]

PR #5332, opened 2026-09-04
Resolve `mt_dotnet_type` to a type that hasn't been registered yet

JonathanS-NTI:
## Problem

Two event types that share a simple name but live in different namespaces collide on a single
`mt_events.type` alias, because Marten derives the alias from `Type.Name` only. `mt_dotnet_type`
exists to disambiguate them on read, and `EventDocumentStorage.Resolve`/`ResolveAsync` does consult
it — but only through `EventGraph.TryGetRegisteredMappingForDotNetTypeName`, which was a linear scan
over mappings that are **already registered**:

Nothing registers event types up front. `StoreOptions.Scan` only registers types matched by an
explicit event-type matcher, `ApplySourceGeneratedManifest` needs a generated manifest, and
**projection registration does not register event types at all**. So, in a store that doesn't call
`AddEventType`, the graph is populated lazily by whichever code path touches a type first — an
append, or a read whose *alias* lookup misses.

A process that has only ever touched one side of a collision therefore gets `null` back from the
lookup above, leaves the wrong mapping in place, and deserializes the other side's rows into the
wrong CLR type. When the two payloads are JSON-compatible — which same-named events in one domain
usually are — this is completely silent: no exception, nothing logged, just an event whose missing
members are defaulted.

## Why it looks intermittent

Registration is per-process and driven by runtime ordering, so the same stored event resolves
correctly on one node and incorrectly on another. A node self-heals as soon as it happens to *append*
the missing type, because appends go through `EventMappingFor(Type)`. Reads never registered
anything, so a read-mostly node (an async daemon that isn't handling the command that writes the
other type) can stay wrong indefinitely — and restarting the projection agent doesn't help, because
the event graph is store-scoped, not agent-scoped.

## Fix

Fall back to `TypeForDotNetName` + `EventMappingFor(Type)` when the scan misses. `EventMappingFor`
self-registers, so the first read of a colliding type also repairs the graph for subsequent reads.

The `catch` preserves the old behavior for a stored type this process genuinely can't load — a
foreign event, or an assembly it doesn't reference — so the caller still falls back to the alias
mapping rather than throwing on read.

`Upcast` targets are unaffected: `EventDocumentStorage` already skips the swap entirely when
`mapping.IsUpcastTarget`, per #4680.

## Behavior change worth a maintainer's eye

`MapEventType` sets no equivalent of `IsUpcastTarget`, which leaves one scenario changed:

| Rename scenario | Before | After |
| --- | --- | --- |
| `MapEventType<Bar>("foo")`, old `Foo` type **deleted** | alias wins — `Foo` is unresolvable | unchanged; `TypeForDotNetName` throws and the `catch` returns null |
| `MapEventType<Bar>("foo")`, old `Foo` type **still in the assembly** | swaps to `Foo` *only if* `Foo` happens to be registered, otherwise alias wins | always swaps to `Foo` |

The ordinary rename (old type removed) is unaffected. The second row is a real change — though note
the existing behavior there is already ordering-dependent, so this makes it consistent rather than
introducing the inconsistency.

Whether "consistently prefer the stored `mt_dotnet_type`" is the right answer, or whether
`MapEventType` should get an `IsUpcastTarget`-style guard the way `Upcast` has, is a call I'd rather
leave to the maintainers.

## Tests

`Bug_XXXX_dotnet_type_disambiguation_requires_registered_mapping`, two facts at different levels:

- `dotnet_type_lookup_finds_a_type_that_has_not_been_registered_yet` — no database. Asserts the
  lookup resolves a type the graph hasn't seen. Fails on `master`, passes with the fix.
- `resolves_by_dotnet_type_when_the_colliding_type_is_not_registered` — end-to-end. A write store
  that knows both colliding types appends the Beta event; a read store that only knows Alpha reads it
  back and, pre-fix, gets an Alpha instance.

The two event types are deliberately JSON-compatible (Beta adds an `int Count`) so that a wrong-type
read fails silently rather than throwing — that's what makes the production failure invisible, so it
belongs in the repro.

## Notes for review

- Also considered, and deliberately **not** included here: `Cache.Fill` is first-wins, so an alias
  collision currently resolves arbitrarily with no warning at bootstrap. An assertion or a startup
  warning for duplicate aliases feels like a separate change.

### DIFF
--- a/src/EventSourcingTests/Bugs/Bug_XXXX_dotnet_type_disambiguation_requires_registered_mapping.cs
+++ b/src/EventSourcingTests/Bugs/Bug_XXXX_dotnet_type_disambiguation_requires_registered_mapping.cs
@@ -1,0 +1,101 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using EventSourcingTests.Bugs.AliasCollisionAlpha;
+using EventSourcingTests.Bugs.AliasCollisionBeta;
+using JasperFx.Events;
+using Marten;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace EventSourcingTests.Bugs
+{
+    /// <summary>
+    /// Marten derives an event's <c>mt_events.type</c> alias from the CLR type's simple name only, so two event
+    /// types that share a name but live in different namespaces collide on one alias. <c>mt_events.mt_dotnet_type</c>
+    /// exists to disambiguate them on read, and <c>EventDocumentStorage.ResolveAsync</c> does consult it — but only
+    /// via <c>EventGraph.TryGetRegisteredMappingForDotNetTypeName</c>, which is a linear scan over mappings that are
+    /// <em>already registered</em>. It never falls back to <c>TypeForDotNetName</c>.
+    ///
+    /// <para>
+    /// Nothing registers event types up front — not even projections — so the graph is populated lazily by whichever
+    /// code path touches a type first (an append, or a read whose alias lookup misses). A process that has only ever
+    /// touched one side of a collision therefore silently deserializes the other side's rows into the wrong CLR type.
+    /// It is silent whenever the two payloads are JSON-compatible, which same-named events in one domain usually are.
+    /// </para>
+    ///
+    /// <para>
+    /// Because registration is per-process and driven by runtime ordering, the same stored event resolves correctly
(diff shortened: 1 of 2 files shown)


JonathanS-NTI:
`mt_dotnet_type` disambiguation is deliberately limited to registered mappings; broadening it breaks documented namespace migration (EventsNamespaceChange). Registration is the signal of intent.

Issue #5347, opened 2026-09-06
Consume JasperFx 2.65.0 to close the three compliance facts fixed by jasperfx#780


jeremydmiller:
Tracking issue for the bump that turns three red compliance facts green. All three were found by #5343 (the first run of the 2.64.0 suites against a real event store) and are already **fixed and merged** upstream in JasperFx/jasperfx#780, but 2.64.0 is the published version Marten pins, so `master` carries them red until a JasperFx release ships.

Nothing to do here but bump `JasperFx*` to the release carrying jasperfx#780 and re-run `src/EventSourcingTests` — no Marten code change is expected.

### 1. `stream_archiving_compliance.capturing_an_archived_event_archives_a_string_identified_stream`

A **product** bug in `JasperFxSingleStreamProjectionBase.ApplyInline`, not a Marten one, and it affects Marten users directly: an `Archived` event appended to a snapshotted stream **with no other handled event in the same batch** left `transformed` null, so an early `continue` skipped `maybeArchiveStream` and the stream stayed live.

Worth recording why this survived so long: **every one** of Marten's local archiving tests in `src/EventSourcingTests/archiving_events.cs` appends `new DEvent()` next to `new Archived(...)`, so not one of them covered the lone-marker case. Reproduced under both Guid and string stream identity, so the identity in the failing fact's name is incidental — it is simply the only one of the four archiving facts that appends the marker alone.

Consider adding a Marten-local regression for the lone-`Archived` case on the bump, since the local file's blind spot is what let this through.

### 2 & 3. `projection_side_effect_compliance.{side_effects_are_suppressed_during_a_rebuild, no_messages_are_published_during_a_rebuild}`

Suite bugs. Both asked the daemon for `nameof(ComplianceWatchtowerProjection)`, assuming a store derives a projection's registered name from the projection type; Marten derives it from the aggregated document type, so both died on *"No registered projection matches the name 'ComplianceWatchtowerProjection'. Available names are 'ComplianceWatchtower'"*. The suite now pins the projection's `Name` explicitly.

### Also in jasperfx#780

[...]

Issue #5344, opened 2026-09-06
Natural keys: three gaps found by NaturalKeyCompliance (jasperfx#764)


jeremydmiller:
Found by #5343, enrolling `NaturalKeyCompliance` from the JasperFx 2.64.0 wave. This is the suite's first run against any real event store, and Marten's `[NaturalKey]` storage half (the `mt_natural_key_X` lookup table from #4788) fails three of its facts. All three are real, and none is a suite defect.

### 1. `a_second_stream_cannot_claim_a_live_natural_key`

The suite expects `JasperFx.Events.DuplicateNaturalKeyException` when a second stream tries to claim a natural key a live stream already holds. Marten throws nothing and lets the claim through, silently repointing the lookup row.

This is the uniqueness ruling settled in jasperfx#764 / jasperfx#772. Marten's lookup table has no unique constraint on the key, so the second claimant wins. The suite reads the mapping back afterwards specifically to catch a store that throws but has *already* repointed the row, so a fix needs both the refusal and the row left intact.

### 2. `an_archived_stream_no_longer_resolves_by_natural_key`

After `ArchiveStream(streamId)` (the explicit call, not an `Archived` event), `FetchLatest<T, TKey>(naturalKey)` still returns the aggregate. Marten's `mt_natural_key_X` row survives archival, so an archived stream stays resolvable by its natural key indefinitely.

Fisher resolves this by joining the streams table rather than copying the flag; Polecat's rebuild test asserts the archived stream's row does not come back. Marten does neither.

### 3. `natural_key_with_a_primitive_string_key`

On a **Guid**-stream-identity store, an aggregate whose natural key is a primitive `string` cannot be fetched by that key:

```
System.InvalidOperationException : This Marten event store is configured to identify streams with Guids
   at Marten.Events.EventGraph.EnsureAsStringStorage(IMartenSession session)
   at Marten.Events.EventStore.FindFetchPlan[TDoc,TId]()
   at Marten.Events.EventStore.FetchForWriting[T,TId](TId id, CancellationToken cancellation)
```

`FindFetchPlan<TDoc, TId>` sees `TId == string` and asserts string *stream* storage, without considering that the `string` here is a natural key rather than a stream key. A strong-typed natural key works (the other facts pass); only the primitive-string case collides with the stream-identity guard.

---

[...]

PR #5350, opened 2026-09-06
Rebuild teardown clears the EF Core table, not mt_doc (#5329)

### LINKED ISSUE #5329: Rebuilding an EF Core projection tries to truncate a non-existent Marten document table
EF Projections persist data in DbContext-managed tables, but rebuilding it through `IProjectionDaemon` still attempts to clear Marten’s conventional document table for TDoc.
Those tables are not created under normal schema generation, so when building the schema via SQL migration and setting `AutoCreateSchemaObjects = AutoCreate.None`, running `IProjectionDaemon.RebuildProjectionAsync` fails with
> Npgsql.PostgresException : 42P01: relation "public.mt_doc_customerorderhistory" does not exist

See failing repro test [here](https://github.com/markotny/marten/commit/7f86538afdb725efa383d7739b6225f1a3f0e1b7).

jeremydmiller:
Fixes #5329

## Two bugs, one cause — and the reported one is the safer half

**The reported bug (loud).** Rebuilding an EF Core projection through `IProjectionDaemon` tried to clear Marten's *conventional* document table for `TDoc`. That table is never created for an EF-persisted projection, so with the schema built by SQL migration and `AutoCreateSchemaObjects = AutoCreate.None`:

```
Npgsql.PostgresException : 42P01: relation "public.mt_doc_customerorderhistory" does not exist
```

**The unreported bug (silent, and worse).** Everywhere the first symptom *didn't* fire — i.e. any store where Marten had auto-created that unused table — the rebuild truncated an empty Marten table, left every real EF Core row in place, and **reported success**. So a rebuild silently replayed on top of stale data. Nothing threw, nothing logged, and the projection looked rebuilt. Before this PR, an EF Core projection's data was never cleared on rebuild at all.

The `42P01` is the better failure of the two: it fails loudly and stops. The silent one is the shape that ships.

Reproduced verbatim from @markotny's repro before touching anything:

```
truncate table efcore_rebuild_5329.mt_doc_customerorderhistory CASCADE
42P01: relation "efcore_rebuild_5329.mt_doc_customerorderhistory" does not exist
```

## Root cause

Every aggregate projection inherits `Options.DeleteViewTypeOnTeardown<TDoc>()` from `JasperFxAggregationProjectionBase`, and Marten's `AsyncOptionsExtensions.Teardown` turned that into a `TruncateTable(TDoc)` resolved through `session.StorageFor(...)`. But an EF Core projection's rows live in a `DbContext`-mapped table, and `DocumentSessionBase.FetchProjectionStorageAsync` already skips `EnsureStorageExistsAsync(typeof(TDoc))` for these types — so `mt_doc_<tdoc>` is never created. Teardown simply never asked the question the write path already asks.

## The fix

Teardown consults `StoreOptions.CustomProjectionStorageProviders` — the same registry the write path uses — and skips a `DeleteDocuments` cleanup whose type is not in Marten document storage. The EF Core projections then register the table they *actually* write via `Options.DeleteDataInTableOnTeardown`, resolved through the same EF-model mapping and schema rule that `AddEntityTablesFromDbContext` uses to create it (factored into a shared `mapToMartenSchema`, so the teardown target and the migrated table cannot drift apart).

Considered and rejected: adding a teardown member to `IProjectionStorage<TDoc,TId>`, or giving `AsyncOptions.CleanUps` a removal API so the EF projection could drop the inherited cleanup. Both need a JasperFx release and a pin bump to land a Marten bug fix; the registry lookup is Marten-side only and reuses a seam that already exists to answer exactly this question.

## Blast radius

Rebuild teardown is shared by every projection type, and skipping too eagerly would leave real Marten projections sitting on stale rows while still reporting success — the same silent-success shape, just pointed at a different table. A conventional projection has no entry in `CustomProjectionStorageProviders`, so nothing about its teardown changes, and that is asserted rather than assumed:

- `Bug_5329...rebuild_of_a_conventional_marten_projection_still_truncates_its_document_table` rebuilds an ordinary Marten projection **out of the same store as the EF Core one**, so the guard is live, and asserts a seeded row that no event produces is gone afterwards.
- `AsyncOptionsTests.teardown_skips_view_types_that_are_not_stored_by_marten` (and its per-tenant twin) pin the sibling type that must *still* be truncated.

Falsified rather than assumed: making the skip unconditional fails exactly the conventional-projection test and leaves the two EF ones green. `DaemonTests.build_aggregate_projection.rebuild_the_projection_clears_state` — the existing regression test for "conventional projection data IS deleted on rebuild" — also passes.

Composite projections route every leaf through the same `teardownProjectionStorage`, so EF members added via `composite.Add(...)` get the same treatment.

## Known limitation

A **per-tenant** rebuild scopes the wipe with `delete from <entity table> where tenant_id = ?`. That column name is not read from the EF model, so it works for a snake_case `DbContext` (as the conjoined-tenancy sample and tests use) but yields `42703` under EF Core's default `TenantId` naming. Strictly better than the previous unconditional `42P01`, but documented in `docs/events/projections/efcore.md` and tracked in #5351 rather than left in a PR thread.

## Found but deliberately not fixed here

The same wrong assumption exists in **schema generation**: registering an EF Core projection makes Marten build a `DocumentMapping` for `TDoc` and create a useless `mt_doc_<tdoc>` table, and it throws `InvalidDocumentException` from every full-schema operation for an EF entity whose key is not named `Id`. Filed as #5351. That change alters which schema objects Marten generates for *all* stores, so its regression surface is the schema/migration suites rather than the daemon suites — a much wider blast radius than a teardown fix, and not something to bundle late in a cycle alongside a concurrent JasperFx/Weasel pin bump.

## Validation

- `Marten.EntityFrameworkCore.Tests` — 48/48 green (45 pre-existing + 3 new), re-run after merging current `master`.
- `DaemonTests` teardown/rebuild areas — `multistream_data_teardown_on_rebuild`, `Bug_5175_composite_member_teardown`, `rebuilding_an_inline_projection`, `AsyncOptionsTests` — green.
- Full `DaemonTests` run compared against an unmodified `origin/master` baseline; the residual failures are pre-existing and unrelated to teardown (high-water detection, gap detectors, dead letters, coordinator lifecycle).
- `src/Marten` and `src/Marten.EntityFrameworkCore` build clean on both `net9.0` and `net10.0`.
- Docs updated; markdownlint and cspell clean.

### DIFF
--- a/docs/events/projections/efcore.md
+++ b/docs/events/projections/efcore.md
@@ -396,6 +396,21 @@ public class MyProjection
 }
 ```
 
+## Rebuilding
+
+`IProjectionDaemon.RebuildProjectionAsync` works normally for EF Core projections. Because the
+projection's data lives in the `DbContext`-mapped table rather than in a Marten document table, the
+rebuild's teardown step clears **that** table -- Marten's conventional `mt_doc_<tdoc>` table is
+neither created nor truncated for an EF Core-backed aggregate.
+
+This matters when you build your schema with a separate migration and run with
+`AutoCreateSchemaObjects = AutoCreate.None`: there is no `mt_doc_<tdoc>` table to find, and the
+rebuild does not go looking for one.
+
+If you need to suppress the teardown entirely -- for instance when the entity table is shared with
+data Marten did not write -- set `Options.TeardownDataOnRebuild = false` in your projection's
+constructor and clear the table yourself.
+
 ## How It Works
 
 Under the hood, EF Core projections:
@@ -404,3 +419,4 @@ Under the hood, EF Core projections:
 2. **Register a transaction participant** so the DbContext's `SaveChangesAsync` is called within Marten's transaction, ensuring atomicity
 3. **Migrate entity tables** through Weasel alongside Marten's own schema objects, so `dotnet ef` migrations are not needed
 4. **Use EF Core change tracking** for insert vs. update detection (detached entities are added; unchanged entities are marked as modified)
+5. **Register the entity table as the rebuild teardown target**, so a rebuild clears the rows EF Core actually wrote
--- a/docs/events/projections/efcore.md
+++ b/docs/events/projections/efcore.md
@@ -411,6 +411,21 @@ If you need to suppress the teardown entirely -- for instance when the entity ta
 data Marten did not write -- set `Options.TeardownDataOnRebuild = false` in your projection's
 constructor and clear the table yourself.
 
+::: warning
+A **per-tenant** rebuild (`RebuildProjectionAsync` scoped to one tenant) scopes the wipe with
+`delete from <entity table> where tenant_id = ?`. That column name is not read from your EF Core
+model, so it only matches a `DbContext` that maps the tenant property to a snake_case `tenant_id`
+column -- as the conjoined multi-tenancy example above does. Under EF Core's default naming the
+column is `TenantId` and the rebuild fails with `42703: column "tenant_id" does not exist`. Map the
+column explicitly:
+
+```csharp
+entity.Property(e => e.TenantId).HasColumnName("tenant_id");
+```
+
+See [#5351](https://github.com/JasperFx/marten/issues/5351). A store-wide rebuild is unaffected.
+:::
+
 ## How It Works
 
 Under the hood, EF Core projections:
--- a/src/DaemonTests/Internals/AsyncOptionsTests.cs
+++ b/src/DaemonTests/Internals/AsyncOptionsTests.cs
@@ -1,9 +1,12 @@
+using System.Collections.Generic;
+using System.Linq;
 using JasperFx.Events.Projections;
 using Marten;
 using Marten.Events.Daemon;
 using Marten.Internal.Operations;
 using Marten.Testing.Documents;
 using NSubstitute;
+using Shouldly;
 using Xunit;
 
 namespace DaemonTests.Internals;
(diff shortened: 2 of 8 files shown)


jeremydmiller:
## Validation note: a shared test database manufactured a convincing false regression

Recording this because it cost real time here and it is the **third** independent instance across the fleet today — a sibling agent hit the same thing while validating #5347, and again while validating #5344.

Validating this PR against the shared `marten_testing` database produced a sequence that looked exactly like a regression in the rebuild path:

| Run | Tree | Tests | Failures |
|---|---|---|---|
| 1 | branch (machine loaded) | 321 | 29 |
| 2 | branch (machine loaded) | 321 | 26 |
| 3 | unmodified `master` | 319 | **0** |
| 4 | branch (quiet machine) | 321 | 2 — set A |
| 5 | unmodified `master` | 319 | **0** |
| 6 | branch, **fix only, my 2 added tests removed** | 319 | 2 — set C |

Branch nonzero on every run, `master` clean on every run. That asymmetry is what makes it convincing, and it survived the obvious first hypothesis: run 6 removed the two tests this PR adds, making the test count identical to `master`, and it *still* failed two — ruling out "extra tests perturb parallel scheduling."

**It was entirely an artifact of the shared database.** Every failure set was disjoint from every other, every failing test passed 3/3 in isolation, and the failures clustered in schema-touching areas (gap detection, dead letters, unknown event types, scoped services) that this diff cannot reach. Runs against the two trees were interleaved against one database, so the apparent branch-vs-master split actually tracked *when* each run happened relative to accumulated schema state — not which tree was under test.

Re-run with each tree on its own freshly created database (`marten_testing_database` env var, which `ConnectionSource` honours):

```
BRANCH   on fresh DB:  321 total, 0 failed, 321 succeeded
BASELINE on fresh DB:  319 total, 0 failed, 319 succeeded
```

Clean on both sides, matching the 319/0 the sibling agent independently reproduced for `DaemonTests` on a fresh database.

**Takeaway for anyone validating locally:** `DaemonTests` and `CoreTests` accumulate cross-test state, and concurrent suites against one database can deadlock and leave residue. Give each full-suite run its own database before reading anything into a failure — otherwise the shared database will hand you a different plausible-looking regression every time, and the disjoint failure sets are the tell.

The structural reason this PR could not have caused those failures, for the record: `DaemonTests.csproj` does not reference `Marten.EntityFrameworkCore`, so `StoreOptions.CustomProjectionStorageProviders` is empty in every one of its stores, the new guard never fires, and teardown queues byte-identical operations to `master`. The direct regression tests — `rebuild_the_projection_clears_state`, `multistream_data_teardown_on_rebuild`, `Bug_5175_composite_member_teardown` — passed in *every* run, including the noisy ones.

Issue #5353, opened 2026-09-06
A failed unit of work still commits its outbox batch


jeremydmiller:
Found while attempting the exception consolidation in #5346, hiding behind the failure that issue is about.

## The finding

`ProjectionSideEffectCompliance.a_unit_of_work_that_fails_publishes_nothing` (from jasperfx#763) fails on two independent counts. The first is the exception type, which #5346 covers. Once that half is addressed by subclassing, the fact fails on its **next** assertion:

```
_outbox.CommittedBatches should be empty but had 1 item
```

So a unit of work that fails still commits an outbox batch. The messages a projection published during a `SaveChangesAsync` that ultimately threw are not rolled back with it.

## Why this matters more than the assertion count suggests

The whole point of the transactional outbox is that published messages and persisted state commit or fail *together*. A failed unit of work that still commits its batch produces exactly the divergence the outbox exists to prevent: downstream consumers receive messages describing a transaction that never landed. That is unrecoverable by retry — the state can be rebuilt, the already-delivered messages cannot be recalled.

It is also the second outbox defect this compliance wave has surfaced in Marten. #5348 fixed `SaveChangesAsync` never calling `IMessageBatch.BeforeCommitAsync` at all — the before-hook was dead code, so the half of the contract that enlists the batch in the transaction never ran. These are plausibly the same root cause seen from two sides, and whoever picks this up should check that first rather than assuming two separate fixes.

## Reproduction

Enrol `ProjectionSideEffectCompliance` (already enrolled on `master` via #5348) and address only the exception-type half of `a_unit_of_work_that_fails_publishes_nothing`; the outbox assertion then fails. The fact is currently red on `master` and tracked as one of the two remaining red compliance facts.

## Note on scope

Filed separately from #5346 deliberately. #5346 is a type-hierarchy question (whether `MartenException` can be preserved while adopting the shared exceptions) that is currently blocked on a maintainer decision; this is a behavioural defect in the outbox that stands on its own and should not wait on that ruling.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5354, opened 2026-09-06
gh-5346: name Marten's own six exceptions from the compliance fixture

### LINKED ISSUE #5346: Consolidate Marten's six event exception types onto the JasperFx.Events copies (jasperfx#756)
The follow-up jasperfx#756 named as "stores will subclass or type-forward in a later node". #5343 hit it twice while bumping to JasperFx 2.64.0, and paid for it with two alias blocks that exist only until this lands.

## What 2.64.0 changed

jasperfx#756 added canonical versions of six exception types that were previously declared once per store:

`UnknownEventTypeException`, `NonExistentStreamException`, `ExistingStreamIdCollisionException`, `EventDeserializationFailureException`, `StreamLockedException`, `DefaultTenantUsageDisabledException`

Each carries superset constructors/properties so every store's current usage stays expressible, plus a protected message-overriding constructor so a store whose copy diverged in wording can subclass **without breaking tests that pin its message**.

## What it cost #5343

1. **Ambiguity everywhere.** Every file importing both `JasperFx.Events` and `Marten.Exceptions` now sees CS0104. #5343 resolved it the API-preserving way — global-using aliases pointing the simple names at Marten's types — in two places:
   - `src/Marten/GlobalUsings.cs` (six aliases)
   - `src/Tests.props` (six `<Using ... Alias=...>` items, covering `EventSourcingTests`, `TenantPartitionedEventsTests`, `MultiTenancyTests`, `DaemonTests`, `DocumentDbTests`, `CoreTests`)

   **Both blocks should be deleted by this issue's fix.**

2. **A red compliance fact.** `projection_side_effect_compliance.a_unit_of_work_that_fails_publishes_nothing` asserts the shared type:

   ```
   Expected ExistingStreamIdCollisionException, directly or aggregated,
   but got Marten.Exceptions.ExistingStreamIdCollisionException: Stream #... already exists
   ```

   Marten throws its own unrelated type, so the shared assertion fails. Making Marten's types **derive from** the shared ones fixes this fact with no message change, which is precisely what the protected message-overriding constructor was added for. More shared suites will assert these types over time, so this will not stay a one-fact problem.

## Scope

Subclass (preferred, preserves the public hierarchy callers already catch) or type-forward. Then delete both alias blocks above. This is a public-API change to Marten's exception hierarchy, which is why #5343 deliberately left it alone rather than folding it into a compliance enrolment.

jeremydmiller:
Closes #5346.

## The ruling

Marten keeps its own six event store exceptions. **No existing Marten behaviour changes.** Adoption
of the exceptions lifted into `JasperFx.Events` by JasperFx/jasperfx#751 is deferred to Marten 10.
The compliance suite is made variable on exception type instead (JasperFx/jasperfx#791).

Adopting the lifted types by subclassing, the way Polecat and Fisher did, would break
`all_exceptions_should_derive_from_MartenException`: C# has single inheritance, so a Marten type
cannot derive from both `MartenException` and the lifted JasperFx type.

## What this changes

One override on `MartenComplianceFixture`, naming Marten's own six:

| Kind | Marten type |
|---|---|
| `UnknownEventType` | `Marten.Exceptions.UnknownEventTypeException` |
| `NonExistentStream` | `Marten.Exceptions.NonExistentStreamException` |
| `ExistingStreamIdCollision` | `Marten.Exceptions.ExistingStreamIdCollisionException` |
| `EventDeserializationFailure` | `Marten.Exceptions.EventDeserializationFailureException` |
| `StreamLocked` | `Marten.Exceptions.StreamLockedException` |
| `DefaultTenantUsageDisabled` | `Marten.Exceptions.DefaultTenantUsageDisabledException` |

**No Marten production code changed** — one test-harness file, 36 added lines, nothing removed.

## This unblocks only half of one fact

`a_unit_of_work_that_fails_publishes_nothing` is still red, and legitimately so. Behind the exception
type assertion sits a second, independent bug: a failed unit of work still commits an outbox batch.
Run locally against jasperfx#791 via `ComplianceSourceDir`, the fact now gets past the type
assertion and fails on:

```
Shouldly.ShouldAssertException : _outbox.CommittedBatches
    should be empty but had
1
    item
```

That is #5353. Not chased here and deliberately not weakened.

## Merge ordering — please read

This does **not** compile against the currently referenced `JasperFx.Events.ComplianceTests`
package, because `ComplianceExceptionKind` and `ExceptionTypeFor` do not exist there yet:

```
src/Marten.Testing/Harness/MartenComplianceFixture.cs(519,43): error CS0246:
  The type or namespace name 'ComplianceExceptionKind' could not be found
```

`Marten.Testing.csproj` references the published package unconditionally — it has no
`ComplianceSourceDir` switch, unlike `EventSourcingTests.csproj`. So merging this before the JasperFx
package bump breaks `master`'s build. **Hold until jasperfx#791 is released and adopted**, the same
ordering as #5347.

## Validation

Built and run against jasperfx#791 via
`-p:ComplianceSourceDir=<jasperfx>/src/JasperFx.Events.ComplianceTests`, on a private database
(`marten_testing_database` pointed at a dedicated `marten_5346_varexc`) rather than shared
`marten_testing` state.

`projection_side_effect_compliance`: 9 total, 8 passed, 1 failed — the #5353 outbox assertion above.

### DIFF
--- a/src/Marten.Testing/Harness/MartenComplianceFixture.cs
+++ b/src/Marten.Testing/Harness/MartenComplianceFixture.cs
@@ -495,6 +495,42 @@ public class MartenComplianceFixture: EventStoreComplianceFixture<IDocumentOpera
     /// </summary>
     public override bool SupportsUnarchiveStream => false;
 
+    /// <summary>
+    ///     Marten names its own six exceptions rather than the ones lifted into JasperFx.Events by
+    ///     jasperfx#751.
+    /// </summary>
+    /// <remarks>
+    ///     <para>
+    ///         Not an oversight and not a gap. Marten enforces
+    ///         CoreTests.all_exceptions_should_derive_from_MartenException:
+    ///         every exception Marten throws must derive from MartenException, so that a caller can
+    ///         catch the whole family in one clause. C# has single inheritance, so a Marten type
+    ///         cannot derive from both MartenException and the lifted JasperFx type -- adopting the
+    ///         shared types by subclassing, the way Polecat and Fisher did, would break that
+    ///         convention. Marten therefore keeps all six in Marten.Exceptions and nominates them
+    ///         here. Revisiting the convention is deferred to Marten 10 (marten#5346).
+    ///     </para>
+    ///     <para>
+    ///         Fully qualified because JasperFx.Events is imported above and declares six types of
+    ///         exactly these names; the unqualified name would silently resolve to the wrong one and
+    ///         the assertion would fail with a type mismatch rather than a compile error.
+    ///     </para>
+    /// </remarks>
+    public override Type ExceptionTypeFor(ComplianceExceptionKind kind) =>
+        kind switch
+        {
+            ComplianceExceptionKind.UnknownEventType => typeof(Marten.Exceptions.UnknownEventTypeException),
+            ComplianceExceptionKind.NonExistentStream => typeof(Marten.Exceptions.NonExistentStreamException),
+            ComplianceExceptionKind.ExistingStreamIdCollision =>
+                typeof(Marten.Exceptions.ExistingStreamIdCollisionException),
+            ComplianceExceptionKind.EventDeserializationFailure =>
+                typeof(Marten.Exceptions.EventDeserializationFailureException),
+            ComplianceExceptionKind.StreamLocked => typeof(Marten.Exceptions.StreamLockedException),
+            ComplianceExceptionKind.DefaultTenantUsageDisabled =>
+                typeof(Marten.Exceptions.DefaultTenantUsageDisabledException),
+            _ => base.ExceptionTypeFor(kind)
+        };
+
     public override async ValueTask DisposeAsync()
     {
         foreach (var disposable in _disposables)


jeremydmiller:
Verified against the 2.66.0 pin, which landed in #5357 (merged).

`Marten.Testing` compiles clean on the new pin — **0 errors**. The `CS0246: ComplianceExceptionKind could not be found` that blocked this PR is resolved by jasperfx#791 shipping in 2.66.0. This branch merges `master` cleanly.

The failure-mode transition on `projection_side_effect_compliance.a_unit_of_work_that_fails_publishes_nothing` is the evidence this PR does its job:

**Before** (2.65.0 pin, and 2.66.0 without this PR) — fails on the exception-type assertion:
```
Expected ExistingStreamIdCollisionException, directly or aggregated, but got
Marten.Exceptions.ExistingStreamIdCollisionException: Stream #... already exists
```

**After** (this branch) — gets past that assertion and fails on the real thing:
```
_outbox.CommittedBatches should be empty but had 1 item
and was [JasperFx.Events.ComplianceTests.RecordingMessageBatch]
```

That second failure is **marten#5353**, the genuine outbox bug, and it stays red. The other 8 tests in the suite pass.

Merging now, after the pin bump, per the required order.

PR #5372, opened 2026-09-10
Mapped concurrency property fails optimistic lock

JurJean:
When mapping the version property using `opts.Schema.For<>().UseOptimisticConcurrency(true).Metadata(m => m.Version.MapTo(x => x.Etag))`, saving the document fails with a ConcurrencyException as demonstrated in the test `closed_shape_optimistic_concurrency_mapped_member_tests.mapped_version_member_should_be_used_as_the_expected_version_on_store` included in this PR.

For a potential fix see this branch:
https://github.com/JurJean/marten/tree/mapped-concurrency-fix

### DIFF
--- a/src/CoreTests/Storage/Identification/closed_shape_optimistic_concurrency_mapped_member_tests.cs
+++ b/src/CoreTests/Storage/Identification/closed_shape_optimistic_concurrency_mapped_member_tests.cs
@@ -1,0 +1,145 @@
+using System;
+using System.Threading.Tasks;
+using JasperFx;
+using Marten;
+using Marten.Exceptions;
+using Marten.Metadata;
+using Marten.Testing.Harness;
+using Shouldly;
+using Xunit;
+
+namespace CoreTests.Storage.Identification;
+
+/// <summary>
+/// The documented contract for a mapped version member —
+/// Metadata(m => m.Version.MapTo(x => x.Etag)) on a UseOptimisticConcurrency(true)
+/// document — is that Store() passes the version number from the document itself
+/// (docs/documents/concurrency.md), i.e. the mapped member should behave exactly
+/// like the IVersioned marker interface does.
+/// These tests pin the gap where that does not hold today.
+/// </summary>
+public class closed_shape_optimistic_concurrency_mapped_member_tests : BugIntegrationContext
+{
+    private DocumentStore mappedMemberStore()
+        => StoreOptions(opts =>
+        {
+            opts.Schema.For<MappedVersionDoc>().UseOptimisticConcurrency(true)
+                .Metadata(m => m.Version.MapTo(x => x.Etag));
+        });
+
+    private DocumentStore markerInterfaceStore()
+        => StoreOptions(opts =>
+        {
+            opts.Schema.For<MarkerVersionedDoc>().UseOptimisticConcurrency(true);
+        });
+
+    [Fact]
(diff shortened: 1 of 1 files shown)


jeremydmiller:
@JurJean I honestly wanted to drop this kind of manual mapping functionality in 9.* because it's been a nonstop pain to support. Let me think on it a bit.

jeremydmiller:
Thank you for this — the diagnosis, the reproduction and the tests are all yours, and they made the fix straightforward to get right.

Superseded by https://github.com/JasperFx/marten/pull/5386, which carries your four tests **verbatim** (including the fourth, which deliberately contrasts the marker-interface path that worked all along) plus the fix and four more cases: the `ConcurrencyChecks.Disabled` guard, two for the subclass decorator, and one for a mapped `Revision` member.

On the fix branch you linked — it was the right shape, and two things moved:

1. **Where the seam lives.** `storeEntity` holds Weasel's `IDocumentStorage<T>`, so reaching the mapped member from there needed either a runtime type test on the storage plus forwarding in every decorator, or an interface member. A decorator that forgets to forward fails *silently*, which is the worst failure mode for a concurrency guard, so the seam went upstream: JasperFx/weasel#592 adds `MappedVersionFor` / `MappedRevisionFor` as default interface members, shipped in Weasel 9.32.0. The Marten side is now virtual dispatch with no cast anywhere.
2. **Two gating details.** The revision getter keys off `Metadata.Revision.Enabled` rather than only `UseNumericRevisions` — `Revision.MapTo(...)` sets `Enabled` on its own, so the narrower check misses a mapped revision member configured without `UseNumericRevisions(true)`. And a revision member can be `int` or `long`, so the getter widens explicitly rather than relying on an implicit conversion.

Also worth recording, because it constrains any fix here: the no-version `Store(session, doc)` overload is *how* `ConcurrencyChecks.Disabled` is implemented, so the mapped member must not reach it. There's a test pinning that now.

Closing this in favour of the superseding PR once that merges — your credit is in the commit message and the PR description.

jeremydmiller:
Closing in favour of #5386, which is now merged as `1484f7483`.

Your four tests are in it **verbatim**, and you are a `Co-authored-by:` on the merge commit — the diagnosis, the reproduction and the tests were yours, and they are what made the fix straightforward to get right.

What changed from the fix branch you linked:

- **The seam moved upstream.** `storeEntity` holds Weasel's `IDocumentStorage<T>`, so reaching a mapped member from there needed either a runtime type test on the storage plus forwarding in every decorator, or an interface member. A decorator that forgets to forward fails *silently*, which is the worst failure mode for a concurrency guard — so JasperFx/weasel#592 added `MappedVersionFor` / `MappedRevisionFor` as default interface members, shipped in Weasel 9.32.0. The Marten side is virtual dispatch with no cast anywhere.
- **Two gating details.** The revision getter keys off `Metadata.Revision.Enabled`, not only `UseNumericRevisions` — `Revision.MapTo(...)` sets `Enabled` on its own, so the narrower check missed a mapped revision member configured without `UseNumericRevisions(true)`. And a revision member may be `int` or `long`, so the getter widens explicitly.
- **One constraint worth recording:** the no-version `Store(session, doc)` overload is *how* `ConcurrencyChecks.Disabled` is implemented, so the mapped member must not reach it. There is a test pinning that now.

Thanks again — this was a real hole in a documented feature.

Issue #5376, opened 2026-09-10
EnumStorage.AsString: a query comparing a [JsonStringEnumMemberName] member silently matches nothing


erdtsieck:
## Summary

With `EnumStorage.AsString`, a LINQ filter comparing an enum member that carries
`[JsonStringEnumMemberName]` never matches the documents that member wrote. No exception — the query
returns nothing, which reads as "no such rows".

Not AOT-specific: it reproduces under a JIT just as well. Found on 9.33.0 while fixing #5374.

## Repro

```csharp
public enum Zorgvorm
{
    Huisarts,

    [JsonStringEnumMemberName("apotheek-houdend")]
    Apotheek
}

public class Aanlevering
{
    public Guid Id { get; set; }
    public Zorgvorm Zorgvorm { get; set; }
}

// store with UseSystemTextJsonForSerialization(EnumStorage.AsString)
session.Store(new Aanlevering { Id = Guid.NewGuid(), Zorgvorm = Zorgvorm.Apotheek });
await session.SaveChangesAsync();

// stored:  {"Id": "...", "Zorgvorm": "apotheek-houdend"}

var found = await session.Query<Aanlevering>().Where(x => x.Zorgvorm == Zorgvorm.Apotheek).ToListAsync();
// found.Count == 0
```

## Where

`EnumAsStringMember.CreateComparison` renders the value with `Enum.GetName`:

```csharp
// src/Marten/Linq/Members/EnumAsStringMember.cs
var stringValue = Enum.GetName(MemberType, constant.Value)!;
return new MemberComparisonFilter(this, new CommandParameter(stringValue, NpgsqlDbType.Varchar), op);
```

`Enum.GetName` is the declared name; `JsonStringEnumConverter` writes the renamed one, and the two only
agree when nothing was renamed. The same holds wherever a query renders an enum for itself rather than
asking the serializer — `DuplicatedField` takes the `EnumStorage.AsString` branch too.

## What a fix looks like

The serializer is the only component that knows what it wrote, and it knows AOT-safely: under a
source-generated resolver the renamed name is baked in at compile time (verified — a boxed
`JsonSerializer.Serialize<object>(Zorgvorm.Apotheek, options)` returns `"apotheek-houdend"` from a
trimmed native binary, while reflection over the enum's fields finds no attribute there).

So: ask `ISerializer` for the rendering rather than calling `Enum.GetName`. That needs a small addition
to the interface — a default interface method keeps every existing implementation compiling:

```csharp
[...]

PR #5377, opened 2026-09-10
Native AOT: write the containment payload in Marten instead of through the consumer's serializer

### LINKED ISSUE #5374: Native AOT: a child-collection filter serializes its jsonb payload through the consumer's resolver and throws on object[]
## Summary

Under Native AOT, any filter over a child collection throws before the query is sent:

```text
System.NotSupportedException: JsonTypeInfo metadata for type 'System.Object[]' was not provided by
TypeInfoResolver of type 'MyApp.ConnectionDocumentsJsonContext'. If using source generation, ensure
that all root types passed to the serializer have been annotated with 'JsonSerializableAttribute'...
```

Measured on Marten 9.33.0 from a published native binary against a real PostgreSQL. The query is ordinary:

```csharp
session.Query<ContractAgreementOverview>()
    .Where(ca => ca.Debtors.Any(d => d.Number == debtor.Number && d.Type == debtor.Type))
```

## Why it throws

`ContainmentWhereFilter` builds the jsonb payload as a `Dictionary<string, object>` of member names to the compared values, and serializes it through the consumer's serializer:

```csharp
// src/Marten/Linq/SqlGeneration/Filters/ContainmentWhereFilter.cs
? _serializer.ToCleanJson(_data)
: _serializer.ToCleanJson(new object[] { _data });
```

`SystemTextJsonSerializer.ToCleanJson` is `JsonSerializer.Serialize(document, _clean)`. Under AOT the resolver is a source-generated `JsonSerializerContext` — the shape the AOT guide asks for — and it carries the consumer's *documents*, not Marten's internal payload shapes. `JsonPathExistsFilter` serializes its dictionary the same way.

That payload is Marten's own; a consumer cannot reasonably be expected to declare it.

## Workaround, and why it is not enough

Declaring the shapes on the context does work today:

```csharp
[JsonSerializable(typeof(object[]))]
[JsonSerializable(typeof(Dictionary<string, object>))]
[JsonSerializable(typeof(DebtorType))]   // and every other enum such a filter compares
```

The enum line is the tell: the leaf values are `object`-typed, so STJ resolves each one by its runtime type, and every enum, date and value type a query happens to compare has to be listed. That is a list nobody can keep correct — it fails at runtime, only under AOT, only for the queries you did not think of.

## What a fix looks like

Write the payload in Marten instead of routing it through the consumer's resolver: a small `Utf8JsonWriter` walk over the dictionary/array shape, with the leaves written directly, and a fall-through to `ISerializer` only for a leaf that is a document type (`.Contains(someRecord)` over a collection of records).

Two things it has to keep honouring, which is why it is more than a one-liner:

- **`EnumStorage`** — an enum leaf must be written as a name or a number to match how the document was stored.
- **Dictionary key casing** — the payload's keys are member names and have to match the stored JSON, so whatever `Casing` does to them now has to keep happening.

Happy to follow up with a PR if that direction is the one you want; I have a native harness that reproduces it in one query.

erdtsieck:
Fixes #5374.

A filter over a child collection compiles to a jsonb containment query, and Marten serialized that query's payload through the store's serializer:

```csharp
// ContainmentWhereFilter.Apply
var json = Usage == ContainmentUsage.Singular
    ? _serializer.ToCleanJson(_data)
    : _serializer.ToCleanJson(new object[] { _data });
```

`_data` is Marten's own `Dictionary<string, object>` of member names to compared values. Under Native AOT the serializer's resolver is a source-generated `JsonSerializerContext` — what the AOT guide asks for — and it carries the consumer's documents, not that dictionary, not the `object[]` wrapper, and not whichever enum the filter compares. So an ordinary `.Any(...)` threw before the query was ever sent:

```text
NotSupportedException: JsonTypeInfo metadata for type 'System.Object[]' was not provided by
TypeInfoResolver of type 'MyApp.MyJsonContext'.
```

`JsonbPayload` writes that shape instead — the dictionary, the array wrapper, the scalars whose rendering is fixed — and hands anything with a converter of its own (an enum, a record, a value object) back to the serializer, which is the only thing that knows how the document was written with it and which has the type because it is part of a document. Six call sites move over: `ContainmentWhereFilter`, `JsonPathExistsFilter` and `ChildCollectionJsonPathCountFilter`, each in both its `Apply` and its compiled-query setter.

## Why this is safe to swap under an existing query surface

A payload that differs by one character matches nothing and reports it as "no rows", so the test is byte equality with the old path: `Bug_5374_containment_payload_json` asserts `JsonbPayload.ToJson(serializer, payload) == serializer.ToCleanJson(payload)` across `Casing` × `EnumStorage`, over a payload holding a nested typed dictionary, an `int`-keyed and an enum-keyed one, `DateOnly`, `DateTimeOffset`, decimals, a null and a record.

Two behaviours came out of writing those assertions rather than out of reading the code, and both are in the writer now:

- A nested dictionary is `Dictionary<TKey, TValue>`, not `Dictionary<string, object>` — that is what `DictionaryMember.PlaceValueInDictionaryForContainment` puts there. An earlier cut matched only the latter, so a `Dictionary<string, string>` fell into the array branch and came out as a list of pairs. `LinqTests` caught it (`x => x.StringDict.Contains(kvp)`), which is why the `IDictionary` branch is now the general one.
- An **enum dictionary key** is written as its member name whatever `EnumStorage` says, because that setting is about values and a key has its own converter. `TimeOnly` went back to the serializer for the mirror-image reason: its converter trims the fractional part.

Against Newtonsoft the two differ by escaping alone — System.Text.Json's encoder escapes an ampersand where Newtonsoft leaves it — and that cannot change a result: Postgres parses the parameter into jsonb, where the two are one character. The test states it that way, with `JsonNode.DeepEquals`.

## Coverage

`Marten.AotRuntimeSmoke` gains the query from the issue plus one over an enum. On master both fail with the `object[]` stack; here:

```text
OK   child collection filter
OK   child collection filter over an enum
Marten AOT runtime smoke OK — every document and event read path ran from a native binary.
```

`LinqTests` is green on net10.0 (1481 passed), as is the new `CoreTests` class.

## Not in scope

While testing the enum branch I found something that is not an AOT problem at all: with `EnumStorage.AsString`, a query comparing an enum member renamed with `[JsonStringEnumMemberName]` silently matches nothing, under a JIT as much as under AOT, because `EnumAsStringMember.CreateComparison` renders the value with `Enum.GetName`. Filed separately as #5376 — the fix wants a small `ISerializer` addition, and that interface is yours to decide on rather than mine.

### DIFF
--- a/docs/configuration/aot-publishing.md
+++ b/docs/configuration/aot-publishing.md
@@ -138,7 +138,7 @@ As of Marten 9.0.0-alpha:
 ### Works (AOT-clean)
 
 - **`UseSystemTextJsonForSerialization`** with the default `JsonSerializerOptions` or a user-supplied one. For best AOT results, pass a source-generated `JsonSerializerContext`.
-- **Document storage** — `Schema.For<TDoc>()` and the entire CRUD / LINQ surface for closed-shape document types (Guid / string / int / long / strong-typed Id strategies), including the containment and JSONPath filters a child-collection query compiles to (#5376).
+- **Document storage** — `Schema.For<TDoc>()` and the entire CRUD / LINQ surface for closed-shape document types (Guid / string / int / long / strong-typed Id strategies), including the containment and JSONPath filters a child-collection query compiles to (#5377).
 - **Event storage** — `StartStream`, `Append`, `FetchStream`, `FetchStreamStateAsync`, `AggregateStreamAsync`, `QueryAllRawEvents`, the async daemon. Reading an event needed a JIT until [#5373](https://github.com/JasperFx/marten/issues/5373): the events table closed its per-column reader over the column's member type at runtime.
 - **Projections** — registered either as a projection type (`Projections.Add<T>(...)`) or, for a self-aggregating type, through the identity-typed `LiveStreamAggregation<TDoc, TId>()` / `Snapshot<TDoc, TId>(...)` overloads: `SingleStreamProjection<TDoc, TId>`, `MultiStreamProjection<TDoc, TId>`, `EventProjection`, `CustomProjection`, and `EventApplier` — the JasperFx.Events source generator emits `[GeneratedEvolver]` dispatchers at compile time for each registration. Marten calls `Options.Projections.DiscoverGeneratedEvolvers(...)` at startup (`src/Marten/DocumentStore.cs:84`) to pick them up.
 - **Compiled queries** registered through `Marten.SourceGenerator` in an assembly marked `[JasperFxAssembly]`.
@@ -261,7 +261,7 @@ NotSupportedException: JsonTypeInfo metadata for type 'System.Object[]' was not 
 TypeInfoResolver of type 'MyApp.MyJsonContext'.
 ```
 
-Fixed by [#5376](https://github.com/JasperFx/marten/pull/5376) — upgrade rather than working around it.
+Fixed by [#5377](https://github.com/JasperFx/marten/pull/5377) — upgrade rather than working around it.
 A filter over a child collection (`x => x.Debtors.Any(d => d.Number == number)`) becomes a jsonb
 containment query, and Marten used to serialize that payload — its own dictionary of member names to the
 values you compared against — through *your* serializer. A source-generated resolver carries your
--- a/docs/configuration/aot-publishing.md
+++ b/docs/configuration/aot-publishing.md
@@ -138,7 +138,7 @@ As of Marten 9.0.0-alpha:
 ### Works (AOT-clean)
 
 - **`UseSystemTextJsonForSerialization`** with the default `JsonSerializerOptions` or a user-supplied one. For best AOT results, pass a source-generated `JsonSerializerContext`.
-- **Document storage** — `Schema.For<TDoc>()` and the entire CRUD / LINQ surface for closed-shape document types (Guid / string / int / long / strong-typed Id strategies).
+- **Document storage** — `Schema.For<TDoc>()` and the entire CRUD / LINQ surface for closed-shape document types (Guid / string / int / long / strong-typed Id strategies), including the containment and JSONPath filters a child-collection query compiles to (#5376).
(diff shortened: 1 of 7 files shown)


jeremydmiller:
Merging this. Thank you — the diagnosis is right, and the test design is the part I want to call out:
asserting byte-equality against the old path across `Casing` × `EnumStorage`, on the grounds that a
payload differing by one character matches nothing and reports it as "no rows", is exactly how this
change had to be verified. The enum-dictionary-key and `TimeOnly` findings falling out of those
assertions rather than out of reading the code makes the point.

Three follow-ups came out of review, tracked in #5385 rather than held against this PR:

1. **Newtonsoft + non-default casing and nested dictionary keys.** `JsonNetContractResolver` sets
   `ProcessDictionaryKeys = true` for CamelCase and SnakeCase, so the old path transformed dictionary
   keys and `JsonbPayload.key()` writes them verbatim. Top-level keys are safe (they are
   `ToJsonKey(casing)` already, so re-casing is idempotent), but a nested user-supplied key copied in
   by `DictionaryMember.PlaceValueInDictionaryForContainment` is not. STJ is unaffected — Marten never
   sets `DictionaryKeyPolicy`. Nothing in CI combines Newtonsoft + non-default casing + dictionary
   containment, which is why the matrix is green. Reasoned from the resolver config, not executed —
   one extra `[InlineData]` on the Newtonsoft test settles it. Likely fix: let a nested *typed*
   dictionary go back through the serializer, since it is a document member type the consumer's
   resolver already carries; only Marten's own `Dictionary<string, object>` and the `object[]` wrapper
   have to be written here.
2. **Documenting what the writer deliberately does not honour** — `NumberHandling.WriteAsString`, a
   custom `decimal`/`DateTimeOffset` converter, a consumer-set `DictionaryKeyPolicy`. That is inherent
   to the fix rather than a defect, but the remarks currently imply only a custom `JavaScriptEncoder`
   differs.
3. **`byte[]` falls into the `IEnumerable` branch** (array of numbers vs base64). Theoretical today —
   no test documents a `byte[]` member — but a one-line guard.

Reviewed with Claude Code.

PR #5378, opened 2026-09-10
EnumStorage.AsString: ask the serializer for the member name, so a renamed enum member can be queried

### LINKED ISSUE #5376: EnumStorage.AsString: a query comparing a [JsonStringEnumMemberName] member silently matches nothing

erdtsieck:
Fixes #5376.

With `EnumStorage.AsString`, a query comparing an enum member renamed with `[JsonStringEnumMemberName]` (or `[EnumMember]` on Newtonsoft) matched nothing. No exception — zero rows, which reads as "no such data":

```csharp
// stored: {"Id": "...", "Zorgvorm": "apotheek-houdend"}
await session.Query<Aanlevering>().Where(x => x.Zorgvorm == Zorgvorm.Apotheek).ToListAsync();
// 0 rows, because the filter went looking for "Apotheek"
```

`EnumAsStringMember.CreateComparison` rendered the value with `Enum.GetName`, and that is the declared name. The serializer is the only component that knows what it actually wrote, so it is now the one asked, through a small cache (`EnumMemberNames`) so it is asked once per member.

Two things worth flagging, because neither is visible from reading the code:

- **The value arrives as the underlying integer as often as the enum.** The C# compiler emits `Convert(x.Member, Int32) == 1` for a comparison against a literal, and `Enum.GetName` happily takes that int. My first cut handed the serializer the boxed `1` and produced `"1"`, which matched nothing at all — the existing `enum_usage` tests caught it. Hence the `Enum.ToObject` before asking.
- **Reflection over the enum's fields is the wrong tool**, though it looks like the obvious one. `field.GetCustomAttribute<JsonStringEnumMemberNameAttribute>()` agrees with the serializer under a JIT and returns *nothing* from a trimmed Native AOT binary, so the query would silently go back to the declared name in exactly the deployment where it is hardest to see. Measured, not assumed: a boxed `JsonSerializer.Serialize<object>(renamed, options)` returns `"apotheek-houdend"` from a native binary while the attribute lookup finds no attribute there.

`Marten.AotRuntimeSmoke` gains the query for that reason. It fails there without this change ("ran, but returned the wrong result") and passes with it.

## Deliberately not in scope: the duplicated column

A duplicated column is written by the document storage, not by the serializer, and it holds the **declared** name. So a query against a duplicated field has to keep comparing against the declared name — changing `DuplicatedField` to match the body would have broken the column, which is why the second test in this PR pins today's behaviour rather than "fixing" it.

That the column and the JSON body disagree for a renamed member is a real inconsistency, and a wider change than this one: it belongs where the column is written. Happy to follow up if you want it, but it is a storage change and not a LINQ one.

## Tests

`LinqTests` green on net10.0 (1481 passed), including the `enum_usage` set that caught the first cut, plus the two new `CoreTests` cases and the native smoke.

### DIFF
--- a/src/CoreTests/Bug_5376_renamed_enum_member.cs
+++ b/src/CoreTests/Bug_5376_renamed_enum_member.cs
@@ -1,0 +1,73 @@
+using System;
+using System.Linq;
+using System.Text.Json.Serialization;
+using System.Threading.Tasks;
+using Marten;
+using Marten.Testing.Harness;
+using Shouldly;
+using Weasel.Core;
+using Xunit;
+
+namespace CoreTests;
+
+// #5376: with EnumStorage.AsString a query rendered the enum with Enum.GetName, which is the declared
+// name. System.Text.Json stores the name from [JsonStringEnumMemberName], so a filter comparing a
+// renamed member matched nothing at all — and said so as "no rows" rather than as an error.
+public class Bug_5376_renamed_enum_member: OneOffConfigurationsContext
+{
+    public enum Zorgvorm
+    {
+        Huisarts,
+
+        [JsonStringEnumMemberName("apotheek-houdend")]
+        Apotheek
+    }
+
+    public class Aanlevering
+    {
+        public Guid Id { get; set; }
+        public Zorgvorm Zorgvorm { get; set; }
+    }
+
+    [Fact]
+    public async Task can_be_found_by_the_member_it_was_stored_as()
+    {
+        StoreOptions(opts =>
+        {
+            opts.UseSystemTextJsonForSerialization(EnumStorage.AsString);
+            opts.Schema.For<Aanlevering>();
+        });
+
+        var id = Guid.NewGuid();
+        theSession.Store(new Aanlevering { Id = id, Zorgvorm = Zorgvorm.Apotheek });
+        await theSession.SaveChangesAsync();
+
+        (await theSession.Json.FindByIdAsync<Aanlevering>(id)).ShouldContain("apotheek-houdend");
+
+        var found = await theSession.Query<Aanlevering>().Where(x => x.Zorgvorm == Zorgvorm.Apotheek).ToListAsync();
+
+        found.ShouldHaveSingleItem().Id.ShouldBe(id);
+    }
+
+    // A duplicated column is written by the document storage rather than by the serializer, so it holds
+    // the declared name - and a query against it has to keep comparing against that, or the column and
+    // the query stop agreeing. The body and the column disagreeing for a renamed member is a wider
+    // change than this one, and it belongs with whatever writes the column.
+    [Fact]
+    public async Task a_duplicated_column_keeps_comparing_against_the_declared_name()
+    {
+        StoreOptions(opts =>
+        {
+            opts.UseSystemTextJsonForSerialization(EnumStorage.AsString);
+            opts.Schema.For<Aanlevering>().Duplicate(x => x.Zorgvorm);
+        });
+
+        var id = Guid.NewGuid();
+        theSession.Store(new Aanlevering { Id = id, Zorgvorm = Zorgvorm.Apotheek });
+        await theSession.SaveChangesAsync();
+
+        var found = await theSession.Query<Aanlevering>().Where(x => x.Zorgvorm == Zorgvorm.Apotheek).ToListAsync();
+
+        found.ShouldHaveSingleItem().Id.ShouldBe(id);
+    }
+}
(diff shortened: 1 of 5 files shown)


jeremydmiller:
Closing in favour of https://github.com/JasperFx/marten/pull/5389, which carries your two tests unchanged and credits you as `Co-authored-by:` on the commit.

Your diagnosis was right and the two findings you flagged in the description are the reason this was straightforward to extend — particularly the AOT one. That `GetCustomAttribute<JsonStringEnumMemberNameAttribute>()` agrees with the serializer under a JIT and returns *nothing* from a trimmed binary is exactly the kind of thing that would have been found much later, and by someone with far less context. The smoke harness makes it permanent.

What the superseding PR adds:

**The collection shapes.** `==` / `!=` is `EnumAsStringMember`, but every `in` / `Contains` shape renders through Weasel's `EnumIsOneOf` / `EnumIsNotOneOf` fragments, which had the identical bug as `stringEntry.ToString()`. Five call sites reach them (`IsOneOf`, `IsNotOneOf`, both `EnumerableContains` branches, `MemoryExtensionsContains`). Fixing only equality would have left `x.Zorgvorm == Apotheek` finding the row while `x.Zorgvorm.In(Apotheek)` silently did not — worse than uniform breakage, because nothing surfaces the inconsistency. Those fragments are Weasel's, so that half went upstream first as JasperFx/weasel#592 and shipped in Weasel 9.32.0.

**Three fixes to the mechanism:**

1. `EnumMemberNames` was a `static ConcurrentDictionary<(Type, string), string>` with the serializer passed as a parameter, so the first store to ask won process-wide. Two stores can legitimately spell the same member differently — `AddMartenStore<T>()`, or a `Configure(...)` that adds a `JsonStringEnumConverter(JsonNamingPolicy.CamelCase)` — and the loser silently queried for the other's spelling. Now keyed per serializer.
2. `Trim('"')` removes the quotes but not the escaping. STJ's default encoder escapes `&`, `<`, `>`, `'`, `+`, so `[JsonStringEnumMemberName("a&b")]` compared the literal `a&b` against the `a&b` that `->>` returns — the same silent-no-rows bug, for a narrower input. Decoded via `Utf8JsonReader` now, which also stays AOT-safe.
3. Adding an optional ctor parameter changes the signature and is binary breaking, so the serializer arrives through an overload instead.

**Your duplicated-column scoping is kept as-is** — the test that pins today's behaviour is still there. The underlying body-vs-column inconsistency is filed as #5388, explicitly low priority.

Thanks for both this and #5377 — the byte-equality test design on that one set the bar for how this was verified.

PR #5390, opened 2026-09-12
Containment payload writer: nested dictionary keys, byte[], and its documented limits

### LINKED ISSUE #5385: Follow-ups from #5377: containment payload writer — Newtonsoft dictionary-key casing, documented caveats, byte[] branch
Follow-ups from the review of #5377 (which fixed #5374 — Marten now writes the jsonb containment /
JSONPath payload itself instead of routing it through the consumer's serializer). #5377 is merged;
these are the loose ends, in priority order.

## 1. Newtonsoft + non-default casing writes nested dictionary keys verbatim

`JsonNetContractResolver` sets `ProcessDictionaryKeys = true` for both `Casing.CamelCase` and
`Casing.SnakeCase` (`src/Marten.Newtonsoft/Services/Json/JsonNetContractResolver.cs:73,77`). The old
`ToCleanJson` payload therefore *transformed* dictionary keys; `JsonbPayload.key()` writes them
verbatim.

Top-level payload keys are unaffected — they are `member.ToJsonKey(casing)` already, so re-casing
`agbCode` / `agb_code` is idempotent. The exposure is a **nested user-supplied dictionary key**,
which `DictionaryMember.PlaceValueInDictionaryForContainment` copies straight from the caller
(`childDict[pair.Key] = pair.Value`):

```csharp
// Newtonsoft + Casing.CamelCase
// stored:  {"kenmerken": {"kleur": "blauw"}}
x => x.Kenmerken.Contains(new KeyValuePair<string, string>("Kleur", "blauw"))
// payload now asks for "Kleur" -> zero rows, no error
```

System.Text.Json is not affected: Marten never sets `DictionaryKeyPolicy`, so STJ wrote keys verbatim
before and after.

**Why CI is green:** `LinqTests/Acceptance/dictionary_usage.cs` runs at default casing, and the only
casing-parameterised child-collection test (`child_collection_queries.cs:89`) uses
`UseSystemTextJsonForSerialization`. Nothing combines Newtonsoft + non-default casing + dictionary
containment.

**Not yet executed** — this is reasoned from the resolver configuration. Adding one `[InlineData]`
with a PascalCase nested key to `Bug_5374_containment_payload_json.writes_what_newtonsoft_wrote_apart_from_escaping`
(parameterised over `Casing`) settles it either way, and should be the first step.

**Suggested fix:** let a *nested typed* dictionary (`Dictionary<TKey, TValue>`) go back through the
serializer rather than through `JsonbPayload`'s `IDictionary` branch. It is a document member type,
so a source-generated resolver already carries it — the AOT problem was only ever Marten's own
`Dictionary<string, object>` and the `object[]` wrapper. That keeps the #5374 fix intact and hands
key rendering back to the component that knows the rules.

## 2. Document what the writer deliberately does not honour

Writing the primitives in Marten necessarily ignores consumer-configured rendering *of those
primitives*: `JsonSerializerOptions.NumberHandling = WriteAsString`, a custom `decimal` /
`DateTimeOffset` converter, a `DictionaryKeyPolicy` set through `Configure(...)`. This is inherent to
the fix — routing leaves back through the serializer is exactly what fails under Native AOT — so it
is a caveat to state, not a defect to repair.

The `JsonbPayload` remarks currently imply that only a custom `JavaScriptEncoder` differs, which
understates it. Worth a sentence there and in `docs/configuration/aot-publishing.md`.

## 3. `byte[]` falls into the array branch

`case IEnumerable` catches `byte[]` and writes an array of numbers where both serializers write a
base64 string. No test documents a `byte[]` member today and the LINQ layer has no byte-array member
handling, so this is theoretical — but it is a one-line `case byte[]` guard, and the same branch
would swallow any value object that happens to implement `IEnumerable`.

Thanks again to @erdtsieck, whose byte-equality-against-the-old-path test design is what makes these
easy to state precisely.

jeremydmiller:
Closes #5385 — the three follow-ups from reviewing #5377, now that the containment payload writer is
on master.

## 1. Newtonsoft + non-default casing silently broke nested dictionary keys

This is the one with teeth, and it was a live bug rather than a tidy-up.

`JsonNetContractResolver` sets `ProcessDictionaryKeys = true` for `Casing.CamelCase` and
`Casing.SnakeCase`, so Newtonsoft cases **dictionary keys**, not just property names. `JsonbPayload`
wrote every key verbatim. The payload's own top-level keys survive that unchanged because they are
already `member.ToJsonKey(casing)` — re-casing them is idempotent — which is exactly why the existing
Newtonsoft test passed and this went unnoticed. A *nested* dictionary is different: its keys come
straight from the caller, since `DictionaryMember.PlaceValueInDictionaryForContainment` copies the
`KeyValuePair` the query supplied.

So under Newtonsoft + CamelCase, a query like
`x => x.Kenmerken.Contains(new KeyValuePair<string,string>("Kleur","blauw"))` asked for `"Kleur"`
against a document stored with `"kleur"`. Zero rows, no error.

**The fix splits the writer by ownership.** Two shapes are Marten's and stay in the writer, because a
consumer's source-generated resolver has no reason to know them: the `Dictionary<string, object>`
payload, and the `object[]` a child-collection filter wraps it in. Everything else came out of the
consumer's document — including a typed `Dictionary<TKey, TValue>` — and goes back through the
serializer, which both knows how it was written and has the type. That keeps #5374 fixed while
handing key rendering back to the component that owns the rules.

System.Text.Json was never affected: Marten does not set `DictionaryKeyPolicy`.

## 2. `byte[]` came out as a list of numbers

`case IEnumerable` caught byte arrays, so a `byte[]` filter value rendered as `[1,2,3,250]` where both
serializers write a base64 string. It now has its own branch ahead of the enumerable case. Theoretical
in the sense that nothing in the test suite compares a `byte[]` member, but it is one line.

## 3. Saying what the writer deliberately does not honour

The class remarks claimed a custom `JavaScriptEncoder` was the only thing not carried across, and that
dictionary keys are written verbatim because `DictionaryKeyPolicy` is never set. The first understated
it and the second was only true for System.Text.Json — it is precisely what made (1) invisible. Both
the remarks and the AOT guide now state the real trade: the scalars written here bypass your
`JsonSerializerOptions`, so a `NumberHandling` setting or a custom converter for `decimal` /
`DateTimeOffset` / `Guid` affects how a document was stored but not how a filter value is rendered.
That is the price of not needing your resolver for Marten's own shapes, and it is worth writing down
rather than discovering.

## Tests

- `Bug_5374_containment_payload_json` — **17/17 green**. The two new
  `writes_what_newtonsoft_wrote_for_a_nested_dictionary_key` cases **fail on master** for CamelCase and
  SnakeCase and pass for Default, which is the bug pinned exactly; the System.Text.Json twin passes
  throughout, pinning that it was never affected. Two `byte[]` cases added, one per serializer.
- The 13 pre-existing byte-equality assertions still hold, which is the assertion that matters most
  here: `payload()` carries three typed dictionaries (`Dictionary<string,string>`,
  `Dictionary<int,string>`, `Dictionary<Soort,decimal>`) that now route through the serializer instead
  of the writer, and they still match `ToCleanJson` exactly.
- `LinqTests` dictionary / child-collection / containment / casing — **264/264 green**. That is the
  real regression surface: this changes how a nested dictionary renders for every dictionary-member
  query, not only the renamed-casing ones.

## Credit

The writer, and the byte-equality-against-the-old-path test design that made all three of these easy
to state and prove, are **@erdtsieck's** from #5377.

### DIFF
--- a/docs/configuration/aot-publishing.md
+++ b/docs/configuration/aot-publishing.md
@@ -267,6 +267,20 @@ containment query, and Marten used to serialize that payload — its own diction
 values you compared against — through *your* serializer. A source-generated resolver carries your
 documents, not that, so the query threw before it was ever sent. Marten now writes the payload itself.
 
+The split is by ownership: Marten writes the shapes that are its own (the payload dictionary, and the
+array a child-collection filter wraps it in) plus the scalar leaves, and hands everything that came out
+of your document — a typed nested dictionary, an enum, a record, a value object — back to your
+serializer, which has those types because they are part of a document.
+
+One consequence worth knowing, because it is the price of not needing your resolver for Marten's own
+shapes: the scalars written here do **not** go through your `JsonSerializerOptions`. A
+`NumberHandling` setting, or a custom converter for one of the scalar types (`decimal`,
+`DateTimeOffset`, `Guid`, …), will not affect how a *filter value* of that type is rendered, even
+though it affects how the document was stored. If you rely on one of those, compare through a member
+whose type Marten hands to the serializer rather than one it renders itself. A custom
+`JavaScriptEncoder` is not a concern: Postgres parses the payload into jsonb, where an escaped and an
+unescaped character are the same character.
+
 ### Secondary store throws `MissingMethodException` at boot
 
 `AddMartenStore<TInterface>` uses `System.Reflection.Emit` to materialize the implementation type at runtime (Marten 9 ripped out the Roslyn-emit subclass path in PR [#4459](https://github.com/JasperFx/marten/pull/4459)). Native AOT supports `Reflection.Emit` only on full-AOT runtimes that include the JIT; some trimming configurations strip it. If a secondary store throws `MissingMethodException` at boot, check:
--- a/src/CoreTests/Bug_5374_containment_payload_json.cs
+++ b/src/CoreTests/Bug_5374_containment_payload_json.cs
@@ -112,6 +112,72 @@ public class Bug_5374_containment_payload_json
         Apotheek
     }
 
+    // #5385: Newtonsoft's contract resolver sets ProcessDictionaryKeys = true for CamelCase and
+    // SnakeCase, so the old ToCleanJson path transformed *dictionary keys* as well as property names.
+    // The payload's own top-level keys survive that unchanged because they are already
+    // member.ToJsonKey(casing) — re-casing them is idempotent — which is exactly why the existing
+    // Newtonsoft test above passes and this gap went unnoticed. A nested dictionary is different: its
+    // keys come from the caller (DictionaryMember.PlaceValueInDictionaryForContainment copies the
+    // KeyValuePair the query supplied), so the serializer would case them and JsonbPayload would not.
+    //
+    // The top-level key here is deliberately already-lowercase so the only thing under test is the
+    // nested one. System.Text.Json is unaffected: Marten never sets DictionaryKeyPolicy.
+    [Theory]
+    [InlineData(Casing.Default)]
+    [InlineData(Casing.CamelCase)]
+    [InlineData(Casing.SnakeCase)]
+    public void writes_what_newtonsoft_wrote_for_a_nested_dictionary_key(Casing casing)
+    {
+        var serializer = new Marten.Services.JsonNetSerializer { Casing = casing };
+        var data = new Dictionary<string, object>
+        {
+            ["kenmerken"] = new Dictionary<string, string> { ["Kleur"] = "blauw" }
+        };
+
+        JsonNode.DeepEquals(
+                JsonNode.Parse(JsonbPayload.ToJson(serializer, data)),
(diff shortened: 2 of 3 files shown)


jeremydmiller:
Retroactive attribution: @erdtsieck is now recorded as `Co-authored-by:` for this work, in commit 64a4ecc91 on master.

It could not go onto this PR's own merge commit (`8c055a0b0`) — a merged commit's trailers cannot be changed without rewriting published history, and three commits had already landed on top of it. GitHub reads `Co-authored-by:` from any commit, so an additive one carries the credit without a force-push.

To be explicit about what is being credited: the `JsonbPayload` writer this PR amended, and the byte-equality-against-the-old-path test design, are both yours from #5377. That design is what made all three follow-ups here easy to state precisely and easy to prove — the Newtonsoft dictionary-key bug in particular went from "I think this is wrong" to a failing test with two `[InlineData]` rows.

PR #5403, opened 2026-09-13
Make partitioned stream archiving idempotent

rmasciarella:
With `UseArchivedStreamPartitioning = true`, a second archive of the same stream raises PostgreSQL `23505` on `mt_streams_archived_pkey`. This also stops an async single-stream projection with `IncludeArchivedEvents = true` when it processes an `Archived` marker after the inline snapshot has archived the stream.

Both `INSERT ... SELECT` statements now exclude archived rows. The function first locks the active stream row with `FOR UPDATE`. The lock matters because two concurrent transactions can still select the same active row if the fix only adds predicates.

This keeps unique violations for a different active stream that reuses an archived stream ID. Suppressing those conflicts with `ON CONFLICT DO NOTHING` would let the later deletes remove the active rows without preserving their stream metadata. A regression test checks that both streams and their events survive the rejected archive.

The tests cover GUID and string IDs, conjoined tenant isolation, overlapping archive transactions, and async projections after inline archival with both Rich and Quick appends. The concurrency tests observe `pg_blocking_pids` before releasing the first transaction. The documentation explains repeated archiving and projections that include archived events.

## Validation

PostgreSQL 17, .NET 10, no test retries:

- On unmodified master `34c72d1`, the new regression class has 10 failures and 1 pass. The passing case checks that a reused stream ID still reports a collision.
- With the fix, the archive, partitioning, and strict stream-identity tests pass 97 cases with each of System.Text.Json and Newtonsoft.Json, including all 11 new cases. Each run has zero failures and skips the same 5 existing compliance cases for unsupported `UnArchiveStream`.
- Markdownlint passes. Cspell reports no issues across all 143 documentation files.

I am opening this as a draft so maintainers can review the locking approach before it is marked ready. CONTRIBUTING.md asks for prior agreement on the approach. The full CI matrix has not been run locally.


jeremydmiller:
Reviewed this with a specific worry about the performance of the SQL changes. **The conclusion went the other way: these statements should be faster than what they replace, not slower.** Writing out the reasoning because it is not the intuitive answer.

## The added predicates buy performance rather than costing it

`is_archived` is the **LIST partition key** on both tables — `StreamsTable.cs:78` and `EventsTable.cs:198` both do `archiving.PartitionByListValues().AddPartition("archived", true)`. The old statements were:

```sql
select ... from mt_streams where id = streamid
select ... from mt_events  where stream_id = streamid
```

With no predicate on the partition key, the planner has to consider **both** partitions. Adding `and is_archived = FALSE` lets it prune to the active one — and on a store that has been archiving for a while, the archived partition is the large one. So the new predicates are a win on the read side, independent of the correctness they are there for.

## The real cost is lock duration, and it is bounded

`perform 1 ... for update` moves the stream-row lock from the trailing `DELETE` up to the top, so it is now held across both `INSERT ... SELECT` statements. That is a genuine change: appends contend for exactly that row — `update mt_streams set version = ...` at `PostgresEventStoreDialect.cs:287`, and the same in `QuickAppendEventFunction.cs:261`.

What keeps it acceptable is that the event copy is index-backed: `pk_mt_events_stream_and_version` is unique on `(tenant_id, stream_id, version, is_archived)`. So the window scales with **the archived stream's own event count**, not with table size, and contention is confined to that one stream — concurrent archives of different streams never meet. For a pathologically large stream it is a longer block than before, but that is the price of the correctness being bought here, and it is the right trade.

## What the review confirmed

- **Leaving `writeSimple` alone is correct, not a half-fix.** `UPDATE ... SET is_archived = TRUE` is idempotent by construction; only the INSERT-then-DELETE partitioned path can collide on `mt_streams_archived_pkey`.
- **The retry semantics are sound under READ COMMITTED.** The blocked second call re-evaluates `is_archived = FALSE` after the first commits, finds the row gone, and no-ops. Re-archiving a stream that is not there takes no lock and does nothing.
- **Preserving the collision for a reused stream id is the right call**, and the fact that there is a test pinning it is the part that convinced me. `ON CONFLICT DO NOTHING` would have let the deletes drop live rows.
- **The concurrency test cannot hang or pass vacuously** — bounded by a linked CTS with `CancelAfter(15.Seconds())`, and it asserts the second archive is still incomplete inside the `pg_blocking_pids` loop rather than just waiting.

## One thing before this goes in: please rebase

CI is green, but the base here is `34c72d1` and master is now `34bd13dcd` — four commits ahead (#5404, #5407, #5410, #5411). The validation numbers in the description ("on unmodified master `34c72d1`, 10 failures and 1 pass") were measured against that older base, and #5410 in particular touched store identity and `SecondaryStoreConfig`. Green on a stale base is not green on what would actually ship, so a rebase and a re-run is worth it before marking this ready.

On the process point in the description: the locking approach is agreed — `FOR UPDATE` on the active row is the right mechanism here, and filtering alone would indeed not have been enough.

jeremydmiller:
@rmasciarella I think this will work. We can catch it in the next release after today, and I'm sure you know that our release cadence is absurdly frequent.

rmasciarella:
@jeremydmiller yes no worries looks like it was faster then i could even notice it wasn't done yet lol, thanks!

Issue #5406, opened 2026-09-13
Codegen output is deterministic as of JasperFx 2.69.2 — regenerate the committed output


jeremydmiller:
Downstream of **JasperFx/jasperfx#832**, fixed in **JasperFx 2.69.2**.

## What changed upstream

`MethodFrameArranger.compileFrames` step 3 enumerated `DependencyGatherer.Dependencies`, an `ImHashMap` keyed by `Frame`. `Frame` does not override `GetHashCode`, so that walk followed **identity hash codes** — arbitrary per object, per process — and two arrangements of the same logical method emitted the same statements in a different order. Step 4's topological sort preserves the relative order of independent frames, so nothing downstream restored it.

jasperfx#196 fixed exactly this for `DependencyGatherer.Variables`, which is why constructor parameters and field declarations in Marten's generated providers have been stable while the statements inside the methods were not.

## What Marten should do

**1. Bump to JasperFx 2.69.2.**

**2. Regenerate the committed generated code — expect a one-time diff.** Marten tracks **14 generated `.cs` files** under `src/**/Internal/Generated/DocumentStorage` (`DocumentDbTests`, `ValueTypeTests` — `RevisionedDocProvider*`, `ClassProvider*`, `DuplicateValueTypeDocProvider*`). Statements inside generated methods may reorder once as they settle onto the deterministic order. That is the fix landing, not a semantic change, and it is the last such diff.

Worth its own commit so the reordering is legible rather than buried in an unrelated change.

**3. The pre-generated-code story gets better.** Marten's `Optimized Artifact Workflow` / pre-generated `Marten.Generated` guidance asks people to commit generated document providers and ship them. Until now, regenerating on a different machine could produce a diff with no semantic content, which makes that workflow harder to trust and harder to review. It is now byte-stable across runs and machines — worth a line in the docs where pre-generation is described, since "your diff is noise, ignore it" was previously the honest answer and no longer is.

## Acceptance

- [ ] `JasperFx` pinned to 2.69.2.
- [ ] Committed generated code regenerated, in a commit of its own.
- [ ] Regenerating twice in a row produces **byte-identical** output.
- [ ] Pre-generation docs mention that output is deterministic, if they currently hedge.

Refs JasperFx/jasperfx#832, JasperFx/jasperfx#833, JasperFx/jasperfx#196.

jeremydmiller:
This isn't really necessary

Issue #5405, opened 2026-09-13
The store-derived Event Model source hardcodes the default model name, so a host that names its model gets two


jeremydmiller:
`AddMarten` registers a `JasperFx.Events.EventModeling.ProjectionEventModelSource` — the store-derived Event Model rung from jasperfx#825 — and leaves its `ModelName` at the default, `ProjectionEventModelSource.DefaultModelName` (`"EventModel"`). `EventModelDiscovery.Assemble` groups descriptors by model name before merging slices, so a host that calls `AddEventModel("Something", …)` ends up with **two** assembled models: its own, and one called `EventModel`. Neither canvas carries both halves — the host's declarations land on one and the store's derived View slices on the other. The user-visible symptom is an assembly that reports more than one model where the host declared one; the reported instance (on the sibling store, see below) was a spec suite failing with `Expected exactly one assembled model, but got [Stoat, EventModel] — assemble first, and check every source names the same model.` Nothing in that message points at the store registration, which is what makes it hard to trace.

JasperFx says so on the property itself, in `ProjectionEventModelSource`:

```csharp
/// <summary>
/// Name of the model these slices contribute to. Must match what the other sources call it, since
/// <see cref="EventModelDiscovery.Assemble" /> groups descriptors by name before merging slices.
/// </summary>
public string ModelName { get; init; } = DefaultModelName;
```

## Affected call sites

Both are in `src/Marten/MartenServiceCollectionExtensions.cs` on the current default branch, and neither passes a model name:

- line 227, in `AddMarten(this IServiceCollection services, Func<IServiceProvider, StoreOptions> optionSource)` — the overload the other four `AddMarten` spellings funnel into:

  ```csharp
  services.AddProjectionEventModelSource(s => [(IEventStore)s.GetRequiredService<IDocumentStore>()]);
  ```

- line 374, in `AddMartenStore<T>(this IServiceCollection services, Func<IServiceProvider, StoreOptions> configure)` — the overload `AddMartenStore<T>(IServiceCollection, Action<StoreOptions>)` funnels into:

  ```csharp
  services.AddProjectionEventModelSource(s => [(IEventStore)s.GetRequiredService<T>()]);
  ```

Worth noting in passing: the comment above the ancillary registration says "Two sources emitting a slice for the same document type is harmless -- slices merge by name, which is what naming them after the document is for." That is true, and it is silently conditional on the two sources naming the same *model* — which is exactly the condition that breaks here.

## The seam already exists upstream

[...]

PR #5411, opened 2026-09-13
Adopt JasperFx 2.69.3 and default the Event Model name to the service name

### LINKED ISSUE #5408: Default the Event Model name to the service name rather than the literal "EventModel"
The store-derived Event Model source falls back to the literal `"EventModel"` when `StoreOptions.EventModelName` is null:

```csharp
var modelName = (store as DocumentStore)?.Options.EventModelName
    ?? ProjectionEventModelSource.DefaultModelName;
// src/Marten/Events/EventModeling/MartenProjectionEventModelSource.cs:59-60
```

`ProjectionEventModelSource.DefaultModelName` is `"EventModel"` (`jasperfx/src/JasperFx.Events/EventModeling/ProjectionEventModelSource.cs:74`). **That is the one default guaranteed to be wrong for every host**, and it is the sole cause of the bug this all started with (marten#5405).

## Every other source defaults to something meaningful

`EventModelDiscovery.Assemble` groups descriptors by name, so the defaults decide whether a host gets one model or several:

| Contributor | Default model name |
|---|---|
| Wolverine chains and HTTP | the service name |
| Bobcat spec assembly | the spec assembly name, overridable by `[assembly: EventModelName]` |
| Bobcat curated file | the file's `model:` value |
| Overlay definition | the defining type's name |
| **The three stores** | **the literal `EventModel`** |

Wolverine's is `wolverine/src/Wolverine/Configuration/EventModeling/WolverineEventModelSource.cs:107`:

```csharp
var model = new EventModelDescriptor(options.ServiceName, slices) { Aggregates = aggregates };
```

So the overwhelmingly common host — Wolverine plus one store — assembles **two** models out of the box, and every host has to restate a name it already declared. That is what marten#5405 was, and threading the name through was a fix for the symptom rather than the default.

Bobcat's `[assembly: EventModelName]` exists for exactly this reason, and its doc comment says so: the attribute is there so a spec assembly can be told to use *the service name*, because that is what the derived model is called.

## The store can read it, lazily, with no new plumbing

`JasperFxOptions` is registered as a singleton (`jasperfx/src/JasperFx/JasperFxServiceCollectionExtensions.cs:87`), and `ServiceName` defaults to the entry assembly name. The source already resolves lazily and is handed the `IServiceProvider`, so the fallback is available at the point it is needed with no capture at registration and no ordering caveat.

## Proposed

```csharp
var modelName = store.Options.EventModelName
    ?? services.GetService<JasperFxOptions>()?.ServiceName
    ?? ProjectionEventModelSource.DefaultModelName;
```

An explicit `EventModelName` still wins, which is what a modular monolith needs when each module's store is genuinely its own bounded context. The literal stays as the last resort for a host with no JasperFx options at all.

Worth settling across the three stores together, since the point is that they agree. Companions: JasperFx/fisher and JasperFx/polecat.

Context: JasperFx/jasperfx#837, on why one service is not one model in either direction.

jeremydmiller:
Adopts JasperFx 2.69.3 and fixes the Event Model default that #5405 was really about.

## JasperFx 2.69.1 → 2.69.3 (all five packages)

- **2.69.2** (jasperfx#832/#833) — codegen output is now deterministic. `MethodFrameArranger.compileFrames` walked an `ImHashMap` keyed by `Frame`, which does not override `GetHashCode`, so step 3 followed identity hash codes and two arrangements of the same logical method emitted the same statements in a different order.
- **2.69.3** (jasperfx#836/#837) — `EventModelSliceDescriptor.Origin` carries the contributing source's subject, so two stores projecting a document of the same simple name leave a `SourceDisagreement` hotspot naming both rather than one silently winning. `EventModelSetDescriptor` pairs a service with every model it hosts.

**Neither needs a Marten change.** `DocumentStore` already builds `EventStoreUsage(Subject, this)`, so every slice is attributed to the store's own subject — which, as of #5410, finally follows `StoreName`.

## Closes #5408 — the default model name is the service name

The store-derived source fell back to the literal `"EventModel"`. That is the one default guaranteed to be wrong for every host: Wolverine names its derived model after `JasperFxOptions.ServiceName`, and so does Bobcat's spec assembly. So the overwhelmingly common host — Wolverine plus one store — assembled **two** models out of the box and had to restate a name it had already declared.

That is what #5405 actually was. Threading a name through `AddMarten` was a fix for the symptom rather than the default.

```csharp
var modelName = (store as DocumentStore)?.Options.EventModelName   // explicit still wins
                ?? serviceName                                      // JasperFxOptions.ServiceName
                ?? ProjectionEventModelSource.DefaultModelName;     // last resort
```

Resolved from the container when the model is assembled, not captured at registration — so `AddJasperFx(opts => opts.ServiceName = …)` may come before or after `AddMarten`. Verified that `ReadHostEnvironment`, the `PostConfigure` that runs last, touches `GeneratedCodeOutputPath` and `ApplicationAssembly` but never `ServiceName`.

## #5406 cannot be done on master, and this PR says so instead of pretending

The issue asks for the committed generated output to be regenerated under a deterministic 2.69.2. **There is nothing on master to regenerate.** `#4438` and `#4461` removed the runtime code-generation pipeline outright, deleted the generated providers, and dropped the `TestCodeGen` build target:

- master tracks **2** files under `Internal/Generated`, both `fingerprint.json`;
- the **14 generated providers the issue counts live on the `8.0` branch**, not master;
- `src/ValueTypeTests/Internal/Generated` does not exist on disk — its `<Compile Remove>` entries are dead references;
- there is no `RuntimeCompiler` reference and no `codegen write` command left to run;
- the docs already state in four places that the step is gone and the folder should be deleted, so the issue's "docs may hedge" item has nothing to fix either.

The version-history note in `Directory.Packages.props` records this so the next reader does not go looking. **See the question in the review comments about how you want #5406 dispositioned.**

## Verification

CoreTests net9.0 **12/12** on the affected class.

Verified load-bearing, not assumed: dropping the `ServiceName` fallback fails **exactly** the two new tests (2 failed / 10 passed); restoring returns 12/12. The discriminating test has the service name and the declared model name disagree on purpose, so a store that merely adopted the declared name would fail it.

Five unrelated failures on the dev machine (advisory-lock contention, a leftover `fkey_mt_events_stream_id` constraint) **fail identically on unmodified master**, checked in a clean worktree — environmental, not this change. CI runs each suite against a fresh database.


jeremydmiller:
### #5406 needs a disposition — its premise does not hold on master

The issue asks to bump to 2.69.2 (done here, at 2.69.3) and then **regenerate the committed generated output**. There is nothing on master to regenerate, and no command left to do it with. Checked rather than assumed:

| Check | Result |
|---|---|
| Files tracked under `Internal/Generated` on `master` | **2** — both `fingerprint.json` |
| Same on `origin/8.0` | **14** — the providers the issue counts |
| `src/ValueTypeTests/Internal/Generated` on disk | does not exist |
| `TestCodeGen` build target | removed by `3ca630a45` |
| `JasperFx.RuntimeCompiler` reference | none anywhere |
| Docs still telling people to run `codegen write` | none — four pages say the opposite |

`ab074aa64` ("Eliminating Runtime Code Generation for the Document Database Features", #4438) deleted the providers, and `3ca630a45` (#4461) removed the pipeline and the build target. The `<Compile Remove="Internal\Generated\...">` entries left behind in `ValueTypeTests.csproj` (12) and `DocumentDbTests.csproj` point at files that no longer exist.

The issue's third item — "pre-generation docs mention determinism, if they currently hedge" — is also moot. The docs do not hedge; `configuration/cli.md`, `devops/devops.md`, `configuration/aot-publishing.md` and `migration-guide.md` all state the step is gone and the folder should be deleted.

**Three ways to go, your call:**

1. **Close #5406 as already handled** — the determinism fix ships here via 2.69.3 and Marten has no generated output to stabilise. The acceptance boxes cannot be ticked because the thing they describe was deleted two releases ago.
2. **Retarget it at the `8.0` branch**, which does track the 14 providers and still has the `TestCodeGen` target, so the regenerate-and-diff exercise is real there.
3. **Narrow it to the dead `<Compile Remove>` entries** — a genuine, small tidy in exactly this area, and the only actionable Marten-side remnant on master. Deliberately not folded into this PR, to keep the bump reviewable.

I did not fold any of these into this PR unasked.

PR #5430, opened 2026-09-14
Stamp patched documents and replaced events with the session's actual instant (#5379)

### LINKED ISSUE #5379: Patching breaks `mt_last_modified` timestamp on non UTC servers
When migrating data using a patch operation, the `mt_last_modified` timestamp gets skewed by the local time zone offset. For example, I'm in UTC+2 and the timestamp is set to two hours in the past (`2026-09-11 08:59:34.489183+02` instead of `2026-09-11 10:59:34.489183+02`). The bug is this line as far as I understand: https://github.com/JasperFx/marten/blob/1850d2d4575af3c8d36acdb9d822794ee644e028/src/Marten/Storage/Metadata/LastModifiedColumn.cs#L27

This is a problem for me as I use `ModifiedSince` in some queries.

Claude seemingly detected the problem and did some analysis [here](https://github.com/JasperFx/marten/issues/5136), but closed that issue.

erikshafer:
Closes #5379.

`session.Patch<T>()` re-stamped `mt_last_modified` with `now() at time zone 'utc'`. That expression strips the offset to a naive timestamp holding UTC wall-clock time, and assigning a naive timestamp to a `timestamptz` column re-interprets it in the session `TimeZone`. On any database whose `TimeZone` is not UTC the stored instant was off by the UTC offset. The insert default (`transaction_timestamp()`) was always right, so a document was stamped correctly on insert and jumped on its first patch, which is what broke the reporter's `ModifiedSince` queries.

This is the narrow patching-path exposure that #5136 recorded when it was closed as latent. The update is statement SQL rather than DDL, so existing stores see no schema delta.

## What changed

- `LastModifiedColumn.WriteMetadataInUpdateStatement` now writes `transaction_timestamp()`, the same expression as the column default.
- `ReplaceEventOperation` had the identical expression against `mt_events.timestamp` (also `timestamptz`, default `now()`). It is reachable through `CompactStreamAsync` and `CompletelyReplaceEvent`, so the `Compacted<T>` snapshot picked up the same skew. It now writes `now()`. I can split this into its own PR if you'd rather keep this one to the document side.

## Tests

CI and the dev container run Postgres at `Etc/UTC`, where the buggy and correct forms are byte-identical, so each test sets `Timezone=` on the Npgsql connection string instead. Both are theories over `Europe/Berlin` (skew reads in the past) and `America/Chicago` (skew reads in the future), and both assert on the stored instant, bracketed by the server-stamped insert below and client UTC now above.

- `PatchingTests/Patching/Bug_5379_patch_keeps_last_modified_correct_on_non_utc_session.cs`, which also checks the reporter's symptom: the patched document shows up in a `ModifiedSince` query.
- `EventSourcingTests/Bugs/Bug_5379_replaced_event_timestamp_on_non_utc_session.cs`.

Before the fix all four cases fail with the expected offsets. After it, PatchingTests, EventSourcingTests and DocumentDbTests are green on net10.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code) beep boop boooooop 🤖


jeremydmiller:
Rebased this onto master and resolved the conflicts — it's mergeable now.

**What conflicted, and why:** the production fix landed independently in #5445 after you opened this, so master already carries a byte-equivalent change to both files — `transaction_timestamp()` in `LastModifiedColumn.WriteMetadataInUpdateStatement` and bare `now()` in `ReplaceEventOperation`. Both conflicting files are resolved to master's side, so this PR is now purely the two test files.

**Your tests are kept, and they are not redundant — they close two real gaps in what #5445 shipped.** That's worth spelling out, because the obvious move here was to close this as a duplicate and it would have been the wrong call:

1. **The event side had no coverage at all.** #5445 changed `ReplaceEventOperation` and tested only the document/patch path. Grepping master, there is no non-UTC timezone test anywhere in `EventSourcingTests` — so compaction re-stamping a surviving event's timestamp was fixed and then held by nothing. `Bug_5379_replaced_event_timestamp_on_non_utc_session` is the only thing covering it.

2. **Your assertions are two-sided; #5445's patch test is not.** It pins one zone (`Europe/Berlin`) and asserts `skew < 1 minute` with no upper bound. Under a *negative* offset the buggy expression stamps the instant into the **future**, which that assertion passes happily. Yours run as a theory over `Europe/Berlin` and `America/Chicago` and bound both ends, so they fail in either direction.

**Verified rather than assumed.** I reintroduced the bug (`timestamp = (now() at time zone 'utc')` in `ReplaceEventOperation`) and **both** zones of your event test failed; restoring the fix makes all four green. That's exactly the property that matters here, since CI runs Postgres at `Etc/UTC` where the two spellings are byte-identical in effect and no test can see the difference.

Nice catch on the two-zone theory — that's the part I'd have gotten wrong.

Issue #5383, opened 2026-09-12
Explorer reads: implement the database-scoped overloads, and decide the tenant predicate by tenancy style rather than cardinality


jeremydmiller:
Downstream of **jasperfx#810**, whose abstraction half landed in JasperFx/jasperfx#817: `IEventStore` now carries database-scoped explorer reads alongside the tenant-scoped ones.

```csharp
Task<IReadOnlyList<StreamSummary>> GetRecentStreamsAsync(IEventDatabase database, int count, string? tenantId, CancellationToken ct);
IAsyncEnumerable<EventRecord>      ReadStreamAsync(IEventDatabase database, string streamId, string? tenantId, CancellationToken ct);
Task<StreamMetadata?>              GetStreamMetadataAsync(IEventDatabase database, string streamId, string? tenantId, CancellationToken ct);
IAsyncEnumerable<EventRecord>      QueryByTagsAsync(IEventDatabase database, IReadOnlyDictionary<string, string> tags, string? tenantId, CancellationToken ct);
Task<IReadOnlyList<ProjectionStatus>> GetProjectionStatusesAsync(IEventDatabase database, string? tenantId, CancellationToken ct);
```

The default delegates on a `Single`-cardinality store and **throws** on a multi-database one, so a single-database Marten store is correct as it stands and a multi-tenanted-databases store refuses until Marten implements them. Nothing breaks at compile time — these are default interface members.

## 1. Implement the database-scoped overloads

Route through `Tenancy.FindOrCreateDatabase` / the database's own session rather than `openExplorerSession()`, so a tool can enumerate `AllDatabases()` and read each one, attributing the answer to the database it came from. That is what CritterWatch#1231 needs: a console over 512 shard databases / 2,173 tenants currently lists recent streams from one database when no tenant is selected, and the result is indistinguishable from a complete answer.

## 2. Decide the tenant predicate by tenancy style, not cardinality — code-read, NOT reproduced

`DocumentStore.EventStoreExplorer.cs` decides between the two tenancy models by cardinality:

```csharp
var spansSeveralDatabases = Options.Tenancy.Cardinality != DatabaseCardinality.Single;
var scopeByColumn = tenantId != null && !spansSeveralDatabases;
```

[...]

Issue #5382, opened 2026-09-12
GetProjectionStatusesAsync throws NotSupportedException on a database-per-tenant store, so the projection registry is unreadable on the store shape that needs it


jeremydmiller:
`GetProjectionStatusesAsync` throws `NotSupportedException` on a **database-per-tenant** store, so the projection registry — "what is registered", which is a store-configuration fact and not a tenancy one — cannot be read on the store shape that most needs it.

Measured against **Marten 9.33.0 / JasperFx.Events 2.67.0**.

## Repro

```csharp
services.AddMarten(opts =>
{
    opts.MultiTenantedDatabasesWithMasterDatabaseTable(configure =>
    {
        configure.ConnectionString = masterConnectionString;
        configure.SchemaName = "public";
        configure.RegisterDatabase("tenant-a", tenantAConnectionString);
        configure.RegisterDatabase("tenant-b", tenantBConnectionString);
    });

    opts.Projections.Add<WidgetTallyProjection>(ProjectionLifecycle.Async);
});

var store = (IEventStore)documentStore;
await store.GetProjectionStatusesAsync(CancellationToken.None);   // NotSupportedException
await store.GetProjectionStatusesAsync(tenantId, CancellationToken.None); // NotSupportedException
```

Both overloads throw. This is pinned downstream — a passing CritterWatch test asserts the throw
(`the_poller_does_not_flag_live_projections_as_orphaned_without_a_registry` in
`db_per_tenant_progression_visibility`), so it is current behaviour at that pin rather than a recollection.

⚠️ **Pointer, not a diagnosis.** I have not read the implementation — my Marten checkout is on 8.x and does
not carry this method, and I did not want to name a throw site I had not seen. The exception TYPE is
verified by the assertion above; I am deliberately not quoting a message I did not capture. Note that the
`NotSupportedException` cannot be the JasperFx default: the tenant-less default on `IEventStore` throws
`NotImplementedException`, and `DocumentStore` explicitly implements both overloads
(`JasperFx-Events-IEventStore-GetProjectionStatusesAsync` appears twice in the shipped assembly).

## Why it matters

The registry answers "which projections are registered, and at what version". That is read off
`StoreOptions.Projections` — the same answer for every database and every tenant in the store. Tenancy
should not be able to make it unanswerable.

`AllShards()` demonstrates the point: it answers **fine** on exactly this store, because it is a registry
read rather than a tenancy-resolving one.

```
AllShards(): [WidgetTally:All]
[...]

PR #5431, opened 2026-09-14
#5383: database-scoped explorer reads, honest store-global reads, and two tenant-scoping defects

jeremydmiller:
Closes the **gap 2** half of #5383. The other parts of that issue — the database-scoped overloads themselves, the store-global read on a multi-database store, and the multi-database compliance arms — are untouched here and the issue should stay open for them.

## The bug

A tenant-scoped explorer read on a **sharded** store returns every co-located tenant's data.

`DocumentStore.EventStoreExplorer.cs` decided whether to filter rows by `tenant_id` from `DatabaseCardinality`, and its own doc comment stated the premise:

> **Database-per-tenant / sharded**: the argument names a physical database, so the session is opened against that tenant's own database and no `tenant_id` filter is needed — every stream in that database already belongs to the tenant.

True of database-per-tenant. **False of `ShardedTenancy`**, which distributes tenants across a *pool* of databases with conjoined events: many tenants share each database, so opening the tenant's database narrows the answer to its shard and no further.

## Reproduced first, because the issue said it had not been

#5383 filed this as a code read and was explicit about it — *"No repro was built — treat this as a pointer, not a diagnosis."* `sharded_explorer_tenant_scoping_5383` is that repro, written before the fix. It co-locates two tenants on one shard with `AddTenantToShardAsync` rather than relying on the hash distribution, so the test is not quietly testing the assigner. On unmodified `master`:

| call | returned |
| --- | --- |
| `GetRecentStreamsAsync(50, "alpha")` | `["beta-beta-one", "alpha-alpha-one"]` |
| `ReadStreamAsync("order-1", "alpha")` | events with tenant ids `["alpha", "beta"]` |

**The second is worse than a leak.** Under conjoined tenancy a stream's identity is `(tenant_id, id)`, so both tenants legitimately use the key `order-1` — and the read interleaves them into **one version-ordered sequence**. That is not a partial answer; it is a coherent-looking stream document that never existed.

## The fix

The two axes #5383 names are independent, and the single boolean could not express the configuration that needs both:

- **Which database do I open** — a question about `DatabaseCardinality`. Unchanged.
- **Do I filter rows once I am there** — a question about the event `TenancyStyle`. Now derived from that.

Four sites: `GetRecentStreamsAsync`, `ReadStreamAsync`, `QueryByTagsAsync`, `GetProjectionStatusesAsync`.

It is a predicate change rather than a new branch because it is safe on every other layout: on a database-per-tenant store with conjoined events every row in the opened database already carries that tenant id, so the predicate is redundant rather than wrong; on a non-conjoined store there is no `tenant_id` to scope by at all. The conjoined single-database and plain database-per-tenant behaviours are unchanged.

The doc comment is **corrected rather than deleted**, so the premise cannot be re-derived from the code later.

## Verification

- The repro fails on `master` and passes here.
- `EventSourcingTests` Explorer + Compliance — **572 passed**, 13 skipped.
- `TenantPartitionedEventsTests` in full — **252 passed**, 2 skipped.

Found while finishing the console half of JasperFx/CritterWatch#1237, which is what asked for these reads to be per-database in the first place.


jeremydmiller:
Pushed `9bcf2f06e` — this now carries **parts 1, 3 and 4** as well as the gap-2 fix it started as. Only part 5 (the multi-database compliance arms) is left on #5383.

## Part 1 — the five `IEventDatabase` overloads

`GetRecentStreamsAsync`, `ReadStreamAsync`, `GetStreamMetadataAsync`, `QueryByTagsAsync` and `GetProjectionStatusesAsync`.

Each read is **parameterised on an optional database** rather than duplicated per scope: one private body per read, with a shared `openExplorerSessionAsync` deciding which database to open — an explicit database wins, else a tenant on a multi-database store resolves to its own, else the default session. Three public scopes, one implementation, so they cannot drift.

⚠️ That helper answers only *which database*. Whether the SQL also carries a `tenant_id` predicate stays the independent question `eventsAreConjoined` answers — the two axes gap 2 is about.

## Part 3 — store-global reads

Listings and single-answer reads diverge, and the split is principled rather than a preference:

| read | store-global on a multi-database store |
| --- | --- |
| `GetRecentStreamsAsync` | **fans out** and merges |
| `QueryByTagsAsync` | **fans out** |
| `ReadStreamAsync` | **refuses**, naming `AllDatabases()` and the overload |
| `GetStreamMetadataAsync` | **refuses** |
| `GetProjectionStatusesAsync` | unchanged — the registry answer #5382 added |

A listing has a well-defined union, and asking each database for the same `count` then keeping the newest of the union is exact rather than approximate: no database can hide a stream newer than one returned. A stream id is unique *within* a database and not across them, so for the single-answer reads there is no merge that produces the one answer the signature returns. What goes away is the option that was actually wrong — answering from whichever database the default session resolved, which is indistinguishable from a complete answer.

## Part 4 — a second defect, found auditing the siblings

`GetStreamMetadataAsync` had **no tenant predicate at all**:

```sql
where id = @stream_id limit 1
```

Under conjoined tenancy `mt_streams` holds one row per `(tenant_id, id)`, so a store where two tenants share a stream id returned **whichever row the scan reached first**, with nothing saying whose it was. It now has all three scopes and the predicate — the same thing polecat#593 fixed outright.

## Tests

`explorer_database_scoped_reads_5383` provisions **two genuinely separate physical databases**. That detail is load-bearing: my first attempt pointed both tenants at one database and the rows duplicated, which would have made every assertion pass for the wrong reason had the expectations been laxer. It pins that each database answers for itself and only itself, that a database-scoped stream read finds the stream in exactly one, that the store-global listing fans out, and that the store-global single-stream and metadata reads refuse by name.

| suite | result |
| --- | --- |
| `EventSourcingTests` Explorer + Compliance | **577 passed**, 13 skipped (was 572 — the five new) |
| `TenantPartitionedEventsTests` | **252 passed**, 2 skipped |
| `MultiTenancyTests` | **172 passed** |

## Left for #5383

Part 5, the multi-database **compliance** arms. Those want a shared fixture in `JasperFx.Events` — a database-per-tenant arm and a sharded arm — rather than Marten-local tests like the ones here, so every store enrols in the same facts. Worth its own change, and the issue should stay open for it.

jeremydmiller:
Pushed `0ff97b08f` — **part 5, which completes #5383.** All five parts plus gap 2 are now on this branch.

Part 5 turned out to be **enrolment rather than authoring**: the arms already ship upstream (jasperfx#822, JasperFx.Events 2.69.0) and Fisher is in both. `EventStoreExplorerCompliance` covers the database-scoped overloads on a *single*-database store, where they can only be asserted to **agree with** the store-global read — vacuously true of a store that ignores the argument. These two arms are where the argument has to mean something.

## The fixture derives the tenancy model rather than taking a flag

Two tenants naming the same database **is** sharding; one database each **is** database-per-tenant. Reading that off the suite's tenant map keeps the fixture honest about which store it actually built, and the two arms differ on exactly that axis.

Three details that keep the enrolment from being vacuous:

- **`SupportsMultipleDatabases` is derived from the config, never hardcoded.** Declaring it unconditionally would make every isolation fact pass vacuously on the single-database configurations — the failure these arms exist to catch, and what the seam's own remarks warn about.
- **`DatabaseForTenantAsync` goes through Marten's own `Tenancy.FindOrCreateDatabase`**, so the suite exercises the resolution the product ships rather than a test-local map.
- **Shard assignment is explicit (`AddTenantToShardAsync`), never the hash distribution.** The sharded arm's point is that `acme` and `globex` are *co-located*; letting an assigner place them would make the isolation facts depend on where it happened to put them.

## Results

Both arms green — **14 tests, 0 skipped**. The zero matters: the suite carries its own guard fact that fails when a fixture declares `SupportsMultipleDatabases` while `AllDatabases()` reports fewer than two, so a skipped or single-database run would not look like this.

**Negative control, run rather than argued.** With `eventsAreConjoined` forced to `false` — the gap-2 behaviour this branch fixes — the sharded arm goes to **3 failed / 4 passed**. So the arm is load-bearing, and gap 2 is now pinned by the *shared* suite every store enrols in rather than only by Marten's own test.

`EventSourcingTests` Explorer + Compliance: **591 passed**, 13 skipped (was 577 before these 14).

## #5383, part by part

| part | where |
| --- | --- |
| gap 2 — sharded tenant predicate | `5ec53d90c`, with the repro written first |
| 1 — database-scoped overloads | `9bcf2f06e` |
| 3 — store-global reads fan out or refuse | `9bcf2f06e` |
| 4 — sibling audit (found `GetStreamMetadataAsync` had no tenant predicate at all) | `9bcf2f06e` |
| 5 — multi-database compliance arms | `0ff97b08f` |

PR #5445, opened 2026-09-15
Stop mt_last_modified jumping by the server's UTC offset when a document is patched

### LINKED ISSUE #5379: Patching breaks `mt_last_modified` timestamp on non UTC servers

jeremydmiller:
Closes #5379. Supersedes the "latent" ruling on #5136 — which predicted exactly this: *"if a non-UTC deployment ever reports odd `mt_last_modified` values from patching."* One did.

## The defect

`mt_last_modified` is `timestamp with time zone`, and the patch path stamped it with `now() at time zone 'utc'`. That expression **strips** the offset to produce a naive timestamp holding UTC wall-clock time, and assigning a naive timestamp back to a `timestamptz` column re-interprets it in the session's `TimeZone`. On a database at UTC+2 the stored instant is two hours in the past — a genuinely different point in time, not a display artifact. Measured here at exactly `01:59:59.996` before the fix.

The column's own DEFAULT was always correct, so a row was stamped right on insert and then **jumped by the offset the first time it was patched**. The reporter hit it running a data migration through `Patch<T>()` and then finding `ModifiedSince()` silently no longer matched — the migration looked like it had not run.

## What changed

Two sites:

- **`LastModifiedColumn.WriteMetadataInUpdateStatement`** → `transaction_timestamp()`, matching the column's DEFAULT. The reported defect.
- **`ReplaceEventOperation`** → bare `now()`. It re-stamps `mt_events.timestamp` during stream compaction and had the identical expression. Reachable through `EventStore.StreamCompacting`, so a live path rather than a latent one. `mt_streams.timestamp` already used bare `now()`; this is now consistent with it.

> ⚠️ **`QuickAppendEventFunction` has the same expression and is deliberately not touched.** #5136 established that `CreateQuickAppendEventsOperation` is assigned in both the Postgres and SQL Server dialects and read by neither, so `mt_quick_append_events` is dead code — and changing the function body would force a schema delta on **every** store to correct an expression nothing executes. That was the maintainer's stated reason for leaving it, and it still holds.

## ⚠️ Two things I got wrong, both caught by running the tests rather than by reading

**I believed every document UPDATE ran the broken expression, and wrote a test saying so. It passed against the unfixed code.** `PatchOperation` is the *only* caller of `WriteMetadataInUpdateStatement`; an ordinary `Store`/`SaveChanges` goes through the generated upsert and takes the column DEFAULT. #5136's own correction says this — I only believed it after the test refused to fail.

That test is kept, renamed `an_ordinary_update_was_never_affected` and documented as passing either way, because *"which writes were wrong"* is the first question an upgrader asks, and the answer — **only patching** — is worth holding rather than re-deriving.

**The first version of this fix went into the wrong checkout entirely** (the `8.0` branch in `~/code/marten` rather than this worktree), which is why the patch test kept failing at exactly two hours against what looked like fixed code. Those stray edits have been reverted there.

## The tests pin a non-UTC TimeZone, and that is the whole point

At UTC the buggy and correct spellings are **byte-identical in effect**, so the entire existing suite — and CI, and the dev container, all at `Etc/UTC` — is structurally incapable of seeing this class of bug. #5136 says so explicitly, and it is why that issue was closed as latent before a user reported it from production.

`the_connection_really_is_not_at_utc` guards the guard: without it, every other fact in the file would pass against the defect.

**Two of the four facts fail without this change and pass with it.**

## Verification

**1113** DocumentDbTests, **133** PatchingTests, **64** event-protection/compaction tests.


BaerMitUmlaut:
Please reconsider blindly merging AI generated code. A wordy multiline comment of why you shouldn't write wrong code that has no reason to be there really isn't helpful but just confusing.

jeremydmiller:
@BaerMitUmlaut Hey man, one more comment like that and I'll happily ban you permanently from the organization just for irritating me. Are we clear? All of this code is heavily curated. Do mistakes happen sometimes? Has bad AI code snuck by sometimes? Occasionally, yes. Do those get addressed pretty quickly anytime we find them? Yes to that as well.

PR #5453, opened 2026-09-15
Name the event type index, and say what adding it costs

erdtsieck:
## The problem

`EnableEventTypeIndex` is recommended for large event stores — its own docs list "projection rebuilds time out" and "millions of events" as the symptoms — but it is added as a plain `CREATE INDEX`:

```csharp
if (events.EnableEventTypeIndex)
{
    Indexes.Add(new IndexDefinition("idx_mt_events_event_type_seq_id") { Columns = ["type", "seq_id"] });
}
```

That holds `ACCESS EXCLUSIVE` on `mt_events` for the whole build. So the stores the option exists for are exactly the ones that cannot afford to create it: on ours it would be a write outage at `db-apply`, before the pods start. And with `UseTenantPartitionedEvents` there is no `CREATE INDEX CONCURRENTLY` to reach for by hand either — PostgreSQL refuses `CONCURRENTLY` on a partitioned parent.

Nothing in the docs says this, on the page that recommends the option.

## What this PR does

Two small things, and deliberately not a third:

- **`EventGraph.EventTypeIndexName`**, beside `HStoreTagIndexName`. The documented way out — ignore the index and own it — currently means hard-coding `"idx_mt_events_event_type_seq_id"` and hoping it stays put.
- **The optimizing page** now says what adding the index costs on a store that already has events, and what the partitioned case means, instead of leaving people to find out at deploy time.

No behaviour change: the index is declared exactly as before.

## What I left out, and why

My first draft added `BuildEventTypeIndexConcurrently`, mirroring `BuildHStoreTagIndexConcurrently` from #5268. It worked, but it would have been the **second** per-index flag on one table, and a third index on `mt_events` would have wanted a third.

Whether an index needs a concurrent build is not a property of the index. It is whether the table is being created — nothing to scan — or altered, where there are rows and the lock matters. Weasel's `TableDelta` already knows which, so I proposed it there instead: **[JasperFx/weasel#594](https://github.com/JasperFx/weasel/pull/594)**. If you take that one, this index needs no flag at all, and `BuildHStoreTagIndexConcurrently` becomes redundant along with the docs that tell people to set it.

Happy to go back to the flag here if you would rather have it in Marten — it is a small change either way, and I did not want to pick that for you.

## Where this comes from

We took four read models out of a composite projection, which gave each of them a narrow event filter for the first time (~18 types out of the whole store) — precisely the case the docs describe, and the point at which we found we could not turn the index on. Our `mt_events` is 3.4 TiB across 512 shards, tenant-partitioned.

## One thing I found next door, not fixed here

`Events.IgnoreIndex(...)` does not take effect when the event tables are **created**, only when they are migrated. `Table.WriteCreateStatement` writes every entry of `Indexes` without consulting `IgnoredIndexes`, so against an empty database both `idx_mt_events_tags` and `idx_mt_events_event_type_seq_id` are created even with both ignored. Measured with both ignored in one store on a fresh database:

```
INDEXES: idx_mt_events_event_type_seq_id, idx_mt_events_tags, pk_mt_event_progression, …
```

That reads as a bug against the DCB docs ("Marten leaves it out of the schema diff in **both** directions — it will neither create it nor treat one you built yourself as drift"), and it is what the constant above is for, so it is worth knowing. It is a Weasel-level fix and unrelated to this change; happy to raise it separately if it is not already known.


erdtsieck:
Question on the shape before you spend review time on it: should this be a flag at all?

This PR adds the *second* per-index opt-in (`BuildHStoreTagIndexConcurrently`, now `BuildEventTypeIndexConcurrently`), and a third index on `mt_events` would want a third. What every one of them really means is "this index is being added to a table that already has rows in it" — and Weasel already knows that, in the one place where it matters:

- `Table.WriteCreateStatement` (Weasel.Postgresql/Tables/Table.cs:142) — `foreach (var index in Indexes) … index.ToDDL(this)`, the table does not exist yet
- `TableDelta.WriteUpdate` (Weasel.Postgresql/Tables/TableDelta.cs:161) — `foreach (var indexDefinition in Indexes.Missing) … ToDDL(Expected)`, the table does

Both call the same `ToDDL`, so `IsConcurrent` today also applies on the create path, where it buys nothing (the table is empty) and where `CONCURRENTLY` is not even legal against a partitioned parent. Inferring it at the delta instead would drop the flags entirely and fix that asymmetry: plain `CREATE INDEX` inline when the table is created, `CONCURRENTLY` when one is added to a table that is already there.

I did not do that here because it is a Weasel change and it is not mine to decide for every consumer. The arguments I can see against making it automatic:

- **The generated script changes shape for everyone.** A concurrent build cannot run inside a transaction, so `db-patch` / `db-dump` output stops being runnable as one block — that is exactly the trade-off #5268 documented. The runtime apply path is fine either way: `PostgresqlMigrator` already splits on the `--WEASEL_INDEX_CREATION_BEGIN` markers and each statement auto-commits.
- **Failure semantics change.** A failed `CONCURRENTLY` leaves an invalid index behind that has to be dropped before a retry, rather than rolling back cleanly.
- Someone with a small table and a maintenance window may genuinely prefer the blocking build.

So: happy to leave this as the flag it is for consistency with #5268, happy to rework it into one store-level switch, or happy to take a shot at the inference in Weasel with an opt-out for the three reasons above — whichever you would rather review. Just say which and I will follow up.

erdtsieck:
Closing this in favour of [JasperFx/weasel#594](https://github.com/JasperFx/weasel/pull/594), which puts the fix where the decision actually is.

What was left here after I dropped the flag was a name constant and a docs paragraph, and neither survives contact with that PR: `Advanced.Migrator` is public, so a consumer can set `BuildIndexesConcurrentlyOnAlter` without Marten changing at all, and what the optimizing page should say about `EnableEventTypeIndex` depends on whether the delta ends up building it concurrently. Writing that page for the current behaviour now would mean rewriting it right after.

Happy to come back with a docs PR once weasel#594 is decided either way — the page recommending this index still does not mention that adding it takes ACCESS EXCLUSIVE, which is what sent me here.

Issue #5457, opened 2026-09-17
Async daemon leaks one Postgres connection per batch for an EfCoreSingleStreamProjection


wpei-infotrack:
`EfCoreProjectionStorage<TDoc,TId,TDbContext>` owns a `DbContext` per tenant and batch, and implements
neither `IDisposable` nor `IAsyncDisposable`. Running an `EfCoreSingleStreamProjection` under
`ProjectionLifecycle.Async` therefore leaks one Npgsql connection per batch until the server refuses new
ones.

### Versions

| | |
| --- | --- |
| `Marten` | 9.37.0 |
| `Marten.EntityFrameworkCore` | 9.37.0 |
| `Npgsql` | 10.0.0 (transitive) |
| `Npgsql.EntityFrameworkCore.PostgreSQL` | 10.0.0 (transitive) |
| TFM | `net10.0` |
| PostgreSQL | 18.3 — `max_connections=100` |

### Reproduction

Reduced from the application where this was observed. The shape is exact; the names are not, and the
reduction itself has not been run standalone.

```csharp
public sealed class Widget
{
    public Guid Id { get; set; }
    public string Name { get; set; } = "";
    public long Version { get; set; }
}

public sealed class WidgetDbContext(DbContextOptions<WidgetDbContext> options) : DbContext(options)
{
    public DbSet<Widget> Widgets => Set<Widget>();

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        var widget = modelBuilder.Entity<Widget>();
        widget.ToTable("widgets", "app");
        widget.HasKey(w => w.Id);
    }
}

public sealed class WidgetProjection : EfCoreSingleStreamProjection<Widget, Guid, WidgetDbContext>
{
    public WidgetProjection() => IncludeType<WidgetRenamed>();

    public override void ConfigureDbContext(DbContextOptionsBuilder<WidgetDbContext> builder) { }

    public override Widget ApplyEvent(
        Widget? snapshot, Guid identity, IEvent @event, WidgetDbContext dbContext, IQuerySession session)
    {
        var widget = snapshot ?? new Widget { Id = identity };
        if (@event.Data is WidgetRenamed renamed) { widget.Name = renamed.Name; }
        widget.Version = @event.Version;
        return widget;
    }
}
```

Registration:

```csharp
services.AddMarten(options =>
    {
        options.Connection(connectionString);
        options.DatabaseSchemaName = "app";
[...]

PR #5466, opened 2026-09-21
Fix #5457 — dispose the DbContext an EF Core projection creates per batch

jeremydmiller:
Addresses #5457. Thanks to @arnelirobles, whose comment on the issue was right on both counts and shaped this — the context really is never disposed, and the backend growth really does not reproduce.

## The defect

`EfCoreProjectionStorage` owns a `DbContext` built once per tenant per batch (`AggregationRunner` → `FetchProjectionStorageAsync` → the factory registered in `RegisterEfCoreStorage`), and **nothing disposed it, on any path**. Not the storage — `IProjectionStorage<,>` in JasperFx declares no disposal contract at all, so the type that owns the context has nothing to hook. Not the daemon, not the session. The only thing disposed was the separately registered `DbContextTransactionParticipant`, and it released just the placeholder `NpgsqlConnection`, never the context.

## The fix

Disposal goes to that participant, because its lifetime already matches the context's exactly: the daemon drains `ProjectionUpdateBatch._transactionParticipants` in `DisposeAsync`, once per batch, on the success and failure paths alike.

**This keeps the fix inside Marten.** The obvious alternative was to widen `IProjectionStorage<,>` upstream with a disposal contract and have `AggregationRunner` call it — architecturally tidier, but it drags Polecat and Fisher through a JasperFx publish cycle to buy behaviour this already gets. Happy to do it that way instead if you'd rather the contract be explicit; see the open question at the bottom.

The context is deliberately **not** torn down at the end of `BeforeCommitAsync` the way the placeholder connection is. The connection is provably finished with there — it has just been swapped out of the context — whereas the context is still reachable by an inline projection whose session has not ended, so releasing it there would be a behaviour change rather than a leak fix.

### Second leak, fixed in the same pass

The live-aggregation fallback in both projection types builds a context and registers a participant only `if (session is ITransactionParticipantRegistrar registrar)`. A plain `IQuerySession` — what `AggregateStreamAsync` passes — is not one, so the participant was never created and **nobody owned the connection**, which `EfCoreDbContextFactory.Create` opens eagerly whenever a schema name is configured (always: `DatabaseSchemaName` defaults to `public`). That branch now disposes what it creates.

## Tests

`Bug_5457_efcore_dbcontext_never_disposed` counts `DbContext` construction against disposal instead of counting backends. That is the point: the reported symptom needs the context to own its connection before it surfaces, while the defect underneath is present either way and pins deterministically, with no pool timing, no `pg_stat_activity` sampling, and no dependence on anyone's PostgreSQL or Npgsql version. On master it reports **5 created / 0 disposed**. It also asserts a context *was* built per batch, so it can't pass vacuously.

`Bug_5457_async_daemon_projection_storage_leak` covers the reported symptom directly — backend growth across 15 daemon batches — and is honest about its status below.

Full `Marten.EntityFrameworkCore.Tests`: 56 passed / 0 failed on **both** net9.0 (EF Core 9 / Npgsql 9) and net10.0 (EF Core 10 / Npgsql 10).

## What this does not do

It does not reproduce the reporter's monotonic backend growth. On PostgreSQL 17.10 with EF Core 10 / Npgsql 10 on net10.0, a 15-batch daemon run sits flat at 2 backends before and after — because the participant *does* release the placeholder connection every batch, exactly as @arnelirobles measured. The connection-count test passed before this change too; it is a guard, not a pin, and I've said so in its doc comment rather than dressing it up.

@wpei-infotrack — the leading hypothesis for your numbers is a `DbContext` that owns its own connection rather than the placeholder Marten hands it: any `OnConfiguring` override, or a `ConfigureDbContext` that calls `UseNpgsql` with a connection string. In that shape the undisposed context held a real backend, and this fix reclaims it. That would also explain why every leaked backend's last statement was the projection's own load. Two things would confirm it:

1. Does your real `WidgetDbContext` override `OnConfiguring`, or does your real `ConfigureDbContext` call `UseNpgsql`? (The reduction in the issue leaves both empty, and that shape does not leak here.)
2. Is the projection multi-tenanted? One context per tenant per batch means a multi-tenanted store multiplies whatever the per-batch cost is.

If the answer to both is no, I'd like to keep #5457 open after this merges — this fixes two real leaks, but it would not yet be the one you hit.

## Open question

Should `IProjectionStorage<,>` get an explicit disposal contract in JasperFx, so storage that owns resources is disposed by the runner rather than by a participant that happens to share its lifetime? That is the structurally correct fix and would cover Polecat and Fisher too. It is a cross-repo change with a publish cycle, so I have not made that call unilaterally.


arnelirobles:
Glad it was useful, and thanks for the credit.

On the open question: I'd keep it in Marten for now. The participant's lifetime does match the context exactly, and widening `IProjectionStorage<,>` to buy behaviour you already get means a JasperFx publish cycle plus Polecat and Fisher moving with it. Worth doing when something else needs that contract, not for this one.

The `ITransactionParticipantRegistrar` branch is the better catch here anyway. That one has no owner at all. My instrumentation only showed the contexts piling up, I never proved the connections were held, so your fix covers more than I could demonstrate.

If @wpei-infotrack answers no to both questions, happy to take another pass at the multi-tenant path. That's the one I never exercised.

Unrelated nit: do you mean to keep the `claude.ai/code/session` link in the PR body? It happened to me too on a few of mine, so just checking it isn't a template leak.

